PR #130, opened 2026-04-08
fix(agent-sec-core): Refactor SKILL.md to executable protocol and align sub-skills

edonyzpc:
## Related Issue

closes #

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [x] `sec-core` (agent-sec-core)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] Multiple / Project-wide

## Checklist

- [x] For `sec-core` (Rust): `cargo clippy -- -D warnings` and `cargo fmt --check` pass
- [x] For `sec-core` (Python): Ruff format and pytest pass

### DIFF
--- a/src/agent-sec-core/README.md
+++ b/src/agent-sec-core/README.md
@@ -21,8 +21,12 @@ As AI Agents gradually gain OS-level execution capabilities (file I/O, network a
 ```
 ┌─────────────────────────────────────────────┐
 │              Agent Application              │
-├─────────────────────────────────────────────┤
-│     Security Decision (Risk Classification) │
+├──────────────────┬──────────────────────────┤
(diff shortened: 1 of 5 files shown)


1570005763 (approved):
LGTM.

Heads up — agent-sec-core is also blocking legitimate skill changes, which is kind of expected for now since we're being conservative on the security side. But going forward, we'll probably need to think about finer-grained risk management around skill signing — like, being able to tell apart a normal incremental update from something actually suspicious, instead of flagging everything the same way.

PR #283, opened 2026-04-22
feat(cosh): support immediate hook activation with extension 

kongche-jbw:
## Description

This PR makes extension hook registration and deregistration take effect **immediately within the running session**, without requiring a restart. Previously, hooks declared in a `cosh-extension.json` were only loaded at startup, so any extension installed or uninstalled at runtime had no effect on the active hook registry.

Key changes:

- **Immediate hook activation**: After a successful `extensions install`, `hookSystem.initialize()` is called so the new extension's hooks are registered in the current session.
- **Immediate hook deactivation**: After `extensions uninstall`, `extensionManager.refreshCache()` is called unconditionally (regardless of whether the hook system is available) to keep the in-memory cache consistent with disk. `hookSystem.initialize()` is then called to deregister the removed extension's hooks.
- **System extension recovery**: Because `refreshCache()` re-scans `SYSTEM_EXTENSIONS_DIR` before user-level extensions, a same-named system extension that was shadowed by the uninstalled user extension will have its hooks restored automatically.
- **Fault isolation**: `refreshCache` and `hookSystem.initialize` each have their own `try/catch`; a failure in either step is logged with an accurate message and does not surface as an install/uninstall error to the user.
- **Local path install fix**: `isOwnerRepoFormat` now explicitly rejects paths starting with `.` or `/`, preventing `extensions install ./local-path` from being misidentified as a GitHub `owner/repo` reference.
- **TOML `run` field**: Commands can now declare a `run` field instead of `prompt` to execute a shell script directly without sending anything to the model. Output is streamed back to the UI as a `message`.

## Related Issue

no-issue: self-contained feature improvement to extension lifecycle management

## Type of Change

## Scope

- [x] `cosh` (copilot-shell)

## Checklist

## Testing

New unit tests added to `extensionsCommand.test.ts` (7 cases):

```
install
  ✓ should call hookSystem.initialize() after installing an extension with hooks
  ✓ should NOT call hookSystem.initialize() when installed extension has no hooks
  ✓ should still report install success when hookSystem.initialize() throws

uninstall
  ✓ should call refreshCache() then hookSystem.initialize() after uninstall
  ✓ should call refreshCache + hookSystem.initialize() even when user extension had no hooks
  ✓ should still report uninstall success when hookSystem.initialize() throws
  ✓ should still call refreshCache() even when getHookSystem() returns undefined
```

All 38 tests pass: `npx vitest run src/ui/commands/extensionsCommand.test.ts`

## Additional Notes

The `run` field in TOML commands is additive and backward-compatible — existing `prompt`-based commands are unaffected. The `example_hook/` directory added in this PR serves as both a functional example and a template for users who want to distribute hooks as an installable extension.

### DIFF
--- a/src/copilot-shell/packages/cli/src/ui/commands/extensionsCommand.test.ts
+++ b/src/copilot-shell/packages/cli/src/ui/commands/extensionsCommand.test.ts
@@ -499,6 +499,111 @@ describe('extensionsCommand', () => {
         expect.any(Number),
       );
     });
+
+    it('should call hookSystem.initialize() after installing an extension with hooks', async () => {
+      mockParseInstallSource.mockResolvedValue({
+        type: 'git',
+        source: 'https://github.com/test/extension',
+      });
+      mockInstallExtension.mockResolvedValue({
+        name: 'test-extension',
+        version: '1.0.0',
+        hooks: { PreToolUse: [{ matcher: 'run_shell_command', hooks: [] }] },
+      });
+      const mockHookInitialize = vi.fn().mockResolvedValue(undefined);
+      const ctx = createMockCommandContext({
+        services: {
+          config: {
+            getExtensions: mockGetExtensions,
+            getWorkingDir: () => '/test/dir',
+            getExtensionManager: () => realMockExtensionManager,
+            getHookSystem: () => ({ initialize: mockHookInitialize }),
+          },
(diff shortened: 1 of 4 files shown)


samchu-zsl (approved):
LGTM, thx!

PR #290, opened 2026-04-22
feat(cosh): add esc key to cancel running slash commands

### LINKED ISSUE #272: [cosh] feat: Add ESC key to cancel running slash commands
### Component

cosh

### Problem Statement

When executing long-running slash commands such as /compress or /summary, users have no way to cancel the operation once it starts. The input remains blocked until the command completes, which can be frustrating when:

- The command takes longer than expected
- The user realizes they didn't want to run that command
- The command appears to be stuck or unresponsive
- Currently, users must wait for the command to finish or force-quit the application, neither of which provides a good user experience.

### Proposed Solution

Add the ability to cancel running slash commands by pressing the ESC key. This feature should:

- Immediately unblock user input when ESC is pressed, without waiting for the underlying operation to complete
- Use AbortController + Promise.race to cancel pending operations cleanly
- Show a "Command cancelled" message to inform the user of the cancellation
- Suppress redundant error messages - when cancelled via ESC, don't show both the abort error and the cancel message

### Alternatives Considered

_No response_

### Additional Context

_No response_

kongld-mq:
## Description

feat(cosh): add esc key to cancel running slash commands

- Use AbortController + Promise.race for immediate unblocking on ESC press
- Pass abortSignal through CommandContext to commands for abort state checking
- Suppress abort error messages in UI and debug console when user cancels
- Suppress debug console errors by using debugLogger instead of console.error
- Add abort checks in compressCommand and summaryCommand catch blocks
- Add i18n support for cancel messages and SummaryMessage component

## Related Issue

closes #272 

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] Multiple / Project-wide

## Checklist


src/copilot-shell/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -87,6 +87,10 @@ vi.mock('../../utils/cleanup.js', () => ({
   runExitCleanup: mockRunExitCleanup,
 }));
 
+vi.mock('./useKeypress.js', () => ({
+  useKeypress: vi.fn(),
```
samchu-zsl: 🔴 **Missing unit coverage for the core feature.**

`useKeypress` is mocked as a no-op `vi.fn()`, so none of the new ESC-cancel logic is actually exercised. Please add tests that cover:

1. Pressing `escape` while `isProcessing === true` invokes `cancelSlashCommand`, which aborts the current `AbortController`, pushes an INFO item with `t('Command cancelled.')`, clears `pendingItem`, and flips `isProcessing` back to `false`.
2. When the controller is aborted mid-`Promise.race`, `handleSlashCommand` returns `{ type: 'handled' }` and does **not** fall into the error branch (no ERROR item, no `SlashCommandStatus.ERROR` event).
3. `CommandContext.abortSignal` is actually threaded into the command action, and `commandToExecute.action` receives a signal whose `.aborted` becomes `true` after ESC.
4. `useKeypress` is only subscribed when `isProcessing === true` (no global leak).

Without these, the feature is effectively untested per `issue-fix.md` Phase 6 and the PR-template checklist item *“I have added tests that prove my feature works.”*

src/copilot-shell/packages/cli/src/ui/commands/compressCommand.ts
```diff
@@ -96,6 +97,11 @@ export const compressCommand: SlashCommand = {
 
       const compressed = await doCompress();
 
+      // If cancelled via ESC, return early (cancelSlashCommand already handled UI)
+      if (abortSignal?.aborted) {
```
samchu-zsl, line 101: 🔴 **`abortSignal` is not propagated into `tryCompressChat`, so this is effectively a soft cancel.**

This post-`await` check only fires *after* the LLM call has already finished — `geminiClient.tryCompressChat(promptId, true)` above does not accept an `AbortSignal` (see `packages/core/src/core/client.ts` `tryCompressChat(prompt_id, force = false)`). Pressing ESC will show `Command cancelled.` immediately in the UI while the underlying LLM request keeps running to completion, wasting tokens and still consuming context.

This is inconsistent with `summaryCommand`, which in this same PR was updated to pass `abortSignal` through to `generateContent` so the network call actually aborts.

Please either:
- Extend `tryCompressChat(prompt_id, force, signal?: AbortSignal)` and thread the signal all the way into the underlying `generateContent` call; or
- If that's out of scope for this PR, document the limitation explicitly in the PR description / a code comment here (e.g. `// NOTE: compression is a soft cancel — the in-flight LLM call is not interrupted`).

src/copilot-shell/packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -137,6 +147,34 @@ export const useSlashCommandProcessor = (
...
+    addItem(
+      {
+        type: MessageType.INFO,
+        text: t('Command cancelled.'),
+      },
+      Date.now(),
+    );
+    setPendingItem(null);
+    setIsProcessing(false);
+  }, [addItem, setIsProcessing]);
```
samchu-zsl: 🟡 **Missing `setPendingItem` in the dependency list.**

`setPendingItem` is used inside the callback but not listed in the deps array. `useState` setters are stable in practice, but the repo enables `react-hooks/exhaustive-deps`; I'd suggest including it explicitly for consistency with other callbacks in this file:

```ts
}, [addItem, setPendingItem, setIsProcessing]);
```

src/copilot-shell/packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -374,10 +417,24 @@ export const useSlashCommandProcessor = (
...
+
+            // Use Promise.race to allow ESC to cancel the command immediately
+            const abortPromise = new Promise<undefined>((resolve) => {
+              abortController.signal.addEventListener(
+                'abort',
+                () => resolve(undefined),
+                { once: true },
+              );
+            });
+            const result = await Promise.race([
```
samchu-zsl, line 429: 🟡 **`Promise.race` returns early, but the underlying action Promise keeps running.**

Once the abort wins the race, the outer code returns `{ type: 'handled' }` and `isProcessing` flips to `false`, but `commandToExecute.action(...)` continues executing in the background. Any `ui.addItem` / `ui.setPendingItem` / `setShellConfirmationRequest` calls that happen *after* the race but before the action finishes will still land in the UI — after `Command cancelled.` has been printed and potentially after the user has already started typing the next command.

This PR mitigates the two common paths (`compressCommand`, `summaryCommand`) by adding explicit `if (abortSignal?.aborted) return` checks, but any other built-in / file / MCP command without those checks will still pollute the UI on cancel. It also means the `finally` block further down still runs, recording this invocation as `SlashCommandStatus.SUCCESS` even though the user cancelled.

Suggest adding a centralized guard — either wrap `ui` in an abort-aware proxy that no-ops writes once the signal fires, or short-circuit writes and the SUCCESS telemetry path once `abortController.signal.aborted` is true — so the contract holds for *every* command, not just the two updated here.

src/copilot-shell/packages/core/src/core/openaiContentGenerator/errorHandler.ts
```diff
@@ -48,7 +51,7 @@ export class EnhancedErrorHandler implements ErrorHandler {
       const logPrefix = context.isStreaming
         ? 'OpenAI API Streaming Error:'
         : 'OpenAI API Error:';
-      console.error(logPrefix, errorMessage);
+      debugLogger.error(logPrefix, errorMessage);
```
samchu-zsl, line 54: ℹ️ **Scope note.**

Switching from `console.error` to `debugLogger.error` is a reasonable hygiene fix for Ink TTY (raw `console.error` corrupts the Ink canvas), and `setDebugLogSession` is wired up in `Config`'s constructor so the log actually lands in `debug.log`. The thrown `enhancedError` is still propagated upward and surfaced to the user through the normal UI error path, so this isn't a silent regression.

However, this change is unrelated to ESC-cancel and would be easier to review / revert independently if it lived in its own PR. At minimum, please call it out explicitly in the PR description so reviewers don't miss the scope change. (No action required if you prefer to keep it bundled.)

src/copilot-shell/packages/core/src/core/openaiContentGenerator/errorHandler.test.ts
```diff
@@ -98,7 +68,7 @@ describe('EnhancedErrorHandler', () => {
       }).toThrow(/Request timeout after 5s.*Troubleshooting tips:/s);
     });
 
-    it('should not log error when suppression is enabled', () => {
+    it('should use custom suppression function', () => {
```
samchu-zsl: 🔴 **Test coverage regression.**

This PR deletes four tests that asserted the logging side effect (`should log error message for non-timeout errors`, `should log streaming error message for streaming requests`, the `mockConsoleError.not.toHaveBeenCalled()` assertion here, and the analogous assertions in the `string errors` / `empty error message` cases) but does **not** replace them with assertions against `debugLogger.error`.

As a result, this renamed test (`should use custom suppression function`) now only verifies that `suppressLogging` is called — it no longer verifies that the suppression *actually prevents logging*. Please restore the behavioral coverage by either:

- Spying on the `debugLogger` instance (e.g. via `vi.mock('../../utils/debugLogger.js', () => ({ createDebugLogger: () => ({ error: vi.fn(), ... }) }))`) and asserting `error` was/was not called; or
- Injecting the logger for testability.

samchu-zsl (commented):
Thanks for the feature! I reviewed the changes against the repo's cosh-dev standards. Overall the direction is good, but a few things would benefit from another pass before merge — left inline comments. (Note: no need to rebase into a single commit since this PR will be squash-merged.)

**Summary**

1. The new ESC cancel flow has **no dedicated unit tests** — `useKeypress` is only mocked as a placeholder, so the whole `abortController → cancelSlashCommand → Promise.race` path is uncovered.
2. `compressCommand` cannot actually stop the in-flight LLM call because `tryCompressChat` does not accept an `AbortSignal`; this is **inconsistent** with `summaryCommand` in the same PR.
3. The switch from `console.error` to `debugLogger.error` in `errorHandler.ts` is reasonable for Ink-TTY hygiene (errors still throw up and are shown via the UI path, and `debugLogger` is backed by a process-wide session set in `Config`), but it is **out of scope** of an ESC-cancel PR, and several related tests were deleted without replacement assertions on `debugLogger.error`.
4. `Promise.race` only returns early from the processor — the underlying action Promise keeps running in the background. Only `compress`/`summary` currently guard their post-abort UI writes; any other built-in / file / MCP command will still pollute the UI after cancel.

Details inline.

src/copilot-shell/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
kongld-mq: Added comprehensive unit tests for ESC cancel flow in `src/ui/hooks/slashCommandProcessor.test.ts`:

**New test block: `ESC Cancel Flow`**

1. `should call useKeypress with isActive based on isProcessing state`
   - Validates useKeypress subscription only active when processing

2. `should pass abortSignal to command action`  
   - Validates abortSignal is passed to CommandContext

3. `should cancel slash command when ESC is pressed during processing`
   - Validates ESC keypress invokes cancelSlashCommand
   - Validates abortController.abort() is called
   - Validates INFO message "Command cancelled." is added
   - Validates isProcessing is set to false

4. `should return handled type when command is cancelled via ESC`
   - Validates Promise.race returns `{ type: 'handled' }` on abort
   - Validates no ERROR item is added

5. `should not cancel if no AbortController exists`
   - Validates ESC does nothing when no command is running

**Implementation:**
- Modified `mockUseKeypress` to capture keypress callback via `vi.hoisted()`
- Added `isProcessing` parameter to `setupProcessorHook` helper
- Tests simulate ESC by calling captured callback with `{ name: 'escape' }`

src/copilot-shell/packages/cli/src/ui/commands/compressCommand.ts
kongld-mq: Extended `tryCompressChat` to accept `AbortSignal` and thread it through to the underlying `generateContent` call, making compression cancellable the same way as `summaryCommand`.

**Changes:**

1. `packages/core/src/core/client.ts`:
   - Added `abortSignal?: AbortSignal` parameter to `tryCompressChat(prompt_id, force, abortSignal?)`

2. `packages/core/src/services/chatCompressionService.ts`:
   - Added `abortSignal?: AbortSignal` parameter to `compress()` method
   - Passed `abortSignal` to `generateContent` via `config: { abortSignal }`

3. `packages/cli/src/ui/commands/compressCommand.ts`:
   - Passed `abortSignal` to `geminiClient.tryCompressChat(promptId, true, abortSignal)`

4. `packages/cli/src/ui/commands/compressCommand.test.ts`:
   - Updated test assertion to expect 3 arguments

**Consistency with summaryCommand:**
- summaryCommand: `geminiClient.generateContent(..., abortSignal, ...)`
- compressCommand: `geminiClient.tryCompressChat(..., abortSignal)` → `compressionService.compress(..., abortSignal)` → `generateContent({ config: { abortSignal } })`

Both commands now properly abort the in-flight LLM call when ESC is pressed.

src/copilot-shell/packages/cli/src/ui/hooks/slashCommandProcessor.ts
kongld-mq: Added `setPendingItem` to the `cancelSlashCommand` useCallback dependency array for consistency with `react-hooks/exhaustive-deps` rule.

src/copilot-shell/packages/cli/src/ui/hooks/slashCommandProcessor.ts
kongld-mq: Added `abortController.signal.aborted` check in the finally block to prevent logging cancelled commands as `SlashCommandStatus.SUCCESS`.

**Root Cause:**
When ESC cancels a command via Promise.race, the finally block runs after the outer code returns `{ type: 'handled' }`. The original condition `!hasError` is true for cancellations (since cancel is not an error), resulting in `SUCCESS` being logged for cancelled commands.

Note on underlying Promise: While the underlying action Promise continues running in the background (JavaScript Promise.race behavior), this fix ensures:

Cancelled commands are NOT logged as SUCCESS in telemetry
The UI cleanup (setIsProcessing(false)) still runs correctly
Commands with explicit abort checks (compress/summary) properly abort their LLM calls
A more comprehensive "abort-aware proxy" solution would require larger architectural changes and is beyond the scope of this PR.

Test added: should not log SUCCESS when command is cancelled via ESC

src/copilot-shell/packages/core/src/core/openaiContentGenerator/errorHandler.test.ts
kongld-mq: Restored behavioral test coverage for `debugLogger.error` logging using vi.mock with hoisted spy.

**Fix:**
1. Added `vi.mock` to mock `createDebugLogger` with spy functions
2. Used `vi.hoisted()` to resolve mock initialization order
3. Restored all deleted test coverage:
   - `should log error message via debugLogger for non-timeout errors`
   - `should log streaming error message via debugLogger for streaming requests`
   - `should not log error when suppression is enabled` (renamed back, added `expect(mockDebugLoggerError).not.toHaveBeenCalled()`)
   - `should handle string errors and log via debugLogger`
   - `should handle null/undefined errors and log via debugLogger`

**Test added:** 5 new test assertions for `debugLogger.error` calls

samchu-zsl (approved):
LGTM, thx!

PR #298, opened 2026-04-23
fix(cosh): sync hook registry on extension enable/disable

### LINKED ISSUE #297: [cosh] bug: `/extensions disable` / `/extensions enable` does not sync hook registry
### Component

cosh

### Bug Description

After running `/extensions disable <name> --scope user` or
`/extensions enable <name> --scope user`, the `/hooks list` command
still shows the old hook state from the extension (hooks from a disabled
extension remain "Enabled"; hooks from a re-enabled extension remain absent).

The hook registry is only initialized once at startup. The
`disableExtension()` / `enableExtension()` calls correctly update
`extension.isActive` in memory and persist the change to
`extension-enablement.json`, but **`HookRegistry.initialize()` is never
re-invoked**, so the in-memory `entries` array stays stale for the entire
session.

### Steps to Reproduce

1. Install an extension with hooks (e.g. `example-hook`)
2. Run `/hooks list` → hooks appear as `✓ Enabled`
3. Run `/extensions disable example-hook --scope user`
4. Run `/hooks list` again → hooks still appear as `✓ Enabled` ← **BUG**

### Expected Behavior

After `/extensions disable`, the disabled extension's hooks should
immediately disappear from `/hooks list`. After `/extensions enable`,
they should immediately reappear.

### Environment

all

### Relevant Log Output

```shell

```

### Additional Context

_No response_

kongche-jbw:
## Description

Fixes a bug where `/extensions disable` and `/extensions enable` did not
update the hook registry in the current session. After disabling an
extension, its hooks remained in `/hooks list` as "Enabled" until
restart; after enabling, the hooks did not appear until restart.

**Root cause:** `disableAction()` / `enableAction()` correctly updated
`extension.isActive` and flushed `extension-enablement.json`, but never
called `hookSystem.initialize()` to rebuild the in-memory
`HookRegistry.entries`. The `installAction()` already handled this
correctly; the same pattern was simply missing for disable/enable.

**Fix:** Call `hookSystem?.initialize()` at the end of both actions,
wrapped in `try/catch` (best-effort — consistent with `installAction`).
`HookRegistry.initialize()` clears and reloads entries, so repeated
calls are safe.

## Related Issue

closes #297 

## Type of Change

## Scope

- [x] `cosh` (copilot-shell)

## Checklist

- [x] I have added tests that prove my fix is effective or that my
      feature works

## Testing

```bash
cd src/copilot-shell
npx vitest run \
  packages/cli/src/ui/commands/extensionsCommand.test.ts
```

New test cases (8 total, disable + enable symmetric):
- `should call hookSystem.initialize() after disabling/enabling`
- `should not crash when hookSystem is not available`
- `should call hookSystem.initialize() exactly once when using --all`
- `should still report success when hookSystem.initialize() throws`

Manual verification:
1. Install `example-hook` via `/extensions install example_hook`
2. Run `/hooks list` → `sandbox-guard` and `sandbox-failure-handler`
   appear as Enabled
3. Run `/extensions disable example-hook --scope user`
4. Run `/hooks list` → hooks no longer appear ✓
5. Run `/extensions enable example-hook --scope user`
6. Run `/hooks list` → hooks reappear ✓
7. Shift+Tab cycles approval mode — no `<Box> inside <Text>` crash ✓

## Additional Notes

The hook priority order (Project=1 > User=2 > System=3 > Extensions=4)
means this fix only affects the Extensions layer; Project/User/System
hooks are unaffected by extension enable/disable.


samchu-zsl (approved):
LGTM, thx!

PR #345, opened 2026-04-25
fix(cosh): unify PreToolUse ask dialog to info type with diff preview

### LINKED ISSUE #344: [cosh] feat: PreToolUse ask hook message invisible when tool has native confirmation dialog
### Component

cosh

### Problem Statement

When a `PreToolUse` hook returns `{ "decision": "ask", "systemMessage": "..." }` for a tool that already has its **own native confirmation dialog** (e.g. `run_shell_command` with `rm` triggers an `exec`-type dialog), the hook message was silently buried in the dialog **title** rather than displayed as prompt body text.

This caused two visible symptoms:

1. **First invocation**: the user saw only the native shell confirmation box — the hook message was appended to the title as `"Confirm Shell Command — ⚠️ Hook: <message>"`, but the title was truncated / visually hidden behind the command preview. Effectively invisible.
2. **Second invocation** (after the user chose "Allow always", which allowlisted the command and suppressed the native dialog): only then did the hook-synthesized `info` dialog appear, showing the message in the body. This made it look like the hook was silently swallowed on the first run.

### Motivation

`PreToolUse` hooks are a primary mechanism for user-defined safety policies (e.g. flag `rm` commands, detect dangerous patterns in file writes). If the hook message is invisible, the entire safety guarantee is broken. The user has no opportunity to review or cancel the operation.

### Proposed Solution

When `hookForceAsk = true`, the hook message **must always appear in the prompt body**, regardless of whether the tool has a native confirmation dialog. The user must be able to read the full message before making a decision.

### Alternatives Considered

_No response_

### Additional Context

_No response_

kongche-jbw:
## Description

When a `PreToolUse` hook returns `ask`, if the target tool already has
its own native confirmation dialog (e.g. `run_shell_command` with `rm`
produces an `exec`-type dialog), the previous implementation only
appended the hook message to the dialog **title**. This made the message
effectively invisible on the first invocation — users would click
through without seeing the hook warning.

This PR unifies the `hookForceAsk` path to always synthesize an `info`
dialog, ensuring the hook message appears in the **prompt body**
regardless of the tool's native confirmation type. For `edit`-type
confirmations (file write/modify), the diff preview is preserved and
rendered below the hook message.

**Changes:**
- `coreToolScheduler.ts`: when `hookForceAsk=true`, always produce an
  `info` dialog; forward native `onConfirm` (with `payload`) to preserve
  side-effects (allowlisting, `AUTO_EDIT` mode); spread `fileName` and
  `fileDiff` from `edit`-type native confirmation.
- `tools.ts`: add optional `payload` to `ToolInfoConfirmationDetails
  .onConfirm`; add optional `fileName`/`fileDiff` fields.
- `ToolConfirmationMessage.tsx`: in the `info` branch, render
  `DiffRenderer` below the prompt when `fileDiff` is present.
- `coreToolScheduler.test.ts`: rewrite `hookForceAsk` test to assert
  `type='info'`, hook message in prompt body, and `onConfirm`
  side-effects.

## Related Issue

closes #344 

## Type of Change

## Scope

- [x] `cosh` (copilot-shell)

## Checklist

- [x] I have added tests that prove my fix is effective or that my
      feature works

## Testing

Unit tests: `coreToolScheduler.test.ts` — 38 tests pass, including the
rewritten `hookForceAsk` test that verifies `type='info'`, prompt body
content, and `ProceedAlways` side-effect (`approvalMode=AUTO_EDIT`).

Manual:
1. Configure `rm-guard.py` as a `PreToolUse` hook (matcher `.*`).
2. Ask the agent to run `rm <file>` — hook message now appears in the
   `info` dialog body on the **first** invocation.
3. Ask the agent to write a file containing `rm ` — hook message and
   diff preview are both shown in the same dialog.

## Additional Notes

Known limitation: when the tool is in `AUTO_EDIT` mode, `shouldConfirm
Execute` returns `false` and no `fileDiff` is available, so the diff
preview will not appear even if the hook triggers `ask`. This edge case
is rare and left as a follow-up.

### DIFF
--- a/src/copilot-shell/packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
+++ b/src/copilot-shell/packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
@@ -343,6 +343,17 @@ export const ToolConfirmationMessage: React.FC<
             ))}
           </Box>
         )}
+        {infoProps.fileDiff && (
+          <Box marginTop={1}>
+            <DiffRenderer
+              diffContent={infoProps.fileDiff}
+              filename={infoProps.fileName ?? ''}
+              availableTerminalHeight={availableBodyContentHeight()}
+              contentWidth={contentWidth}
+              settings={settings}
+            />
+          </Box>
+        )}
       </Box>
     );
   } else {
--- a/src/copilot-shell/packages/core/src/core/coreToolScheduler.test.ts
+++ b/src/copilot-shell/packages/core/src/core/coreToolScheduler.test.ts
@@ -2326,11 +2326,12 @@ describe('CoreToolScheduler Sequential Execution', () => {
   });
 
   // ─────────────────────────────────────────────────────────────────────────
-  // Regression: Bug 2 — hookForceAsk must AUGMENT native confirmationDetails
-  //   (preserve type + original onConfirm side-effects), not replace them
-  //   with a bare info dialog.
+  // Regression: hookForceAsk must always synthesize an 'info' dialog that
+  //   shows the hook message in the prompt body (not just the title), while
+  //   still forwarding the original onConfirm so native side-effects
+  //   (e.g. setApprovalMode, allowlisting) are preserved.
   // ─────────────────────────────────────────────────────────────────────────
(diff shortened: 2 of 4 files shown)


samchu-zsl (approved):
LGTM, thx!

PR #387, opened 2026-04-27
feat(cosh): concatenate hook systemMessages with [name] prefix

kongche-jbw:
## Description

Previously, when multiple hooks returned `systemMessage` for the same event
(e.g. `PreToolUse`), `HookAggregator.mergeWithOrLogic` used last-write-wins
semantics — the final message was overwritten by whichever hook ran last,
even if that hook didn't trigger the `ask` / `block` decision.

This caused a confusing UX where hook A raised `ask` with message "A", but
a later allow-hook C silently replaced the message with "C", so the user
saw "C" in the confirmation dialog without any hint that A was the one
that actually triggered the prompt.

This PR changes the aggregation rule to **concatenate** every non-empty
`systemMessage` as `[hookName] message` on separate lines, so users can
see exactly which hook contributed which message. The change applies
uniformly to both parallel and sequential hook execution since both
paths funnel into the same `aggregateResults` entry point.

## Related Issue

no-issue: internal UX polish for hook message attribution

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] Multiple / Project-wide

## Checklist

## Testing

Added 4 new unit tests under
`mergeWithOrLogic - systemMessage concatenation` in hookAggregator.test.ts:

- Single hook fast path: passes through message untouched (no `[name]`
  prefix), preserving backward compatibility.
- Multiple hooks: messages concatenated as `[rm-guard] rm detected\n[audit] logged`;
  `decision=ask` wins over `allow`.
- Missing `name` field: falls back to the command basename, e.g.
  `/usr/local/bin/guard.py --strict` → `[guard.py] ...`.
- Skip rules: hooks with `systemMessage === undefined` or `''` do not
  contribute to the concatenation.

Run:

```
cd src/copilot-shell/packages/core
npx vitest run src/hooks/hookAggregator.test.ts
```

Result: **28/28 passed** (4 new + 24 existing).
Type check: `npx tsc --noEmit` clean.

## Additional Notes

- Single-hook behavior is unchanged (no `[name]` prefix) to keep existing
  UX identical for the most common case.
- Behavior is identical for parallel and sequential hook execution — the
  aggregator is the single merge point; `hookRunner` differences don't
  affect message attribution.
- Name resolution priority: `hookConfig.name` → basename of the first
  token in `command` → fallback `'hook'`.

### DIFF
--- a/src/copilot-shell/packages/core/src/hooks/hookAggregator.test.ts
+++ b/src/copilot-shell/packages/core/src/hooks/hookAggregator.test.ts
@@ -203,6 +203,157 @@ describe('HookAggregator', () => {
     });
   });
 
+  describe('mergeWithOrLogic - systemMessage concatenation', () => {
+    it('should pass through single systemMessage untouched (fast path)', () => {
+      const results: HookExecutionResult[] = [
+        {
+          hookConfig: {
+            type: HookType.Command,
+            command: '/path/to/guard.py',
+            name: 'rm-guard',
+          },
+          eventName: HookEventName.PreToolUse,
+          success: true,
+          output: { decision: 'ask', systemMessage: 'danger detected' },
+          duration: 100,
+        },
+      ];
+
+      const result = aggregator.aggregateResults(
+        results,
+        HookEventName.PreToolUse,
+      );
+      // Single hook keeps its original systemMessage, no `[name]` prefix.
+      expect(result.finalOutput?.systemMessage).toBe('danger detected');
+    });
+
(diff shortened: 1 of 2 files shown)


samchu-zsl (approved):
LGTM, thx!

PR #418, opened 2026-04-29
fix(cosh): block initial chat on skill/subagent first-load discovery

samchu-zsl:
## Description

On small instances (e.g. 2c2g), the very first request sent to the model contained the placeholder description `Loading available skills...` instead of the real skill list, so the model never invoked any skill. Root cause: `SkillTool` (and `TaskTool`) kicked off their first `refreshSkills()` / `refreshSubagents()` in fire-and-forget fashion in the constructor. When disk IO was slow or the remote Skill-OS server was unreachable, the tool function declarations were emitted before the list was populated.

This PR makes the cold-start path deterministic by:

1. Exposing a public `initPromise` on both tools and awaiting it from `Config.createToolRegistry` and `GeminiClient.startChat` (covers every chat-creation path).
2. Adding a monotonic `refreshSeq` race guard plus a `hasLoadedOnce` flag so transient errors no longer overwrite a previously-good list with `[]`.
3. Switching `SkillTool` to local-first cold start (remote skills are fetched in the background after `initPromise` resolves), so a slow Skill-OS server cannot block startup.
4. Splitting the single 30s `RemoteSkillRegistry` timeout into a 10s metadata timeout and a 60s download timeout.
5. Adding a `notify` flag to `SkillManager.refreshCache` to suppress re-entrant change-listener storms when `listSkills` lazily refreshes its own cache.
6. Defensive `?? []` guards so a misbehaving manager (or test mock cleared by `vi.clearAllMocks`) cannot leave `availableSkills` / `availableSubagents` as `undefined`.

## Related Issue

no-issue: discovered during 2c2g instance smoke test; reproduced via network capture showing the placeholder description in the very first function declaration sent to the model.

## Type of Change

- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)

## Checklist

## Testing

```
cd src/copilot-shell
npm run format         # prettier clean
npm run build          # 3 packages built
npm run lint           # eslint clean
npm run typecheck      # tsc clean
npm run test           # 178 test files / 3863 tests / 0 failed
```

Three new regression tests under `packages/core/src/tools/skill.test.ts > cold-start race conditions`:
- `should expose initPromise that resolves after first load completes` — guarantees the description is no longer the placeholder once `initPromise` resolves.
- `should call listSkills with includeRemote=false on cold start` — locks in the local-first contract.
- `should preserve previously loaded skills when a refresh fails` — protects against transient IO errors clobbering the list.

## Additional Notes

The original sequential `Promise.all` parallelization in `SkillManager.refreshCache` was reverted because an existing test (`should load skills from multiple custom directories`) relies on `readdir` call ordering. The race-condition fix does not depend on it; per-level scan latency is small compared to the network and disk waits we already removed from the cold-start path.

### DIFF
--- a/src/copilot-shell/packages/core/src/config/config.ts
+++ b/src/copilot-shell/packages/core/src/config/config.ts
@@ -1833,6 +1833,26 @@ export class Config {
       registerCoreTool(LspTool, this);
     }
 
+    // Wait for SkillTool / TaskTool first-load to settle so that the
+    // initial chat request sees populated skill/subagent descriptions
+    // instead of the "Loading available skills..." placeholder.
+    const skillToolInstance = registry.getTool(SkillTool.Name);
+    if (skillToolInstance instanceof SkillTool) {
+      try {
+        await skillToolInstance.initPromise;
+      } catch (error) {
+        console.warn('[Config] SkillTool initial discovery failed:', error);
+      }
+    }
+    const taskToolInstance = registry.getTool(TaskTool.Name);
+    if (taskToolInstance instanceof TaskTool) {
+      try {
+        await taskToolInstance.initPromise;
+      } catch (error) {
+        console.warn('[Config] TaskTool initial discovery failed:', error);
+      }
+    }
+
     await registry.discoverAllTools();
     console.debug('ToolRegistry created', registry.getAllToolNames());
     return registry;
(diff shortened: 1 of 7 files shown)


samchu-zsl:
> [!CAUTION]

KaiLongZhou (approved):
LGTM

PR #435, opened 2026-05-06
fix(cosh): emit allow decision reason to UI when systemMessage is absent

### LINKED ISSUE #421: [cosh] bug: [cosh] PreToolUse hook allow decision reason field is silently ignored and not displayed in UI
### Component

cosh

### Bug Description

When a `PreToolUse` hook returns an **allow** decision with a `reason` field (i.e., `{decision: "allow", reason: "some warning message"}`), the `reason` string is silently discarded and **never displayed** to the user in the terminal UI.

This is misleading for hook authors who naturally expect `reason` to be visible when returned alongside an allow decision. The current behavior makes it impossible to surface informational warnings to the user via the `reason` field on allow decisions.

**Root cause**: In [`coreToolScheduler.ts` (lines 894–903)](https://github.com/alibaba/anolisa/blob/main/src/copilot-shell/packages/core/src/core/coreToolScheduler.ts#L894-L903), only the `systemMessage` field is emitted to the UI for non-blocking decisions. The `reason` field is only consumed when the decision is `deny`/`block` (via `getEffectiveReason()`). There is no code path that renders `reason` for allow decisions.

**Affected code pattern** (e.g., in `skill_ledger_hook.py`):
```python
def _allow_with_reason(reason: str) -> str:
    """Return an allow decision with a warning reason for display."""
    return json.dumps({"decision": "allow", "reason": reason}, ensure_ascii=False)
```
The returned `reason` is parsed and aggregated by the hook system (in `hookAggregator.ts`), but ultimately never shown to the user.

### Steps to Reproduce

1. Register a `PreToolUse` hook that outputs:
   ```json
   {"decision": "allow", "reason": "Warning: skill signature could not be verified"}
   ```
2. Trigger any tool invocation in cosh.
3. Observe that the tool executes successfully, but the reason string is **not** displayed anywhere in the terminal.

### Expected Behavior

When a `PreToolUse` hook returns `{decision: "allow", reason: "..."}`, the `reason` should be displayed to the user in the terminal, similar to how `systemMessage` is currently rendered. At minimum, one of the following should happen:

1. **Option A**: Treat `reason` as a fallback for `systemMessage` — if `systemMessage` is absent but `reason` is present on an allow decision, display `reason` in the UI.
2. **Option B**: Document clearly in `reference.md` that `reason` is **only** for deny decisions and that hook authors **must** use `systemMessage` for informational messages on allow decisions.

### Environment

- OS: macOS 14.1 (development), Anolis OS 8 (target)
- Component version: copilot-shell (main branch, latest)
- Node.js version: 20.x

### Relevant Log Output

```shell
# No error output — the reason is silently ignored.
# Hook output is parsed correctly (visible in debug logs) but never emitted to terminal.
```

### Additional Context

- The `reference.md` documentation states that `reason` is "The feedback/error message provided when a decision is `deny`", but does not explicitly warn that it is **ignored** for allow decisions.
- The `hookAggregator.ts` correctly concatenates `reason` fields from multiple allow hooks, suggesting the field was intended to be meaningful for allow decisions as well.
- Integration tests in `hooks.test.ts` test the pattern `{decision: 'allow', reason: '...'}` without verifying that the reason is actually displayed to the user.

samchu-zsl:
### Problem

`PreToolUse` hook messages reached the terminal through a single merged string. This caused four concrete regressions:

- `{ decision: "allow", reason }` without `systemMessage` → message silently dropped.
- Auto-approved tools (`SkillTool`, YOLO mode) → notification stored but never rendered.
- Multiple hooks on one call → messages concatenated into one blob, losing per-hook attribution.
- `decision: "ask"` → the dialog reused the first hook's `systemMessage` as its prompt, letting hooks override system-level UX copy.

All four share the same root cause: the display layer tries to reconstruct per-hook context from a post-merge blob, after the information it needs has already been flattened away.

### Changes

The fix is a structured, per-hook notification channel that replaces the merged string end-to-end. The unchanged parts of the pipeline — hook matching, execution, decision evaluation, tool invocation — are untouched.

**1. Aggregator — produce structured notifications (`hookAggregator.ts`, `hooks/types.ts`)**
Introduce `HookNotification { hookName, message, decision? }` and have `aggregateResults()` return a `notifications: HookNotification[]` alongside the merged decision. Each entry's `message` resolves as `systemMessage ?? reason`, so `allow + reason` is preserved without any special-casing downstream. Hooks with neither field are omitted.

**2. Scheduler — emit early, expose to dialog (`coreToolScheduler.ts`, `tools/tools.ts`)**
Emit the notifications list via `outputUpdateHandler` as a structured `HookNotificationDisplay` chunk *before* the merged decision is applied, so delivery no longer depends on the final state transition. `ToolConfirmationInfo` gains `hookNotifications` and `mergedDecision` fields so the same data is available when a dialog is shown. The dead `liveOutput` field on `ValidatingToolCall` is removed.

**3. CLI propagation — sticky across every state (`useReactToolScheduler.ts`, `ui/types.ts`)**
A type guard recognizes the structured chunk and appends it to a per-tool-call list on both a synchronous ref (for history snapshots) and React state (for live display). The list is carried as a sticky field across `validating` / `scheduled` / `waiting` / `executing` / `completed` / `cancelled`, which is what finally closes the gap for tools that never enter `awaiting_approval`.

**4. Terminal rendering — per-hook boxes with merged-decision dimming (`ToolMessage.tsx`)**
Each entry renders as a standalone box between the tool header and the tool output / confirmation dialog: left-side colored border, icon + bold hook name, indented message. Colors go through `decisionToStyle(item.decision, item.mergedDecision)`: when the merged outcome is `block`/`deny`, non-blocking hook boxes are dimmed so they don't visually compete with the denied outcome, while blocking hooks keep full color and all messages stay readable.

**5. Ask-dialog — fixed, localized prompt (`ToolConfirmationMessage.tsx`, `i18n/locales/*.js`)**
The dialog always shows `A hook requires your confirmation to proceed.`, wrapped in `t()` for i18n (zh: `有 Hook 需要您的确认方可继续。`). The hook's own `systemMessage` is no longer pulled into the dialog — it surfaces above the dialog as a regular per-hook box, keeping the same display contract as the allow / deny paths.

**6. Docs (`hooks/docs/reference.md`, `hooks/docs/writing-hooks.md`)**
Align `systemMessage` / `reason` semantics with the new rendering model, and document the fixed ask-dialog prompt plus the merged-decision dimming rule.

## Related Issue

closes #421

## Type of Change

## Scope

- [x] `cosh` (copilot-shell)

## Checklist

## Testing

```bash
cd src/copilot-shell
npm run format && npm run build && npm run lint && npm run typecheck
npm test
# All tests pass, including new regression coverage for:
# - hookAggregator: per-hook notifications carry hookName + decision; message falls back to reason when systemMessage is absent
# - coreToolScheduler: notifications emit before the merged decision applies (allow / ask / block paths)
# - coreToolScheduler: ask-dialog uses the fixed prompt regardless of the hook's systemMessage
# - useToolScheduler: hookNotifications propagate as a sticky field across validating / executing / completed states
# - useToolScheduler: mergedDecision=block dims non-blocking per-hook boxes while keeping blocking ones fully colored
```

## Additional Notes

Hook protocol is unchanged — hooks that only set `systemMessage` behave as before. The difference is that every hook now gets its own labeled, independently-colored box instead of a merged string, and `allow + reason` is no longer silently dropped.

### DIFF
--- a/src/copilot-shell/hooks/docs/reference.md
+++ b/src/copilot-shell/hooks/docs/reference.md
@@ -38,15 +38,15 @@ All hooks receive these common fields via `stdin`:
 
 Most hooks support these fields in their `stdout` JSON:
 
-| Field                | Type      | Description                                                                                                                                              |
-| :------------------- | :-------- | :------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `systemMessage`      | `string`  | Displayed immediately to the user in the terminal.                                                                                                       |
-| `suppressOutput`     | `boolean` | If `true`, hides internal hook metadata from logs/telemetry.                                                                                             |
-| `continue`           | `boolean` | If `false`, stops the entire agent loop immediately.                                                                                                     |
-| `stopReason`         | `string`  | Displayed to the user when `continue` is `false`.                                                                                                        |
-| `decision`           | `string`  | `"allow"`, `"deny"` (alias `"block"`), `"ask"`, or `"approve"`.                                                                                          |
-| `reason`             | `string`  | The feedback/error message used for `"deny"`/`"block"` decisions and stop-like flows. For user-visible allow/approve/ask messaging, use `systemMessage`. |
-| `hookSpecificOutput` | `object`  | Event-specific output fields (see individual event sections).                                                                                            |
+| Field                | Type      | Description                                                                                                                                                                                                                         |
+| :------------------- | :-------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `systemMessage`      | `string`  | Shown to the user as a per-hook notification box labeled with the hook name, independent of the tool confirmation dialog.                                                                                                           |
+| `suppressOutput`     | `boolean` | If `true`, hides internal hook metadata from logs/telemetry.                                                                                                                                                                        |
+| `continue`           | `boolean` | If `false`, stops the entire agent loop immediately.                                                                                                                                                                                |
+| `stopReason`         | `string`  | Displayed to the user when `continue` is `false`.                                                                                                                                                                                   |
+| `decision`           | `string`  | `"allow"`, `"deny"` (alias `"block"`), `"ask"`, or `"approve"`.                                                                                                                                                                     |
+| `reason`             | `string`  | The feedback/error message used for `"deny"`/`"block"` decisions and stop-like flows. For user-visible allow/approve/ask messaging, prefer `systemMessage`; if omitted, `reason` is used as the fallback text for the notification. |
+| `hookSpecificOutput` | `object`  | Event-specific output fields (see individual event sections).                                                                                                                                                                       |
 
 ---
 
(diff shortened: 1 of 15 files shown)


samchu-zsl:
> [!CAUTION]

samchu-zsl:
Manual verification successful. Tested three PreToolUse hooks (allow, ask, deny) and observed their behavior when calling tools:

1. **Deny**: Blocks tool execution and displays all hook information.image
3. **Ask**: Shows a confirmation prompt with all hook information.image
4. **Allow**: Proceeds while keeping hook information displayed.image

kongche-jbw (approved):
lgtm，thx

PR #465, opened 2026-05-09
feat(cosh): add auto memory background extraction system

samchu-zsl:
## Description

Add a background Auto Memory extraction system that scans past user sessions and extracts reusable skills/memory patches via a subagent. Includes fixes for subagent hanging on tool confirmation, session telemetry pollution, and patch validation failures. All extraction thresholds are now configurable via `settings.json`.

Additionally, fix the private memory loading gap: approved private memory (`MEMORY.md`) is now auto-loaded into runtime agent context via `extensionContextFilePaths`, closing the loop where extracted memories were written but never consumed by future sessions.

## Related Issue

no-issue: new feature implementation

## Type of Change

- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)

## Checklist

## Testing

- TypeScript type check passes for both `core` and `cli` packages
- Lint and prettier pass via pre-commit hooks
- Manual testing: enable `autoMemory.enabled = true`, verify extraction runs, patches are normalized and retained

### Key implementation decisions:

1. **Subagent tool confirmation**: Background subagents without `eventEmitter` now auto-approve `awaiting_approval` tool calls to prevent indefinite hanging.

2. **Telemetry isolation**: `ChatRecordingService` filters out auto-memory subagent events by prompt_id prefix, preventing session pollution that would cause future runs to re-scan their own output.

3. **Patch validation with retry**: A `postToolUse` hook validates `.patch` files on write. If `Diff.parsePatch()` fails, an error message is returned to the model which can then regenerate. A `normalizePatchContent()` utility fixes common LLM formatting issues (trailing empty lines, incorrect hunk header counts).

4. **Configurable thresholds**: 7 new settings under `autoMemory.*`:
   - `cooldownSeconds` (1800) — min seconds between runs
   - `sessionMinMessages` (10) — min user messages for eligibility
   - `sessionMinIdleSeconds` (10800) — min idle time for eligibility
   - `sessionMaxPerRun` (10) — max new sessions per run
   - `sessionIndexLimit` (50) — max sessions in agent index
   - `agentTimeoutSeconds` (1800) — agent max runtime
   - `agentMaxTurns` (30) — agent max conversation turns

5. **Private memory context injection**: `getExtensionContextFilePaths()` now includes the project-level `MEMORY.md` (from `getProjectMemoryTempDir()`) when it exists, so approved private memory is automatically loaded into future sessions via `loadServerHierarchicalMemory`. The `@import` mechanism in `processImports` also expands sibling references.

### Known TODOs (marked as `TODO(auto-memory)` in code):

- Auto Memory-extracted skills land in `getProjectSkillsMemoryDir()` which is NOT scanned by `SkillManager` — needs a scan path addition or directory change.
- Skills `*.patch` files have no user-facing approve/apply mechanism (`/memory inbox approve` only handles memory patches).

### DIFF
--- a/src/copilot-shell/packages/cli/src/config/config.ts
+++ b/src/copilot-shell/packages/cli/src/config/config.ts
@@ -1024,6 +1024,21 @@ export async function loadCliConfig(
       ? { baseUrl: settings.skillOS.baseUrl }
       : undefined,
     customSkillPaths: settings.skills?.customPaths ?? [],
+    autoMemory: settings.autoMemory?.enabled
+      ? {
+          enabled: true,
+          model: settings.autoMemory.model,
+          baseUrl: settings.autoMemory.baseUrl,
+          apiKey: settings.autoMemory.apiKey,
+          cooldownSeconds: settings.autoMemory.cooldownSeconds,
+          sessionMinMessages: settings.autoMemory.sessionMinMessages,
+          sessionMinIdleSeconds: settings.autoMemory.sessionMinIdleSeconds,
+          sessionMaxPerRun: settings.autoMemory.sessionMaxPerRun,
+          sessionIndexLimit: settings.autoMemory.sessionIndexLimit,
+          agentTimeoutSeconds: settings.autoMemory.agentTimeoutSeconds,
+          agentMaxTurns: settings.autoMemory.agentMaxTurns,
+        }
+      : undefined,
     summarizeToolOutput: settings.model?.summarizeToolOutput,
     ideMode,
     chatCompression: settings.model?.chatCompression,
--- a/src/copilot-shell/packages/cli/src/config/settingsSchema.ts
+++ b/src/copilot-shell/packages/cli/src/config/settingsSchema.ts
@@ -1237,6 +1237,127 @@ const SETTINGS_SCHEMA = {
       },
     },
   },
+  autoMemory: {
+    type: 'object',
+    label: 'Auto Memory',
+    category: 'Advanced',
+    requiresRestart: false,
+    default: {},
+    description:
+      'Configuration for automatic memory extraction from past sessions.',
+    showInDialog: false,
+    properties: {
+      enabled: {
+        type: 'boolean',
+        label: 'Enable Auto Memory',
+        category: 'Advanced',
+        requiresRestart: false,
+        default: false,
+        description:
+          'Enable background extraction of durable memory and reusable skills from past sessions.',
+        showInDialog: true,
+      },
+      model: {
+        type: 'string',
+        label: 'Auto Memory Model',
+        category: 'Advanced',
+        requiresRestart: false,
+        default: undefined as string | undefined,
+        description:
+          'Model to use for memory extraction. If not set, uses the currently configured model.',
+        showInDialog: false,
+      },
+      baseUrl: {
+        type: 'string',
+        label: 'Auto Memory Base URL',
+        category: 'Advanced',
+        requiresRestart: false,
+        default: undefined as string | undefined,
+        description:
+          'API base URL override for memory extraction. If not set, uses the current endpoint.',
+        showInDialog: false,
+      },
+      apiKey: {
+        type: 'string',
+        label: 'Auto Memory API Key',
+        category: 'Advanced',
+        requiresRestart: false,
+        default: undefined as string | undefined,
+        description:
+          'API key override for memory extraction. If not set, uses the current credentials.',
+        showInDialog: false,
+      },
+      cooldownSeconds: {
+        type: 'number',
+        label: 'Extraction Cooldown',
+        category: 'Advanced',
+        requiresRestart: false,
+        default: 1800,
+        description: 'Minimum seconds between two extraction runs.',
+        showInDialog: false,
+      },
+      sessionMinMessages: {
+        type: 'number',
+        label: 'Session Min Messages',
+        category: 'Advanced',
+        requiresRestart: false,
+        default: 10,
+        description:
+          'Minimum number of user messages a session must have to be eligible for extraction.',
+        showInDialog: false,
+      },
+      sessionMinIdleSeconds: {
+        type: 'number',
+        label: 'Session Min Idle Time',
+        category: 'Advanced',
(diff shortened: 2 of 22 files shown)


samchu-zsl:
> [!CAUTION]

kongche-jbw:
Thanks for the implementation. I reviewed PR #465 with focus on correctness, regression risk, and operational safety. I found 4 issues that should be addressed before merge.

1. High: Inbox approval applies patches using raw header paths, which can diverge from validated paths  
Location: memoryCommand.ts 449, memoryCommand.ts 475, memoryPatchUtils.ts 153  
Why this matters: Validation normalizes patch headers (including stripping a/ and b/ prefixes), but approval later reads/writes using patch.newFileName or patch.oldFileName directly. This creates a mismatch where a patch can pass validation but be applied to a different path during approval.  
Risk: Wrong target writes or failed applies in real usage.  
Recommendation: Reuse the same canonical target-resolution function in approval that is used during validation.

2. High: All candidate sessions are marked as processed even if the subagent did not actually read/process them  
Location: memoryService.ts,648 memoryService.ts 651, memoryService.ts  664
Why this matters: The current logic marks every candidate as processed by design, even when extraction exits early, times out, or only partially processes data.  
Risk: Permanent loss of extraction coverage; some sessions may never be considered again.  
Recommendation: Track actually-read session files or only mark sessions processed when there is explicit evidence they were handled.

3. Medium: Extracted skills are written to a directory that is not scanned by SkillManager  
Location: memoryService.ts 429, skill-manager.ts  553
Why this matters: Skills generated by auto-memory may never be discovered at runtime.  
Risk: Feature appears to run successfully but has no effective impact for users.  
Recommendation: Either add this directory to SkillManager scanning/watching or write extracted skills into an already-scanned level.

4. Medium: Inbox patch file is deleted even when some hunks fail to apply  
Location: memoryCommand.ts 467, memoryCommand.ts  479
Why this matters: Partial failure still leads to unconditional deletion of the source patch file.  
Risk: Data loss and no recovery path for failed hunks.  
Recommendation: Delete only when all hunks apply successfully, or keep residual failed hunks in the inbox for retry.

Open questions  
Do we want extraction state semantics to represent attempted sessions or successfully processed sessions? Current implementation mixes both.

## Overall assessment  
The feature direction is strong, but there are two correctness-critical issues and two important reliability gaps. I recommend fixing these before merge.

samchu-zsl:
@kongche-jbw Thanks for the review. Fixes pushed as an amend to the same commit (`08f1726` → `7d16ca3`); all 4 issues addressed except M3 which is intentionally deferred — detail below.

### 1. H1 — raw header paths during approval ✅
Approval now reuses the paths already resolved by `validateInboxMemoryPatchFile`. `ValidateInboxMemoryPatchFileResult` was extended to return `{ patches, parsed }` (the same `ValidatedSkillPatchHeader[]` produced during validation, with `a/`/`b/` prefixes stripped and absoluteness enforced), and `/memory inbox approve` iterates those instead of `patch.newFileName` / `patch.oldFileName`. Validation and application now traverse the exact same path.

### 2. H2 — candidates always marked processed ✅
`createExtractionHooks` now accepts an optional `{ chatsDir, onSessionRead }`. On `read_file` / `read_many_files` success it maps the absolute path back to a `sessionId` via a new exported `extractSessionIdFromChatFilePath` helper (requires the file to be a direct child of `chatsDir` and match the `{UUID}.jsonl` pattern). `memoryService` accumulates these into `actuallyReadSessions` and builds `processedSessions` as:

```
candidateSessions.filter(s =>
  actuallyReadSessions.has(s.sessionId) ||
  getSessionAttemptCount(state, s) + 1 >= MAX_ATTEMPTS_BEFORE_SKIP /* = 3 */
)
```

So a candidate is marked processed only if the subagent actually read it, or it has failed 3 attempts (safety valve against permanently-unreadable sessions). Partial / early-exit runs leave untouched candidates pending for the next run. This also answers the open question: semantics are now "successfully processed" plus a bounded-attempt skip.

### 3. M3 — skills written to an un-scanned directory ⏭️ deferred
Acknowledged; integrating the auto-memory skill output into `SkillManager` scanning is a larger follow-up and will be handled in a separate PR so this one stays focused on extraction + inbox flow.

### 4. M4 — inbox patch deleted on partial failure ✅
Approval is now all-or-nothing per file. Every hunk is applied in-memory first; `fs.unlink` runs only when every hunk succeeds and every write succeeds. On any failure the patch file is retained with a user-visible reason so retry is possible. I also tightened `validateInboxMemoryPatchFile` to require exactly one `ParsedDiff` per inbox file, which makes per-file semantics unambiguous.

### Tests
Added coverage for all of the above (39 new cases across 4 test files):
- `sessionAdapter.test.ts` — `extractSessionIdFromChatFilePath`, `getSessionAttemptCount`
- `memoryPatchUtils.test.ts` — validation success / reject paths incl. single-diff enforcement and allowed-root checks
- `skillExtractionAgent.test.ts` — hook session-read tracking (read_file / read_many_files / failure / no-options / non-read tools)
- `memoryCommand.test.ts` — `/memory inbox approve` apply-and-unlink, validation-reject retain, applyPatch-false retain, write-failure retain, kind filtering, empty state

`tsc --noEmit` (core + cli), `vitest run`, and `eslint` all pass.

KaiLongZhou (approved):
LGTM

kongche-jbw:
Overall this looks much better, and the major correctness issues appear to be addressed.  
two non-blocking follow-up improvements that can be considered in a later iteration:

1. Dismiss behavior for invalid inbox patches  
It may be helpful to align dismiss behavior with the new approve flow. Since invalid patch files are now intentionally retained for retry, allowing dismiss to also clear invalid entries would improve cleanup ergonomics and avoid stale inbox artifacts.

2. Auto-generated skill discoverability  
As a follow-up enhancement, we could make sure skills generated by auto-memory are discoverable by the runtime skill loading/scanning path. This would complete the end-to-end loop from extraction to actual reuse.

These are improvement suggestions rather than merge blockers.

kongche-jbw (approved):

PR #540, opened 2026-05-15
fix(sec-core): detect scan-pii module mode via subprocess

1570005763:
## Summary
- make scan-pii e2e module-mode detection use the same subprocess import path that the test later executes
- skip module mode in RPM environments where only the installed agent-sec-cli binary is importable

## Tests
- uv run --project agent-sec-cli ruff check --config agent-sec-cli/pyproject.toml tests/e2e/cli/test_scan_pii_e2e.py
- uv run --project agent-sec-cli pytest tests/e2e/cli/test_scan_pii_e2e.py -v
- git diff --check

no-issue: CI fix split out from PIIChecker hook PR #539


edonyzpc (approved):
LGTM

PR #542, opened 2026-05-15
feat: improve aliyun-ecs skill score (72% → 88%)

yogesh-tessl:
Hey @Ziqi002 👋

the OS skills library is a really thoughtful addition to ANOLISA. 24 skills organised across 7 categories from security to system admin, and the bilingual documentation shows you're thinking about accessibility for a wide audience.

ran your skills through `tessl skill review` at work and found some targeted improvements for `aliyun-ecs`. Here's the before/after:

| Skill | Before | After | Change |
|-------|--------|-------|--------|
| aliyun-ecs | 72% | 88% | +16% |

improves the `aliyun-ecs` skill score by adding an explicit trigger clause to the frontmatter description, removing a redundant API quick-reference section that duplicated the command tables, and adding verification guidance for destructive operations (instance deletion, disk deletion).

picked `aliyun-ecs` because it had the most improvement headroom among the non-failing skills - 72% baseline with clear, additive fixes that don't restructure anything.

<details>
<summary>Changes made</summary>

- **Added "Use when..." trigger clause** to the frontmatter description - adds terms like 阿里云, ECS, 云服务器, 弹性计算, aliyun CLI so Claude knows when to select this skill (description score: 83% → 100%)
- **Removed duplicate "常用API速查" section** (42 lines) - every API listed there was already in the command tables above, pure redundancy
- **Added verification notes for destructive operations** - `DeleteInstance` and `DeleteDisk` sections now remind Claude to confirm targets before executing and verify status after
- **Quoted the description string** in frontmatter to follow standard YAML format

</details>

---

quick honest disclosure. I work at https://github.com/tesslio where we build tooling around skills like these. Not a pitch, just saw room for improvement and wanted to contribute.

if you want to self-improve your skills, or define your own scenarios to pressure test, just ask your agent (Claude Code, Codex, etc.) to evaluate and optimize your skill with Tessl. Ping me [@yogesh-tessl](https://github.com/yogesh-tessl), if you hit any snags.

### DIFF
--- a/src/os-skills/aliyun/aliyun-ecs/SKILL.md
+++ b/src/os-skills/aliyun/aliyun-ecs/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: aliyun-ecs
 version: 1.0.0
-description: 通过aliyun命令行管理阿里云弹性计算服务(ECS)。用于查询或创建实例、启动/停止/重启实例、管理磁盘/快照/镜像/安全组/密钥对/弹性网卡、查询状态以及故障排查等工作流。
+description: "通过aliyun命令行管理阿里云弹性计算服务(ECS)。用于查询或创建实例、启动/停止/重启实例、管理磁盘/快照/镜像/安全组/密钥对/弹性网卡、查询状态以及故障排查等工作流。当用户提到阿里云、ECS、云服务器、弹性计算、aliyun CLI，或需要管理云实例、磁盘、快照、安全组等资源时使用。"
 layer: application
 lifecycle: operations
 ---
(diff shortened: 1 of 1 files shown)


yogesh-tessl:
hey @Ziqi002, just a gentle nudge in case this got buried, let me know if you need any changes!
No rush at all if this is in a review pipeline.

jfeng18:
Nice work on the trigger-clause pattern — adding "Use when..." keywords is a great technique for improving skill discovery.

One thought on the DeleteInstance template: the updated command includes `--Force true` by default, which bypasses instance protection and performs a hard shutdown (cache data lost per the API docs). Since the LLM tends to copy command templates directly, users might force-delete a running instance unintentionally.

Would it be safer to keep the default template as `--InstanceId <实例ID>` (without --Force), and mention `--Force true` as an optional flag in the verification note for cases where the instance is stuck?

yogesh-tessl:
Great catch @jfeng18, you're absolutely right. Since the LLM copies command templates verbatim, defaulting to `--Force true` was risky exactly as you describe (it skips instance protection and forces a shutdown, which can lose cache data).

Just pushed a fix: the `DeleteInstance` template is back to the safe default `--InstanceId <实例ID>`, and `--Force true` is now documented as an optional flag in the verification note, only for cases where an instance is stuck in the running state.

Thanks for the careful review 🙏

Ziqi002:
I think there are few modifications in this PR, and valuable commonly used API shortcuts have been removed.
And the main improvement comes from deleting valuable commonly used API shortcuts so reducing tokens.
Not allowed to merge.

PR #545, opened 2026-05-15
feat(cosh): surface UserPromptSubmit and PostToolUse hook reason in UI

### LINKED ISSUE #535: [cosh] enhancement: surface reason for UserPromptSubmit and PostToolUse hooks
### Component

cosh

### Feature Description

Follow-up to #421. That issue covered the same visibility problem for `PreToolUse`: hook output can include a `reason`, but the user-facing terminal UI may not display it.

It would be useful to provide the same ability for `UserPromptSubmit` and `PostToolUse` hooks, so hook authors can surface meaningful explanations, warnings, or post-check results through the `reason` field instead of losing that message or having to rely on a different field.

Current behavior appears inconsistent across hook events: `PreToolUse` has a requested path for displaying `reason`, while `UserPromptSubmit` and `PostToolUse` do not have an obvious equivalent user-visible path.

### Use Cases

- `UserPromptSubmit`: show why a submitted prompt was accepted with warnings, rejected, or otherwise handled by a policy hook.
- `PostToolUse`: show post-execution warnings, review summaries, compliance notices, or recovery guidance after a tool has run.

### Steps to Reproduce

1. Register a `UserPromptSubmit` hook that returns a structured response containing `reason`, for example:
   ```json
   {"decision": "allow", "reason": "Prompt accepted, but policy check found a warning"}
   ```
2. Or register a `PostToolUse` hook that returns a structured response containing `reason`, for example:
   ```json
   {"decision": "block", "reason": "Tool output requires follow-up review before continuing"}
   ```
3. Trigger the matching hook in cosh.
4. Observe that the hook result is processed, but the `reason` text is not consistently surfaced in the terminal UI.

### Expected Behavior

When a `UserPromptSubmit` or `PostToolUse` hook returns a `reason`, cosh should display that reason to the user in the terminal UI where it is semantically meaningful, similar to the behavior requested for `PreToolUse` in #421.

At minimum, one of the following would be helpful:

1. Treat `reason` as a fallback for `systemMessage` when a hook result is non-blocking and no explicit `systemMessage` is provided.
2. Ensure blocking or feedback-producing results for `UserPromptSubmit` and `PostToolUse` render `reason` in the same user-visible style as other hook feedback.
3. If `reason` is intentionally unsupported for these hook events, document that clearly in the hook reference and point authors to the supported field.

### Environment

- Component version: copilot-shell main branch, latest
- Node.js version: 20.x

### Relevant Log Output

```shell
# No error output. The hook response can be parsed, but the reason is not visibly shown to the user.
```

### Additional Context

- This is intended as a follow-up to #421 for the other hook events where `reason` is also useful.
- Displaying `reason` would make hook behavior easier to understand and debug, especially for policy, security, audit, and post-tool review hooks.

kongche-jbw:
## Description

Follow-up to #421. `UserPromptSubmit` and `PostToolUse` hooks could return
a `reason` (or `systemMessage`) on non-blocking decisions, but the
terminal UI never showed it — leaving informational warnings invisible.

This PR mirrors the #421 PreToolUse path:

- `firePostToolUseEvent()` now carries aggregator `notifications[]` on
  the returned output, and `coreToolScheduler` emits each one via
  `outputUpdateHandler(callId, {hookName, hookMessage, decision, mergedDecision})`
  before applying the merged block / ask / allow outcome.
- `UserPromptSubmit` non-blocking allow / approve now yields a
  `HookSystemMessage` event using `systemMessage ?? reason`. Block
  (`Error` event) and ask (`UserPromptConfirmation`) paths are unchanged
  because they already render the message.
- `additionalContext` LLM injection is unchanged — the new path is
  UI-only.

## Related Issue

closes #535

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] Multiple / Project-wide

## Checklist

- [x] Lock files are up to date (\`package-lock.json\` / \`Cargo.lock\`)

## Testing

\`\`\`
npx vitest run packages/core/src/core/client.test.ts \
                packages/core/src/core/coreToolScheduler.test.ts \
                packages/core/src/hooks/hookSystem.test.ts
→ 3 files passed, 122/122 tests passed
\`\`\`

Type check on the three changed test files is clean (pre-existing
third-party \`@types\` errors for \`simple-git\` / \`google-auth-library\` /
\`@opentelemetry/exporter-metrics-otlp-http\` are unrelated to this PR).

Manual verification scenarios:

- \`UserPromptSubmit\` returning \`{decision: "allow", reason: "..."}\` —
  reason now renders in the terminal as a \`HookSystemMessage\`.
- \`PostToolUse\` returning \`{decision: "allow", reason: "..."}\` —
  reason renders as a per-hook notification on the existing tool
  message UI; tool still completes successfully.
- \`PostToolUse\` returning \`{decision: "block", reason: "..."}\` —
  notification renders with \`mergedDecision: "block"\` *and* the
  response sent to the LLM is replaced with the reason.

## Additional Notes

The non-interactive JSON output adapter
(\`BaseJsonOutputAdapter.processEvent\` handling of \`HookSystemMessage\`)
is intentionally deferred; #535 scopes to the terminal UI.

### DIFF
--- a/src/copilot-shell/packages/core/src/core/client.test.ts
+++ b/src/copilot-shell/packages/core/src/core/client.test.ts
@@ -30,6 +30,7 @@ import {
   GeminiEventType,
   Turn,
   type ChatCompressionInfo,
+  type ServerGeminiStreamEvent,
 } from './turn.js';
 import { getCoreSystemPrompt } from './prompts.js';
 import { DEFAULT_QWEN_FLASH_MODEL } from '../config/models.js';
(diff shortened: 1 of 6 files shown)


samchu-zsl (approved):
## Summary

LGTM — clean follow-up to #421 with strong test discipline and explicit scope control.

The diff splits cleanly: ~60 lines of business logic, ~600 lines of tests, all of which sit on top of types and event kinds (`HookSystemMessage`, `notifications?: Array<…>`, `OutputUpdateHandler`) that PR #421 already introduced. The risk surface is small.

### Why the structure works

- **PostToolUse path** (`coreToolScheduler.ts:1512-1547`) is a near-perfect mirror of the existing PreToolUse emission at `:866-900`: the `mergedDecision` precedence (block/deny → ask → per-hook decision), the "emit notifications BEFORE applying the merged outcome" rule, and the same `(callId, {hookName, hookMessage, decision, mergedDecision})` payload shape. Future readers won't have to learn two different patterns.
- **UserPromptSubmit allow path** (`client.ts:622-633`) is correctly gated on `!isBlockingDecision() && !shouldStopExecution() && !isAskDecision()`, so it never double-renders alongside the existing `Error` event (block, `client.ts:577-583`) or `UserPromptConfirmation` event (ask, `client.ts:594`). The four negative tests in `client.test.ts` (block, ask, allow-with-no-message, regression for #535) lock this in.
- **Aggregator → output transport** (`hookSystem.ts:197-201`) uses `result.notifications?.length` to avoid attaching empty arrays. `hookAggregator.ts:135` already normalizes to `undefined` when there are none, so the two ends agree.

- `DefaultHookOutput.notifications` is declared in `types.ts:165-166` and inherited by `PostToolUseHookOutput`, so the assignment at `hookSystem.ts:200` is type-safe.
- `GeminiEventType.HookSystemMessage` already exists at `turn.ts:67` (introduced by PR #421); this PR adds new yield sites but no new event kinds.
- The `if (message)` guard before yielding HookSystemMessage correctly suppresses the empty-allow case (covered by `does not emit HookSystemMessage when allow has no message at all`).
- Block path inside UserPromptSubmit *does* fold the reason into its `Error` event (`UserPromptSubmit hook blocked processing: ${getEffectiveReason()}`), as the PR description claims.

### Notes (non-blocking)

1. **`systemMessage` / `reason` precedence is inconsistent across paths**. The new allow path uses `systemMessage ?? reason` (systemMessage wins), and the corresponding test (`prefers systemMessage over reason when both are present on allow`) locks that in. The existing ask path at `client.ts:594` does the opposite: `reason || systemMessage || 'Prompt reviewed and approved'`. Both choices are defensible in isolation, but having them differ means a hook author writing both `systemMessage` and `reason` will see *different* strings depending on whether the decision is allow vs ask. Worth a follow-up to converge on one direction (and update the ask path or document the asymmetry); not blocking this PR.

2. **non-interactive / acp modes intentionally deferred**. The PR description is upfront that `BaseJsonOutputAdapter.processEvent` handling of `HookSystemMessage` is out of scope for #535. That's fine for now, but it does mean a hook author who emits an informational `reason` will see it in the terminal and *not* see it via the JSON adapter, which is a subtle inconsistency that hook docs should call out — or, better, a follow-up issue should track JSON-adapter coverage so the gap doesn't get rediscovered later.

Otherwise, ship it.

PR #547, opened 2026-05-15
fix(cosh): correct read_file arg key in auto-memory session hook

kongche-jbw:
## Description

The auto-memory extraction subagent's `postToolUse` hook reads
`payload.args['absolute_path']` to track which chat session files were
actually consumed, but the `read_file` tool's real schema uses `file_path`
(see `tools/read-file.ts`). The wrong key returns `undefined`, so
`onSessionRead` never fires for `read_file` reads. The
`actuallyReadSessions` set in `memoryService.ts` stays empty, and the
"only mark processed after read" logic causes the same candidate sessions
to be retried run after run until `MAX_ATTEMPTS_BEFORE_SKIP` — burning LLM
calls. This was likely inherited from gemini-cli where the tool used
`absolute_path`. The accompanying test fed the mock payload with the same
wrong key, so it stayed green and missed the regression.

## Related Issue

no-issue: trivial bug fix surfaced during review of cba22cb

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] Multiple / Project-wide

## Checklist

## Testing

Red-green verification on `skillExtractionAgent.test.ts`:

1. Apply the test fix only (mock payloads use `file_path`) — the
   `reports sessionId when read_file reads a chat session file` test
   FAILS, asserting `[]` instead of `[uuidA]`. This proves the test
   would have caught the original bug.
2. Apply the source fix — all 6 tests in
   `skillExtractionAgent.test.ts` pass.

## Additional Notes

The `read_many_files` branch on the next line (`payload.args['paths']`)
was already correct — only the `read_file` branch had the wrong key.

### DIFF
--- a/src/copilot-shell/packages/core/src/services/autoMemory/skillExtractionAgent.test.ts
+++ b/src/copilot-shell/packages/core/src/services/autoMemory/skillExtractionAgent.test.ts
@@ -37,7 +37,7 @@ describe('createExtractionHooks session-read tracking', () => {
     });
     await hooks.postToolUse?.(
       makePayload('read_file', {
-        absolute_path: path.join(chatsDir, `${uuidA}.jsonl`),
+        file_path: path.join(chatsDir, `${uuidA}.jsonl`),
       }),
     );
     expect(reads).toEqual([uuidA]);
@@ -68,7 +68,7 @@ describe('createExtractionHooks session-read tracking', () => {
       onSessionRead: (sid) => reads.push(sid),
     });
     await hooks.postToolUse?.(
-      makePayload('read_file', { absolute_path: '/etc/passwd' }),
+      makePayload('read_file', { file_path: '/etc/passwd' }),
     );
     expect(reads).toEqual([]);
   });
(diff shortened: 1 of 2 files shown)


samchu-zsl (approved):
## Summary

LGTM — clean, well-scoped trivial fix.

The change correctly aligns the autoMemory `postToolUse` hook with the real `read_file` tool schema (`ReadFileToolParams.file_path`, see `packages/core/src/tools/read-file.ts:33,168`). The `read_many_files` branch is untouched and remains correct (`paths: string[]`).

- `tools/read-file.ts` schema confirms `file_path` is the only key (`required: ['file_path']`); the file contains no `absolute_path` literal anywhere.
- `useGeminiStream.ts:1359` reads `args['file_path']` for the same tool, so the new key is consistent with the rest of the codebase.
- The remaining `absolute_path` occurrence at `read-many-files.ts:399` is unrelated — it is the input object passed to `getProgrammingLanguage(...)`, a telemetry helper, not the tool's input schema.
- After the fix, no `absolute_path` literal remains in the autoMemory module on PR HEAD (verified via `git show` of both `skillExtractionAgent.ts` and `skillExtractionAgent.test.ts`).
- Red-green logic holds: applying only the test diff while keeping the old source key would yield `reads === []` vs expected `[uuidA]`, failing `reports sessionId when read_file reads a chat session file` — exactly as the PR description claims.

Nice catch on the regression that the original test masked.

PR #548, opened 2026-05-15
fix(cosh): preserve user dirs in auto-memory workspace cleanup

kongche-jbw:
## Description

The auto-memory background extraction adds `chatsDir` and `memoryDir` to
the shared `WorkspaceContext`, then on cleanup calls
`setDirectories(getInitialDirectories())`. `initialDirectories` is a
snapshot frozen at constructor time and is never updated, so any
directory the user added at runtime via the `/directory` slash command
gets wiped when the background extraction finishes — and any subsequent
file access in those user directories silently fails the workspace
boundary check.

This change introduces `WorkspaceContext.removeDirectory(path)` and
makes the cleanup remove only the two directories auto-memory itself
added, leaving every other directory in the context untouched. The
extracted directories are tracked individually as they are added, which
also closes a partial-add leak: previously, if `addDirectory(chatsDir)`
succeeded but the subsequent `mkdir`/`addDirectory(memoryDir)` threw,
`addedExtraDirs` stayed `false` and the `chatsDir` entry leaked.

## Related Issue

no-issue: bug fix surfaced during review of cba22cb

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] Multiple / Project-wide

## Checklist

## Testing

`workspaceContext.test.ts` gains six TDD-driven tests covering
`removeDirectory`: removal, leaving initial directories intact, no-op
on non-member, symlink resolution, listener notification, and throwing
on a non-existent path. Red-green confirmed: all six tests fail with
`removeDirectory is not a function` before the implementation, and pass
after. Full suite for the touched modules: 60/60 pass
(`workspaceContext` 40 + `autoMemory/*` 20).

## Additional Notes

This fix is independent of #547 (the `read_file`/`absolute_path` arg-key
bug); both issues live in the auto-memory subsystem introduced in
cba22cb but touch disjoint files.

### DIFF
--- a/src/copilot-shell/packages/core/src/services/autoMemory/memoryService.ts
+++ b/src/copilot-shell/packages/core/src/services/autoMemory/memoryService.ts
@@ -558,14 +558,17 @@ export async function startAutoMemoryExtraction(
     );
 
     // Add chats and memory directories to workspace context so the extraction
-    // agent can read session files and write patches/skills via tools
+    // agent can read session files and write patches/skills via tools.
+    // Track each successfully added directory so cleanup removes only what we
+    // added, leaving any user-added directories (e.g. via /directory) intact.
     const workspaceContext = config.getWorkspaceContext();
-    let addedExtraDirs = false;
+    const extraDirs: string[] = [];
     try {
       workspaceContext.addDirectory(chatsDir);
+      extraDirs.push(chatsDir);
       await fs.mkdir(memoryDir, { recursive: true });
       workspaceContext.addDirectory(memoryDir);
-      addedExtraDirs = true;
+      extraDirs.push(memoryDir);
     } catch {
       debugLogger.warn(
         `Could not add chats/memory directories to workspace context`,
@@ -606,11 +609,16 @@ export async function startAutoMemoryExtraction(
       await subagent.runNonInteractive(context, abortController.signal);
     } finally {
       chatRecordingService?.removeExcludedPromptIdPrefix(excludePrefix);
-      // Remove the temporarily added directories
-      if (addedExtraDirs) {
-        workspaceContext.setDirectories(
-          workspaceContext.getInitialDirectories(),
-        );
+      // Remove only the directories we added; never touch directories that
+      // were already present or added by the user during this run.
+      for (const dir of extraDirs) {
+        try {
+          workspaceContext.removeDirectory(dir);
+        } catch (e) {
+          debugLogger.warn(
+            `Failed to remove auto-memory directory ${dir} from workspace context: ${e}`,
+          );
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


samchu-zsl (approved):
## Summary

LGTM — surgical fix with sound reasoning and good test coverage.

The two bugs are well diagnosed and the fix is precise:

- **User-dir wipe**: `setDirectories(getInitialDirectories())` was using the constructor-time snapshot, so any directory the user added at runtime via `/directory` was lost when extraction finished. The new `extraDirs[]`-driven cleanup only removes what auto-memory itself added.
- **Partial-add leak**: previously, `addDirectory(chatsDir)` succeeding followed by `mkdir`/`addDirectory(memoryDir)` throwing left `addedExtraDirs === false`, so cleanup didn't run and `chatsDir` leaked. Tracking each successful add individually closes that hole.

- `WorkspaceContext` is re-exported from `packages/core/src/index.ts:93`, so the new `removeDirectory` is public API. The signature is purely additive and `grep` finds no external callers besides the new tests and `memoryService.ts`.
- `removeDirectory` is implemented as the exact mirror of `addDirectory` (same `resolveAndValidateDir` path, including `realpathSync`, plus the `if (!delete) return; notify();` shortcut). Symmetry checks out at `workspaceContext.ts:81-90` vs `addDirectory` above it.
- Partial-add reasoning holds: if `mkdir(memoryDir)`/`addDirectory(memoryDir)` throws, `extraDirs === [chatsDir]` and the `finally` loop calls `removeDirectory(chatsDir)`, draining the leak.
- The inner `try/catch` around `removeDirectory` in the `finally` block is the right call — a missing/replaced directory shouldn't break the other cleanup.
- Copyright header on `workspaceContext.ts` (`Copyright 2025 Google LLC`) is correctly preserved; diff is ~13/212 lines, well under the 50% threshold.
- Six TDD tests cover the new method end-to-end (removal, initial intact, no-op on non-member, symlink resolution, listener notification, throw on non-existent).

### Note (non-blocking, optional)

`removeDirectory` mirrors `addDirectory` deliberately, which means it accepts *any* resolvable directory — including `cwd` or anything in `initialDirectories`. For auto-memory's actual call site this is fine (it only ever passes its own `chatsDir`/`memoryDir`), and treating both methods symmetrically is a defensible design choice. Just flagging that since `WorkspaceContext` is public API, a future caller could in principle remove the workspace root and silently break boundary checks. If you'd rather harden the API up front (e.g. refuse to remove paths in `initialDirectories`, or expose a narrower `removeAddedDirectory` instead), this is a fine moment to do it. Equally fine to leave as-is and treat it as the caller's contract.

Otherwise, ship it.

PR #550, opened 2026-05-18
fix(cosh): atomic stale-lock takeover and async IO in memory hooks

kongche-jbw:
## Description

Fix two correctness bugs in the auto-memory background extraction system introduced in cba22cb:

1. `postToolUse` hook used sync `readFileSync` / `writeFileSync` inside an `async` callback, blocking the event loop on every patch write.
2. Stale-lock recovery in `tryAcquireLock` had a TOCTOU window between the `isLockStale` check and the `unlink + recreate` sequence, allowing two instances to both think they hold the lock.

The lock takeover now uses an atomic `rename` + `stat.ino` verification, with a `vanished` outcome that lets the caller retry the fast path safely when a concurrent takeover-winner has just moved the stale lock away. Sync FS calls in the hook are replaced with `fs/promises`.

## Related Issue

no-issue: address code-review findings on commit cba22cb

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)
- [ ] `sec-core`
- [ ] `skill`
- [ ] `sight`
- [ ] `tokenless`
- [ ] Multiple / Project-wide

## Checklist

- [x] I have added tests that prove my fix is effective
- [x] For \`cosh\`: tests pass (40 tests across 4 files in autoMemory)
- [ ] For \`cosh\`: Lint and type check — not run in this session

## Testing

\`\`\`bash
cd src/copilot-shell/packages/core
npx vitest run src/services/autoMemory/
\`\`\`

40 tests pass:
- \`memoryService.test.ts\` (14, new): fresh acquire, live-lock rejection, dead-PID/aged-lock takeover, malformed JSON, concurrent contention with single-winner invariant, \`isLockStale\` and \`releaseLock\` cases.
- \`skillExtractionAgent.test.ts\` (12, +6): valid patch normalized via async IO, invalid hunk → \`additionalContent\`, paths outside \`.inbox/\` ignored, non-\`.patch\` files ignored, missing file silent, upstream failure skipped.

## Additional Notes

The takeover relies on POSIX \`rename\` atomicity and \`stat.ino\` stability — holds on Linux (ext4/btrfs/xfs), macOS (APFS), and Windows NTFS within a single volume. On exotic filesystems with unstable inodes, the inode check weakens but \`rename\` atomicity alone still prevents the dominant race.

### DIFF
--- a/src/copilot-shell/packages/core/src/services/autoMemory/memoryService.test.ts
+++ b/src/copilot-shell/packages/core/src/services/autoMemory/memoryService.test.ts
@@ -1,0 +1,174 @@
+/**
+ * @license
+ * Copyright 2025 Qwen Code
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'node:fs/promises';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import { tryAcquireLock, isLockStale, releaseLock } from './memoryService.js';
+
+const FAR_FUTURE_PID = 2 ** 22; // Reasonably guaranteed not to be alive on POSIX.
+
+describe('memoryService lock management', () => {
+  let tmpDir: string;
+  let lockPath: string;
+
+  beforeEach(async () => {
+    tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), 'memsvc-lock-test-'));
+    lockPath = path.join(tmpDir, '.extraction.lock');
+  });
+
+  afterEach(async () => {
+    await fs.rm(tmpDir, { recursive: true, force: true });
+  });
+
+  describe('tryAcquireLock', () => {
+    it('acquires a lock on a fresh path', async () => {
+      expect(await tryAcquireLock(lockPath)).toBe(true);
+      const content = await fs.readFile(lockPath, 'utf-8');
+      const parsed = JSON.parse(content);
+      expect(parsed.pid).toBe(process.pid);
+      expect(typeof parsed.startedAt).toBe('string');
+    });
+
+    it('returns false when a live-process lock already exists', async () => {
+      await fs.writeFile(
+        lockPath,
+        JSON.stringify({
+          pid: process.pid,
+          startedAt: new Date().toISOString(),
+        }),
+      );
+      expect(await tryAcquireLock(lockPath)).toBe(false);
+    });
+
+    it('reclaims a stale lock whose owner PID is dead', async () => {
+      await fs.writeFile(
+        lockPath,
+        JSON.stringify({
+          pid: FAR_FUTURE_PID,
+          startedAt: new Date().toISOString(),
+        }),
+      );
+      expect(await tryAcquireLock(lockPath)).toBe(true);
+      const parsed = JSON.parse(await fs.readFile(lockPath, 'utf-8'));
+      expect(parsed.pid).toBe(process.pid);
+    });
+
+    it('reclaims an aged lock even when the owner PID is alive', async () => {
+      const ancient = new Date(Date.now() - 60 * 60 * 1000).toISOString();
+      await fs.writeFile(
+        lockPath,
+        JSON.stringify({ pid: process.pid, startedAt: ancient }),
+      );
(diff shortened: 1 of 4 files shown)


samchu-zsl (approved):
## Summary

LGTM — solid race-condition fix with thorough test coverage and honest platform notes.

The two correctness bugs are independently real and the fixes are well-targeted:

### 1. Sync IO inside `async` `postToolUse` (#1, #13)

The hook lives on the hot path of every patched write_file, so `readFileSync` / `writeFileSync` were blocking the Node event loop on each call. The PR replaces both with `await fs.readFile` / `await fs.writeFile` from `node:fs/promises`. Verified `git show pr-550:skillExtractionAgent.ts` is clean of any remaining sync FS API.

### 2. Stale-lock TOCTOU (#4 concurrency)

The old `isLockStale` check + `releaseLock` (`unlink`) + `tryAcquireLock` recursion had a classic check-then-act window: two callers could both see "stale", both unlink, both `O_EXCL`-create. The new flow is sound:

- Fast path (`O_EXCL`) is unchanged and is the primary serializer.
- On `EEXIST`, takeover uses **atomic** `rename(lockPath, trashPath)` — a single dirent source, only one caller's rename can succeed.
- After rename, `stat(trashPath).ino === inspection.stat.ino` is the **identity token**: if a fresh lock was created in the gap between `inspect` (stat + readFile) and `rename`, we'd be holding the wrong file's inode and we restore via `link` (best-effort) before returning `'busy'`.
- `'vanished'` (rename → ENOENT) lets the loser of a takeover race retry the fast path safely; combined with `retries=1`, the recursion is bounded (`'taken'` and `'vanished'` both consume the budget; `'busy'` returns false directly).

### Race trace (verified against the diff)

- A and B both observe stale lock at ino=X.
- A: `rename` succeeds, ino-check passes, retries the fast path, wins `O_EXCL`.
- B: `rename` returns `ENOENT` → `'vanished'` → retries fast path → hits `EEXIST` (A's lock) with `retries=0` → returns `false`. ✅ single winner.
- Adversarial: between A's stat and rename, another process replaces the lock with a fresh one. A's rename grabs the fresh ino=Y; ino-check fails; `link(trashPath, lockPath)` restores it; A returns `'busy'`. ✅ no double-takeover.

- `tryAcquireLock` / `isLockStale` / `releaseLock` are still exported but a repo-wide grep finds no consumer outside `services/autoMemory/`. The refactor is internal-only — no API ripple.
- `isLockStale` is now a thin wrapper over `inspectLock` and preserves all four legacy stale-conditions (read fail / JSON malformed / dead PID / aged).
- 14 lock-management tests + 6 patch-validation tests are well-aimed; the "concurrent contention with single-winner invariant" test (`Promise.all([tryAcquireLock, tryAcquireLock])`, expects exactly 1 winner) directly asserts the property the rewrite is meant to give.
- Copyright header on the new `memoryService.test.ts` (`Copyright 2025 Qwen Code`) matches the existing convention in `services/autoMemory/` (all other `.test.ts` files there use the same header).
- The platform note in the PR description is accurate: POSIX `rename` atomicity holds on Linux/macOS/Windows-NTFS within a volume; `stat.ino` is stable on the same filesystems. The PR explicitly acknowledges that pathological FSes weaken the inode check while leaving rename atomicity intact — this is the right framing.

### Notes (non-blocking)

1. **Lint / typecheck not run locally** — the PR checklist itself flags this. Worth confirming the CI checks come back green before merge; type-wise the diff looks clean (`import type { Dirent, Stats }`, `error: unknown` guards, `TakeoverOutcome` union exhaustive at the call site), but a local `npm run lint && npm run typecheck` would close the loop.
2. **`inspectLock` does `stat` then `readFile`** as two separate syscalls, so the `Stats` it returns and the `LockInfo` it parsed could in principle come from different inodes if the file is replaced in between. The takeover's `movedStat.ino === inspection.stat.ino` check turns this into a `'busy'` outcome (i.e. it's safe by construction), but a one-line JSDoc on `inspectLock` documenting *why* the returned `Stats.ino` is the load-bearing identity token would help future readers.

Otherwise, ship it.

PR #661, opened 2026-05-28
feat(sight): agent crash full-stack observability

jfeng18:
## Summary

Agent 进程层级全栈可观测：血缘树 + 精准 crash 检测 + blast radius 量化 + token 按层级归因 + CLI crash report + `token --by-type` 聚合查询。

agent 崩了，系统知道崩没崩、为什么崩、影响多大、崩的时候在干什么、哪个子进程在烧钱。

## Strategic positioning

This PR is the AgentSight foundation for the OS-middle-layer recoverability and task-level observability substrate: process supervision, exit-reason capture, Agent/SubAgent/Tool lineage, crash blast-radius reporting, and token attribution by process type.

It deliberately does **not** implement the scheduling/control-plane pieces from the broader Agent Infra roadmap: no cgroup QoS control, no LSM enforcement, no transparent MaaS cache proxy, and no closed-loop task recovery orchestration. Those can build on this observability substrate later.

## Problem

1. **crash 检测靠猜**：用 pending LLM calls 判断 crash，SIGSEGV 在两次调用间 → 漏判，正常退出 SSE 没收完 → 误判
2. **子进程完全不可见**：SubAgent/Tool 崩溃静默丢弃（scanner 只认 root agent）
3. **token 无层级归因**：不知道哪个子 agent 在烧钱，无法按 Agent/SubAgent/Tool 分层分析
4. **启动时已有 agent 不在 lineage tree**：scanner 初始扫描发现的 agent 没有 lineage 节点

## Solution

| 层 | 改动 | 效果 |
|---|---|---|
| BPF 采集 | procmon.h 加 `exit_code`，BPF_CORE_READ task_struct | crash/normal 从内核事实判定 |
| 血缘树 | lineage/mod.rs：pid→ppid 树 + Agent/SubAgent/Tool 分类 + root_agent_ancestor | 进程层级结构化 |
| crash 判定 | classify_exit() 纯函数替代 pending-calls 猜测 | 假阴性/假阳性修复 |
| blast radius | ProcessType → total_session_loss/partial/recoverable | 量化崩溃影响，severity 按类型分级 |
| SubAgent/Tool crash | Exit handler 扩展：lineage 分类的子进程 signal crash 也检测 | 子进程不再静默 |
| token 归因 | LLMCall/PendingCallInfo 加 process_type，SQLite+Logtail 全链路 | 按层级分析消耗 |
[...]

PR #665, opened 2026-05-31
feat(sight): observe-only BPF LSM audit probe (socket_connect + file_open)

jfeng18:
## What

Adds an opt-in, **observe-only** BPF LSM probe (`lsmaudit`) that audits two security-relevant actions for traced Agent families:

- **`lsm/socket_connect`** — every outbound connection attempt (destination IP:port), regardless of protocol or port.
- **`lsm/file_open`** — every file opened (basename + access mode), deduped per `(pid, inode)` via an LRU map.

Both hooks filter to traced families via the shared `traced_processes` map (`is_pid_traced`) and emit through the shared ring buffer (`EVENT_SOURCE_LSM`). Events surface as structured audit log lines:

```
[LSM] connect pid=… comm=… dst=IP:PORT
[LSM] open    pid=… comm=… mode=r|w|rw path=…
```

## Observe-only

Strictly observe-only: every program returns the incoming `ret` unchanged, so no operation is ever denied. The LSM attach point leaves room for future enforcement.

## Opt-in

Off by default. Enable with `--enable-lsm-audit` (or the matching config field). Requires BPF LSM active (`bpf` in `/sys/kernel/security/lsm`); otherwise the probe is skipped with a warning.

## Correctness fixes folded in (commit `fix(sight): lsmaudit BPF correctness + leak guard`)

A self-review of the as-pushed branch caught three issues, all fixed in the latest commit:

1. **path[] info leak** — `bpf_ringbuf_reserve` returns a slot that may carry residue from a previously-submitted record on the shared ring buffer (sslsniff TLS plaintext, filewrite content). The original code only zeroed `path[0]`, leaving `path[1..256]` uninitialized. Memset the whole record after reserve in both hooks.
2. **file_open dedup-before-emit** — the `(pid, inode)` LRU was inserted *before* reserve / read, so any failure permanently suppressed that file. Move the dedup mark to after a successful submit; `bpf_ringbuf_discard` the slot on path-read failure.
3. **socket_connect addrlen unchecked** — LSM fires before the protocol-level length check; if `addrlen < sizeof(struct sockaddr_in[6])` the destination fields are stack residue. Drop the event in that case.

Plus an always-on Rust-side leak detector: `LsmEvent::from_bytes` scans `path[]` past the null terminator and `log::warn!`s on any non-zero byte — permanent regression net for the producer-side memset invariant.

## Testing

|---|---|---|
| **已 E2E**(real-machine, kernel 6.6.102, BPF LSM active) | ✓ | Both hooks attach; traced agent's connects/opens captured with correct fields; observe-only proven (all attempts succeed); 173 events on the post-fix sanity run, 0 leak-detector warnings, paths clean. |
| **仅单测**(`cargo test --lib lsmaudit`) | ✓ | 8/8: gate (cacheability), IPv4/IPv6 + port byte-order decode, file_open parsing, RawLsmEvent C↔Rust layout, **leak detector positive** (garbage tail triggers warn + decoder still cleanly extracts path), **leak detector negative** (clean tail stays silent). 447 lib tests, zero regressions. |
| **未 E2E**(documented gap) | — | The leak class (residue from ringbuf wrap) is not reproducible in a synthetic test in seconds — the 32 MiB ring rarely wraps. A reverse-test with the memset removed produced 158 events and 0 leak warnings on the synthetic load. The fix is correct by construction (defense-in-depth, parallel to filewrite's pattern); the detector exists to catch the real case in long-running production. |

## Notes

- The event struct is `lsm_audit_event` to avoid colliding with the kernel's own `enum lsm_event` in `vmlinux.h`.
- `file_open` records the basename only (full path is future work — `bpf_d_path` is allowlist-rejected for `lsm/file_open`).
- Other probes likely share the same path-tail-residue pattern (filewrite's `buf[16384]`). Out of scope for this PR; tracked as a separate follow-up.

Independent of #661–#668.

### DIFF
--- a/src/agentsight/build.rs
+++ b/src/agentsight/build.rs
@@ -61,7 +61,11 @@ fn main() {
 
     // Generate tcpsniff skeleton (no header — reuses sslsniff.h event format)
     generate_skeleton(&mut out, "tcpsniff");
-    
+
+    // Generate lsmaudit skeleton and bindings
+    generate_skeleton(&mut out, "lsmaudit");
+    generate_header(&mut out, "lsmaudit");
+
     // generate_header(&mut out, "frametypes");
     // generate_header(&mut out, "errors");
     // generate_header(&mut out, "stackdeltatypes");
--- a/src/agentsight/src/bin/cli/trace.rs
+++ b/src/agentsight/src/bin/cli/trace.rs
@@ -22,6 +22,11 @@ pub struct TraceCommand {
     #[structopt(long)]
     pub enable_filewatch: bool,
 
+    /// Enable LSM security-audit probe (observe-only: records outbound
+    /// connections and file opens for traced agents; requires BPF LSM active)
+    #[structopt(long)]
+    pub enable_lsm_audit: bool,
+
     /// Path to JSON configuration file
     #[structopt(short, long, default_value = "/etc/agentsight/config.json")]
     pub config: String,
@@ -72,6 +77,7 @@ impl TraceCommand {
         let config = AgentsightConfig::new()
             .set_verbose(self.verbose)
             .set_enable_filewatch(self.enable_filewatch)
+            .set_enable_lsm_audit(self.enable_lsm_audit)
             .set_config_path(std::path::PathBuf::from(&self.config));
 
         // Create AgentSight (auto-attaches probes and starts polling)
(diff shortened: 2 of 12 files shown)


jfeng18:
Closing: observability gap is real (no probe on main sees file reads or pre-connect attempts), and the BPF code is clean and deployable on our 6.6 kernels. But three independent problems: (1) zero consumers — handle_lsm_event writes to log::info! and discards; (2) unsolved volume architecture — file_open generates thousands of events per build vs audit store designed for tens of LLM calls; (3) no demonstrated demand from roadmap or customers.

The BPF core (lsmaudit.bpf.c + lsmaudit.h + lsmaudit.rs, ~511 lines) is cherry-pickable from git history. When enterprise audit requirements materialize: design volume-appropriate storage first → define the consumer → then cherry-pick the sensor core and write fresh integration wiring.

PR #667, opened 2026-05-31
fix(sight): sync FFI header + drift guard + example

jfeng18:
## What

The hand-maintained C declarations in `cbindgen.toml`'s `after_includes` block had drifted from the actual FFI surface in `src/ffi.rs`:

- declared the deleted `agentsight_config_add_domain_rule`
- missing the newer `agentsight_config_add_https` / `agentsight_config_add_http`
- `examples/agentsight_example.c` still called the deleted symbol (would not link)

Root cause: cbindgen 0.27 does not recognize Rust 2024's `#[unsafe(no_mangle)]` attribute and silently skips every exported function, so the toml has `item_types = ["structs"]` and the function list is pasted manually — with no mechanical check.

## Changes (3 commits)

1. **`chore(sight): sync FFI header with current ffi.rs`** — re-syncs the toml `after_includes` block to match the actual 16 FFI exports; rewrites the NOTE to separate the two independent workarounds (`item_types=["structs"]` is required even after a cbindgen upgrade, to keep unrelated `pub const`s out of the public header).
2. **`feat(sight): add drift guard for FFI header`** — `build.rs` extracts every `pub [unsafe] extern "C" fn agentsight_*` from `src/ffi.rs` and every `<type> agentsight_*(` declaration from the generated header, then panics with the symmetric difference (missing-in-header / stale-in-header) when the two NAME sets disagree. No `.github/workflows` change needed. Also adds a crate-local `.gitignore` for the generated `include/`.
3. **`fix(sight): update FFI example for current API`** — replaces the deleted `agentsight_config_add_domain_rule` call with `agentsight_config_add_https` and adds `-I./include` to the docstring `gcc` command.

## What is intentionally NOT changed

- `src/ffi.rs` — actual exports unchanged; this PR only realigns the header.
- cbindgen version — kept at 0.27; bumping is a separate effort, with TODOs in `cbindgen.toml` enumerating the cleanup steps for when it lands.
- `include/agentsight.h` — still generated, not tracked.
- `.github/workflows/` — drift detection runs in `build.rs`, not CI YAML.

## Testing

- **Forward**: `cargo build` clean.
- **Reverse (3 modes)**: drift guard correctly panics with a precise `missing / stale` list for (a) deleted declaration, (b) renamed function (count preserved — the count-only iteration would have silently passed), (c) duplicate-with-drop (count preserved). All three reverse cases tried in-tree before commit; restored after.
- **Release**: `cargo build --release` clean; `gcc -I./include -L./target/release -lagentsight` links the example; `nm` shows all 13 example-used FFI symbols resolved against `libagentsight.so` (including the new `agentsight_config_add_https`).
- **ECS sanity** (kernel 6.6.102): example runs for 5s under root, loads BPF objects via libbpf without crash, exits cleanly on timeout. No segv / panic / abort.

## Known limits (called out in code + commit)

The drift guard checks NAMES only — signature drift (return type, parameter types/order/count) is **not** detected automatically. Both the panic message and the cbindgen.toml note say this verbatim, so a green guard cannot be misread as a full ABI assertion.

Independent of #661–#666.

### DIFF
--- a/src/agentsight/.gitignore
+++ b/src/agentsight/.gitignore
@@ -1,0 +1,3 @@
+# Generated C header — written on every `cargo build` by build.rs via cbindgen.
+# Consumers regenerate locally; do NOT commit.
+/include/
--- a/src/agentsight/build.rs
+++ b/src/agentsight/build.rs
@@ -99,4 +99,130 @@ fn main() {
         .write_to_file(&header_path);
     println!("cargo:rerun-if-changed=src/ffi.rs");
     println!("cargo:rerun-if-changed=cbindgen.toml");
+
+    // Drift guard: cbindgen 0.27 silently skips `#[unsafe(no_mangle)]`, so the
+    // function declarations in cbindgen.toml are hand-maintained. Compare the
+    // export NAMES on both sides and fail the build if they disagree.
+    //
+    // NOTE: this guard checks NAMES only — it does NOT verify return types or
+    // parameter signatures. A maintainer changing only a return type or arg
+    // list (no name change) will pass this check but still produce an ABI
+    // mismatch. Keep cbindgen.toml's hand-written signatures in lockstep with
+    // src/ffi.rs by hand until cbindgen learns the new attribute.
+    check_ffi_header_drift(&crate_dir, &header_path);
+}
+
+/// Extract every `agentsight_<ident>` that appears as a function declaration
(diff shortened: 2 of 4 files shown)


jfeng18:
FFI header sync fix (161 lines). Ensures cbindgen output stays consistent with Rust API changes.

chengshuyi (approved):

PR #663, opened 2026-05-30
chore(sight): fix HTTP/2 Huffman decode + dead-code/warning cleanup + docs

jfeng18:
## Summary

Independent polish pass on `sight` (off `main`, no dependency on the lineage/scheduler PRs). Four focused commits:

1. **fix: decode HTTP/2 HPACK Huffman headers** — `ParsedHttp2Frame::huffman_decode()` returned a `"<huffman:N bytes>"` placeholder, so any Huffman-encoded HTTP/2 header (`:path`, `:status`, content-type, header values) was unreadable, breaking path classification / SSE detection for HTTP/2 LLM-API traffic. Now delegates to the already-imported `hpack` crate's RFC 7541 decoder, with RFC test vectors.
2. **chore: remove dead code, deprecated APIs, unused imports** — delete ~265 lines of compiler-confirmed dead token-computation code in `analyzer/unified.rs`, the unused `comm_to_string` / `get_pending_calls_for_pid`, and two `todo!()` public stubs in `tokenizer/llm_tok.rs` (latent panics); replace deprecated `base64::encode`; make `set_ffi_sender` `pub(crate)` (was leaking a crate-internal type); drop unused imports.
3. **chore: silence generated-code warnings** — scope an `#[allow(...)]` to each probe's `mod bpf { include!(...) }` block (bindgen/skeleton output) and clear the last hand-written warnings. The crate now builds **warning-free** (lib + tests + bins).
4. **docs: sync probe/Event/test-index docs** — AGENTS.md, integration-tests/README, and the ebpf-probes design doc had drifted behind the 7 merged probes / 6 Event variants.

## Test plan

- [x] 439 unit tests pass, zero regressions; new RFC 7541 Huffman decode tests
- [x] `cargo build` clean: **0 warnings** across lib + tests + bins
- [x] Pure userspace (parsing/cleanup/docs) — no eBPF runtime change, so no kernel E2E required

> Follow-up (separate PR): a `cargo fmt` + `cargo clippy --fix` pass to make `sight` pass `clippy -D warnings`, then add the fmt/clippy/FFI gate to the agentsight CI job (the repo PR template already expects it; CI currently runs only `cargo test`).

Generated with [Claude Code](https://claude.com/claude-code)

---

## Update (polish)

**Huffman fix split out** to #670 — `fix(sight): decode HPACK Huffman headers`. Force-pushed to drop commit `63fad98` from this stack so the diff is now scope-clean (polish/cleanup/docs only).

---

[...]

PR #670, opened 2026-06-01
fix(sight): decode HPACK Huffman headers

jfeng18:
## What

`ParsedHttp2Frame::huffman_decode()` previously returned the literal placeholder string `<huffman:N bytes>`, so any Huffman-encoded HTTP/2 header literal (`:path`, `:status`, `content-type`, request/response header values) came out unreadable. This broke path classification and SSE detection for HTTP/2 LLM-API traffic.

Delegate to the already-imported `hpack` crate's canonical Huffman decoder (RFC 7541 Appendix B), falling back to a lossy view of the raw bytes on a decode error. Adds RFC 7541 test vectors.

## Why a separate PR

Originally bundled into #663 (polish) — that mixed a behavior fix with cleanup, violating "PR scope: one PR = one goal". Splitting it out gives the fix its own review surface and makes back-porting / cherry-picking trivial. Once this PR lands, #663 will rebase onto main and the duplicate commit drops out automatically.

## Testing

- **仅单测** (`cargo test --lib parser::http2`):
  - `test_huffman_decode_rfc7541_vectors` — RFC 7541 Appendix B C.4.1 `"www.example.com"`, C.4.2 `"no-cache"`, C.6.1 `":status 302"`.
  - `test_huffman_decode_invalid_falls_back_to_lossy` — discriminating signal: `assert!(!out.starts_with("<huffman:"))` flips red if the fix is reverted.
  - 17 http2 frame tests + full lib (439 tests) all pass.

- **未 E2E**: real HTTP/2 LLM-API traffic through SSL probe → frame parser → aggregator → audit pipeline is **not** exercised here; deferred to a follow-up real-traffic test on the AgentSight ECS.

Independent of #661–#668 except #663, from which this commit was extracted.

### DIFF
--- a/src/agentsight/src/parser/http2/frame.rs
+++ b/src/agentsight/src/parser/http2/frame.rs
@@ -323,15 +323,16 @@ impl ParsedHttp2Frame {
         (result, pos + length - start)
     }
 
-    /// Simple Huffman decoder for HPACK
+    /// Decode an HPACK Huffman-encoded string (RFC 7541 Appendix B) using the
+    /// hpack crate's canonical decoder. On decode error (corrupt/truncated data)
+    /// we fall back to a lossy view of the raw bytes rather than a placeholder,
+    /// so callers always get the most readable result available.
     fn huffman_decode(data: &[u8]) -> String {
-        // HPACK uses a specific Huffman code. For simplicity, we'll use
-        // a basic approach - in production, use the hpack crate's decoder
-        // which handles this correctly.
-        // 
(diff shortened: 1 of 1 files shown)


Daydreamer-Li (approved):
LGTM — fixes a real bug where Huffman-encoded HTTP/2 headers were unreadable, restoring SSE detection, path classification, and status code parsing; no new dependencies, scoped to the stateless decode path only, with RFC 7541 test coverage.

PR #671, opened 2026-06-02
fix(tokenless): fixes code review findings

shiloong:
## Description

refactor(tokenless): rename openclaw plugin Name to Tokenless and ID to tokenless
chore(tokenless): fix formatting and clippy warnings after cherry-pick
chore(tokenless): misc cleanups — serialization, permissions, dual trust
fix(tokenless): reserve truncation marker length in response compressor
fix(tokenless): add input size limit and validate db path
fix(tokenless): recover from poisoned mutex in stats recorder instead of failing
fix(tokenless): harden env-fix install paths with uid trust check and divert stderr to log
fix(tokenless): anchor home lookup on getpwuid_r and trust-check candidate binaries
fix(tokenless): propagate env-fix subprocess failures instead of returning stdout
fix(tokenless): bound SchemaCompressor recursion to prevent stack overflow
fix(tokenless): use mktemp in tests and safe home expansion
chore(tokenless): add tokenizer tests, improve chrono config, warn on poison clear
fix(tokenless): secure resolveBinaryPath and improve binary cache invalidation
fix(tokenless): add subprocess returncode checks and extract shared hook utilities
fix(tokenless): secure shell variable interpolation in env-fix and hooks

## Related Issue

fixes code review findings

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional change)
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `sec-core` (agent-sec-core)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [x] `tokenless` (tokenless)
- [ ] `memory` (agent-memory)
- [ ] Multiple / Project-wide

## Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [x] My code follows the project's code style
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly
- [ ] For `cosh`: Lint passes, type check passes, and tests pass
- [ ] For `sec-core` (Rust): `cargo clippy -- -D warnings` and `cargo fmt --check` pass
- [ ] For `sec-core` (Python): Ruff format and pytest pass
- [ ] For `skill`: Skill directory structure is valid and shell scripts pass syntax check
[...]

PR #676, opened 2026-06-02
fix(tokenless): compress-schema on array input

### LINKED ISSUE #697: [tokenless] bug: compress-schema gives zero compression on array input
When `tokenless compress-schema` receives a JSON array (e.g. OpenAI `tools` field), it falls through to the single-object branch which has no `function` key, producing zero compression.

**Reproduce:**
```bash
echo '[{"type":"function","function":{"name":"f","description":"short","parameters":{"type":"object","properties":{"x":{"type":"string"}}}}}]' | tokenless compress-schema
```
Output is identical to input (0% savings).

**Root cause:** CLI only checks `if batch` but not `value.is_array()`.

**Fix:** PR #676

jfeng18:
## What

Two fixes for the tokenless CLI compression commands:

### 1. `compress-schema` silently did nothing on array input

OpenAI's `tools` field is always a JSON array `[{type: "function", function: {...}}, ...]`, but `compress-schema` only compressed single objects. Passing the array as a whole `Value` skipped all compression because the top-level `Value` has no `"function"` key.

**Fix**: auto-detect array input (`value.is_array()`) and compress each element individually — same path as `--batch`.

**Result**: realistic schema input 747 → 582 bytes (22% saved); old binary returned input unchanged.

### 2. Silent fallback when compression does not reduce size

All three compression subcommands (`compress-schema`, `compress-response`, `compress-toon`) silently output the original input when the compressed result is not smaller. Users had no way to know compression was skipped.

**Fix**: add `eprintln!` diagnostics to stderr so callers know compression was skipped and why.

## Testing

|---|---|---|
| **已 E2E** (ECS, old vs new binary) | ✓ | Same realistic schema: old binary returns 747 bytes unchanged; new binary returns 582 bytes (22% saved) with title/examples removed. stderr warns correctly when no-op. |
| **仅单测** (`cargo test`) | ✓ | All 60 existing tests pass. |
| **新 shell test** (discriminating) | ✓ | `run-all-tests.sh` Test 1.5: pipes a JSON array with title+examples, asserts both are removed. Reverting the `is_array()` condition makes this test fail. |
| **未验证** | — | `--batch` flag interaction; 100+ tool arrays; deeply nested schemas; hook integration (compress_schema_hook.py stderr propagation). |

## Notes

- Does NOT conflict with #671 (security hardening) — that PR adds recursion limits and mutex recovery but does not touch array detection or stderr diagnostics.
- The 3 pre-existing `clippy::uninlined_format_args` errors in `tokenless-schema` are a baseline issue on origin/main (1.89 lint tightening); not introduced by this change.
- `--batch` flag is now semantically redundant for array inputs (auto-detected); consider deprecating in a follow-up.

Fixes #697

### DIFF
--- a/src/tokenless/crates/tokenless-cli/src/main.rs
+++ b/src/tokenless/crates/tokenless-cli/src/main.rs
@@ -327,6 +327,10 @@ fn run() -> Result<(), (String, i32)> {
             let before_tokens = estimate_tokens_from_bytes(input.len());
             let after_tokens = estimate_tokens_from_bytes(after_compact.len());
             let output_text = if after_tokens >= before_tokens {
+                eprintln!(
+                    "tokenless: response compression did not reduce size ({} -> {} est. tokens), outputting original",
+                    before_tokens, after_tokens
+                );
                 input.clone()
             } else {
                 result_json.clone()
@@ -447,6 +451,10 @@ fn run() -> Result<(), (String, i32)> {
             let before_tokens = estimate_tokens_from_bytes(input.len());
             let after_tokens = estimate_tokens_from_bytes(output.len());
             let display = if output.is_empty() || after_tokens >= before_tokens {
+                eprintln!(
+                    "tokenless: TOON encoding did not reduce size ({} -> {} est. tokens), outputting original JSON",
+                    before_tokens, after_tokens
+                );
                 input.clone()
             } else {
                 output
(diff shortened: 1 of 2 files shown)


shiloong (dismissed):
LGTM, Thx!

jfeng18:
Hi — this one has been approved and rebased onto latest main (post-#671). Is there anything blocking merge? Happy to address any remaining feedback.

shiloong (approved):
LGTM, Thx!

PR #679, opened 2026-06-02
 feat(sight):Add FFI input_message_delta field and unify the per-round increment algorithm

beartyson-tech:
## Background

`agentsight_read` previously exposed only the full request messages over FFI, so consumers could not get the
per-round delta; meanwhile each "latest round" computation was implemented separately and inconsistently, and
dropped the user's actual question under Anthropic and OpenClaw traffic.

## Scope of Changes

1. **FFI delta field**: Adds `input_message_delta` to `AgentsightLLMData`, populated by `build_llm_data` and
   delivered via the `agentsight_read` callback; the existing full `request_messages` is left unchanged.
2. **Single source for the increment**: Adds `genai::semantic::latest_round_input_messages`; SQLite
   (`try_insert_event` and `complete_pending`), FFI, SLS (`logtail.rs`), and the pending record (`builder.rs`)
   all switch to it.
3. **Fix dropped user question**: Anchors on the first message of the trailing run of text-bearing user messages,
   correctly handling Anthropic tool results encoded as `role=user` and OpenClaw's runtime-context user message
   appended after the question.
4. **Docs and example synced**: Updates `docs/design-docs/c-ffi-api.md` and `examples/agentsight_example.c`.

## Known Issues / Caveats

- `AgentsightLLMData` gains fields appended at the end; `agentsight.h` is generated by cbindgen at build time, so
  downstream must rebuild, and consumers compiled against the old layout face ABI incompatibility.
- `extract_last_user_raw` / `user_query` / fingerprint still use the old "last text user" logic, so under
  OpenClaw's dual-user pattern `user_query` may still be the runtime-context metadata; not fixed in this PR.

### DIFF
--- a/src/agentsight/docs/design-docs/c-ffi-api.md
+++ b/src/agentsight/docs/design-docs/c-ffi-api.md
@@ -70,6 +70,11 @@ typedef struct {
     /* 工具定义（JSON 数组字符串） */
     const char* tools;                /* LLMRequest.tools 序列化 JSON 数组; 无工具时为 "[]" */
     uint32_t    tools_len;
+
+    /* 增量输入消息（JSON 数组字符串）：与 SQLite genai_events.input_messages 同一套算法，
+       去掉 system 消息后，保留从最后一个 user 消息开始（含）到末尾的部分 */
+    const char* input_message_delta;
+    uint32_t    input_message_delta_len;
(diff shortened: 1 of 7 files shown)


Daydreamer-Li (approved):
LGTM

PR #753, opened 2026-06-06
fix(tokenless): harden hook exit-code handling + trust model consistency

### LINKED ISSUE #749: [tokenless] bug(tokenless): tool_ready_hook conflates recommended and required deps in post-fix re-scan
## Summary

`tool_ready_hook.sh` accumulates both **required** (line 300) and **recommended** (line 317) missing dependencies into the same `MISSING_DEP_JSONS` array. After the fix attempt, the re-scan loop (lines 362–368) iterates over all entries and unconditionally sets `HAS_REQUIRED_MISSING=true` for any still-missing dep — including recommended-only ones.

This causes the PARTIAL path (line 403 and 422) to be skipped, and the tool falls through to the `NOT_READY` / `decision:block` path, even though only recommended deps are missing and the tool should be functional in degraded mode.

## Code flow

```
Line 287: MISSING_DEP_JSONS="[]"         # init
Line 300: MISSING_DEP_JSONS += required  # required missing deps added
Line 317: MISSING_DEP_JSONS += recommended  # recommended missing deps ALSO added
Line 351: missing_count = length(MISSING_DEP_JSONS)  # counts both
Line 362-368: for each in MISSING_DEP_JSONS:
                 if still missing → HAS_REQUIRED_MISSING=true  # ← BUG: recommended treated as required
Line 403: if ! $HAS_REQUIRED_MISSING → PARTIAL  # skipped because of recommended dep
Line 449: → NOT_READY / decision:block           # tool unnecessarily blocked
```

## Impact

Any tool with optional (recommended) dependencies, on a system where auto-install fails (network issues, permission denied, missing package manager), gets unnecessarily blocked instead of running in degraded/PARTIAL mode.

## Suggested fix

Either maintain separate arrays (`MISSING_REQUIRED_JSONS` / `MISSING_RECOMMENDED_JSONS`), or only iterate over required deps when setting `HAS_REQUIRED_MISSING` in the re-scan loop.

## Affected file

`src/tokenless/adapters/tokenless/common/hooks/tool_ready_hook.sh`, lines 300, 317, 362–368, 403, 422

### LINKED ISSUE #750: [tokenless] bug(tokenless): rewrite_hook does not handle unexpected rtk exit codes
## Summary

`rewrite_hook.py` lines 126–134 only handle rtk exit codes 1 and 2 (with `skip()`). The documented protocol covers codes 0–3, but any other exit code (e.g., 101=Rust panic, 127=binary not found via TOCTOU, 137=OOM kill) falls through and uses `proc.stdout.strip()` as the replacement command.

When rtk crashes, stdout is typically empty. The empty string does not equal the original command, so `skip()` is not called. The hook then replaces the user's valid command with an empty string.

## Reproduction

```python
# Simulate rtk crashing with exit code 127
proc = subprocess.Popen(['/tmp/fake_rtk_exit127.sh', 'rewrite', 'ls -la'],
                        stdout=subprocess.PIPE, stderr=subprocess.PIPE)
stdout, _ = proc.communicate()
# returncode=127, stdout=b''
# rewritten = '' != 'ls -la' → skip() NOT called
# updated_input["command"] = ''  ← user command silently replaced with empty string
```

## Suggested fix

Change line 126 from:
```python
if proc.returncode in (1, 2):
    skip()
```
to:
```python
if proc.returncode not in (0, 3):
    skip()
```

This treats any unexpected exit code as passthrough, matching the documented protocol.

## Affected file

`src/tokenless/adapters/tokenless/common/hooks/rewrite_hook.py`, lines 126–134

### LINKED ISSUE #751: [tokenless] bug(tokenless): Rust check_parent_dir is fail-open, shell _check_parent is fail-closed
## Summary

The Rust trust check `check_parent_dir` in `env_check.rs` (lines 77–88) returns `true` (trusted) when `fs::symlink_metadata` on the parent directory fails. The shell equivalent `_check_parent` in `tool_ready_hook.sh` (lines 55–71) returns `false` (untrusted) in the same scenario.

Both files carry explicit `KEEP IN SYNC` comments referencing each other, confirming this inconsistency is unintentional.

## Code comparison

**Rust (fail-open):**
```rust
fn check_parent_dir(parent: &std::path::Path) -> bool {
    if let Ok(parent_meta) = fs::symlink_metadata(parent) {
        let parent_uid = parent_meta.uid();
        if parent_uid != current_uid() && parent_uid != 0 {
            return false;
        }
        if parent_meta.mode() & 0o002 != 0 {
            return false;
        }
    }
    true  // ← Err case: returns trusted
}
```

**Shell (fail-closed):**
```bash
po=$(stat -c "%u" "$pd" 2>/dev/null || echo "-1")
# stat failure → owner="-1" → fails ownership check → return 1 (untrusted)
```

Note: the file-level `symlink_metadata` check at line 107 of the same Rust file correctly returns `false` on `Err`, showing the developer intended fail-closed semantics.

## Suggested fix

```rust
fn check_parent_dir(parent: &std::path::Path) -> bool {
    let parent_meta = match fs::symlink_metadata(parent) {
        Ok(m) => m,
        Err(_) => return false,  // fail-closed, matching shell behavior
    };
    // ... existing checks ...
}
```

## Affected files

- `src/tokenless/crates/tokenless-cli/src/env_check.rs`, lines 77–88
- `src/tokenless/adapters/tokenless/common/hooks/tool_ready_hook.sh`, lines 55–71 (reference)

jfeng18:
## Summary

- **rewrite_hook.py**: use blocklist (`not in (0, 3)`) instead of allowlist (`in (1, 2)`) for rtk exit codes; add empty-stdout guard and `warn()` on unexpected codes. Prevents replacing user's command with empty string when rtk crashes.
- **tool_ready_hook.sh**: fix post-fix re-scan conflating recommended deps with required deps. A missing recommended dep no longer incorrectly blocks the tool — it correctly enters PARTIAL (degraded) mode.
- **env_check.rs**: change `check_parent_dir` to fail-closed on stat error, matching the shell `_check_parent` behavior. Both files carry `KEEP IN SYNC` comments.

## Verification

| Fix | Test |
|-----|------|
| rewrite_hook | 9/9 exit-code scenarios pass (0,1,2,3,101,127,137,empty stdout,unchanged cmd) |
| tool_ready_hook | E2E: old code → `decision:block` on recommended-only missing; fixed → `PARTIAL` |
| env_check.rs | Rust binary: stat failure on unreachable dir → old returns `true`, fixed returns `false`; normal path: no regression |
| All | `cargo test` 25/25, `cargo check` OK, `bash -n` OK, `ast.parse` OK |
| All | Adversarial workflow review: 7 agents, all APPROVED, 0 blocking issues |

Closes #749, closes #750, closes #751

### DIFF
--- a/src/tokenless/adapters/tokenless/common/hooks/rewrite_hook.py
+++ b/src/tokenless/adapters/tokenless/common/hooks/rewrite_hook.py
@@ -123,8 +123,10 @@ def main() -> None:
     #   3 = Ask/Default verdict (rewrite available but permission model requires
     #       user confirmation; in non-interactive hook context, treat as valid
     #       rewrite since the intent is token optimization, not permission gating)
-    if proc.returncode not in (0, 3):
+    if proc.returncode not in (0, 1, 2, 3):
         warn(f"rtk rewrite exited with unexpected code {proc.returncode}")
+        skip()
+    if proc.returncode in (1, 2):
         skip()
     rewritten = proc.stdout.strip()
     if not rewritten or rewritten == cmd:
--- a/src/tokenless/adapters/tokenless/common/hooks/rewrite_hook.py
+++ b/src/tokenless/adapters/tokenless/common/hooks/rewrite_hook.py
@@ -123,10 +123,11 @@ def main() -> None:
     #   3 = Ask/Default verdict (rewrite available but permission model requires
     #       user confirmation; in non-interactive hook context, treat as valid
     #       rewrite since the intent is token optimization, not permission gating)
-    if proc.returncode in (1, 2):
+    if proc.returncode not in (0, 3):
+        warn(f"rtk rewrite exited with unexpected code {proc.returncode}")
         skip()
     rewritten = proc.stdout.strip()
-    if rewritten == cmd:
+    if not rewritten or rewritten == cmd:
         skip()
 
     # 7. Build response
--- a/src/tokenless/adapters/tokenless/common/hooks/tool_ready_hook.sh
+++ b/src/tokenless/adapters/tokenless/common/hooks/tool_ready_hook.sh
@@ -305,6 +305,8 @@ for i in $(seq 0 $((req_count - 1))); do
   esac
 done
 
+REQUIRED_MISSING_COUNT=$(echo "$MISSING_DEP_JSONS" | jq 'length' 2>/dev/null || echo 0)
+
 # Check recommended deps
 rec_count=$(echo "$RECOMMENDED" | jq 'length')
 missing_count_rec=0
@@ -363,7 +365,9 @@ if [ "$missing_count" -gt 0 ] && [ -n "$FIX_SCRIPT" ] && [ -x "$FIX_SCRIPT" ]; t
         binary=$(echo "$MISSING_DEP_JSONS" | jq -r ".[$i].binary")
         if ! resolve_binary "$binary" >/dev/null 2>&1; then
             STILL_MISSING="${STILL_MISSING} ${binary}"
-            HAS_REQUIRED_MISSING=true
+            if [ "$i" -lt "$REQUIRED_MISSING_COUNT" ]; then
+                HAS_REQUIRED_MISSING=true
+            fi
         fi
     done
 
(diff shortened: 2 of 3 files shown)


shiloong (approved):
LGTM，thx!

PR #755, opened 2026-06-06
fix(sec-core): restore validation error diagnostic in event insert

### LINKED ISSUE #752: [sec-core] bug(sec-core): validation error silently swallowed in SecurityEventRepository.insert()
## Summary

Commit `8eadfe4` removed the `stderr` print from the `except (ValueError, TypeError)` block in `SecurityEventRepository.insert()` without adding any replacement diagnostic (e.g., `logger.warning`).

When `_event_values(event)` raises because event fields are malformed, the exception with the actual cause is now discarded. The caller `SqliteEventWriter.write()` sees `insert()` return `False` and logs a generic `RuntimeError("sqlite write was skipped")` via `_log_drop`, which contains zero information about what validation failed.

## Before (had diagnostic)

```python
except (ValueError, TypeError) as exc:
    print(
        f"[security_events] invalid event params: {exc}",
        file=sys.stderr,
    )
    return False
```

## After (diagnostic lost)

```python
except (ValueError, TypeError):
    return False
```

The caller then creates a synthetic exception with no root cause:
```python
if not inserted:
    self._log_drop(event, RuntimeError("sqlite write was skipped"), "insert")
```

Every other failure path in `SqliteEventWriter.write()` preserves the original exception in `_log_drop`, making this the only path where the root cause is lost.

## Suggested fix

Re-add a diagnostic, ideally via `logging.warning` instead of `print`:
```python
except (ValueError, TypeError) as exc:
    logger.warning("invalid event params: %s", exc)
    return False
```

## Affected file

`src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/security_events/repositories.py`, line 47

jfeng18:
## Summary

- Restore diagnostic logging in `SecurityEventRepository.insert()` except block
- Commit `8eadfe4` removed `stderr` print without adding replacement — the actual `ValueError`/`TypeError` message was silently discarded
- The caller only saw a generic `RuntimeError("sqlite write was skipped")` with no root cause
- Now logs via `logger.warning("invalid event params: %s", exc)` using lazy formatting

## Verification

- `ast.parse`: syntax OK
- Adversarial workflow review: APPROVED, 0 issues
- Not E2E testable — change is diagnostic-only (adds log output on error path, does not alter control flow)

Closes #752

### DIFF
--- a/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/security_events/repositories.py
+++ b/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/security_events/repositories.py
@@ -44,7 +44,8 @@ class SecurityEventRepository:
         """Insert an event. Returns False for invalid or skipped writes."""
         try:
             values = self._event_values(event)
-        except (ValueError, TypeError):
+        except (ValueError, TypeError) as exc:
+            logger.warning("invalid event params: %s", exc)
             return False
         session_factory = self._store.session_factory(raise_on_error=True)
         if session_factory is None:


jfeng18:
Gentle ping @RemindD — one-line fix restoring swallowed validation diagnostics. Let me know if anything needs adjusting.

PR #783, opened 2026-06-08
fix(sight): don't cache empty owner-account-id

jfeng18:
## Summary

Replace the owner-account-id `OnceLock` with a small `CachedUid` (`Mutex<Option<String>>`) that caches **only a non-empty** fetch result. A transient ECS metadata failure returns empty without being cached, so a later call retries instead of permanently serving an empty owner-account-id — which previously poisoned the cache for the whole process and silently dropped the SLS multi-tenant attribution key from every later log. `get_owner_account_id()` now returns `String`.

## Not changed
- instance-id's `OnceLock` (hostname fallback, never empty).
- The config-watcher / startup reaction to a missing uid — whether a running agent should abort vs. retry on a config-change-time uid blip is a separate concern, left for a follow-up (the config watcher now lives in the testable `decide_sls_config_change` structure on `main`, so that change belongs there).

## Tests

`cached_uid_caches_success_only`: a scripted fetcher (fail → uid → fail) asserts the empty result is NOT cached (a recovery re-fetch returns the real uid), the non-empty IS cached (a later call short-circuits even though the fetcher would now return empty), and exactly two fetches happen. Reverting the fix fails the test. Local (toolchain 1.89.0): clang BPF build, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, 677 tests — all green.

### DIFF
--- a/src/agentsight/src/background.rs
+++ b/src/agentsight/src/background.rs
@@ -223,7 +223,7 @@ pub(crate) fn start_config_watcher(
                 let action = handle_config_event(
                     &content,
                     &sls_activated,
-                    || crate::genai::instance_id::get_owner_account_id().to_string(),
+                    crate::genai::instance_id::get_owner_account_id,
                     encryption_pem.as_deref(),
                     trace_enabled,
                     &pending_logtail,
--- a/src/agentsight/src/genai/instance_id.rs
+++ b/src/agentsight/src/genai/instance_id.rs
@@ -3,17 +3,50 @@
 //! Provides a shared function to resolve the current machine's instance ID,
 //! used by both SLS PutLogs uploader and Logtail file exporter.
 //!
-//! Both `get_instance_id()` and `get_owner_account_id()` are cached via
-//! `OnceLock` — the metadata HTTP call is only made once per process lifetime.
+//! `get_instance_id()` is cached via `OnceLock` (it always resolves to a
+//! non-empty value thanks to the hostname fallback). `get_owner_account_id()`
+//! uses a `CachedUid` that caches ONLY a successful (non-empty) fetch: a
+//! transient metadata failure returns empty without being cached, so a later
+//! call retries instead of permanently serving an empty owner-account-id.
 
+use std::sync::Mutex;
 use std::sync::OnceLock;
 use std::time::Duration;
 
 /// ECS metadata 请求超时（连接 + 读取均为 1 秒）
 const METADATA_TIMEOUT: Duration = Duration::from_secs(1);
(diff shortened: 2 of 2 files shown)


jfeng18:
Gentle ping @chengshuyi — this is a small, focused bug fix (minimal diff). Happy to address any feedback. No rush.

jfeng18:
**CI note**: diff-cover flags 3/4 lines at 25% — the uncovered lines are inside the config-watcher thread closure (retry logic on uid fetch failure), which is unreachable by unit tests. The single covered line is in `instance_id.rs` (the new `fetch_owner_account_id` function).

This is a 3-line behavior change: `process::exit(1)` → `continue` (retry on next config change). Verified via ECS E2E.

chengshuyi:
## Review Comment

整体 LGTM，修复了真实的生产问题，范围控制得当，测试充分。一个建议：

### 建议：增加重试上限，避免非 ECS 环境反复超时

当前实现中，如果 metadata 永久不可达（非 ECS 环境、metadata 服务宕机等），每次调用 `get_owner_account_id()` 都会发 HTTP 请求并等待 1 秒超时后返回空。在 `logtail.rs:247` 路径上（每次 LLM 调用导出都会触发），这会导致每次 export 多等 1 秒。

建议加一个失败计数器，连续 N 次失败后停止重试：

```rust
struct CachedUid {
    cell: Mutex<Option<String>>,
    failures: AtomicUsize,
}

const MAX_RETRIES: usize = 3;

fn get_or_fetch(&self, fetch: impl FnOnce() -> String) -> String {
    let mut guard = self.cell.lock().unwrap_or_else(|e| e.into_inner());
    if let Some(v) = guard.as_ref() {
        return v.clone();
    }
    if self.failures.load(Ordering::Relaxed) >= MAX_RETRIES {
        return String::new();
    }
    let v = fetch();
    if !v.is_empty() {
        *guard = Some(v.clone());
    } else {
        self.failures.fetch_add(1, Ordering::Relaxed);
    }
    v
}
```

这样连续失败 3 次后认定 metadata 不可用，不再重试，避免热路径上反复卡 1 秒超时。

### Nit

`background.rs:226` 中 `.to_string()` 多余（返回值已是 `String`），可顺手移除。

chengshuyi (approved):

jfeng18:
Closing: the two commits in this PR (`don't cache empty owner-account-id` + `cap ECS metadata UID fetch retries`) were brought into main via #1029 (merged 2026-06-23). No additional changes needed.

PR #789, opened 2026-06-08
fix(sight): extract token usage from non-streaming and H2 responses

### LINKED ISSUE #788: [sight] bug(sight): genai_events records token=0 for non-streaming LLM calls despite valid usage in API response
## Summary

`genai_events.db` records `input_tokens=0, output_tokens=0` for non-streaming LLM API calls (both HTTP/1.1 and HTTP/2), even when the API response includes valid `usage` data. SSE streaming calls without `stream_options.include_usage=true` also show `token=0` because the tokenizer fallback (`compute_tokens_manually`) does not support HTTP/2 streams.

## Root Cause

`src/agentsight/src/analyzer/unified.rs`, token extraction match (~line 435):

```rust
let mut token_result = match result {
    AggregatedResult::SseComplete(pair) => { ... }        // SSE — handled ✅
    AggregatedResult::ResponseOnly { .. } => { ... }       // orphan SSE — handled ✅
    _ => None,  // HttpComplete AND Http2StreamComplete land here ❌
};
```

Both `HttpComplete` (HTTP/1.1) and `Http2StreamComplete` (HTTP/2) fall through to `_ => None`. The existing `TokenParser::parse_json()` correctly handles `{"usage":{"prompt_tokens":N}}` — it is simply never called for these variants.

Additionally, `compute_tokens_manually()` (~line 636) only matches `SseComplete`, so the tokenizer fallback also does not work for HTTP/2 SSE streams without usage data.

Three gaps total:
1. `HttpComplete` — no token extraction from response JSON body
2. `Http2StreamComplete` — no token extraction from response JSON body
3. `compute_tokens_manually` — no tokenizer fallback for `Http2StreamComplete`

## How to Reproduce on ECS

```bash
# Start agentsight, inject test PID into traced_processes map, then:
curl -s https://dashscope.aliyuncs.com/compatible-mode/v1/chat/completions \
  -H "Authorization: Bearer $API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"model":"qwen-plus","messages":[{"role":"user","content":"hi"}],"max_tokens":3}'

# API returns: "usage":{"prompt_tokens":9,"completion_tokens":3,"total_tokens":12}
# genai_events.db shows: input_tokens=0, output_tokens=0, total_tokens=0
```

## Reproduced

Tested on ECS (kernel 6.6.102+) with DashScope qwen-plus. Confirmed token=0 for both HTTP/2 non-streaming and HTTP/1.1 non-streaming calls (--http1.1). Also confirmed SSE without include_usage shows token=0 due to missing tokenizer fallback for H2.

jfeng18:
## Summary

- Add `HttpComplete` and `Http2StreamComplete` token extraction from JSON response body
- Extend `compute_tokens_manually` to support `Http2StreamComplete` for tokenizer fallback
- New helper `extract_token_from_json_body` to avoid code duplication

Fixes #788.

## What changed

| Location | Change |
|----------|--------|
| `unified.rs:443-456` | Add `HttpComplete` + `Http2StreamComplete` match arms to token extraction |
| `unified.rs:605-628` | New `extract_token_from_json_body` helper |
| `unified.rs:636-680` | `compute_tokens_manually` now handles `Http2StreamComplete` via `SseData` enum |
| `unified.rs:744-755` | Output token counting unified for `ParsedSseEvent` and `Vec<Value>` |

## What's NOT changed

- `SseComplete` path — untouched, no regression
- `TokenParser::parse_json` — already correct, just not called before
- `GenAIBuilder` / storage layer — no changes needed
- No new dependencies

## Root Cause

Token extraction match only handled `SseComplete` and `ResponseOnly`. Both `HttpComplete` (HTTP/1.1) and `Http2StreamComplete` (HTTP/2) fell to `_ => None`. `TokenParser::parse_json()` correctly parses `{"usage":{"prompt_tokens":N}}` — it was simply never called for these variants.

## Verification (6 scenarios on ECS, kernel 6.6.102+)

| Scenario | Before | After |
|----------|--------|-------|
| Non-streaming HTTP/2 | 0/0/0 | **9/3/12** (exact from API) |
| Non-streaming HTTP/1.1 | 0/0/0 | **10/1/11** (exact) |
| SSE H2 with include_usage | 10/1/11 | 10/1/11 (no regression) |
| SSE H1 with include_usage | — | 10/3/13 (no regression) |
| SSE H2 without include_usage | 0/0/0 | **31/1/32** (tokenizer estimate) |
| 401 error | 0/0/0 | 30/0/30 (input estimate, no crash) |

### Discriminating signal

```
# Before: API returns usage:{prompt_tokens:9} but DB has 0
genai_events: qwen-plus|0|0|0

# After: DB matches API response
genai_events: qwen-plus|9|3|12
token_records: qwen-plus|openai|9|3
```

### DIFF
--- a/src/agentsight/src/analyzer/unified.rs
+++ b/src/agentsight/src/analyzer/unified.rs
@@ -445,7 +445,7 @@ impl Analyzer {
             return results;
         }
 
-        // 2. Token analysis - extract from SSE events
+        // 2. Token analysis - extract from SSE events or non-streaming JSON body
         let mut token_result = match result {
             AggregatedResult::SseComplete(pair) => {
                 let pid = pair.request.source_event.pid;
@@ -456,6 +456,20 @@ impl Analyzer {
                 let pid = response.pid();
                 let comm = response.parsed.source_event.comm_str();
                 self.extract_token_from_sse(&response.sse_events, pid, &comm)
+            }
+            AggregatedResult::HttpComplete(pair) => {
+                let pid = pair.request.source_event.pid;
+                let comm = pair.request.source_event.comm_str();
+                self.extract_token_from_json_body(
+                    pair.response.parsed.json_body().as_ref(),
+                    pid,
+                    &comm,
+                )
+            }
+            AggregatedResult::Http2StreamComplete(stream) => {
+                let pid = stream.pid();
+                let comm = stream.comm();
+                self.extract_token_from_json_body(stream.response_json_body().as_ref(), pid, &comm)
             }
             _ => None,
         };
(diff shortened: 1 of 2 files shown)


jfeng18:
**CI note**: diff-cover reports 38% (45/73 lines uncovered). Breakdown:

- `extract_token_from_json_body` (lines 632-649): new method, needs `Analyzer` + `TokenParser` — can add unit tests
- `compute_tokens_manually` Http2 branch (lines 663-690) + SSE chunk iteration (785-799): requires global tokenizer initialization, harder to unit-test
- `HttpComplete`/`Http2StreamComplete` match arms (lines 469-472): inside `analyze_aggregated`, needs full pipeline

This PR predates the diff-cover gate (#840). Will add unit tests for `extract_token_from_json_body` to improve coverage. The tokenizer-dependent paths are structurally similar to the existing `SseComplete` path which is also uncovered.

chengshuyi (approved):
LGTM — 根因准确，非流式 HttpComplete/Http2StreamComplete 的 usage 字段之前从未被提取。extract_token_from_json_body 实现干净，测试覆盖全面（含 pipeline 集成测试和 6 场景 E2E）。

PR #791, opened 2026-06-08
fix(sec-core): degrade prompt scan to L1 when ML model not downloaded

### LINKED ISSUE #790: [sec-core] bug(sec-core): prompt_scan fails on every scan when ML model not downloaded (no graceful degradation)
## Summary

On ECS, every `prompt_scan` event fails (`result=failed`) because the Llama-Prompt-Guard-2 model was never downloaded, yet `MLClassifier` is treated as a mandatory, always-available detector. The scanner then raises at scan time on every prompt, leaving prompts effectively unscanned.

`agent-sec-cli events --event-type prompt_scan` on a fresh ECS shows all events `failed` with `error_type=ModelLoadError`.

## Root Cause

`MLClassifier` (L2) does **not** override `is_available()`; it inherits `DetectionLayer.is_available()` which unconditionally returns `True` (`detectors/base.py`). So the scanner believes L2 is available even when the model files are absent. At scan time, `detect()` → `classify()` → `ModelManager._resolve_local_model_path()` raises `ModelLoadError` ("Model not available locally, run scan-prompt warmup").

Because `ml_classifier` is not in `_OPTIONAL_DETECTORS` (`scanner.py`), there is no graceful degradation — the whole scan errors out instead of falling back to L1 (regex).

Note: torch/transformers ARE installed; only the model files are missing. This is distinct from #698/#680 (which change the cosh hook's fail-open→fail-ask behavior but do not fix the underlying model-availability gap).

## How to Reproduce on ECS

```bash
# Ensure model is not downloaded
rm -rf ~/.cache/prompt_scanner/models

# Standard mode (L1+L2) on any prompt
agent-sec-cli scan-prompt --mode standard --format text --text "test"

# Result: verdict=error, "Detector 'ml_classifier' is not available" / ModelLoadError
# Every prompt_scan event recorded as failed.
```

## Suggested Fix

Two coupled changes (must land together):

1. **`MLClassifier.is_available()` override** — return False when torch/transformers missing OR the model is not downloaded (reuse a new `ModelManager.is_model_downloaded()` predicate), mirroring `SemanticDetector`'s override pattern.

2. **Add `ml_classifier` to `_OPTIONAL_DETECTORS`** — so when L2 is unavailable, the scanner skips it and degrades to L1 (regex) with a warning, instead of erroring on every scan.

Why coupled: changing only #1 makes things worse — `is_available()=False` on a mandatory detector raises `LayerNotAvailableError` in the constructor (eager crash) instead of the current lazy scan-time error.

Also: `PromptScanner.warmup()` must bypass the `is_available()` gate (build detectors directly from config), otherwise warmup can never download a model that is currently unavailable (chicken-and-egg).

## Impact

Until fixed, prompt injection scanning is non-functional on any host where `scan-prompt warmup` was never run — which is the default after install (no install-time or daemon auto-warmup exists). The security layer silently does nothing.

jfeng18:
## Summary

Fixes prompt injection scanning being non-functional on any host where the ML model was never downloaded (the default after install). Previously every `prompt_scan` errored with `ModelLoadError` instead of degrading to L1.

Fixes #790.

## Root Cause

`MLClassifier` inherited `is_available()->True`, so the scanner treated L2 (ML) as mandatory-and-available even when the Llama-Prompt-Guard-2 model was absent. At scan time `detect()` raised `ModelLoadError`, and because `ml_classifier` was not in `_OPTIONAL_DETECTORS`, the whole scan errored — no fallback to L1 (regex).

(torch/transformers ARE installed; only the model files were missing. Distinct from #680, which changes the cosh hook's fail-open→fail-ask behavior but not this model-availability gap.)

## Changes

| File | Change |
|------|--------|
| `detectors/ml_classifier.py` | Add `is_available()` override: probe torch/transformers importability + model presence |
| `models/model_manager.py` | New `is_model_downloaded()` predicate (reused by `_resolve_local_model_path`) |
| `scanner.py` | Add `ml_classifier` to `_OPTIONAL_DETECTORS` → skip to L1 when unavailable |
| `scanner.py` | `warmup()` bypasses the `is_available()` gate (builds detectors from config) so it can download a currently-unavailable model |

**Coupling note:** the `is_available()` override and the `_OPTIONAL_DETECTORS` change must land together — changing only the former makes a False detector raise `LayerNotAvailableError` in the constructor (eager crash) instead of degrading.

## What's NOT changed

- cosh hook / `prompt_scanner_hook.py` — that's #680's domain
- Daemon preload (#786) — separate auto-warmup path
- No install-time auto-warmup added (out of scope; tracked separately)
- Pre-existing isort disorder in `model_manager.py` (pydantic import) left untouched — exists on main, not in scope

## Verification (ECS, kernel 6.6.102+, Python 3.11)

| Scenario | Before | After |
|----------|--------|-------|
| Model absent, `--mode standard` | `verdict=error`, ModelLoadError | L1 fallback, warns + skips L2, no error |
| `scan-prompt warmup` | silently no-op (chicken-and-egg) | downloads model (1.04G) |
| Model present, injection text | n/a | DENY, jailbreak, 99.9% confidence |
| Model present, benign text | n/a | PASS |
| Unit tests | — | 287 passed |
| black 26.3.1 (CI version) | — | clean |

### Discriminating signal
```
# Before (ml_classifier mandatory): verdict=error "Detector 'ml_classifier' is not available"
# After: PASS via L1, log "Detector 'ml_classifier' ... will be skipped"
```

### DIFF
--- a/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/prompt_scanner/detectors/ml_classifier.py
+++ b/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/prompt_scanner/detectors/ml_classifier.py
@@ -52,13 +52,19 @@ class MLClassifier(DetectionLayer):
         return "ml_classifier"
 
     def is_available(self) -> bool:
-        """Return True only if torch/transformers are importable AND the
-        model is downloaded locally.
+        """Return True only if torch/transformers are importable AND the model
+        is loadable — either already resident in RAM or downloaded on disk.
 
         Returns False (instead of raising) when dependencies are missing or
-        the model has not been fetched yet (e.g. ``scan-prompt warmup`` never
-        ran), so the scanner can skip L2 and degrade to L1 rather than
-        failing every scan.
+        the model cannot be loaded (e.g. ``scan-prompt warmup`` never ran), so
+        the scanner can skip L2 and degrade to L1 rather than failing every
+        scan.
+
+        Availability checks RAM residency *before* disk: ``load_model`` serves
+        a resident model from its RAM fast path even if the on-disk cache was
+        evicted afterwards, so gating on disk alone would under-report and
+        silently drop L2 while a scan would in fact have run L2 — diverging
+        from the daemon's in-memory ``degraded`` signal.
         """
         import importlib.util
 
@@ -68,8 +74,10 @@ class MLClassifier(DetectionLayer):
         ):
             return False
 
-        return self._classifier._manager.is_model_downloaded(
-            self._classifier._model_name
+        manager = self._classifier._manager
+        model_name = self._classifier._model_name
+        return manager.is_model_loaded(model_name) or manager.is_model_downloaded(
+            model_name
         )
 
     def warmup(self) -> None:
--- a/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/prompt_scanner/detectors/ml_classifier.py
+++ b/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/prompt_scanner/detectors/ml_classifier.py
@@ -51,6 +51,27 @@ class MLClassifier(DetectionLayer):
     def name(self) -> str:
         return "ml_classifier"
 
+    def is_available(self) -> bool:
+        """Return True only if torch/transformers are importable AND the
+        model is downloaded locally.
+
+        Returns False (instead of raising) when dependencies are missing or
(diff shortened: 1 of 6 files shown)


jfeng18:
Hi @haosanzi, this PR degrades to L1 (regex) scanning when the ML model is unavailable, instead of skipping the scan entirely. As the prompt scan module owner, could you take a look when you have time?

PR #801, opened 2026-06-09
fix(cosh): report API errors in all output formats, not only text

jfeng18:
## What

In non-interactive mode (`cosh -p`), API error events (401, 429, etc.) were silently swallowed in JSON and STREAM_JSON output formats — `isError: false`, exit 0, empty stderr. Programmatic callers had zero indication of failure.

## Root Cause

`nonInteractiveCli.ts:284` gated error detection on `outputFormat === OutputFormat.TEXT`. In JSON/STREAM_JSON modes, the `GeminiEventType.Error` event was consumed by `adapter.processEvent()` as normal text content, the loop finished normally, and `emitResult({isError: false})` was called.

## Fix

Remove the `outputFormat === OutputFormat.TEXT &&` condition (one line). Error events now throw in all output formats. The existing catch block already correctly calls `adapter.emitResult({isError: true, errorMessage})` for all formats — it just was never reached in JSON/STREAM_JSON mode.

TEXT mode behavior is unchanged (still writes to stderr + exit 1, as before).

## Testing

- ✅ 30 existing tests pass (TEXT mode behavior unchanged)
- ✅ 2 new discriminating tests added (JSON + STREAM_JSON): verify error propagates + stderr receives message
- ✅ prettier + eslint clean
- ✅ 6-agent adversarial review: 5 findings → 0 confirmed (all refuted)

## Impact

| Mode | Before | After |
|------|--------|-------|
| TEXT | stderr + exit 1 | unchanged |
| JSON | stderr empty, `isError: false`, exit 0 | stderr + `isError: true`, exit 1 |
| STREAM_JSON | stderr empty, `isError: false`, exit 0 | stderr + `isError: true`, exit 1 |

### DIFF
--- a/src/copilot-shell/packages/cli/src/nonInteractiveCli.test.ts
+++ b/src/copilot-shell/packages/cli/src/nonInteractiveCli.test.ts
@@ -899,6 +899,74 @@ describe('runNonInteractive', () => {
     expect(processStderrSpy).toHaveBeenCalled();
   });
 
+  it('should not leave an unterminated assistant message on API error in stream-json with partial messages', async () => {
+    // Regression for the #801 review: with stream-json + includePartialMessages,
+    // an API error must neither be fed to the adapter (which would emit fresh
+    // partial blocks) nor skip finalize on the partial message already streamed
+    // this turn — either leaves consumers with an open content_block / missing
+    // message_stop right before the error result.
+    (mockConfig.getOutputFormat as Mock).mockReturnValue(
+      OutputFormat.STREAM_JSON,
+    );
+    (mockConfig.getIncludePartialMessages as Mock).mockReturnValue(true);
+    setupMetricsMock();
+
+    const events: ServerGeminiStreamEvent[] = [
+      // Assistant text streams first (opens message_start + a content block)...
+      { type: GeminiEventType.Content, value: 'Partial answer before failure' },
+      // ...then the API errors mid-stream.
+      {
+        type: GeminiEventType.Error,
+        value: { error: { message: '429 rate limit exceeded', status: 429 } },
+      },
+    ];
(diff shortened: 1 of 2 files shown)


kongche-jbw:
Thanks for fixing the silent error path here. I think there is one edge case worth addressing before we merge.

With `--output-format stream-json --include-partial-messages`, the new path still calls `adapter.processEvent(event)` before checking `GeminiEventType.Error`. For an API error event, `StreamJsonOutputAdapter` can therefore emit partial assistant events such as `message_start`, `content_block_start`, and `content_block_delta`; then the throw skips `adapter.finalizeAssistantMessage()`, so the stream never emits the matching `content_block_stop` / `message_stop` events.

That leaves event-stream consumers with an unclosed assistant message immediately before the final error result. A small fix might be to handle `GeminiEventType.Error` before passing it to the adapter, or otherwise ensure the partial message is finalized before throwing. It would also be useful to add a regression test for the `stream-json + includePartialMessages=true + API error` case.

jfeng18:
Thanks @kongche-jbw, good catch — fixed in `a87ab4c`.

The `GeminiEventType.Error` event is now handled **before** `adapter.processEvent()`, so it no longer emits partial blocks of its own, and `finalizeAssistantMessage()` is called **before** the `throw`, so any blocks already streamed this turn get their `content_block_stop` / `message_stop`. `finalizeAssistantMessage()` is idempotent and emits nothing when no message was started, so the error-first case stays clean as well.

Added a regression test (`stream-json` + `includePartialMessages` + content-then-error) asserting the emitted stream events are balanced (every `content_block_start` / `message_start` has its matching stop); reverting the fix fails it (`content_block_stop` = 0).

kongche-jbw (approved):
lgtm thx

PR #804, opened 2026-06-09
feat(sight): add ReactiveExporter for observe→act pipeline

jfeng18:
## What

The first piece of **dynamic orchestration** in ANOLISA: agentsight can now **react** to observed events, not just record them.

`ReactiveExporter` is a new `GenAIExporter` that inspects each LLM call event for critical signals and triggers actions:

- **Critical interruption** (crash / OOM / SIGKILL in error field) → spawn `ws-ckpt checkpoint` to save workspace state
- **Token waste advisory** (>50K input tokens with zero prompt caching) → log a recommendation

## Why

From the north-star ("max power inference + dynamic workflow orchestration") gap analysis: agentsight scores 2/10 on "dynamic orchestration" because modules only observe — they never act. This PR is the 0→1: the first time a module's observation triggers another module's action.

## Design

- **Non-blocking**: `export()` does `try_send` on a bounded channel (cap 32). Background thread runs `ws-ckpt`.
- **Timeout-protected**: `try_wait` poll loop with 10s deadline + kill. A stuck ws-ckpt never blocks the thread or Drop.
- **Debounced**: at most 1 checkpoint per interval (default 30s). Prevents storm cascades.
- **Graceful**: ws-ckpt not installed → `new()` returns None → not registered (zero cost).
- **Default disabled**: needs explicit config. Zero impact on existing deployments.

## Config

```json
{ "reactive": { "enabled": true, "debounce_secs": 30, "workspace": "/root" } }
```

## Testing

- ✅ 8 unit tests (detection, advisory, disabled, integration: ws-ckpt spawn + timeout + debounce + Drop)
- ✅ 984 full regression
- ✅ ECS E2E: registers in trace mode, zero false positives on normal traffic
- ✅ ECS integration test: crash event → spawn → timeout → kill → debounce → Drop clean (13.2s)
- ✅ Adversarial review (7 agents): 1 confirmed finding (timeout hang) → fixed

- ✅ diff-cover 82% (gate 80%): extracted checkpoint decision logic into pure functions for testability

### DIFF
--- a/src/agentsight/src/config.rs
+++ b/src/agentsight/src/config.rs
@@ -236,6 +236,8 @@ struct JsonFullConfig {
     #[serde(default)]
     deadloop: Option<JsonDeadloop>,
     #[serde(default)]
+    reactive: Option<JsonReactive>,
+    #[serde(default)]
     cgroup_filter_enabled: Option<bool>,
     #[serde(default)]
     cgroup_ids: Option<Vec<u64>>,
@@ -747,6 +749,14 @@ pub struct AgentsightConfig {
     // --- Runtime Resource Limits ---
     /// Bounded channel capacities, pending queue limits, etc.
     pub runtime_limits: RuntimeLimits,
+
+    // --- Reactive Exporter Configuration ---
+    /// Enable the reactive exporter (observe→act pipeline)
+    pub reactive_enabled: Option<bool>,
+    /// Minimum seconds between consecutive checkpoint actions
+    pub reactive_debounce_secs: Option<u64>,
+    /// Workspace path for ws-ckpt checkpoint
+    pub reactive_workspace: Option<String>,
 }
 
 impl Default for AgentsightConfig {
@@ -821,6 +831,11 @@ impl Default for AgentsightConfig {
 
             // Runtime resource limits
             runtime_limits: RuntimeLimits::default(),
+
+            // Reactive exporter defaults (disabled by default)
+            reactive_enabled: None,
+            reactive_debounce_secs: None,
+            reactive_workspace: None,
         }
     }
 }
@@ -959,6 +974,13 @@ impl AgentsightConfig {
             if let Some(count) = dl.kill_after_count {
                 self.deadloop_kill_after_count = count;
             }
+        }
+
+        // 解析 reactive exporter 配置
+        if let Some(ref r) = parsed.reactive {
+            self.reactive_enabled = r.enabled;
+            self.reactive_debounce_secs = r.debounce_secs;
+            self.reactive_workspace = r.workspace.clone();
         }
 
         // Parse cgroup filter settings
--- a/src/agentsight/src/config.rs
+++ b/src/agentsight/src/config.rs
@@ -350,6 +350,17 @@ pub struct JsonRuntime {
     pub sls_logtail_path: Option<String>,
 }
 
+/// Reactive exporter configuration section
+#[derive(serde::Deserialize, Clone, Debug)]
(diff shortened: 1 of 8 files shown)


jfeng18:
## E2E verification: full signal chain on live ECS

Triggered a real RetryStorm (6 identical invalid API calls → 5+ auth_error in one conversation) with the reactive exporter enabled and ws-ckpt daemon running:

```
--- interruption_events ---
auth_error|b59780476b41bbe7f6e4461c7199e096
retry_storm|b59780476b41bbe7f6e4461c7199e096

--- ws-ckpt snapshots ---
WORKSPACE            SNAPSHOT                         CREATED             MESSAGE
---------------------------------------------------------------------------------
/root/test-workspace auto-20260609T140312-retry_storm 2026-06-09 14:03:12 reactive: retry_storm (conv=b59780476b41bbe7f6e4461c7199e096)

--- reactive log ---
[2026-06-09T14:03:12Z INFO  agentsight::genai::reactive] [reactive] checkpoint created: auto-20260609T140312-retry_storm
```

Full chain verified: eBPF capture → interruption detection → RetryStorm insert → notify_interruption → background thread → ws-ckpt spawn → **real btrfs snapshot created on disk**.

jfeng18:
Closing this. After a deep re-review (including reading ws-ckpt's own design docs), the issue isn't the wiring — it's the direction of this specific action.

**Why the direction doesn't hold (verified against the code + ws-ckpt design):**

1. **Reactive checkpoint-on-failure is temporally backwards for recovery.** ws-ckpt's recovery model is *proactive per-turn* snapshots (`ws-ckpt/docs/ws-ckpt-design.md:12`, `plugin-design.md:11,52`: a snapshot at every agent turn boundary, so a known-good *pre-incident* state always exists). This ReactiveExporter checkpoints *when a failure is observed* — by which point the good state is already gone; it captures post-incident/damaged state.
2. **That recovery capability already exists, in the right place.** The per-turn auto-checkpoint is done harness-side by the ws-ckpt openclaw/hermes plugins, which know turn boundaries and own the workspace. A reactive checkpoint in the observability layer is a worse, mistimed duplicate.
3. **Timing collides with the effector.** For live-agent loop paths (retry_storm/dead_loop), ws-ckpt rejects checkpoints under active writes (`WriteLockConflict`) — exactly when this fires. The plugin avoids that by running at quiescent turn boundaries.

Also confirmed as-shipped: the `agent_crash` trigger keys on `call.error` (`reactive.rs:306`), which is only populated from HTTP-4xx LLM bodies (`call_builder.rs:144`) — a real crash never produces that string, and the same PR's real OOM detector (`was_pid_oom_killed` → `AgentCrash` → store) doesn't call the effector. And the checkpoints have no consumer (no rollback reader anywhere). So the loop is open at both ends.

**What's still valid and carries forward:**
- The ambition (observe→act closed loop) is right — it's the north-star direction.
- The engineering harness (bounded channel, background thread, 10s timeout+kill, 30s debounce, default-off, graceful ws-ckpt-absent) is reusable.
- The signals (crash / retry_storm / dead_loop / cost) are valuable — as *emitted signals for consumers*, not for self-triggering a checkpoint.
- A valid future observe→act for agentsight would keep recovery proactive/harness-side, and have agentsight's reaction be *notify / gate / kill* (which it already partially does via `/restart` + sec-core) rather than checkpoint.

Closing rather than parking so it doesn't read as a merge-ready "observe→act" capability it isn't. Reopenable if we revisit; the harness is worth salvaging for the right shape.

PR #662, opened 2026-05-29
feat(sight): agent activity monitor via schedmon BPF

jfeng18:
## Summary

Replaces the cgroup-based idle-burst-idle scheduler with a **pure observability** activity monitor. The BPF scheduling probes are retained; the cgroup actuation layer is removed.

### Why the change

Reviewer feedback + kernel code analysis identified fundamental flaws in the cpu.idle toggle approach:

1. **Architectural race**: userspace round-trip (BPF event → daemon → cgroup write) is slower than the kernel's first scheduling decision after wakeup
2. **Active anti-pattern**: `cpu.idle=1` sets weight to `WEIGHT_IDLEPRIO=3` (kernel/sched/fair.c:14141), making waking tasks unable to preempt non-idle tasks (line 9087) until userspace restores `cpu.idle=0`
3. **No-op when idle**: setting cpu.idle on sleeping tasks has no scheduling effect (nothing on runqueue)

Correct approach for CPU acceleration/high-density: static `cpu.weight` in container spec, or future `sched_ext` BPF scheduler (in-kernel, zero round-trip). ECS already has `CONFIG_SCHED_CLASS_EXT=y`.

### What's kept

- **schedmon BPF probes** (`tp_btf/sched_switch` + `sched_wakeup`): per-tid sleep/wakeup detection
- **Activity state machine**: per-family idle/active tracking with debounce
- **Metrics**: idle_to_active_count, active_to_idle_count, duration tracking
- **All unit tests** (state machine correctness)

### What's removed

- All cgroup operations (create/remove/write cpu.idle/cpu.weight/migrate pid)
- `active_weight`, `cgroup_root` config fields
- Drop/cleanup/sweep logic
- README zero-intrusion scoping docs (no longer needed — purely observational)

**Stacked on #661** (lineage tree). Review top commits only.

## Test plan

- [x] 466 unit tests pass, 9 activity monitor tests
- [x] Workflow adversarial review: PASS
- [ ] ECS E2E: verify idle/active log transitions with real agent workload
- [ ] Reviewer sign-off

PR #822, opened 2026-06-10
feat(sight): add agent activity monitor via schedmon BPF

jfeng18:
Supersedes #662.

## Summary

- Replace cgroup-based idle-burst scheduler with observation-only activity monitor
- Add `schedmon` BPF probe (`tp_btf/sched_switch` + `sched_wakeup`) for per-thread sleep/wakeup tracking
- Add `ActivityMonitor` state machine with per-family idle/active tracking and configurable debounce
- No cgroup operations, no `cpu.idle` toggle — CPU policy belongs in container spec

**Why**: The original design (#662) set `cpu.idle=1` on idle agents, which drops CFS weight to 3 (vs 1024 normal). On API response wakeup, the thread cannot preempt any non-idle task and may starve 10+ms — the opposite of the intended acceleration.

## Changed files

| File | What |
|------|------|
| `schedmon.bpf.c` / `schedmon.h` | New BPF tracepoint programs |
| `probes/schedmon.rs` | Rust wrapper with shared map reuse |
| `scheduler/mod.rs` | ActivityMonitor (337 lines, 9 tests) |
| `probes/probes.rs` | SchedMon integration + ring buffer dispatch |
| `unified.rs` | Event handling + process lifecycle wiring |
| `config.rs` | `activity_monitor.enabled` + `idle_threshold_ms` |
| `event.rs` / `build.rs` / `lib.rs` / `probes/mod.rs` / `parser/unified.rs` | Plumbing |

## Test plan

- [x] Unit tests: 553 passed, 0 failed (full suite)
- [x] Scheduler tests: 9/9 passed (state machine)
- [x] Schedmon tests: 3/3 passed (probe wrapper)
- [x] ECS E2E: BPF probe loads, Sched events flow through ring buffer to ActivityMonitor
- [x] ECS E2E: Cosh agent auto-discovered, Sched events received for traced process
- [x] Adversarial workflow review: 0 blocking issues

### DIFF
--- a/src/agentsight/build.rs
+++ b/src/agentsight/build.rs
@@ -61,6 +61,10 @@ fn main() {
 
     // Generate tcpsniff skeleton (no header — reuses sslsniff.h event format)
     generate_skeleton(&mut out, "tcpsniff");
+
+    // Generate schedmon skeleton and bindings
+    generate_skeleton(&mut out, "schedmon");
+    generate_header(&mut out, "schedmon");
     
     // generate_header(&mut out, "frametypes");
     // generate_header(&mut out, "errors");
--- a/src/agentsight/src/bpf/common.h
+++ b/src/agentsight/src/bpf/common.h
@@ -23,6 +23,7 @@ typedef enum {
     EVENT_SOURCE_FILEWATCH = 4, // File watch events (filewatch)
     EVENT_SOURCE_FILEWRITE = 5, // File write events (filewrite)
     EVENT_SOURCE_UDPDNS = 6,   // UDP DNS query events (udpdns)
+    EVENT_SOURCE_SCHED  = 7,   // Scheduler monitor events (schedmon)
 } event_source_t;
 
 // Common event header - every ringbuffer event MUST start with this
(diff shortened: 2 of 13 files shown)


jfeng18:
Supersedes #662 — reworked from cgroup-based scheduler to observation-only activity monitor per review feedback on CFS weight starvation.

jfeng18:
Closing: the downstream consumer (idle-burst scheduler #662 → ReactiveExporter #804) was killed. The observation dimension (CPU activity state) is real, but BPF per-sched_switch is over-engineered — /proc/{pid}/stat polling at 1-20Hz provides the same Active/Idle binary signal the 50ms hysteresis threshold implies is sufficient.

If CPU activity state is needed in the future (e.g. for DeadLoop verify arm or utilization dashboard), start with the cheap /proc approach and only escalate to BPF if polling proves insufficient.

Design decisions preserved in git history: preemption filtering (voluntary sleep only), per-family multi-thread semantics, LRU dedup, 50ms hysteresis, BTF availability check.

PR #830, opened 2026-06-10
fix(sight): replace hand-rolled new_id with uuid v4

jfeng18:
## Summary
- Replace `new_id()` (timestamp XOR stack address, zero entropy) with `uuid::Uuid::new_v4()` (122-bit CSPRNG)
- Add `assert_ne!(id1, id2)` to `test_new_id_uniqueness` (previously only checked format, never asserted uniqueness)
- Fix `rustfmt` on `interruption/types.rs`

## Details

The previous `new_id()` generated 32-char hex IDs using nanosecond timestamp XOR'd with a stack variable address. This provided zero actual entropy — the stack address is deterministic within the same thread/call depth, and the second half was the raw timestamp. The doc comment on `InterruptionEvent.interruption_id` already claimed "UUID v4 hex, 32 chars" but the implementation was not UUID v4.

`uuid` is already a transitive dependency (via moka, debugid, llm-tokenizer) so this adds zero compile cost — only activates the `v4` feature for CSPRNG-based ID generation.

## Test plan
- [x] Local 541/541 tests pass
- [x] ECS 541/541 tests pass
- [x] `test_new_id_uniqueness` now asserts `id1 != id2`
- [x] Output format unchanged (32 lowercase hex chars)
- [x] `rustfmt` clean on `interruption/types.rs`

### DIFF
--- a/src/agentsight/Cargo.lock
+++ b/src/agentsight/Cargo.lock
@@ -269,6 +269,7 @@ dependencies = [
  "typed-arena",
  "uname",
  "ureq 2.12.1",
+ "uuid",
 ]
 
 [[package]]
--- a/src/agentsight/Cargo.toml
+++ b/src/agentsight/Cargo.toml
@@ -68,6 +68,7 @@ symbolic-demangle = { version = "12.10.0", default-features = false, features = 
 thiserror = "1.0.63"
 typed-arena = "2.0.2"
 uname = "0.1.1"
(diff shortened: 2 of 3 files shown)


chengshuyi (approved):
LGTM — 零熵的手写 ID 替换为 uuid v4（CSPRNG），uuid 已是传递依赖无额外开销，测试补了唯一性断言。

PR #852, opened 2026-06-11
feat(anolisa): add raw backend install and component lifecycle plumbing

ikunkun-sys:
## Description

  This PR adds the first component-first install/uninstall path for ANOLISA.

  - Add `repo.toml` backend configuration loading, backend selection, package-name mapping, and raw backend URL
  resolution.
  - Implement raw backend install: fetch distribution index, resolve version/target, derive artifact URL from the code-
  owned raw layout, verify sha256, install manifest-declared files, and persist installed state.
  - Add component-first uninstall while keeping legacy capability uninstall as a fallback during the migration away
  from capabilities.
  - Record install provenance with `install_backend` / `distribution_source`, and reject reinstalling an already-
  installed component through a different backend.
  - Harden lifecycle/install plumbing: uninstall rollback behavior, owned-path validation, symlink install support,
  central-log coverage, and explicit install lock release on drop.

  ## Ship Note
 ### Limitations
  - Only the `raw` backend has an executor today. `yum` and `npm` are config-valid but return `NOT_IMPLEMENTED` when
  selected for install.
  - `purge` uninstall is still plan-only; real execution is gated and returns `NOT_IMPLEMENTED`.
  - `install` does not start/enable services or run health checks yet.
  - Upgrade semantics are not implemented. Reinstall through the same backend is allowed, but full version transition
  behavior is still incomplete.
  - Cross-backend reinstall is intentionally rejected. Users must uninstall first before switching a component from one
  backend to another.
  - `$releasever` is not auto-detected yet and must be set explicitly under `[vars]` when used in yum URLs.
  - The raw artifact still requires `sha256` in the distribution index; missing checksums are treated as publisher/
  config errors.

## Testing
cargo fmt --all -- --check
cargo check --workspace --all-targets
cargo test --workspace

image
image
image
image
image
image


src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -1,19 +1,1281 @@
...
+    // processes that serialize on the lock within the same second.
+    let lock_ts = Utc::now();
+    let operation_id = format!(
+        "op-install-{}-{}",
+        lock_ts.format("%Y%m%d%H%M%S"),
+        lock_ts.timestamp_subsec_nanos()
+    );
+
+    let runner = InstallRunner::new(layout);
+    let outcome = runner
```
kongche-jbw, line 489: P1：这里如果 `runner.install_files(...)` 在“部分文件已经写入后”失败，会直接通过 `map_err` 返回，调用方拿不到已写入文件列表，因此不会执行 rollback。比如 tar.gz 先写入 regular file，随后创建 symlink / hash referent 失败时，state 不会写入，但前面落盘的文件会残留，形成半安装状态。

建议把 rollback 放到 `InstallRunner::install_files` 内部：记录每个已成功写入/创建的文件，只要后续任一步失败就 best-effort 删除；或者让错误携带 partial outcome 给上层统一 rollback。测试上可以补一个场景：regular file 已写入后 symlink referent 缺失，最终 regular file 也必须被清理。

ikunkun-sys: Done:
  - install_files 在 regular files 已写入后，如果后续 symlink 创建或 hash referent 失败，会回滚已经写入的文件。
  - install_tar_gz 多文件写入时，如果中间某个文件失败，会回滚此前已写入的文件。
  - write_dest_atomic 现在会先校验 mode，再落 .tmp；set_permissions 失败时也会清理 .tmp。
  - 补了回归测试，覆盖“regular file 已落盘后 symlink 失败必须清理”的场景。

kongche-jbw (approved):
lgtm

Issue #870, opened 2026-06-12
[anolisa-cli] Extend install/uninstall to all distributed components (beyond tokenless)


kongche-jbw:
## Background

Currently `anolisa install` and `anolisa uninstall` have end-to-end logic only for the **tokenless** component. However, the monorepo ships 6 other runtime components under `src/` that are already published or planned for distribution:

| Component | Path | Runtime Manifest |
|-----------|------|------------------|
| agentsight | `src/agentsight/` | `manifests/runtime/agentsight.toml` |
| agent-sec-core | `src/agent-sec-core/` | `manifests/runtime/agent-sec-core.toml` |
| agent-memory | `src/agent-memory/` | `manifests/runtime/agent-memory.toml` |
| cosh (copilot-shell) | `src/copilot-shell/` | `manifests/runtime/cosh.toml` |
| os-skills | `src/os-skills/` | `manifests/runtime/os-skills.toml` |
| ws-ckpt | `src/ws-ckpt/` | `manifests/runtime/ws-ckpt.toml` |

## Goal

Extend the `anolisa install <component>` and `anolisa uninstall <component>` commands to support **all 6 remaining distributed components** with the same level of completeness as tokenless.

### In scope
- **Fresh install** flow: artifact download → file extraction → state tracking → central log
- **Uninstall** flow: state lookup → owned file deletion → state cleanup → central log
- Support both `--install-mode user` and `--install-mode system` where applicable
- `--dry-run` mode for all new components
- JSON output (`--json`) for all new components
- Unit tests covering install/uninstall for each component

### Out of scope (future work)
- Takeover of existing installations (adopt already-installed components)
- Upgrade/update logic (`anolisa update <component>`)
- Adapter install/uninstall for these components (covered by separate issues)
- Service enable/disable (systemd unit activation)

## Acceptance Criteria

- `anolisa install <component>` works for all 6 components when the distribution index has a valid entry
- `anolisa uninstall <component>` cleanly removes all ANOLISA-owned files and state
- `installed.toml` accurately reflects the install/uninstall operations
- Central log records each operation with correct `LogKind` and `Severity`
- `--dry-run` prints the plan without executing
- `cargo test` passes with new test cases for each component
- All code, comments, and documentation in English

kongche-jbw:
[...]

PR #931, opened 2026-06-15
feat(anolisa-cli): add install --all to install every available component

### LINKED ISSUE #924: [anolisa-cli] feat: Add 'install --all' to install all distributed components in one command
## Context

Today `anolisa install` requires a component name (e.g. `anolisa install tokenless`).
A user setting up a fresh host therefore has to invoke it once per component,
look up the component list manually, and stitch the results together.

We need a single-command entry point that installs every distributed component in
one shot, on top of the per-component install support that #870 is delivering.

This is complementary to #870, not a duplicate:
- **#870**: make every component individually installable (`anolisa install <component>`).
- **This issue**: provide `anolisa install --all` as a batch entry point on top of #870.

## Implementation

- Add a mutually-exclusive flag `--all` to the `install` subcommand, accepted in
  place of the positional `<component>` argument.
- Resolve the target component set from the component catalog
  (whatever `anolisa ls` already shows as `available`), filtered by the host
  platform (os/arch) so unsupported components are silently skipped with a
  warning, not a hard failure.
- Reuse the existing per-component install path; iterate sequentially so failures
  on one component don't corrupt others' state.
- Apply existing flags consistently:
  - `--install-mode user|system` applies to every component.
  - `--dry-run` prints the full plan (component list + per-component actions)
    without executing.
  - `--json` returns one envelope summarising per-component status
    (`installed | skipped | failed` + reason).
- Failure policy: continue on error by default; print a summary table at the end
  and exit non-zero if any component failed. Add `--fail-fast` for the opposite
  behaviour.
- Mirror the symmetric form on `anolisa uninstall --all` for parity.

## Acceptance Criteria

- [ ] `anolisa install --all` installs every catalog-listed, platform-supported
      component on a clean host.
- [ ] `anolisa install --all --dry-run` prints the full plan without side effects.
- [ ] `anolisa install --all --json` emits a structured per-component result list.
- [ ] One component's failure does not block the rest; final exit code reflects
      whether any failure occurred.
- [ ] `--all` and a positional `<component>` are mutually exclusive (clap-level
      validation, not runtime check).
- [ ] `anolisa uninstall --all` works symmetrically.
- [ ] Unit tests cover: all-success, partial-failure, platform-skip, dry-run,
      and `--fail-fast` paths.

## References

- #870 — per-component install/uninstall support (prerequisite)
- `src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs` — current
  single-component install handler
- `src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs` — catalog
  resolution logic to reuse

---

## Original (zh-CN)

我想实现 anolisa install --all 或者 install all 这个参数

casparant:
Closes #924.

## Summary

Add `--all` flag to `anolisa install`, enabling batch installation of all catalog-listed components in one command.

## Changes

- `--all` is mutually exclusive with the positional `COMPONENT` argument (clap ArgGroup).
- Resolves target set from the component catalog (only `status: available` entries).
- Iterates sequentially; one failure does not block the rest by default.
- `--fail-fast` stops on first failure.
- Human mode: prints `==> <component>` progress + summary line.
- JSON mode: emits `{total, installed, failed, skipped, items[]}` envelope.
- Exit non-zero when any component failed.

## Testing

All 46 existing install/list/status/uninstall tests pass. New behavior depends on a live catalog; integration testing is deferred to #870 completion (per-component install support).

## Relationship to #870

- **#870**: make every component individually installable.
- **This PR**: provide the batch entry point on top of #870.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
@@ -16,7 +16,7 @@
 //! services are recorded in state with `enabled: false`.
 
 use clap::Parser;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::path::{Path, PathBuf};
@@ -47,7 +47,7 @@ use crate::repo_config::{
     HostVars, RepoConfig, RepoConfigError, normalize_override_url, raw_artifact_url, raw_index_url,
     raw_relative_root,
 };
-use crate::response::{CliError, render_json};
+use crate::response::{CliError, render_json, render_json_with_status};
 
 const COMMAND: &str = "install";
 
@@ -56,10 +56,21 @@ const COMMAND: &str = "install";
 // convention), so the auto-generated CLI-version flag must be disabled
 // to free the name. `anolisa --version` still works at the top level.
 #[command(disable_version_flag = true)]
+#[command(group(
+    clap::ArgGroup::new("target")
+        .required(true)
+        .args(["component", "all"]),
+))]
 pub struct InstallArgs {
     /// Component name to install
     #[arg(value_name = "COMPONENT")]
-    pub component: String,
+    pub component: Option<String>,
+    /// Install every component listed in the catalog (mutually exclusive with COMPONENT)
+    #[arg(long, conflicts_with_all = ["component", "version", "package"])]
+    pub all: bool,
+    /// With --all, stop on the first failure instead of continuing
+    #[arg(long, requires = "all")]
+    pub fail_fast: bool,
     /// Install a specific version instead of the latest in the channel
     #[arg(long, value_name = "VERSION")]
     pub version: Option<String>,
@@ -157,8 +168,26 @@ struct InstallResultPayload {
 }
 
 pub fn handle(args: InstallArgs, ctx: &CliContext) -> Result<(), CliError> {
-    let command = format!("install {}", args.component);
-    let component = args.component.clone();
+    if args.fail_fast && !args.all {
+        return Err(CliError::InvalidArgument {
+            command: COMMAND.to_string(),
+            reason: "--fail-fast is only meaningful with --all".to_string(),
+        });
+    }
+    if args.all {
+        return handle_all(args, ctx);
+    }
+    // clap ArgGroup guarantees at least one of `component` / `--all`; with
+    // `--all` ruled out above, `component` is necessarily Some.
+    let component = args
+        .component
+        .clone()
+        .expect("clap ArgGroup ensures component is set when --all is absent");
+    handle_one(component, args, ctx)
+}
+
+fn handle_one(component: String, args: InstallArgs, ctx: &CliContext) -> Result<(), CliError> {
+    let command = format!("install {component}");
 
     let layout = common::resolve_layout(ctx);
     let env = anolisa_env::EnvService::detect();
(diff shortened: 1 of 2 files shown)


Forrest-ly (commented):
## Code Review: feat(anolisa-cli): add install --all

Overall the implementation is clean and well-structured. The `--all` / `--fail-fast` CLI surface, the ArgGroup mutual exclusivity, and the human/JSON dual-mode output are all solid. Below are the items I think deserve attention before merge.

---

### 1. `AllCatalogV1` / `AllCatalogEntry` duplicates `list.rs` types (Medium)

`list.rs` already defines `ComponentCatalogV1` + `ComponentEntry` with the same `schema_version` + `components` shape. The PR's comment says "keeping it local avoids a cross-module type dependency", but both files now parse the **same** remote catalog JSON with two separate struct definitions. This will silently diverge when `list.rs` adds fields or changes semantics.

**Suggestion**: Extract a shared `catalog_v1` module (e.g. under `commands/tier1/` or in `common`) with a single `ComponentCatalogV1` that both `list.rs` and `install.rs` import. `install.rs` can still project only the fields it needs, but the deserialization contract stays unified. This also eliminates the risk of `AllCatalogEntry` silently accepting `schema_version != 1` if `list.rs` is updated to support v2.

---

### 2. `handle_all` constructs `InstallArgs` with `package: None` — silently drops `--package` (Medium)

When `handle_all` iterates components, it creates a synthetic `InstallArgs` per component with `package: None`. But `--package` is already declared `conflicts_with_all = ["component"]` won't fire here — `--all` itself doesn't conflict with `--package` in the clap config, yet the value is silently discarded. This is a minor foot-gun: a user running `anolisa install --all --package foo` would get no error and no effect.

**Suggestion**: Either (a) add `"all"` to `--package`'s `conflicts_with` so clap rejects it upfront, or (b) document in the `--package` arg that it's incompatible with `--all`.

---

### 3. `handle_one` takes `InstallArgs` by value but only needs selected fields (Low)

`handle_one(component: String, args: InstallArgs, ctx: &CliContext)` takes full ownership of `InstallArgs`, but `handle_all` has already cloned/constructed the per-component args. This works, but it means `handle_all` must construct a complete `InstallArgs` (including fields it doesn't use like `version`, `package`) just to satisfy the signature. Not a blocker, but if `InstallArgs` grows, this becomes fragile.

**Suggestion**: Consider refactoring `handle_one` to accept only the fields it needs (or a small `OneInstallConfig` struct), making the contract explicit. Can be deferred to a follow-up.

---

### 4. `AllSummaryItem.status` is `&'static str` — no compile-time exhaustiveness (Low)

The `status` field uses string literals `"installed" | "failed" | "skipped"`, but there's no enum to guarantee exhaustiveness. If a new status is added later, the `filter(|i| i.status == "installed")` counting logic won't catch it at compile time.

**Suggestion**: Consider a small `enum AllItemStatus { Installed, Failed, Skipped }` with `Display`/`Serialize` impls. Low priority, but would prevent typo bugs.

---

### 5. Empty component names are filtered after deserialization (Low)

```rust
.filter(|n| !n.trim().is_empty())
```

This is defensive, but a catalog entry with an empty/whitespace `name` is almost certainly a bug in the catalog itself. Silently skipping it hides data-quality issues.

**Suggestion**: Log a warning (at least in non-quiet mode) when an entry is skipped for this reason, or return an error since a malformed catalog is a signal worth surfacing.

---

### 6. `first_error` only retains the *first* failure (Nit)

When multiple components fail, only the first error is propagated. The rest are visible in JSON output (`items[]`), but in human mode the user sees only the summary counts and one error. This is acceptable for v1, but worth noting — the user has no way to see *which* specific components failed in human mode beyond the counts.

**Suggestion**: In the human-mode summary, consider printing the list of failed component names, e.g. `failed=2 (tokenless, sight)`. Can be a follow-up.

---

### Minor positives

- ArgGroup mutual exclusivity is the right approach — much cleaner than manual validation.
- Reusing `handle_one` for each component avoids logic duplication.
- The JSON envelope with `total/installed/failed/skipped/items` is well-structured.
- `schema_version` check is good forward-compatibility hygiene.

casparant:
Thanks for the thorough review @Forrest-ly. Addressed inline:

**#1 Duplicated catalog types** — Agreed. Will extract a shared `catalog_v1` module in a follow-up PR once #870 lands (it will also need the catalog types), so we unify all three consumers together.

**#2 `--package` not conflicting with `--all`** — Actually already handled: `--all` declares `conflicts_with_all = ["component", "version", "package"]`, so `anolisa install --all --package foo` is rejected by clap at parse time. No change needed.

**#3 `handle_one` signature** — Valid, deferred. Will revisit when InstallArgs grows.

**#4 `&'static str` status** — Valid, deferred to the same follow-up as #1 (enum + exhaustiveness).

**#5 Silent skip on empty names** — Fixed in 73feb41. Now emits a warning to stderr when a catalog entry has an empty name.

**#6 Failed names not visible in human mode** — Fixed in 73feb41. Summary now prints `failed=2 (tokenless, sight)` with the actual component names.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -233,6 +256,199 @@ pub fn handle(args: InstallArgs, ctx: &CliContext) -> Result<(), CliError> {
...
+        let per_args = InstallArgs {
+            component: Some(name.clone()),
+            all: false,
+            fail_fast: false,
+            version: None,
+            backend: args.backend.clone(),
+            repo: args.repo.clone(),
+            package: None,
+        };
+        match handle_one(name.clone(), per_args, ctx) {
```
kongche-jbw: P1：这里在 batch 循环里直接复用 `handle_one(..., ctx)`，但当用户传 `--json` 时，`handle_one` 会为每个组件先各自 `render_json`，循环结束后 `handle_all` 又会输出一次汇总 JSON。这样 `anolisa install --all --json` 会打印多个顶层 JSON object，不再是一个合法 JSON 文档。建议把单组件安装核心逻辑拆成只返回 payload/result、不负责渲染的内部函数，由 `handle_one` 和 `handle_all` 分别做最终输出。

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -233,6 +256,199 @@ pub fn handle(args: InstallArgs, ctx: &CliContext) -> Result<(), CliError> {
...
+            }),
+            Err(err) => {
+                let reason = err.reason().to_string();
+                items.push(AllSummaryItem {
+                    component: name.clone(),
+                    status: "failed",
+                    reason: Some(reason),
+                });
+                if args.fail_fast {
+                    return Err(err);
```
kongche-jbw: P2：`--fail-fast --json` 这里会直接返回单组件错误，main 最终渲染的是 error envelope，而不是 PR 描述承诺的 `{total, installed, failed, skipped, items[]}` batch summary。机器调用方会因为是否 fail-fast 得到两种完全不同的 JSON shape。建议 fail-fast 也先构造已处理 items 的 summary；可以通过非零退出码或 summary 内的 failed count 表达失败。

kongche-jbw:
测试覆盖建议：`install --all` 这类 batch CLI 行为不一定需要依赖 live catalog 才能测。建议至少补纯单测或 file:// repo 测试覆盖 `--all/--fail-fast` clap 互斥、catalog schema/status/empty-name 解析，以及 `--json` 只输出一个 batch summary envelope。当前两个 JSON 输出问题都可以被这类测试直接抓到。

casparant:
Added in 69af448: 9 offline unit tests covering:

- **Clap validation**: `--all` + COMPONENT mutual exclusivity, `--all` + `--package` conflict, `--all` + `--version` conflict, `--all` standalone parse, `--all --fail-fast` parse
- **Catalog parsing**: status=available filtering, schema_version mismatch detection, empty-name skipping

Note: `--fail-fast` without `--all` is syntactically accepted by clap (ArgGroup + `requires` interaction) but semantically harmless — the flag is never read when `all=false`. Documented in-test.

Total install-related tests: 50 passing.

casparant:
Fixed both inline comments in ede7b8d:

**P1 (double JSON output)**: `handle_all` now creates a suppressed `CliContext` (`quiet=true, json=false`) for per-component calls. `handle_one` never renders to stdout during batch; only the single batch summary goes out.

**P2 (fail-fast JSON shape)**: `--fail-fast` now `break`s the loop instead of `return Err()`. The partial summary (items processed up to the failure) is always emitted in the same `{total, installed, failed, skipped, items[]}` envelope. Non-zero exit code still signals failure.

kongche-jbw:
Thanks for the update. This fixes the per-component JSON rendering part, but I think there is still one remaining JSON-output issue in the failure path.

`handle_all()` now renders the batch summary first, but if any component failed it still returns `Err(first_error)` at the end. The top-level `main` handler will then render another error envelope, so `install --all --json` can still emit two top-level JSON objects on failure: the batch summary plus the error response.

There is also a smaller dry-run semantics issue: with `--dry-run --all`, a successful per-component dry-run currently gets counted as `installed`, even though no files/state are written. It may be clearer to mark those items as planned/skipped, or add `dry_run: true` to the batch summary and avoid counting them as installed.

casparant:
Both issues fixed in c0c2ffb.

**1. Double JSON on failure path**

Added `render_json_with_status(command, ok, data)` in `response.rs`. In `handle_all`, JSON mode now calls this with `ok = failed == 0` and **returns `Ok(())`** immediately after. The top-level `main` handler never reaches `render_error`, so only a single JSON envelope is emitted — even on partial failure.

JSON shape on failure:
```json
{
  "ok": false,
  "schema_version": 1,
  "command": "install --all",
  "data": {
    "total": 3, "installed": 2, "planned": 0,
    "failed": 1, "skipped": 0, "dry_run": false,
    "items": [
      {"component": "tokenless", "status": "failed", "reason": "..."}
    ]
  },
  "warnings": [],
  "error": null
}
```

Trade-off: exit code is 0 in JSON mode even with failures (since we return `Ok`). JSON consumers should check the `ok` field and `failed` count. Human mode still returns `Err` for non-zero exit code.

**2. Dry-run semantics**

`--dry-run --all` now:
- Marks successful items as `"planned"` (not `"installed"`)
- Sets `dry_run: true` in the batch summary payload
- Reports `planned=N` instead of `installed=N` in human-mode summary

This makes it explicit that no files/state were written.

kongche-jbw:
Thanks for the latest update. The new shape is definitely moving in the right direction: the per-component JSON output is now suppressed, the failure path emits a single JSON envelope, and the dry-run case now reports `planned` instead of `installed`, which is much clearer.

I noticed a few remaining behavior/contract details that may still be worth tightening:

1. `install --all --json` now emits only one JSON object on failure, but the process exits with code `0` even when the payload has `ok=false` and `failed > 0`. The PR description still says failures should exit non-zero, and this is also useful for shell/CI callers. It may be possible to keep the single JSON envelope while still returning a non-zero exit code, for example by returning a command-level batch error that does not trigger a second render.

2. With `--fail-fast --json`, `total` currently means “processed items” rather than the full target set. In human mode, the same scenario reports the full catalog target count. For example, if 3 components are available and the first one fails, JSON reports `total=1`, while human output reports `total=3`. It might be clearer to keep `total = names.len()` and mark the unprocessed items as `skipped`, or introduce a separate `processed` field.

3. `anolisa install tokenless --fail-fast` is still accepted and silently ignored. Since `--fail-fast` only has meaning for batch install, rejecting it unless `--all` is present would make the CLI contract easier to understand. Otherwise users may assume it changes single-component install behavior.

Overall, the main JSON rendering direction looks much better now; these are mostly about making the final CLI contract consistent for automation and user expectations.

casparant:
All three points fixed in 1df70f3.

**1. Non-zero exit code in JSON mode**

Added `CliError::BatchPartial` variant. `handle_all` now returns `Err(BatchPartial)` after rendering the batch summary. `render_error` recognises this variant and **skips the JSON render** (batch summary already emitted) but still returns exit code 1.

Result: `install --all --json` with failures emits exactly one JSON object **and** exits non-zero.

```
$ anolisa install --all --json; echo $?
{ "ok": false, "data": { "total": 3, "failed": 1, ... } }
1
```

**2. `total` semantics with `--fail-fast`**

`total` now always equals `names.len()` (the full target set). When `--fail-fast` stops early, unprocessed components are added as `skipped` items with `reason: "--fail-fast: not attempted"`. This aligns JSON and human output.

Example with 3 components, first fails:
```json
{ "total": 3, "failed": 1, "skipped": 2,
  "items": [
    {"component": "a", "status": "failed", ...},
    {"component": "b", "status": "skipped", "reason": "--fail-fast: not attempted"},
    {"component": "c", "status": "skipped", "reason": "--fail-fast: not attempted"}
  ] }
```

**3. `--fail-fast` without `--all`**

Now rejected at runtime in `handle()`:
```
$ anolisa install tokenless --fail-fast
error[INVALID_ARGUMENT]: --fail-fast is only meaningful with --all
```

Updated test from `install_fail_fast_without_all_is_harmless` → `install_fail_fast_without_all_is_rejected` (asserts `INVALID_ARGUMENT` error).

kongche-jbw (approved):
LGTM thx

PR #938, opened 2026-06-15
fix(sight): add stop signal to background threads

jfeng18:
## Summary

- Three background threads (`genai-stale-scanner`, `config-watcher`, `token-collector-watcher`) ran infinite loops without checking the `running` flag. After `shutdown()`/`Drop`, they survived indefinitely — the stale-scanner held `Arc<GenAISqliteStore>` preventing the DB handle from being freed, undermining the WAL checkpoint contract (added in 559c918).
- Create `running` Arc<AtomicBool> before thread spawns, pass clones to all three
- stale-scanner: split `sleep(60)` into `60×sleep(1)` with stop check each second
- config-watcher: replace blocking `for event in rx` with `recv_timeout(1s)` guarded by stop flag
- token-collector-watcher: change `loop` to `while stop.load()`
- Remove dead `sls_activated` struct field (`#[allow(dead_code)]`, never read via `self` — truth lives in thread-owned clones)

## What changed / what didn't

**Changed (1 file, +24/-10):**
- `unified.rs`: thread lifecycle + dead field removal

**Not changed:**
- Thread business logic inside loops — only loop conditions changed
- shutdown() logic (already sets running=false, no new join/await)
- `sls_activated` constructor local + Arc::clone to threads — only the **struct field** removed

**Known tradeoff:** No JoinHandle storage / explicit join — threads exit on their own within ≤1s of `running=false`. Storing JoinHandles would require changing 3 static methods to return handles + 3 new struct fields — deferred to a follow-up if needed.

## Test plan

- [x] 573 unit tests pass (local + ECS, rustc 1.89)
- [x] `cargo clippy -D warnings` pass (unified.rs zero errors)
- [x] ECS E2E: `agentsight trace` → cosh agent → SIGTERM → process **exits cleanly** (previously would hang due to detached threads)
- [x] Adversarial workflow review (5 dimensions, 5 agents): shutdown race / scanner timing / recv_timeout correctness / dead field safety / running flag ordering — all PASS

### DIFF
--- a/src/agentsight/src/unified.rs
+++ b/src/agentsight/src/unified.rs
@@ -94,9 +94,6 @@ pub struct AgentSight {
     pid_agent_name_cache: HashMap<u32, String>,
     /// HTTP domain patterns from config, used for runtime DNS-based tcpsniff target addition
     http_domains: Vec<String>,
-    /// Flag indicating SLS Logtail has been activated (irreversible)
-    #[allow(dead_code)]
-    sls_activated: Arc<AtomicBool>,
     /// Mailbox for watcher thread to deposit a dynamically-created LogtailExporter
     pending_logtail: Arc<Mutex<Option<Box<dyn GenAIExporter>>>>,
     /// DeadLoop auto-kill: enabled flag
@@ -452,6 +449,9 @@ impl AgentSight {
         let pending_logtail: Arc<Mutex<Option<Box<dyn GenAIExporter>>>> =
             Arc::new(Mutex::new(None));
 
+        // Create `running` flag early so background threads can observe shutdown.
+        let running = Arc::new(AtomicBool::new(true));
+
         // Spawn config file watcher for runtime hot-reload
         if let Some(ref cfg_path) = config.config_path {
             Self::start_config_watcher(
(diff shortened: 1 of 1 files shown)


jfeng18:
**CI note**: The diff-cover 80% gate will flag this PR — the changed lines are all inside thread spawn closures (`new()` lines 474-493, `start_config_watcher` line 948, `start_token_collector_watcher` line 1079), which only execute inside detached background threads. These paths are structurally unreachable by unit tests since they require a fully-constructed `AgentSight` with live eBPF probes.

The changes are purely loop-condition modifications (`loop {}` → `while stop.load()` / `recv_timeout` with stop check) — no new business logic. Verified via ECS E2E: `agentsight trace` → cosh agent → SIGTERM → process exits cleanly (previously would hang due to detached threads).

chengshuyi (approved):
## Code Review

**维度覆盖**：硬性规则 ✓ | eBPF N/A | FFI N/A | Footprint ✓ | 流水线测试 N/A | 文档同步 ✓

✓ 全部 6 个维度检查通过，未发现问题。

- `running` flag 在线程 spawn 前创建，3 个线程共享同一 flag
- config-watcher: 阻塞 `for event in rx` → `recv_timeout(1s)` + stop guard
- stale-scanner: `sleep(60)` → 60×`sleep(1)` + stop check，最大退出延迟 ≤1s
- token-collector-watcher: `loop` → `while stop.load(SeqCst)`
- `sls_activated` dead field 移除合理，局部变量和 Arc::clone 保留

LGTM 👍

Issue #962, opened 2026-06-16
[anolisa] feat(anolisa): enforce safe uninstall semantics for rpm-observed components


kongche-jbw:
## Goal
Prevent ANOLISA from deleting preinstalled system RPMs by default.

## Scope
- For `rpm-observed`, default uninstall only removes ANOLISA state.
- Add `--remove-system-package` to explicitly delegate removal to dnf.
- Keep `raw-managed` uninstall behavior unchanged.
- For `rpm-managed`, uninstall delegates to dnf remove.

## Acceptance Criteria
- `anolisa uninstall <component>` on `rpm-observed` does not run dnf remove.
- `--remove-system-package` is required to remove observed system RPMs.
- Dry-run clearly shows whether package removal will happen.

Issue #960, opened 2026-06-16
[anolisa] feat(anolisa): detect RPM state drift and provide repair path


kongche-jbw:
## Goal
Detect when users modify RPM packages outside ANOLISA and expose repair/forget options.

## Scope
- `status` compares ANOLISA state with rpmdb reality.
- Manual `dnf update` produces `Drifted`.
- Manual `rpm -e` produces `Missing`.
- Add repair logic to refresh state from rpmdb when package identity is still valid.
- Add forget path to remove stale ANOLISA state without touching RPM.

## Acceptance Criteria
- Manual RPM version changes are shown as drift.
- Missing RPM packages are shown as missing, not silently reinstalled.
- `repair` refreshes state when safe.
- `forget` removes ANOLISA state without package operations.

Issue #958, opened 2026-06-16
[anolisa] feat(anolisa): adopt existing system RPM components as rpm-observed


kongche-jbw:
## Goal
In system mode, detect preinstalled component RPMs and adopt them as `rpm-observed` instead of replacing them with raw.

## Scope
- During `install` or `status`, detect RPM packages mapped to ANOLISA components.
- Mapping order: manifest `[backends.rpm].package`, `repo.toml` package map, RPM provides metadata, default package naming.
- If a unique RPM match exists, write or refresh `rpm-observed` state.
- Do not download raw artifacts or overwrite RPM-managed files.
- Emit clear human output explaining the adoption.
- Implement backend resolution priority: explicit --backend flag > existing state > system RPM presence > repo.toml default_backend.
- Add --backend raw|rpm optional parameter to install command.

## Acceptance Criteria
- With `default_backend = raw`, a preinstalled system RPM is still adopted as `rpm-observed`.
- No raw install happens when a matching system RPM exists.
- Ambiguous RPM matches are reported and are not adopted silently.

PR #967, opened 2026-06-16
feat(anolisa): model RPM-observed ownership in installed state

### LINKED ISSUE #956: [anolisa] feat(anolisa): model RPM-observed ownership in installed state
## Goal
Add state representation for components that already exist as system RPMs and are observed/adopted by ANOLISA without taking file ownership.

## Scope
- Add explicit ownership/provenance model: `raw-managed`, `rpm-managed`, `rpm-observed`.
- Reuse or extend existing `ObjectStatus::Adopted`, `managed`, `adopted`, and `install_backend` fields.
- Record RPM package name, EVR/version, arch, repo/source when available.
- Ensure `rpm-observed` means ANOLISA does not own package removal by default.

## Acceptance Criteria
- State can represent a system RPM adopted by ANOLISA.
- `rpm-observed` objects are distinguishable from raw installs and RPM installs initiated by ANOLISA.
- Existing raw install state remains backward compatible.

ikunkun-sys:
## Description

Add explicit provenance/ownership model to `InstalledObject` for distinguishing `raw-managed`, `rpm-managed`, and `rpm-observed` components.

This implements the ownership model defined in `raw_rpm_lifecycle_proposal.md` §5:

- **`raw-managed`**: Installed via ANOLISA raw backend; ANOLISA manages owned files and may remove them on uninstall.
- **`rpm-managed`**: Installed via ANOLISA-delegated RPM backend (`dnf install`); file transactions owned by rpm/dnf, uninstall delegates to `dnf remove`.
- **`rpm-observed`**: Pre-existing system RPM adopted/observed by ANOLISA. ANOLISA does not own package removal by default — `uninstall` only removes the state record unless `--remove-system-package` is passed.

Key design decisions:
- New fields (`ownership`, `rpm_metadata`) are `Option` with `serde(default)`, so pre-v3 state files load unchanged.
- `effective_ownership()` provides a legacy heuristic fallback for pre-v3 files using existing `managed`/`adopted`/`install_backend` fields.
- `save()` auto-upgrades `schema_version` to current, so old files migrate on next write.
- The install path explicitly writes `Some(Ownership::RawManaged)` instead of relying on heuristic fallback.

## Related Issue

closes #956

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] Multiple / Project-wide

## Checklist

## Testing

- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace` (390 tests passed)

## Additional Notes

New state model tests cover:
- Ownership enum semantics (`owns_removal`, `is_rpm`)
- Explicit ownership field roundtrip (raw/rpm-managed/rpm-observed)
- Legacy heuristic fallback for pre-v3 state files
- Serde snake_case serialization
- Backward compatibility (pre-v3 files without ownership fields still deserialize)


Forrest-ly (commented):
## Code Review: feat(state): model RPM-observed ownership in installed state

Overall this is a well-structured, incremental change. The `Ownership` enum design is clean, backward compatibility is properly handled, and test coverage is thorough. Below are specific findings by severity.

---

### Summary

| Aspect | Assessment |
|--------|------------|
| Design quality | ✅ Clean three-variant ownership model with clear semantics |
| Backward compat | ✅ `Option` + `serde(default)` on new fields; pre-v3 files load unchanged |
| Construction sites | ✅ All 13+ `InstalledObject` literals updated with new fields |
| Test coverage | ✅ Comprehensive — enum helpers, legacy heuristics, serde roundtrips, schema upgrade |
| Documentation | ✅ Doc comments follow repo conventions (why not what, intra-doc links) |
| Wiring | ⚠️ New API defined but not yet consumed by uninstall path |

---

### [Medium] `effective_ownership()` heuristic conflation

**File:** `state.rs` — `effective_ownership()`

```rust
if (self.adopted || !self.managed)
    && self.install_backend.as_deref() == Some("rpm")
{
    return Ownership::RpmObserved;
}
```

The condition `self.adopted || !self.managed` treats `managed = false` as semantically equivalent to adoption. While in the current codebase `managed = false` only co-occurs with `adopted = true`, this heuristic encodes an implicit assumption that could misclassify future states — e.g., a component explicitly set to `managed = false` for "frozen/manual" reasons but never adopted via RPM.

**Suggestion:** Consider tightening to `self.adopted` only, or add a doc comment documenting the invariant that `managed = false` implies adoption in the current data model. If the OR is intentional (covering legacy files where `adopted` wasn't set), a comment explaining that rationale would help.

---

### [Low] New API not yet wired into uninstall decision path

**Files:** `lifecycle.rs` (planner: `plan_owned_files` L383–407, executor: L972), `uninstall.rs`

`owns_removal()` and `is_rpm_observed()` are well-defined but have zero non-test consumers. The actual uninstall removal decision still uses per-file `OwnedFile.owner` (`Anolisa` / `External` / `Unknown`). This means:

- An `rpm_observed` component whose files are marked `FileOwner::Anolisa` would still have those files removed on uninstall today.
- The `--remove-system-package` flag described in the `Ownership::RpmObserved` doc comment doesn't exist yet.

This is acceptable for an incremental data-model-first PR, but the doc comment on `RpmObserved` currently describes behavior that isn't implemented:

> ANOLISA does **not** own package removal by default — `uninstall` only removes the ANOLISA state record unless `--remove-system-package` is passed.

**Suggestion:** Either soften the doc comment to indicate this is the intended future behavior, or add a tracking note (issue/TODO with owner per repo conventions) for the follow-up PR that wires ownership into the planner.

---

### [Low] `save()` stamps schema version but doesn't migrate `ownership`

**File:** `state.rs` — `save()`

```rust
let mut snapshot = self.clone();
snapshot.schema_version = STATE_SCHEMA_VERSION;
```

After a v2 file is loaded and saved, `schema_version` becomes 3, but `ownership` remains `None` in the persisted file (because `skip_serializing_if = "Option::is_none"`). This means every subsequent load must re-run `effective_ownership()` heuristic indefinitely.

Consider populating `ownership` from `effective_ownership()` during save for a one-time forward migration:

```rust
for obj in &mut snapshot.objects {
    if obj.ownership.is_none() {
        obj.ownership = Some(obj.effective_ownership());
    }
}
```

This would also make the persisted state self-documenting. If the intentional design is to keep `ownership = None` for legacy objects, a brief comment explaining why would help.

---

### [Info] Template now has two `[[objects]]` entries

**File:** `templates/installed-state.toml`

The template grew from one object to two (original + rpm-observed example). If any test or tooling parses the template expecting exactly one `[[objects]]` block, this could break. Worth verifying that the template isn't loaded programmatically.

---

### [Info] `RpmMetadata.evr` is optional

Only `package_name` is required in `RpmMetadata`; `evr`, `arch`, and `source_repo` are all `Option`. For RPM packages, EVR is typically always available from `rpmdb`. Consider whether `evr` should be required (non-optional) when `rpm_metadata` is populated, to avoid incomplete metadata records that complicate version comparison during `repair`/`update`.

---

### Security Assessment

No security concerns. Changes are purely additive to in-memory data models with no new I/O, network calls, privilege escalation, or unsafe code. The state file is written atomically via the existing `save()` path. Backward compatibility is well-handled.

---

### Verdict

Solid foundation for the RPM ownership model. Recommend addressing the doc comment accuracy on `RpmObserved` (Medium) before merge, and tracking the uninstall wiring as a follow-up. The heuristic and save-migration suggestions are non-blocking improvements.

ikunkun-sys:
Thanks for the thorough review @Forrest-ly! Pushed a follow-up (comment/doc only, **no logic change**) addressing the findings. Point by point:

### Fixed

**[Medium] `RpmObserved` doc describes unimplemented `--remove-system-package`**
Reworded to separate the invariant this PR actually guarantees (`owns_removal() == false`) from the *intended future* CLI behavior. The doc now points to the lifecycle contract in `raw_rpm_lifecycle_proposal.md` §11 and explicitly notes the `uninstall` / `--remove-system-package` wiring is a follow-up (tracked in #962), not implemented by this state-model change.

### Kept, with rationale comments added

**[Medium] `effective_ownership()` heuristic conflation (`adopted || !managed`)**
Kept the condition; added a rationale comment. Key point: the heuristic runs **only when `ownership` is `None`**, i.e. on pre-v3 files — every v3 write sets `ownership` explicitly, so future writes never reach it and cannot be misclassified. Tightening to `adopted`-only would actually *regress*: a pre-v3 adopted RPM may have been recorded via `managed = false` (the "external, do not mutate" marker) without the `adopted` bool set, and would then be misread as `rpm-managed`.

**[Low] `save()` stamps `schema_version` but doesn't migrate `ownership`**
Intentional — added a comment explaining why. `effective_ownership()` is a *guess* derived from filesystem-side fields. Persisting it on every save would bake a possibly-wrong guess into the file permanently and make it indistinguishable from a verified value, defeating the `None` = "unknown/unverified" distinction. Authoritative ownership is written only when actually known — by `install` (raw-managed) or by `adopt`/`repair` after an rpmdb query (#958/#961). The per-load cost is a few field comparisons, not I/O.

### Not changed (verified / reasoned)

**[Info] Template now has two `[[objects]]`**
Verified no breakage. The only programmatic consumer is the `parse_template_round_trip` test, which uses `.find(kind == Component)` + `!objects.is_empty()` and does not assume a single object.

**[Info] `RpmMetadata.evr` optional**
Keeping it optional. The #956 scope records "EVR/version, arch, repo/source **when available**" and the proposal treats them as best-effort; `package_name` is the stable identity. Making `evr` required would make any record missing it fail to deserialize, conflicting with the backward-compat / lenient-load posture. Drift detection (#960) can treat "evr unknown" as a repair trigger rather than requiring it at load time.

---

Checks all green: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets` (no warnings), `cargo doc --no-deps`, `cargo test -p anolisa-core`.

kongche-jbw (approved):
lgtm

Issue #961, opened 2026-06-16
[anolisa] feat(anolisa): add adopt and repair CLI commands for component lifecycle


kongche-jbw:
## Goal
Expose explicit CLI commands for RPM adoption and state repair.

## Scope
- Add `anolisa adopt <component>`.
- Add `anolisa repair <component>`.
- Ensure commands respect install mode/scope.
- `adopt` only creates `rpm-observed` when a unique existing RPM is found.
- `repair` does not switch backend.

## Acceptance Criteria
- `anolisa adopt copilot-shell` records an existing RPM as `rpm-observed`.
- `anolisa repair copilot-shell` fixes `Drifted` / `Missing` where possible.
- Ambiguous or unsafe repair paths fail with clear messages.

Issue #959, opened 2026-06-16
[anolisa] feat(anolisa): implement RPM-backed update for observed and managed system components


kongche-jbw:
## Goal
Allow `anolisa update <component>` to manage system RPM upgrades through dnf/rpm.

## Scope
- For `rpm-observed` and `rpm-managed`, run RPM update flow: `rpmdb query -> dnf repo query -> dnf update -> refresh state`.
- Require root/system scope for RPM updates.
- If the user is not root, return an actionable error.
- Do not switch backend during update.
- Refresh ANOLISA state from rpmdb after a successful update.

## Acceptance Criteria
- `sudo anolisa update <component>` updates a system RPM component through dnf.
- Non-root update of a system RPM fails with a clear message.
- Raw components continue using raw update behavior.
- User raw override updates only user raw, not the shadowed system RPM.

PR #993, opened 2026-06-17
feat(anolisa): adopt pre-installed system RPMs as rpm-observed

### LINKED ISSUE #958: [anolisa] feat(anolisa): adopt existing system RPM components as rpm-observed

ikunkun-sys:
## 1. Description

When a component is already installed as a system RPM, `anolisa install`
(system mode) now **adopts** it as `rpm-observed` state instead of replacing it
with a raw install, and `status` reports it. This lets ANOLISA coexist with
image-preinstalled RPM components rather than fighting the package manager.

Concretely:

- `install` records the existing RPM as `rpm-observed` (ownership
  `RpmObserved`, status `Adopted`) — reading `rpmdb` and writing ANOLISA state
  only. It never downloads an artifact, runs `dnf`, or claims ownership of the
  RPM's files (`owns_removal() == false`).
- `status` reports an installed-but-untracked component as `observed`, and an
  adopted one as `adopted`, with `rpm_package` / `rpm_evr` / `rpm_source_repo`
  details. `status` stays strictly read-only.
- The pre-release `yum` backend name is renamed to `rpm` throughout
  (`KNOWN_BACKENDS`, bundled `templates/repo.toml`, fixtures). The CLI is
  unreleased, so no alias is kept.

## 2. Design Note

**Two-layer backend decision.** Layer 1 picks a backend *name*
(`--backend` > existing state > system-RPM presence > `default_backend`); layer 2
picks the *action* from `(backend, rpmdb hit, install mode)`. Decoupling them
means a component already adopted as `rpm-observed` re-routes to an **adopt
refresh** on the next `install`, instead of being blocked by the raw trunk. The
adopt branch bypasses `select_backend` / `resolved_base_url`, so it works with a
raw-only `repo.toml` (no `[backends.rpm]` table required).

**Package-name resolution**, highest precedence first:

1. CLI `--package <name>`
2. manifest `[backends.rpm].package`
3. repo.toml `[backends.rpm].package_map[component]`
4. RPM `provides` reverse-lookup — `rpm -q --whatprovides 'anolisa-component(<component>)'` (deduped)
5. default naming `anolisa-<component>`

The first three short-circuit to one candidate; `provides` may return several or
none. **Ambiguous providers**, **multi-version drift**, and **user-mode
`--backend rpm`** are rejected with a clear error — never silently adopted.

**`status` projection.** Adopted (`rpm-observed`) rows skip the raw-layout
manifest health probe, so they stay `adopted` rather than being spuriously
escalated to `degraded`/`failed` by checks that assume a raw install layout.

## 3. Ship Note

- **Scope:** install-path auto-adopt + read-only `status` Observed reporting
  only. Out of scope and tracked separately: delegated `dnf install` (#959),
  drift/repair of adopted rows (#960), and an explicit `adopt` subcommand (#961).
- **`provides` lookup (mapping level 4) is dormant.** It depends on a virtual
  provide `Provides: anolisa-component(<name>)` that does **not yet exist** on
  the packaging side. The chain tolerates this: an absent/empty result silently
  falls through to default naming, so this change is not blocked on packaging.
  Until that provide ships, adoption relies on levels 1-3 and default naming.
- **Explicit `--backend rpm` against a not-yet-installed package** returns
  `not_implemented` (pointing at #959) rather than installing via `dnf`.
- **`status` aggregate view** (no component name) does not scan rpmdb for
  Observed components yet — only a named-component query does. Full-scan is left
  to #960/#963.
- The orthogonal `yum-repo` artifact-source enum in the distribution index is
  intentionally left untouched by the `yum` -> `rpm` rename.

## 4. Test

Deterministic coverage via injected fakes (`FakeCommandRunner` / `dyn PackageQuery`),
no live rpmdb required:

- **platform (+9):** `installed_origin` and `what_provides_installed` — reponame
  parse, empty -> `None`, provides dedup / multiple-distinct / not-provided ->
  empty, query failure -> error.
- **core (+3):** manifest `[backends.rpm].package` parse, `rpm_package()`
  accessor, serialize round-trip (empty backends omitted).
- **install (+19):** package-name precedence (all 5 levels); situation probe
  (adoptable / absent / ambiguous / multi-version); adopt writes `rpm-observed`
  state with empty owned files; dry-run writes nothing; adopt refresh overwrites
  EVR; origin-lookup failure degrades to `None` without failing adopt; explicit
  rpm not-installed -> `not_implemented`; ambiguous / user-mode / backend-switch
  rejections; `--all` batch status mapping.
- **status (+4):** `observed` projection from rpmdb; `None` when absent;
  `package_map` honored; rpm-observed row skips manifest health escalation.

Full workspace `fmt` / `clippy -D warnings` / `check` / `test` / `doc --no-deps`
all pass.

close #958.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
@@ -8,8 +8,16 @@
 //! from the raw repository root, resolve an artifact, then execute by
 //! downloading it with mandatory sha256 verification, loading the install
 //! contract, installing the declared files, and recording state plus a
-//! central-log audit entry. `yum` / `npm` backends are selectable but their
-//! executors are NOT_IMPLEMENTED.
+//! central-log audit entry.
+//!
+//! The `rpm` backend additionally supports **adopt** (issue #958): in system
+//! mode, when a component is already present as a system RPM, `install`
+//! records it as `rpm-observed` state without downloading or running
+//! `dnf install`. The backend decision is two-layered — pick a backend name
+//! (`--backend` > existing state > system RPM presence > `default_backend`),
+//! then pick an action by `(backend, rpmdb hit, install mode)`. Delegated
+//! `dnf install` for not-yet-installed RPM components is out of scope (#959),
+//! and `npm` remains NOT_IMPLEMENTED.
 //!
 //! Deliberately out of scope for this milestone: execution-policy gating,
 //! pre/post hooks, health checks, and service start/enable. Installed
@@ -31,21 +39,23 @@ use anolisa_core::lock::InstallLock;
 use anolisa_core::path_safety::validate_owned_path;
 use anolisa_core::state::{
     FileOwner, InstallMode as StateInstallMode, InstalledObject, InstalledState, ObjectKind,
-    ObjectStatus, OperationRecord, OwnedFile, Ownership, ServiceRef,
+    ObjectStatus, OperationRecord, OwnedFile, Ownership, RpmMetadata, ServiceRef,
 };
 use anolisa_core::{
     ArtifactType, ComponentManifest, DistributionEntry, DistributionIndex, FileKind, ResolveQuery,
     expand_layout_placeholders,
 };
 use anolisa_platform::fs_layout::FsLayout;
+use anolisa_platform::pkg_query::{PackageInfo, PackageQuery, PackageQueryError};
+use anolisa_platform::rpm_query::RpmPackageQuery;
 use chrono::{SecondsFormat, Utc};
 
 use crate::color::Palette;
 use crate::commands::common;
-use crate::context::CliContext;
+use crate::context::{CliContext, InstallMode};
 use crate::repo_config::{
-    HostVars, RepoConfig, RepoConfigError, normalize_override_url, raw_artifact_url, raw_index_url,
-    raw_relative_root,
+    BackendConfig, HostVars, RepoConfig, RepoConfigError, normalize_override_url, raw_artifact_url,
+    raw_index_url, raw_relative_root,
 };
 use crate::response::{CliError, render_json, render_json_with_status};
 
@@ -74,7 +84,7 @@ pub struct InstallArgs {
     /// Install a specific version instead of the latest in the channel
     #[arg(long, value_name = "VERSION")]
     pub version: Option<String>,
-    /// Backend override (raw | yum | npm); defaults to repo.toml default_backend
+    /// Backend override (raw | rpm | npm); defaults to repo.toml default_backend
     #[arg(long, value_name = "BACKEND")]
     pub backend: Option<String>,
     /// One-off base_url override for the selected backend
@@ -167,6 +177,33 @@ struct InstallResultPayload {
     warnings: Vec<String>,
 }
 
+/// What `handle_one` did, so `--all` can distinguish a fresh install from an
+/// RPM adopt in its batch summary (§7.5). The dry-run vs real distinction is
+/// layered on by the caller from [`CliContext::dry_run`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum InstallOutcome {
+    /// A raw install (downloaded + placed files, or its dry-run preview).
+    Installed,
+    /// An existing system RPM recorded as `rpm-observed` (or its dry-run
+    /// preview); no bytes fetched, no owned files written.
+    Adopted,
+}
+
+/// Source that decided the backend name in layer 1 (§4). Only used to phrase
+/// conflict errors; the action is chosen by layer 2 from `(backend, rpmdb,
+/// mode)`, independent of how the name was picked.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum BackendSource {
+    /// Explicit `--backend`.
+    Explicit,
+    /// Component already in state; backend follows its recorded provenance.
+    ExistingState,
+    /// State miss; system mode + rpmdb hit selected `rpm`.
+    SystemRpm,
+    /// None of the above; fell back to `default_backend`.
+    Default,
+}
+
 pub fn handle(args: InstallArgs, ctx: &CliContext) -> Result<(), CliError> {
     if args.fail_fast && !args.all {
         return Err(CliError::InvalidArgument {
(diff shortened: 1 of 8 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -254,12 +385,411 @@ fn handle_one(component: String, args: InstallArgs, ctx: &CliContext) -> Result<
...
+    component: &str,
+    args: &InstallArgs,
+    repo_config: &RepoConfig,
+    ctx: &CliContext,
+    query: &dyn PackageQuery,
+    command: &str,
+) -> Result<RpmSituation, CliError> {
+    let manifest = load_optional_manifest(ctx, component);
+    let rpm_backend = repo_config.backends.get("rpm");
+    let candidates = rpm_package_candidates(
```
kongche-jbw: [P1] I think this should also treat a missing rpm binary as `Absent` at the candidate-resolution step. `probe_rpm_situation()` documents that non-RPM hosts fall back to the default backend, but `rpm_package_candidates()` calls `what_provides_installed()` before `query_installed()`, so `CommandMissing` is converted into `rpm query failed: command not found: rpm` here and the raw fallback is skipped. This can break system-mode installs on hosts/images without rpm tooling. Could we catch `PackageQueryError::CommandMissing` from this call and fall back to default naming or `RpmSituation::Absent`?

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -254,12 +385,411 @@ fn handle_one(component: String, args: InstallArgs, ctx: &CliContext) -> Result<
...
+        return Ok(InstallOutcome::Adopted);
+    }
+
+    // Acquire the lock, then load state inside it so a concurrent writer is
+    // not clobbered — mirrors `execute_raw`'s ordering.
+    let _lock = InstallLock::acquire(&layout.lock_file).map_err(|err| CliError::Runtime {
+        command: command.to_string(),
+        reason: format!("failed to acquire install lock: {err}"),
+    })?;
+    let mut state =
```
kongche-jbw, line 665: [P1] Could we rerun the backend-compatibility guard after this lock-protected state reload? The raw path reloads state under the install lock and calls `ensure_component_backend_compatible()` again before writing, but the rpm adopt path reloads and then directly `upsert_object()`s. If a concurrent raw install wins the lock first, this adopt path can still overwrite the component as `rpm-observed`, losing the backend provenance guarantee.

ikunkun-sys: Good catch — fixed in f053d74. `execute_adopt` now re-runs `ensure_component_backend_compatible(state, component, "rpm")` on the state reloaded **under the install lock**, mirroring `execute_raw`. The adopt-refresh case (existing `rpm-observed`) still matches and proceeds; a concurrent raw install that won the lock first is now rejected with `INVALID_ARGUMENT` instead of being clobbered. Added a test that drives `execute_adopt` against a pre-seeded raw record and asserts the raw provenance survives.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
ikunkun-sys: Thanks for flagging this — the path changed in c6b6d44 (pushed shortly after your review), so it no longer surfaces as the confusing `rpm query failed: command not found` runtime error.

We deliberately chose **warn-and-exit** here rather than treating `CommandMissing` as `Absent` and falling back to raw. In system mode the probe exists precisely to decide between *adopt* and *raw*; when it can't run at all we'd rather fail loud than silently pick a backend the caller didn't ask for. `install --backend raw <component>` bypasses the probe entirely and installs fine on hosts without rpm tooling, so the non-RPM path is one explicit flag away. The error message names that escape hatch.

You're right that a host with no rpm/dnf has no RPMs to clobber, so a silent raw fallback would also be safe in practice — we went with explicit-opt-in over implicit-fallback as a design preference. If you'd rather non-RPM images 'just work' without the flag, I'm happy to switch `CommandMissing` back to `RpmSituation::Absent`; it's your call as maintainer.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -183,23 +220,131 @@ pub fn handle(args: InstallArgs, ctx: &CliContext) -> Result<(), CliError> {
...
     let layout = common::resolve_layout(ctx);
     let env = anolisa_env::EnvService::detect();
-
-    // — Resolution chain: repo.toml → backend → base_url → package. —
     let repo_config = RepoConfig::load(&layout).map_err(|err| repo_config_err(err, false))?;
+    let installed = common::load_installed_state(ctx, COMMAND)?;
+
+    // ── Layer 1: pick the backend name + its source (§4). ──
+    //
+    // Priority: explicit --backend > existing state > system RPM presence
```
kongche-jbw, line 255: [P3] Small follow-up: this comment seems stale after the latest change. In system mode with no explicit `--backend` and fresh state, we now do probe rpm first, and missing rpm/dnf intentionally fail-fast with the `--backend raw` hint. Could we update this wording so future readers do not assume the default raw path never touches rpm/dnf?

ikunkun-sys: Done

kongche-jbw (approved):
lgtm

PR #994, opened 2026-06-17
refactor(sight): keep sls config decision pure

jfeng18:
## Description

`decide_sls_config_change` — a function named and documented as a pure decision
returning an `SlsConfigAction` — also performed a process-global side effect: it
called `genai::logtail::set_dynamic_logtail_path(...)`, mutating the shared
`DYNAMIC_LOGTAIL_PATH`. That contradicts the function's own enum doc, which says
side effects are carried out by the thread shell "so the decision logic stays
pure and testable."

This moves the dynamic-path update into the caller, `handle_config_event`, whose
match arms now own it: `Deactivated` → clear, `Activate` / `Reactivated` → set.
`decide_sls_config_change` keeps only the caller-owned `sls_activated`
test-and-set (the primitive that distinguishes first-activation from
reactivation) and no longer touches any cross-module global state.

## Related Issue

no-issue: realigns the SLS config decision function with its own documented
contract (side effects belong in the shell). No behavior change.

## Type of Change

- [x] Refactor (no functional change)

## Scope

- [x] `sight` (agentsight)

## Key Changes

- `src/background.rs`: remove the two `set_dynamic_logtail_path` calls from
  `decide_sls_config_change`; add them to `handle_config_event`'s `Deactivated` /
  `Activate` / `Reactivated` arms (Activate sets the path before building the
  exporter, preserving the prior ordering). Doc comments updated to match.

## Behavior preservation

Every `SlsConfigAction` variant leaves `DYNAMIC_LOGTAIL_PATH` in the identical
final state as before; only the call-site moved. `sls_activated` has no
production reader that correlates it with the path (it is only `.swap`-ed in
`decide` and `.store`-d at startup), so reordering the swap and the path-set is
unobservable.

## Checklist

- [x] `cargo +1.89.0 fmt --all --check` pass
- [x] `cargo +1.89.0 clippy --all-targets -- -D warnings` pass
- [x] tests pass (661 lib tests)
- [x] I have added tests

## Testing

**已单测（判别性，mutation 验证）:**

- New `test_dynamic_path_side_effect_is_in_handler_not_decider`: asserts `decide`
  leaves the global path untouched, and that `handle` sets it on activation and
  clears it on deactivation.
- Augmented `test_handle_event_reactivate_no_new_exporter`: asserts reactivation
  overwrites a stale path to the new value.
- All four side-effect sites are mutation-covered: reverting decide-purity, or
  deleting any of the three handler set-sites (Deactivated / Activate /
  Reactivated), makes a test fail. Verified by hand-applied mutations.
- A module-level mutex serializes the global-path tests (no flaky parallel race).

**已 ECS live E2E（kernel 6.6.102, real running watcher）:**

Ran `agentsight trace --config <file>` on ECS and drove real config-file edits;
the running config-watcher fired all three handler set-sites (uid redacted):

- Activate → `Dynamic logtail path set: /tmp/a.log` + "SLS Logtail activated".
- Reactivated → `Dynamic logtail path set: /tmp/b.log` + "re-activated with
  path=/tmp/b.log" — the overwrite is exercised in a live watcher.
- Deactivated → `Dynamic logtail path cleared` + "SLS Logtail deactivated".

No process abort. Confirms the moved side-effect works end-to-end in a running
watcher, not just in unit tests.

### DIFF
--- a/src/agentsight/src/background.rs
+++ b/src/agentsight/src/background.rs
@@ -54,8 +54,9 @@ pub(crate) fn path_matches_target(paths: &[PathBuf], target: &Option<OsString>) 
 /// Decision produced by [`decide_sls_config_change`]: what the config watcher
 /// should do in response to a parsed `runtime.sls_logtail_path` value.
 ///
-/// Side effects (process exit, exporter construction, mailbox write) are carried
-/// out by the thread shell so the decision logic stays pure and testable.
+/// Side effects (process exit, exporter construction, mailbox write, dynamic
+/// logtail-path update) are carried out by the thread shell so the decision
+/// logic stays pure and testable.
 #[derive(Debug, PartialEq)]
 pub(crate) enum SlsConfigAction {
     /// Field missing / parse error, or empty path while already inactive: no-op.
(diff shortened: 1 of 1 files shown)


chengshuyi (approved):
LGTM — 重构正确，行为保持一致，测试覆盖充分。

PR #995, opened 2026-06-17
feat(anolisa): RPM-backed update and delegated dnf install

### LINKED ISSUE #959: [anolisa] feat(anolisa): implement RPM-backed update for observed and managed system components

ikunkun-sys:
## Description

Adds the `anolisa update <component>` surface and wires two RPM/dnf-backed
lifecycle paths:

- **RPM update** for `rpm-observed` / `rpm-managed` components — upgrades the
  package in place via `dnf`, then refreshes ANOLISA state from rpmdb.
- **Delegated install** for not-yet-present RPM components — delegates the file
  transaction to `dnf install` and records `rpm-managed` ownership (the first
  code path to produce `Ownership::RpmManaged`).

The CLI surface changes from `update runtime <component>` to a direct positional
`update <component>`; `self` / `all` remain subcommands, kept mutually exclusive
with the positional via `args_conflicts_with_subcommands`.

## Design Note

- **Platform-layer split.** A new `PackageTransaction` trait (`install` +
  `update`) is the write-side mirror of the existing read-only `PackageQuery`.
  `RpmTransaction<R: CommandRunner>` implements it by running `dnf <verb> -y
  <pkg>` through the injectable `CommandRunner`, sharing one `run_dnf` helper and
  classifying spawn-vs-exit faults the same way as `PackageQuery`. The trait is
  object-safe, so the CLI holds `&dyn PackageTransaction` and injects a fake in
  tests instead of shelling out to a live `dnf`.
- **Privilege and state refresh stay in the CLI.** The platform layer only
  spawns the transaction and classifies the outcome; root-gating, the two-layer
  backend decision, and the post-transaction rpmdb refresh live in the CLI
  consumer.
- **Update never switches backend.** The flow is `rpmdb query -> dnf repo query
  -> dnf update -> refresh state`; ownership and `install_backend` are preserved
  across the upgrade.
- **Delegated-install routing.** In system mode an absent RPM component
  (explicit `--backend rpm`, or `default_backend = rpm`) routes through
  `RpmSituation::Absent { package }` to `dnf install`, injected via
  `RpmExec { query, txn, is_root }`. `dnf` owns the file transaction, so
  RPM-owned paths stay out of ANOLISA owned-files (`managed=true`,
  `adopted=false`).
- **Lock-time guards.** Before persisting an update the package identity is
  re-checked under the install lock; a refresh failure after a successful `dnf
  update` errors out (pointing at `anolisa repair`) rather than recording a
  stale EVR.

## Ship Note

- **Raw-backend update is not implemented.** Raw-trunk components keep the
  existing not-yet-implemented planner boundary; only the RPM path is wired here.
- **Real runs require system mode + root.** RPM update and delegated install
  execute only under `--install-mode system` as root; `--dry-run` previews
  without touching the host. `sudo anolisa update <component>` alone does **not**
  auto-switch `install_mode` to system — it is resolved from the flag (default
  `user`), not from euid. This is a known cross-cutting limitation shared with
  the merged #958 adopt path, to be fixed in a separate issue.
- **RPM only.** Other backends (npm, etc.) are out of scope for this change.

## Test

- Platform-layer unit tests for `RpmTransaction` install/update: success,
  already-satisfied no-op, non-zero exit → `TransactionFailed`, and spawn faults
  (missing binary / permission denied).
- CLI tests for the update flow and delegated install, driven by an injected
  fake (`FakeRpm` / `FakeInstaller`) so no live `dnf` is required — covering a
  successful RPM update, refresh-failure-after-successful-update errors while
  leaving state intact, persist refusing when the package identity changed under
  the lock, and delegated install recording `rpm-managed`.
- Full gate green: `cargo fmt --all -- --check`,
  `cargo clippy --workspace --all-targets -- -D warnings`,
  `cargo test --workspace` (613 passing), `cargo doc --workspace --no-deps`.

Closes #959.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
@@ -1,31 +1,53 @@
-//! `anolisa update` — unified update surface (launch spec §7.3).
+//! `anolisa update` — unified update surface.
 //!
-//! Three subcommands:
+//! Three forms:
+//! - `update <component>` - update one ANOLISA-managed component.
 //! - `update self` - update the `anolisa` CLI binary only.
-//! - `update runtime <COMP|all>` - update one or all ANOLISA-managed
-//!   runtime components.
 //! - `update all` - update every ANOLISA-managed runtime, osbase, and
 //!   adapter object.
 //!
-//! Explicit invariant (spec §7.3, decision §11.2): `update all` does
-//! **not** include CLI self-update. The binary swap never shares a
-//! transaction with component updates. Self-update is reachable via
-//! both `anolisa update self` and `anolisa self update`.
+//! The component is a positional argument; `self` / `all` are subcommands
+//! (kept mutually exclusive with the positional via
+//! `args_conflicts_with_subcommands`). A component literally named `self` or
+//! `all` would be shadowed by the subcommand — those are reserved.
+//!
+//! Explicit invariant: `update all` does **not** include CLI self-update. The
+//! binary swap never shares a transaction with component updates.
+//!
+//! `update <component>` implements the **RPM** update path (issue #959): for
+//! `rpm-observed` and `rpm-managed` components it runs the flow
+//! `rpmdb query -> dnf repo query -> dnf update -> refresh ANOLISA state`,
+//! gated on root for the real run. It never switches backend —
+//! ownership/`install_backend` are preserved. Raw components stay on the
+//! not-yet-implemented planner boundary, and `update all` remains
+//! `NOT_IMPLEMENTED` pending the raw distribution resolver.
 
 use std::io::Read;
 use std::path::Path;
 use std::time::Duration;
 
+use chrono::{SecondsFormat, Utc};
 use clap::{Parser, Subcommand};
 use serde::Serialize;
 
+use anolisa_core::central_log::{CentralLog, LogKind, LogRecord, LogStatus, Severity};
+use anolisa_core::lock::InstallLock;
 use anolisa_core::self_update::{self, ProgressFn, SelfUpdateOutcome};
+use anolisa_core::state::{ObjectKind, OperationRecord, Ownership};
+use anolisa_platform::pkg_query::{PackageInfo, PackageQuery, PackageQueryError};
+use anolisa_platform::pkg_transaction::{PackageTransaction, PackageTransactionError};
+use anolisa_platform::privilege;
+use anolisa_platform::rpm_query::RpmPackageQuery;
+use anolisa_platform::rpm_transaction::RpmTransaction;
 
 use crate::color::Palette;
 use crate::commands::common;
 use crate::context::CliContext;
 use crate::repo_config::RepoConfig;
 use crate::response::{self, CliError};
+
+/// Command label for JSON envelopes and error routing.
+const COMMAND: &str = "update";
 
 const CLI_CHANGELOG_URL: &str = "https://agentic-os.sh/#anolisa-cli-changelog";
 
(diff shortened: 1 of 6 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
```diff
@@ -44,51 +66,572 @@ const DEFAULT_REPO_CONFIG_URL: &str =
...
+/// Dispatches the selected `anolisa update` form.
 ///
 /// # Errors
 ///
-/// Returns [`CliError`] when the selected update operation fails or is not
-/// implemented yet.
+/// Returns [`CliError`] when the selected update operation fails, no target is
+/// given, or the operation is not implemented yet.
 pub fn handle(args: UpdateArgs, ctx: &CliContext) -> Result<(), CliError> {
     bootstrap_repo_config(ctx);
```
kongche-jbw: [P2] I think we should move this bootstrap after target validation, or only run it in the branches that actually need repo config. Since `command` is now optional, plain `anolisa update` reaches `handle()` and triggers the bootstrap before returning `INVALID_ARGUMENT`. I reproduced `anolisa update --dry-run` printing the repo-config download message before the usage error, and a real run can also hit the network / try to write config for an invalid invocation.

kongche-jbw: [P2] Related to this bootstrap path: the default OSS repo.toml currently still uses the old `[backends.yum]` spelling, while this build only accepts `raw/rpm/npm`. I reproduced bootstrap failing with `unknown backend 'yum' in repo config (known: raw, rpm, npm)`. If the OSS object is not updated before this ships, upgraded users without a local repo.toml will keep skipping bootstrap. Could we either update the published file as part of the ship step or point the bootstrap URL at a versioned config that already uses `[backends.rpm]`?

ikunkun-sys: Agreed — this is a regression from making the surface a positional argument. Before, `command` was a required subcommand, so a bare `anolisa update` was a clap parse error and never reached `handle()`; now `(None, None)` reaches it and the top-of-function bootstrap fired before validation.

Fixed by moving `bootstrap_repo_config` out of the top of `handle()` and into only the two branches that actually run an update (`self` and `<component>`). The invalid `(None, None)` path and the not-yet-implemented `update all` now return without any network/config side effect. Added a regression test `bare_update_errors_before_any_bootstrap` (bare `update` -> INVALID_ARGUMENT, no repo.toml written). Amended into 0e37964.

ikunkun-sys: Good catch on the ship risk, but I don't think there's a code change to make in this PR for it:

- The in-repo source template `templates/repo.toml` already uses `[backends.rpm]` (it was updated with the yum->rpm rename in #958). The stale `[backends.yum]` only lives in the *published* OSS object at the bootstrap URL, which is a release artifact outside this repo.
- The bootstrap already fails safe: `fetch_and_write_repo_config` validates the download via `RepoConfig::from_toml_str` before writing, so a yum-spelled published file is rejected and degrades to a warning — it never lands a broken config on the host. As you note, the effect is just that bootstrap keeps being skipped, not a corrupted state.
- Adding a `yum` backend alias would reverse #958's deliberate 'no alias kept' decision (the CLI is unreleased), so I'd rather not.

So the fix is the ship step you suggested — republish the OSS object from the current template

src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
```diff
@@ -44,51 +66,582 @@ const DEFAULT_REPO_CONFIG_URL: &str =
...
+    // `args_conflicts_with_subcommands` guarantees `command` and `component`
+    // are never both set, so a present subcommand always wins.
+    //
+    // Bootstrap the repo config only inside the branches that actually run an
+    // update: with `component` now optional, a bare `anolisa update` (or a
+    // not-yet-implemented `update all`) must fail validation without first
+    // reaching out to the network or writing config.
+    match (args.command, args.component) {
+        (Some(UpdateCommands::SelfBin), _) => {
+            bootstrap_repo_config(ctx);
```
kongche-jbw, line 116: Thanks for moving the bootstrap call after target validation; the no-target `update` path looks fixed now. One remaining concern is that the default bootstrap URL still does not seem compatible with the current parser. With an isolated HOME, `anolisa --json update self` still prints `warning: repo config bootstrap skipped: downloaded config invalid: unknown backend 'yum' in repo config (known: raw, rpm, npm)`, because the published `repo.toml` at the default URL still contains `[backends.yum]`. Could we either update the published config to `[backends.rpm]` or temporarily tolerate the old `yum` spelling here? Otherwise upgraded users without a local repo.toml will continue to miss the bootstrap config.

kongche-jbw (approved):
lgtm

PR #763, opened 2026-06-07
fix(sight): reduce MAX_BUF_SIZE from 4MB to 32KB

jfeng18:
## Summary

Fixes #759. MAX_BUF_SIZE was 8*512*1024 = 4MB (comment said 512KB). Each SSL/TCP event reserved ~4MB from a 32MB ring buffer, allowing only ~7 concurrent events. Reduced to 32KB (covers max TLS record of 16KB with margin), increasing capacity to ~1000 concurrent events.

## Changes

- `sslsniff.h`: `(8 * 512 * 1024)` -> `(32 * 1024)` with corrected comment
- `sslsniff.rs`: Update 2 misleading '512KB' comments to '32KB'
- `parser.rs`: Update 1 misleading '512KB' comment

## Verification

- E2E on ECS: `bpftool prog dump xlated` confirmed reserve size changed from 4,194,384 to ~32KB
- cargo check + cargo test passed
- BPF verifier masks in sslsniff.bpf.c and tcpsniff.bpf.c verified correct with 32KB

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #759, opened 2026-06-07
[sight] bug(sight): MAX_BUF_SIZE is 4MB not 512KB, ring buffer wastes 99.6% per event


jfeng18:
## Summary

MAX_BUF_SIZE in sslsniff.h is 8*512*1024 = 4MB, but the comment says 512KB. Each SSL/TCP event reserves ~4MB from a 32MB ring buffer, allowing only ~7 concurrent events.

## Root cause

```c
// sslsniff.h:9
#define MAX_BUF_SIZE (8 * 512 * 1024)  // 512KB eBPF buffer size (kernel limit)
```

8 * 512 * 1024 = 4,194,304 = **4MB**, not 512KB. Comment is wrong by 8x.

- struct probe_SSL_data_t embeds u8 buf[MAX_BUF_SIZE] = 4MB inline
- All 5 bpf_ringbuf_reserve calls use sizeof(*data) ~ 4,194,384 bytes
- Ring buffer is 32MB (common.h:9), max ~7 concurrent events
- Typical TLS payload max 16KB, 99.6% wasted per event

## Evidence

- `bpftool prog dump xlated`: r2 = 4194384 (compiled reserve size)
- `bpftool map show`: ringbuf max_entries = 33554432 (32MB)
- 8+ core machines under SSL load will overflow ring buffer

## Suggested fix

Reduce MAX_BUF_SIZE to 32KB:

```c
#define MAX_BUF_SIZE (32 * 1024)  // 32KB max capture per SSL/TCP event
```

Or use dynamic-size reservation via bpf_ringbuf_output.

PR #1000, opened 2026-06-18
fix(sight): tiered SSL ring buffer reservations

jfeng18:
## Summary

Replace the fixed **4 MiB-per-record** SSL ring-buffer reservation with **per-tier reservations** sized to the payload, so small SSL calls no longer pad the shared 32 MiB ring to the worst case and get dropped under burst (#759) — while a single SSL call up to 4 MiB is still captured **whole** (no truncation).

**Supersedes #763** (my own PR, which shrank `MAX_BUF_SIZE` to a flat 32 KiB). As raised in review on #763, a flat 32 KiB cap truncates large request/response bodies; tiering keeps the full 4 MiB ceiling without padding every record. I'll close #763 once this lands.

## What changed

- **`sslsniff.h`** — four per-tier record types (`probe_SSL_data_{small,medium,large,t}` = 16 KiB / 128 KiB / 512 KiB / 4 MiB) sharing one byte-for-byte header (`SSL_DATA_HEADER_FIELDS`).
- **`sslsniff.bpf.c`** — `SSL_EMIT_TIERED` reserves `sizeof(smallest tier that fits len)`. The copy length is re-clamped to the tier **immediately before** `bpf_probe_read_user`, behind an `asm` barrier so clang can't drop the clamp as dead code (the outer tier branch already bounds it) while the verifier still gets a fresh in-register bound — otherwise the program is rejected at load with `R2 min value is negative` once the bound is lost across the `bpf_get_current_comm` register spill. Handshake records reserve header-only.
- **`probes/sslsniff.rs`** — `from_bytes()` decodes the shared header by offset and slices the payload by the `buf_size` field (never `data.len()`), so a variable-length / still-full-size record decodes correctly. Both consumers use it.
- **`tcpsniff.bpf.c`** — a second `EVENT_SOURCE_SSL` producer sharing this header; it now initializes the new shared `truncated` field (the ring reservation is not zeroed), and the consumer only warns when `truncated` is set **and** `len` exceeds the cap.

## Verification

**✅ E2E — real 6.6 kernel (6.6.102):**
- BPF object loads (all 4 tier programs + handshake verify; `main` also loads — the `pids[]` CO-RE relocation "invalid insn" is a pre-existing dead-branch artifact present on `main` too).
- Live capture: a single **300 KiB** `SSL_write` captured **whole** in one record (`len=307200`, `buf=307200`, `truncated=0`), decoded and parsed as a complete POST; small records (handshake, response chunks) too; **0** truncation warnings.

**✅ Unit (toolchain 1.89, exact CI command):** 669 tests, incl. 6 discriminating `from_bytes` tests (all tier shapes, truncation, `buf_size` clamp, header-only handshake, per-field sentinels). `clippy --all-targets -- -D warnings` clean; `fmt --all --check` clean.

**◑ tcpsniff field-init:** the BPF object loads + attaches (verifier-safe); the fix is correct by construction (`truncated = 0` + the consumer guard) and the `from_bytes` truncated path is E2E-exercised via sslsniff. Live tcpsniff **capture** was not triggered in a loopback E2E (0 stored HTTP rows — a tcpsniff capture-path matter orthogonal to this field-init), so the no-false-warning behavior was not exercised against live tcpsniff records.

## Notes

- Kernel floor stays 5.8 (no dynptr / `bpf_loop` introduced).
- tcpsniff still reserves the full 4 MiB; giving it per-tier reservations is intentionally a follow-up, not in this PR.

### DIFF
--- a/src/agentsight/src/bpf/sslsniff.bpf.c
+++ b/src/agentsight/src/bpf/sslsniff.bpf.c
@@ -78,8 +78,67 @@ int BPF_UPROBE(probe_SSL_rw_enter, void *ssl, void *buf, int num) {
     return 0;
 }
 
+/* Emit ONE SSL record using the per-tier record TYPE (e.g. probe_SSL_data_small)
+ * whose buf[] is exactly TIER bytes. The reservation is sizeof(struct TYPE) — a
+ * true compile-time constant — and the payload is clamped to TIER and copied
+ * WITHIN that type's buf: the only shape the verifier accepts. (Reserving fewer
+ * bytes than sizeof(struct) and then writing into a LARGER typed buf[] is
+ * REJECTED with EACCES at load — confirmed on 6.6.) A small SSL call reserves a
+ * small type, so it no longer pads the shared ring to the 4 MiB worst case
+ * (#759). `truncated` is set only when the payload exceeds the chosen tier
+ * (possible only at the top 4 MiB tier). */
+#define SSL_EMIT_ONE(TYPE, TIER, src_, len_, rw_, ts_, delta_, pid_, tid_, uid_, ssl_, ishs_) \
+    do {                                                                        \
+        struct TYPE *_d = bpf_ringbuf_reserve(&rb, sizeof(struct TYPE), 0);     \
+        if (!_d)                                                                \
+            break;                                                              \
+        _d->source = EVENT_SOURCE_SSL;                                          \
+        _d->timestamp_ns = (ts_);                                              \
+        _d->delta_ns = (delta_);                                               \
+        _d->pid = (pid_);                                                       \
+        _d->tid = (tid_);                                                       \
+        _d->uid = (uid_);                                                       \
+        _d->len = (u32)(len_);                                                  \
+        _d->rw = (rw_);                                                         \
+        _d->is_handshake = (ishs_);                                             \
+        _d->ssl_ptr = (ssl_);                                                   \
+        _d->truncated = ((u32)(len_) > (u32)(TIER)) ? 1 : 0;                    \
+        bpf_get_current_comm(&_d->comm, sizeof(_d->comm));                      \
+        /* Re-clamp the copy length to the tier right before the read. Two      \
+         * verifier traps must be dodged: (1) a length computed earlier is      \
+         * spilled across bpf_get_current_comm() and reloaded as an UNBOUNDED   \
+         * scalar; (2) the outer tier-selection already proved len_ <= TIER, so \
+         * clang folds a plain `if (_n > TIER)` away as dead code -- yet the     \
+         * verifier does NOT carry that bound across the spill, so the clamp     \
+         * still has to execute. The asm barrier makes _n opaque to clang so the \
+         * clamp is emitted, handing bpf_probe_read_user a fresh umax=TIER bound \
+         * (TIER is a compile-time constant). Without it the load is rejected:   \
+         * "R2 min value is negative, either use unsigned or 'var &= const'". */ \
+        u32 _n = (u32)(len_);                                                   \
+        asm volatile("" : "+r"(_n));                                            \
+        if (_n > (u32)(TIER))                                                   \
+            _n = (u32)(TIER);                                                   \
+        int _rc = (src_) ? bpf_probe_read_user(&_d->buf, _n, (const char *)(src_)) : -1; \
+        if (_rc) { _d->buf_filled = 0; _d->buf_size = 0; }                      \
+        else     { _d->buf_filled = 1; _d->buf_size = _n; }                     \
+        bpf_ringbuf_submit(_d, 0);                                              \
+    } while (0)
+
+/* Pick the smallest tier (and its record type) that holds `len_` and emit one. */
+#define SSL_EMIT_TIERED(src_, len_, rw_, ts_, delta_, pid_, tid_, uid_, ssl_, ishs_)       \
+    do {                                                                                   \
+        u32 _l = (u32)(len_);                                                              \
+        if (_l <= SSL_TIER_SMALL)                                                          \
+            SSL_EMIT_ONE(probe_SSL_data_small, SSL_TIER_SMALL, src_, len_, rw_, ts_, delta_, pid_, tid_, uid_, ssl_, ishs_);   \
+        else if (_l <= SSL_TIER_MEDIUM)                                                    \
+            SSL_EMIT_ONE(probe_SSL_data_medium, SSL_TIER_MEDIUM, src_, len_, rw_, ts_, delta_, pid_, tid_, uid_, ssl_, ishs_); \
+        else if (_l <= SSL_TIER_LARGE)                                                     \
+            SSL_EMIT_ONE(probe_SSL_data_large, SSL_TIER_LARGE, src_, len_, rw_, ts_, delta_, pid_, tid_, uid_, ssl_, ishs_);   \
+        else                                                                              \
+            SSL_EMIT_ONE(probe_SSL_data_t, MAX_BUF_SIZE, src_, len_, rw_, ts_, delta_, pid_, tid_, uid_, ssl_, ishs_);         \
+    } while (0)
+
 static int SSL_exit(struct pt_regs *ctx, int rw) {
-    int ret = 0;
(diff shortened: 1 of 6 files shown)


src/agentsight/src/probes/sslsniff.rs
```diff
@@ -93,6 +93,66 @@ impl SslEvent {
...
+    /// 4 MiB `buf` array on the stack if materialized — and (3) slice the payload
+    /// by the `buf_size` FIELD, never by `data.len()` (which over-counts for any
+    /// still-full-size record such as tcpsniff's 4 MiB reservation).
+    pub fn from_bytes(data: &[u8]) -> Option<Self> {
+        type R = bpf::probe_SSL_data_t;
+        let hdr = mem::offset_of!(R, buf);
+        if data.len() < hdr {
+            return None;
+        }
+        let u32_at = |off: usize| u32::from_ne_bytes(data[off..off + 4].try_into().unwrap());
```
chengshuyi: **[硬性规则]** 非测试代码使用了 `unwrap()`（此处及下面两行共 3 处）。

AGENTS.md §0 硬性规则："禁止在非测试代码中使用 `unwrap()` 或 `expect()`"。

虽然 `try_into()` 从精确 4/8 字节切片转 `[u8; N]` 理论上不会失败（前面 `data.len() < hdr` 已保证），但按字面规则仍需修复。

**建议**（任选）：
```rust
// 方案 A：直接构造数组，完全避免 try_into
let u32_at = |off: usize| u32::from_ne_bytes([data[off], data[off+1], data[off+2], data[off+3]]);

// 方案 B：unwrap_or 降级（语义上等价，因为永远不会走 or 分支）
let u32_at = |off: usize| u32::from_ne_bytes(data[off..off + 4].try_into().unwrap_or([0; 4]));
```
推荐方案 A，零开销且完全消除 panic 路径。

chengshuyi (commented):
## AgentSight Code Review — 6 维度审查

**变更文件**: `sslsniff.bpf.c`, `sslsniff.h`, `tcpsniff.bpf.c`, `probes.rs`, `sslsniff.rs`
**Diff 大小**: +337 / -112（净 ~449 行）

| 维度 | 结果 |
|------|------|
| 硬性规则合规 | ⚠️ 1 finding — `unwrap()` x3 in non-test `from_bytes` |
| eBPF 安全 | ℹ️ 2 findings（低优先级建议） |
| FFI 边界 | N/A（无 FFI 变更） |
| Footprint Ladder | ✓ 通过（扩展现有模块，级别 1-2） |
| 流水线测试覆盖 | ✓ 通过（6 个单元测试 + E2E 验证） |
| 文档同步 | ✓ 通过 |

### 总结

整体设计优秀：tiered reservation 解决了 #759 的 ring buffer 浪费问题，同时保留了 4 MiB 完整捕获能力（不退步 #763）。asm barrier 绕过 verifier 的 register spill 问题有清晰注释。测试覆盖充分。

**阻塞项**：Finding #1（`unwrap()` in non-test code）需按 AGENTS.md 硬性规则修复。其余为建议性改进。

src/agentsight/src/bpf/sslsniff.bpf.c
```diff
@@ -364,8 +359,10 @@ int BPF_URETPROBE(probe_SSL_do_handshake_exit) {
     if (ret <= 0)  // handshake failed
         return 0;
 
-    /* reserve space in ring buffer */
-    struct probe_SSL_data_t *data = bpf_ringbuf_reserve(&rb, sizeof(*data), 0);
+    /* Handshake records carry no payload: reserve ONLY the header (no buf), so a
+     * handshake costs ~one header in the ring instead of the prior 4 MiB reserve. */
+    struct probe_SSL_data_t *data =
+        bpf_ringbuf_reserve(&rb, __builtin_offsetof(struct probe_SSL_data_t, buf), 0);
```
chengshuyi, line 365: **[eBPF — 低优先级]** `__builtin_offsetof` 作为 `bpf_ringbuf_reserve` 的 size 参数。

clang 在 BPF target 下会将 `__builtin_offsetof(struct probe_SSL_data_t, buf)` 折叠为立即数（编译时常量），所以 5.8 verifier 应该没问题。PR 描述中提到 E2E 在 6.6 上验证通过。

建议在注释中补充一句："`__builtin_offsetof` is folded to an immediate by clang, satisfying the verifier's constant-size requirement on all supported kernels (>= 5.8)."，方便后续维护者理解为何这里可以不用 `sizeof`。

非阻塞，纯文档建议。

src/agentsight/src/bpf/sslsniff.bpf.c
```diff
@@ -78,8 +78,67 @@ int BPF_UPROBE(probe_SSL_rw_enter, void *ssl, void *buf, int num) {
...
+        int _rc = (src_) ? bpf_probe_read_user(&_d->buf, _n, (const char *)(src_)) : -1; \
+        if (_rc) { _d->buf_filled = 0; _d->buf_size = 0; }                      \
+        else     { _d->buf_filled = 1; _d->buf_size = _n; }                     \
+        bpf_ringbuf_submit(_d, 0);                                              \
+    } while (0)
+
+/* Pick the smallest tier (and its record type) that holds `len_` and emit one. */
+#define SSL_EMIT_TIERED(src_, len_, rw_, ts_, delta_, pid_, tid_, uid_, ssl_, ishs_)       \
+    do {                                                                                   \
+        u32 _l = (u32)(len_);                                                              \
```
chengshuyi, line 130: **[eBPF — 低优先级]** `SSL_EMIT_ONE` 宏无 `_n > 0` 防御。

当前所有调用路径在外层已保证 `len > 0`（`if (len <= 0) return 0`），所以无实际风险。但如果未来有人以 `len_=0` 误调此宏，`bpf_probe_read_user` 以 size=0 调用的行为在不同内核版本上不一致。

可选加固：
```c
if (_n > 0) {
    int _rc = (src_) ? bpf_probe_read_user(&_d->buf, _n, (const char *)(src_)) : -1;
    if (_rc) { _d->buf_filled = 0; _d->buf_size = 0; }
    else     { _d->buf_filled = 1; _d->buf_size = _n; }
} else {
    _d->buf_filled = 0; _d->buf_size = 0;
}
```

或在宏注释中标明前置条件 `/* Precondition: len_ > 0 */`。非阻塞。

jfeng18:
Thanks for the thorough 6-dimension review @chengshuyi. Fixes in f296410:

**#1 (unwrap in `from_bytes`) — fixed.** Replaced the three closures with direct `from_ne_bytes([...])` array builds (value-identical, native endianness); `fmt` + `clippy --all-targets -D warnings` + the from_bytes decode tests pass on 1.89. Heads-up: `u64_at` reads 8 bytes, so it needs an 8-element array — the 4-element form in the suggestion only fits `u32`/`i32`. For the record these `unwrap`s were actually infallible and OOB-safe (the `data.len() < hdr` guard plus `buf` being the last struct member prove every header read is in-bounds), so it was a hard-rule/lint fix rather than a latent panic — but it's a §0 rule, so it's gone.

**#2 (offsetof reserve comment) — holding off.** The reasoning is already documented in `sslsniff.h:13-15` (names `offsetof(probe_SSL_data_t, buf)` as the compile-time-constant reserve size) plus the local comment at `sslsniff.bpf.c:362-363`, so a new one would duplicate it. Happy to append a half-line there if you'd prefer.

**#3 (`_n > 0` guard) — agree there's no risk; declining the guard.** Both call sites gate on `if (len <= 0) return 0` and clamp to `[1, TIER]`, so `_n == 0` is unreachable. And `bpf_probe_read_user` is a benign no-op at size 0 anyway: in our 6.6 tree it's `ARG_CONST_SIZE_OR_ZERO` and `copy_from_user_nofault` short-circuits — stable since v5.5, below our 5.8 floor. Adding the branch would be unreachable code inside the `asm volatile`-tuned macro (extra conditionals risk perturbing the clang-DCE-vs-verifier balance). Let me know if you disagree.

Also removed the now-unused `from_bpf` and repointed the eBPF event-routing doc (`ebpf-probes.md`) to `from_bytes` — this PR is what orphaned it, and it still assumed the pre-tiering layout, so cleaned it up here rather than deferring.

chengshuyi (approved):

PR #1009, opened 2026-06-18
feat(anolisa): rpm drift status with repair and forget commands

### LINKED ISSUE #960: [anolisa] feat(anolisa): detect RPM state drift and provide repair path

ikunkun-sys:
## Description

Adds the final piece of the RPM lifecycle loop: `anolisa status` now adjudicates each in-state RPM component against the live rpmdb, and two new commands let the user reconcile what it reports.

- `status` compares a tracked RPM component's recorded EVR against rpmdb. A manual `dnf update` renders the component as `drifted`; a manual `rpm -e` renders it as `missing`. ANOLISA never silently reinstalls — it reports the divergence and points at a remediation.
- `anolisa repair <component>` refreshes ANOLISA state from rpmdb (version / EVR / arch / source repo) when the package is still installed, and backfills metadata for legacy rows. It runs no `dnf` / `rpm` mutation.
- `anolisa forget <component>` drops the ANOLISA state record with no package operation and no file removal.

## Design Note

`drifted` and `missing` are computed wire-status labels rendered at `status` time — alongside the existing `observed` label and the integrity/manifest health escalations — **not** new `ObjectStatus` variants. Drift is a comparison between persisted EVR and live rpmdb EVR, so persisting it would contradict the invariant that `status` is strictly read-only and never writes `installed.toml`. No schema bump.

Drift adjudication overrides only a clean projection (`installed` / `adopted`); it never demotes a record that integrity/manifest health already escalated to `failed` / `degraded`, nor a deliberately `disabled` one, since those carry a more-severe signal a drift label must not mask.

`repair` reuses the injected-`PackageQuery` pattern (so it can run without dnf), the adopt candidate chain for legacy metadata backfill, and the same lock + reload-and-revalidate persistence as `update`. `forget` mirrors `uninstall`'s adapter-claim guard. `Ownership::label()` was centralized in `anolisa-core` so every command renders the provenance string identically.

## Ship Note

- `repair` supports RPM-backed components only; a raw component returns `NOT_IMPLEMENTED`.
- `repair` on a package that is gone from rpmdb refuses and redirects to `forget` — it cannot conjure a removed package back.
- `forget` performs no package operation: for RPM the package stays installed; for raw the owned files remain on disk (use `uninstall` for file teardown).
- Out of scope, left for later issues: `uninstall --remove-system-package`, migration / downgrade flows.

## Test

- Drift unit tests (injected `PackageQuery` fakes): EVR-match → no drift, EVR-mismatch → `drifted`, `rpm -e` → `missing`, unexpected/multi-version output → drift carrying the backend's own detail, missing tooling → recorded status preserved, an already-escalated (`failed`) record is never demoted, and non-RPM rows are untouched.
- `repair`: drift refresh updates EVR and preserves ownership, keep-prior-source-repo on a failed origin lookup, missing-package → forget pointer, unexpected-output refusal, missing tooling, raw → `NOT_IMPLEMENTED`, absent → `INVALID_ARGUMENT`, dry-run writes nothing, legacy metadata backfill.
- `forget`: drops the object and records the op, unknown component → `INVALID_ARGUMENT`, refuses with an enabled adapter claim, dry-run leaves state untouched.
- Quality gate green: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (anolisa-cli 215 tests), `cargo doc --workspace --no-deps`.
- Also exercised end-to-end against a real host rpmdb: seeded a stale EVR → `status` `drifted` → `repair` refreshed to live EVR → `adopted`; uninstalled the package → `missing` → `repair` refused toward `forget` → `forget` dropped the state.

Closes #960.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/forget.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/forget.rs
@@ -1,0 +1,542 @@
+//! `anolisa forget <component>` — drop a component's ANOLISA state record
+//! without touching the underlying package or files.
+//!
+//! `forget` is the escape hatch for stale state: after a manual `rpm -e` (the
+//! `missing` case from `anolisa status`), or whenever the operator wants ANOLISA
+//! to stop tracking a component, `forget` removes the `installed.toml` object
+//! and records the operation. It performs **no** package operation — no
+//! `dnf remove`, no `rpm -e` — and deletes no files. An observed/managed RPM
+//! stays installed in rpmdb; a raw component's owned files stay on disk (use
+//! `anolisa uninstall` to remove those).
+
+use chrono::{SecondsFormat, Utc};
+use clap::Parser;
+use serde::Serialize;
+
+use anolisa_core::central_log::{CentralLog, LogKind, LogRecord, LogStatus, Severity};
+use anolisa_core::lock::InstallLock;
+use anolisa_core::state::{ObjectKind, OperationRecord};
+
+use crate::color::Palette;
+use crate::commands::common;
+use crate::context::CliContext;
+use crate::response::{CliError, render_json};
+
+/// Command label for JSON envelopes and error routing.
+const COMMAND: &str = "forget";
+
+/// Arguments for `anolisa forget <component>`.
+#[derive(Debug, Parser)]
+pub struct ForgetArgs {
+    /// Component whose ANOLISA state record should be dropped
+    #[arg(value_name = "COMPONENT")]
+    pub component: String,
+}
+
+/// Wire shape for a `forget <component>` result (`--json`) and its dry-run
+/// preview.
+#[derive(Serialize)]
+struct ForgetPayload {
+    component: String,
+    /// Provenance label of the dropped record, for the audit trail.
+    ownership: &'static str,
+    install_mode: String,
+    /// Whether the state record was actually removed (false on dry-run).
+    forgotten: bool,
+    dry_run: bool,
+    /// `None` on dry-run (nothing recorded).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    operation_id: Option<String>,
+}
+
+/// Dispatch `forget <component>`: drop the ANOLISA state record, run no package
+/// operation.
+///
+/// # Errors
+///
+/// Returns [`CliError`] when the component is absent, still has enabled adapter
+/// receipts, or the state write fails.
+pub fn handle(args: ForgetArgs, ctx: &CliContext) -> Result<(), CliError> {
+    let target = args.component.as_str();
+    let command = format!("forget {target}");
+    let installed = common::load_installed_state(ctx, COMMAND)?;
+
+    let obj = installed
+        .find_object(ObjectKind::Component, target)
+        .ok_or_else(|| CliError::InvalidArgument {
+            command: command.clone(),
+            reason: format!(
+                "component '{target}' is not installed — nothing to forget (run `anolisa status` to see what is tracked)"
+            ),
+        })?;
+    let ownership_label = obj.effective_ownership().label();
+
+    // Adapter receipts must be released before the component is dropped:
+    // silently orphaning a registered plugin is worse than refusing. This guard
+    // is a fast-fail and the dry-run preview; `persist_forget` re-checks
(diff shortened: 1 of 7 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/forget.rs
```diff
@@ -0,0 +1,484 @@
...
+    let layout = common::resolve_layout(ctx);
+    let _lock = InstallLock::acquire(&layout.lock_file).map_err(|err| CliError::Runtime {
+        command: command.to_string(),
+        reason: format!("failed to acquire install lock: {err}"),
+    })?;
+    let mut state = common::load_installed_state(ctx, command)?;
+
+    // Re-validate under the lock: a concurrent uninstall/forget may have already
+    // dropped the object between the pre-lock read and here.
+    if state
```
kongche-jbw: [P1] `forget` revalidates that the component still exists after acquiring the install lock, but it does not re-check adapter claims in the locked state. A concurrent `adapter enable` between the pre-lock read and this locked removal can leave an orphaned adapter receipt after the component object is deleted. Please move the claim check into this locked section as well (load state under lock -> find component -> check `adapter_claims_for_component()` -> remove), and ideally return the ownership label from the locked object so the response does not rely on the pre-lock read.

src/anolisa/crates/anolisa-cli/src/commands/tier1/forget.rs
```diff
@@ -0,0 +1,484 @@
...
+        color.muted(format!("({})", payload.ownership)),
+    );
+    println!(
+        "    {} ANOLISA stopped tracking this component; no package operation was performed",
+        color.label("note:"),
+    );
+    // Tailor the residue reminder to what forget deliberately left behind.
+    if payload.ownership == "raw-managed" {
+        println!(
+            "    {} any ANOLISA-owned files remain on disk; run 'anolisa uninstall {}' to remove them",
```
kongche-jbw: [P2] After `forget` succeeds for a raw-managed component, this hint points users to `anolisa uninstall`, but the state object and file inventory have already been removed. In that state `uninstall` usually cannot perform the safe file teardown anymore. Please make the wording explicit that files may remain and ANOLISA no longer has the inventory after `forget`, or tell users to choose `uninstall` instead when they want file removal.

src/anolisa/crates/anolisa-cli/src/commands/tier1/forget.rs
ikunkun-sys: Confirmed the race. The authoritative claim guard now runs inside `persist_forget` under the install lock (reload → check `adapter_claims_for_component()` → remove), so an `adapter enable` landing between the pre-lock read and the removal can no longer orphan a receipt. The check in `handle` is kept only as a fast-fail / dry-run preview, and the response now takes its ownership label from the object removed under the lock rather than the pre-lock read. Added `persist_forget_rechecks_adapter_claim_under_lock`, which calls `persist_forget` directly (bypassing the fast-fail) so a regression that drops the locked check fails the test while the `handle`-level test would still pass. Fixed in 006a6b2.

src/anolisa/crates/anolisa-cli/src/commands/tier1/forget.rs
ikunkun-sys: Agreed — `forget` drops the file inventory, so `uninstall` can no longer perform the teardown. Reworded the note: ANOLISA-owned files remain on disk, the inventory is gone so `uninstall` can no longer remove them, delete them manually, and use `uninstall` instead of `forget` next time you want ANOLISA to remove files. Fixed in 006a6b2.

kongche-jbw (approved):
lgtm thx

PR #1013, opened 2026-06-18
feat(anolisa): add explicit adopt command for system RPMs

### LINKED ISSUE #961: [anolisa] feat(anolisa): add adopt and repair CLI commands for component lifecycle

ikunkun-sys:
## Description

Adds `anolisa adopt <component>`, the explicit counterpart to install's implicit system-mode adoption (#961). It reads rpmdb and records a unique installed RPM as `rpm-observed` ANOLISA state, running no `dnf`/`rpm` transaction. The command is system-scope only.

Behavior:
- Unique installed RPM → recorded as `rpm-observed` / `Adopted`.
- Already `rpm-observed` → idempotent refresh (EVR/arch/source_repo).
- Already `rpm-managed` → refused, pointing at `repair`.
- Already `raw-managed` → refused, pointing at `uninstall`.
- Absent / ambiguous / multi-version → refused with actionable messages.
- `--package <name>` pins the RPM when a component maps to several candidates.

The `repair` command shipped earlier already satisfies #961's repair acceptance criteria; this PR adds a regression test pinning that a managed refresh keeps ownership/backend, but makes no change to repair's logic.

## Design Note

The adopt core is reused from the install path rather than duplicated: `probe_rpm_situation`, `execute_adopt`, and `RpmSituation` become `pub(crate)`, and `probe_rpm_situation` is decoupled from `RepoConfig` (takes `Option<&BackendConfig>`) so `adopt` resolves the rpm backend best-effort.

Ownership safety: `adopt` has a pre-lock gate that refuses a component already tracked under a non-observed ownership. Because `rpm-managed` and `rpm-observed` share the `"rpm"` backend label, the existing lock-held `ensure_component_backend_compatible` cannot tell them apart, so `execute_adopt` now also re-checks ownership **under the lock** and refuses to downgrade a managed component to `rpm-observed`. This closes a TOCTOU window where a concurrent managed install could land between the pre-lock read and the lock, and as a bonus fixes the same latent silent-downgrade on `install` re-installing an `rpm-managed` component.

## Ship Note

- Ownership migration stays out of scope: `adopt` refuses to convert `raw-managed`/`rpm-managed` → `rpm-observed` rather than performing it (left for a follow-up).
- Known consistency item: `adopt`/`repair`/`status` share a best-effort `RepoConfig::load(&layout).ok()`. A **missing** `repo.toml` correctly degrades to "no rpm backend config", but a **malformed** one is silently ignored (its `[backends.rpm].package_map` tier is dropped) rather than surfaced as an error the way `install` does. The right fix is a shared loader helper applied to all three commands; deferred to keep this PR scoped.
- `execute_adopt` mints `op-install-…` operation ids even when invoked as `adopt`; the audit record's `command` field still reads `adopt <component>`, and no consumer keys off the id prefix, so this is cosmetic.

## Test

- Unit (adopt): success → `rpm-observed`/`Adopted` with correct `rpm_metadata` + operation record; idempotent refresh; refuse raw-managed / rpm-managed / user-mode / absent / ambiguous / multi-version; `--dry-run` writes nothing; `--package` override; CLI parse.
- Unit (install): `execute_adopt` refuses to downgrade a concurrent `rpm-managed` record under the lock; re-installing an `rpm-managed` component via the full install entrypoint refuses instead of downgrading; JSON envelope reports the bare verb (`adopt`, not `install`).
- Unit (repair): regression test pinning that a managed refresh keeps `ownership`/`install_backend`.
- Real-host rpmdb end-to-end: `adopt` an installed package via `--package` → `status` shows `adopted`/`rpm-observed` → idempotent re-adopt → `repair` refreshes → `forget` drops state (RPM left installed); plus refusal paths (user mode, absent) and `--json` envelope consistency.

Closes #961

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
@@ -186,7 +186,7 @@ struct InstallResultPayload {
 /// RPM adopt in its batch summary (§7.5). The dry-run vs real distinction is
 /// layered on by the caller from [`CliContext::dry_run`].
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum InstallOutcome {
+pub(crate) enum InstallOutcome {
     /// A raw install (downloaded + placed files, or its dry-run preview).
     Installed,
     /// An existing system RPM recorded as `rpm-observed` (or its dry-run
@@ -301,8 +301,14 @@ fn handle_one_with_exec(
             // layer 2, rather than being rejected by the raw trunk.
             (label.to_string(), BackendSource::ExistingState)
         } else if ctx.install_mode == InstallMode::System {
-            let situation =
-                probe_rpm_situation(&component, &args, &repo_config, ctx, exec.query, &command)?;
+            let situation = probe_rpm_situation(
+                &component,
+                args.package.as_deref(),
+                repo_config.backends.get("rpm"),
+                ctx,
+                exec.query,
+                &command,
+            )?;
             if matches!(situation, RpmSituation::Absent { .. }) {
                 // Absent + no `--backend`: fall through to the default backend.
                 // If that is `rpm`, layer 2 re-probes and delegates a `dnf
@@ -435,7 +441,7 @@ fn handle_raw_install(
 // ── rpm adopt path (#958) ───────────────────────────────────────────
 
 /// Result of probing whether `component` is present as a system RPM (§5/§7.1).
-enum RpmSituation {
+pub(crate) enum RpmSituation {
     /// Exactly one candidate package name, installed once — ready to adopt.
     Adoptable {
         /// Resolved package name (the candidate that hit rpmdb).
@@ -469,18 +475,17 @@ enum RpmSituation {
 /// back to raw rather than treat it as [`Absent`].
 ///
 /// [`Absent`]: RpmSituation::Absent
-fn probe_rpm_situation(
+pub(crate) fn probe_rpm_situation(
     component: &str,
-    args: &InstallArgs,
-    repo_config: &RepoConfig,
+    cli_override: Option<&str>,
+    rpm_backend: Option<&BackendConfig>,
     ctx: &CliContext,
     query: &dyn PackageQuery,
     command: &str,
 ) -> Result<RpmSituation, CliError> {
     let manifest = load_optional_manifest(ctx, component);
-    let rpm_backend = repo_config.backends.get("rpm");
     let candidates = match rpm_package_candidates(
-        args.package.as_deref(),
+        cli_override,
         manifest.as_ref(),
         rpm_backend,
         query,
@@ -601,7 +606,14 @@ fn route_rpm_adopt(
 
     let situation = match situation {
         Some(s) => s,
-        None => probe_rpm_situation(component, args, repo_config, ctx, exec.query, command)?,
+        None => probe_rpm_situation(
+            component,
+            args.package.as_deref(),
+            repo_config.backends.get("rpm"),
+            ctx,
+            exec.query,
+            command,
+        )?,
     };
 
     match situation {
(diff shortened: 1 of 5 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -711,6 +723,26 @@ fn execute_adopt(
...
 
+    // Backend compatibility is necessary but not sufficient: rpm-managed and
+    // rpm-observed share the "rpm" backend label, so the check above passes for
+    // a component ANOLISA actively manages. Adopt may only create a new record
+    // or refresh an existing rpm-observed one — it must never downgrade a managed
+    // component to observed and silently drop ANOLISA's removal authority
+    // (`owns_removal`). `adopt`'s pre-lock gate refuses this for the common case;
+    // re-checking here under the lock closes the window where a concurrent
+    // managed install lands between that read and this acquisition.
+    if let Some(existing) = state.find_object(ObjectKind::Component, component)
```
kongche-jbw: [P1] This locked check prevents downgrading rpm-managed/raw-managed records, but an existing `rpm-observed` record can still be silently repointed to a different RPM package. For example, a component already adopted from `pkg-a` can be re-run with `adopt <component> --package pkg-b`; this passes the ownership check and the later `upsert_object()` overwrites `rpm_metadata.package_name` to `pkg-b`. That is a package-identity migration, not an idempotent refresh. Please reject under the lock when an existing rpm-observed object has a non-empty `rpm_metadata.package_name` different from the probed `package`/`info.name`, and ask the user to `forget` then adopt if they really want to switch packages. A regression test for observed package A + `--package` B would pin this.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -806,17 +838,29 @@ fn execute_adopt(
...
+/// is "install" through the install trunk and "adopt" through the explicit
+/// command — not a hardcoded "install".
+fn adopt_envelope_verb(command: &str) -> &str {
+    match command.split(' ').next() {
+        Some(verb) if !verb.is_empty() => verb,
+        _ => COMMAND,
+    }
+}
+
+fn render_adopt(ctx: &CliContext, command: &str, payload: &AdoptResultPayload) {
```
kongche-jbw, line 911: [P2] This renderer still prints `Adopted as rpm-observed` even when `payload.dry_run` is true. The previous line says nothing was recorded, so the follow-up sentence reads like the state write already happened. Please branch the wording here: dry-run should say something like `Would adopt as rpm-observed...`, while the real run can keep `Adopted...`.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
ikunkun-sys: Good catch — fixed in 8dabc33. Under the lock, when an existing record is rpm-observed we now also refuse if its `rpm_metadata.package_name` is non-empty and differs from the probed `info.name`, so `upsert_object` can no longer silently repoint the package identity. The error names both packages and steers the user through `anolisa forget <component>` → adopt to switch packages on purpose; an idempotent same-package re-adopt still refreshes. Added regression test `adopt_refuses_repointing_observed_to_different_package` (observed `anolisa-copilot-shell` + `--package anolisa-other` → INVALID_ARGUMENT, state and EVR unchanged).

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
ikunkun-sys: Fixed in 8dabc33. `render_adopt` now branches the action line on `payload.dry_run`: dry-run prints `Would adopt as rpm-observed. ANOLISA will not replace it with raw.` while the real run keeps `Adopted as rpm-observed. ...`, so it no longer contradicts the `(dry-run — nothing recorded)` suffix on the line above.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -688,7 +700,7 @@ fn execute_adopt(
     };
 
     if ctx.dry_run {
```
kongche-jbw, line 750: [P2] The real execution path now rejects repointing an existing rpm-observed component from package A to package B, but dry-run returns here before that identity check runs. That means `adopt <component> --package pkg-b --dry-run` can print `Would adopt...` while the real command would fail because state already records `pkg-a`. Please run the same package-identity validation before this dry-run return (non-locked is fine for preview; keep the locked check below for TOCTOU), and add a regression test for observed package A + `--package` B + `--dry-run` returning `INVALID_ARGUMENT`.

ikunkun-sys: Fixed in 2a6e543. Extracted the package-identity check into a shared `refuse_observed_repoint(...)` and now run it *before* the dry-run return (non-locked read of installed state, for preview/fast-fail), keeping the locked call below for TOCTOU. So `adopt <component> --package pkg-b --dry-run` now returns INVALID_ARGUMENT instead of printing `Would adopt...`, matching the real run. Added regression test `adopt_dry_run_refuses_repointing_observed_to_different_package` (observed `anolisa-copilot-shell` + `--package anolisa-other` + `--dry-run` → INVALID_ARGUMENT, state untouched).

kongche-jbw (approved):
lgtm thx

PR #1018, opened 2026-06-18
feat(anolisa): enforce safe uninstall semantics for rpm-observed components

### LINKED ISSUE #962: [anolisa] feat(anolisa): enforce safe uninstall semantics for rpm-observed components

ikunkun-sys:
## Description

Makes `anolisa uninstall` ownership-driven so a preinstalled system RPM is never removed by default. Today uninstall is ownership-blind: every component goes through core's `LifecyclePlan`/`execute_plan`, which only removes ANOLISA-owned files and drops state — it never calls `dnf`. That made `rpm-observed` *accidentally* safe (no owned files, so only state was dropped) with no way to explicitly remove the package, and left `rpm-managed` packages in rpmdb even though they should be delegated to `dnf remove`.

New behavior, keyed on `effective_ownership()`:

| ownership | `uninstall` | `uninstall --remove-system-package` |
|---|---|---|
| `raw-managed` | remove ANOLISA-owned files + drop state (**unchanged**) | no-op for raw → warn, proceed as default |
| `rpm-managed` | `dnf remove` + drop state (`owns_removal() == true`) | same as default (flag redundant, harmless) |
| `rpm-observed` | **drop state only, no dnf** (`owns_removal() == false`) | `dnf remove` + drop state |

Acceptance: ① rpm-observed never runs `dnf remove` by default; ② removing an observed system package requires an explicit `--remove-system-package`; ③ dry-run states whether the package will be removed (`package_removal`: `dnf remove` / `state only` / `already absent`).

## Design Note

- **dnf delegation stays at the CLI layer.** Core's `lifecycle.rs` does not shell out to dnf. The RPM path branches inside `uninstall.rs`, mirroring `update.rs`'s `update_rpm_component` pattern (inject `query`/`txn`/`is_root`, dnf delegation + root gating + error mapping). Raw components keep the existing `LifecyclePlan`/`execute_plan` path with zero behavior change; only `Uninstall` on an `is_rpm()` component is rerouted (`--purge` is untouched, still plan-gated).
- **Removal decision recomputed under the lock.** `remove_package = locked_ownership.owns_removal() || --remove-system-package` is evaluated from the *locked* ownership, never the pre-lock read. If a concurrent op flipped the component from rpm-managed to rpm-observed under the same package name, the stale pre-lock value would still say "remove" and a default uninstall would wrongly `dnf remove` a preinstalled RPM — the exact guarantee rpm-observed exists to provide. The locked path also re-checks that `rpm_metadata.package_name` is unchanged (TOCTOU identity guard), consistent with adopt's `refuse_observed_repoint`.
- **Lock held across the dnf remove (unlike install/update, which release before dnf).** The destructive path runs entirely under the install lock: `acquire → reload → in-lock adapter-claim re-check → (optional dnf remove) → remove_object → record op → save → central log`. The adapter guard must take effect *before* the irreversible `dnf remove`, otherwise a concurrent `adapter enable` could slip in, remove the package, then refuse to drop state — orphaning a registered plugin. A comment documents why this differs from install/update.
- **Tooling-missing guidance is ownership-aware on both probes.** Whether rpm/dnf is found absent by the rpmdb query *or* by `dnf remove` itself (rpm present, dnf absent), the error steers an owns-removal component at `anolisa forget` and an rpm-observed removal at dropping `--remove-system-package` — both paths share one `tooling_missing_err` helper so the guidance can't diverge.
- Adds `PackageTransaction::remove` (delegates to `dnf remove`); a `PackageDisposition` three-state enum keeps "already absent" (the package already gone via a manual `rpm -e`) distinct from "intentionally kept" and "removed".

## Ship Note

- **`rpm-managed` default is now `dnf remove`** (previously state-only, leaving the package in rpmdb). Unified through `owns_removal()`. In scope, but flagged as a behavior change; happy to split it out if you'd rather keep this PR to rpm-observed safety only.
- **Missing handled gracefully:** with removal intended, the package is `query_installed`-checked first; if it's already gone, that's not an error — drop the stale state and report `already absent`.
- **Known limitation — the save-after-remove window:** `dnf remove` runs before `state.save()`, so a state-write failure *after* a successful removal leaves the package gone but the ANOLISA record present — a Missing drift the user can resolve with `status` / `forget` / `repair`. This is inherent to a two-phase destructive op (an external rpmdb change cannot be made atomic with a local state write), so it is flagged here rather than papered over.
- **`--remove-system-package` does not stack with `--purge`** (purge path unchanged, still gated); on a raw component the flag warns and proceeds as default.

## Test

- **Unit (uninstall):** rpm-observed default drops state with `remove_calls == 0`; `--remove-system-package` + root runs `dnf remove`; same flag non-root refuses with a sudo hint and no dnf; rpm-managed default delegates `dnf remove`; dry-run touches nothing and reports the right `package_removal`; already-absent drops state only; dnf failure keeps state; rpm-query tooling missing points at `forget` (managed) vs dropping the flag (observed); rpm-present / dnf-absent (the `txn.remove` `CommandMissing` branch) gives the same `forget` guidance instead of the generic message; raw + flag warns but succeeds; enabled-adapter refusal (incl. in-lock re-check); removal decision recomputed under the lock; package-metadata-missing points at `repair`; decision matrix + disposition label/intent mapping; flag parse.
- **Unit (platform):** `RpmTransaction::remove` success → ok; non-zero exit → `TransactionFailed`.
- **Full gate green:** `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, `cargo doc --workspace --no-deps`.
- **Real-host rpmdb (dry-run, no root):** a seeded rpm-observed component shows `package removal: state only`; with `--remove-system-package --dry-run` shows `dnf remove`. The real removal path is covered by the `FakeRpm` unit tests.

Closes #962

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/uninstall.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/uninstall.rs
@@ -1,17 +1,33 @@
-//! `anolisa uninstall <COMPONENT>` (with optional `--purge`).
+//! `anolisa uninstall <COMPONENT>` (with optional `--purge` /
+//! `--remove-system-package`).
 //!
-//! The CLI face of [`anolisa_core::execute_plan`] for component
-//! teardown. Two surfaces:
+//! Teardown is **ownership-driven** (`raw_rpm_lifecycle_proposal.md` §11):
 //!
-//!   * `--dry-run` — render the [`LifecyclePlan`] (human or JSON) and
-//!     return without touching the filesystem.
-//!   * default — execute the plan: only ANOLISA-owned files are
-//!     removed; external residue is preserved and surfaced as
-//!     warnings.
+//!   * `raw-managed` — the CLI face of [`anolisa_core::execute_plan`]:
+//!     only ANOLISA-owned files are removed, external residue is kept.
+//!     Unchanged by the RPM lifecycle work.
+//!   * `rpm-managed` — delegates `dnf remove` then drops ANOLISA state
+//!     (`Ownership::owns_removal()` is `true`).
+//!   * `rpm-observed` — a preinstalled system RPM: the default uninstall
+//!     drops **only** the ANOLISA state record and never runs dnf
+//!     (`owns_removal()` is `false`). Removing the system package
+//!     requires the explicit `--remove-system-package` override.
 //!
-//! `--purge` widens the scope from "uninstall" to "uninstall + drop
+//! The removal decision collapses to
+//! `ownership.owns_removal() || --remove-system-package`. The RPM path
+//! mirrors `update`'s dnf delegation (injected query/transaction, root
+//! gate) and `forget`'s state drop; the raw path is untouched.
+//!
+//! Two surfaces apply to every ownership:
+//!
+//!   * `--dry-run` — render the plan (human or JSON), touching nothing.
+//!     For RPM components it states whether package removal will happen.
+//!   * default — execute.
+//!
+//! `--purge` widens the raw scope from "uninstall" to "uninstall + drop
 //! ANOLISA-owned config/cache/state fragments". External modifications
-//! are always refused regardless of `--purge`.
+//! are always refused regardless of `--purge`. `--purge` keeps its
+//! existing (plan-only) path and is independent of the RPM routing.
 //!
 //! `--force` is parsed today as a wire stub (the spec calls it out)
 //! but the executor does not yet branch on it. We surface a warning so
@@ -36,11 +52,20 @@
 //! `anolisa-core::lifecycle::check_destructive_execute_gate`; see the
 //! docstring there for the lift conditions.
 
+use chrono::{SecondsFormat, Utc};
 use clap::Parser;
 
+use anolisa_core::central_log::{CentralLog, LogKind, LogRecord, LogStatus, Severity};
+use anolisa_core::lock::InstallLock;
+use anolisa_core::state::{OperationRecord, Ownership};
 use anolisa_core::{
     LifecycleError, LifecycleOperation, LifecycleOutcome, LifecyclePlan, ObjectKind, execute_plan,
 };
+use anolisa_platform::pkg_query::{PackageQuery, PackageQueryError};
+use anolisa_platform::pkg_transaction::{PackageTransaction, PackageTransactionError};
+use anolisa_platform::privilege;
+use anolisa_platform::rpm_query::RpmPackageQuery;
+use anolisa_platform::rpm_transaction::RpmTransaction;
 
 use crate::color::Palette;
 use crate::commands::common;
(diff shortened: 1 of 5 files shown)


kongche-jbw:
整体设计方向是对的：PR 把 `uninstall` 从原来的文件计划驱动调整为 ownership 驱动，和 raw/rpm lifecycle 设计里的矩阵一致。`rpm-observed` 默认只 drop ANOLISA state、不删除系统 RPM，`rpm-managed` 默认委托 `dnf remove`，这个行为边界在 Ship Note 里也说清楚了。

实现上我重点看了真实执行路径：在 install lock 内重新 load state、重新计算 locked ownership、重新校验 package identity 和 adapter claim，然后再决定是否执行 `dnf remove`。这个顺序能覆盖并发修改下的关键安全点，也符合 adapter 设计里首版阻止 uninstall 而不是自动级联 disable 的原则。测试覆盖了 observed default、显式 remove、managed default、already absent、非 root、dnf failure、adapter guard 和 locked recompute，核心路径比较完整。

一个建议补到 PR description / Ship Note：`dnf remove` 成功后才保存 ANOLISA state，如果 `state.save()` 失败，会留下“RPM 已删除但 state 仍存在”的 Missing drift。这个窗口不是阻塞问题，后续可以通过 `status/forget/repair` 处理，但作为 destructive package transaction 的已知限制，写出来会更清楚。

src/anolisa/crates/anolisa-cli/src/commands/tier1/uninstall.rs
```diff
@@ -247,6 +343,463 @@ fn lifecycle_err_to_cli(command: &str, err: LifecycleError) -> CliError {
...
+        operation_id: Some(operation_id),
+    };
+    render_uninstall_rpm(ctx, &payload);
+    Ok(())
+}
+
+/// Map a [`PackageTransactionError`] from `dnf remove` onto a CLI runtime error.
+fn txn_remove_err(err: PackageTransactionError, command: &str) -> CliError {
+    match err {
+        PackageTransactionError::CommandMissing { command: bin } => CliError::Runtime {
```
kongche-jbw, line 711: [P2] This branch loses the ownership-aware fallback described in the PR body when `rpm` is available but `dnf` is missing. The `query_installed()` `CommandMissing` path distinguishes rpm-managed (`forget`) from rpm-observed + `--remove-system-package` (drop the flag), but `txn.remove()` maps `PackageTransactionError::CommandMissing` here to a generic “cannot remove the RPM package” message. Consider passing `component` / `locked_ownership` / `remove_system_package` into this mapper, or handling `CommandMissing` at the call site, so the `dnf`-missing case gives the same actionable guidance as the rpm-query-missing case.

ikunkun-sys: Good catch — fixed in da4de54. The `dnf`-missing case (rpm present, dnf absent, so the rpmdb query succeeds but `dnf remove` then reports `CommandMissing`) now gets the same ownership-aware guidance as the rpm-query-missing branch.

I didn't thread `component`/`locked_ownership`/`remove_system_package` through `txn_remove_err`: that mapper also handles `PermissionDenied`/`TransactionFailed`, which don't need the context, so coupling a generic error mapper to uninstall-specific state felt wrong. Instead I handled `CommandMissing` at the call site and pulled the guidance into one helper:

```rust
fn tooling_missing_err(command, bin, package, component, locked_ownership) -> CliError {
    let alt = if locked_ownership.owns_removal() {
        format!("run `anolisa forget {component}` to drop ANOLISA state without a package operation")
    } else {
        "re-run without --remove-system-package to drop ANOLISA state only".to_string()
    };
    // "cannot remove '{package}': {bin} not found on PATH — install rpm/dnf, or {alt}"
}
```

Both the `query_installed` and the `txn.remove` `CommandMissing` branches route through it, so the two messages can't drift. Added a regression test `uninstall_rpm_managed_dnf_missing_points_at_forget` (rpm present, dnf absent → `forget` guidance, `remove_calls == 1`, state intact).

ikunkun-sys:
感谢 review,两点都已处理(da4de54):

1. **save-after-remove 窗口**:确实存在 —— `dnf remove` 在 `state.save()` 之前执行,save 失败会留下「包已删除但 ANOLISA state 仍在」的 Missing drift。已按建议补进 Ship Note,作为这个两阶段 destructive op 的已知限制(外部 rpmdb 变更无法与本地 state 写入原子化),后续可由 `status` / `forget` / `repair` 收敛,非阻塞。

2. **行级 P2(dnf 缺失时的引导)**:抽出共享 helper `tooling_missing_err`,让 rpmdb 查询缺失与 `dnf remove` 缺失两条路径给出同样的 ownership 感知引导(owns-removal → `forget`,rpm-observed → 去掉 `--remove-system-package`),不再回退到通用消息;并补了回归测试。细节见该行下的回复。

全量 gate(fmt / clippy `-D warnings` / `test --workspace` / doc)仍全绿。

kongche-jbw (approved):
lgtm thx

PR #1019, opened 2026-06-18
test(anolisa): add MVP coverage for image-preinstalled RPM lifecycle

### LINKED ISSUE #963: [anolisa] test(anolisa): add MVP coverage for image-preinstalled RPM lifecycle
## Goal
Add tests for the target MVP flow: image preinstalls RPM, ANOLISA adopts and manages lifecycle.

## Scope
- Test preinstalled RPM adoption with `default_backend = raw`.
- Test update path delegates to RPM backend and refreshes state.
- Test user-mode raw override shadows system RPM.
- Test uninstall forget behavior for `rpm-observed`.
- Test drift/missing status after simulated manual dnf/rpm changes.

## Acceptance Criteria
- MVP behaviors are covered by unit/integration tests without requiring real dnf in CI.
- Tests use fake package manager/rpmdb seams.
- Raw install tests remain passing.

ikunkun-sys:
## Description

Closes out the raw_rpm_lifecycle series (#956–#963) with MVP-flow coverage for the image-preinstalled RPM lifecycle: an image ships an RPM, ANOLISA adopts it and manages its lifecycle. The per-command behaviors already have unit tests from #956–#962; this PR adds one cross-command end-to-end test that feeds each command's *real* persisted output into the next on a single evolving fake rpmdb world, catching the state-contract and rpmdb-evolution regressions the per-command tests assume away by seeding their own start state.

Scope coverage:
- ① adopt under `default_backend = raw` — adopts the preinstalled RPM as `rpm-observed` instead of fetching a raw artifact.
- ② update delegates to the RPM backend and refreshes the recorded EVR, keeping `rpm-observed` ownership.
- ④ rpm-observed uninstall (no `--remove-system-package`) drops only ANOLISA state and leaves the system RPM; `forget` drops state with no package operation.
- ⑤ drift / missing status after a simulated out-of-band `dnf`/`rpm` change.
- ③ user-mode raw shadowing a system RPM — covered by the existing `update::tests::user_raw_override_does_not_touch_system_rpm` unit test (see Design Note).

This is a test-only PR; no production behavior changes.

## Design Note

The end-to-end test is an inline `#[cfg(test)]` module (`commands/tier1/mvp_lifecycle.rs`), because `anolisa-cli` is a bin-only crate (no `lib.rs`) whose private/`pub(crate)` items and test fakes are unreachable from an external `tests/` integration test.

To drive each command from one module, a few per-command injection seams are relaxed to `pub(crate)` (`adopt_with_query`, `update_component_with_deps`, `handle_with_deps`, `probe_rpm_drift`/`RpmDrift`, `handle_one_with_exec`/`RpmExec`); these are pure visibility widenings with no behavior change. A single `RpmWorld` implements both `PackageQuery` and `PackageTransaction` and mutates as commands act on it, so `update` advances the rpmdb view and `simulate_dnf_upgrade` models an out-of-band change the drift probe must catch.

The value is the cross-command state contract: the test feeds adopt's real written state into update, update's into the drift probe, and that into uninstall, so it guards the fields adopt writes against what update/status/uninstall read — something per-command tests cannot do, since each seeds its own start state.

## Ship Note

- Functional coverage of ①–⑤ already exists per-command (#956–#962); this PR's incremental value is the cross-command regression guard, not new functional coverage. The end-to-end module was deliberately scoped to the single highest-value chain; the `missing` and `forget` single-hop cases stay mapped to their existing unit tests (`status::probe_rpm_drift_detects_missing`, `forget::forget_drops_object_and_records_operation`).
- ③ is not re-done end-to-end: `common::resolve_layout` binds user mode to the real `$HOME`, so a command-layer user-mode test cannot be isolated without polluting the home dir or mutating process-global `XDG_*` (which would race other tests). It stays covered by the `update` unit test.
- Two real-environment observations surfaced during validation, out of #963 scope (noted, not fixed): the default component→RPM name convention (`anolisa-cosh`) does not match the real repo package (`copilot-shell`), so `adopt` needs `--package copilot-shell`; and `forget` on a component already gone from rpmdb still prints "the RPM package remains installed", which is inaccurate in the missing case.

## Test

- Unit / integration (CI, no real dnf): `cargo test -p anolisa-cli` — 261 passed, including the new `mvp_lifecycle::lifecycle_adopt_update_drift_then_uninstall` end-to-end test on a fake, evolving rpmdb world.
- Real Alibaba Cloud Linux 4 end-to-end (real dnf/rpmdb): built `anolisa 0.1.10` for x86_64, deployed it to a live Alinux 4 box, and ran the full MVP flow against the real `alinux4-agentic` repo (component `cosh` ↔ RPM `copilot-shell`, versions 2.0.0–2.5.0): `dnf install copilot-shell-2.3.0` → `adopt cosh --package copilot-shell` records `rpm-observed` (evr 2.3.0, `managed=false`) → `update` delegates to dnf, advancing both rpmdb and state EVR to 2.5.0 → out-of-band `dnf downgrade` to 2.4.0 → `status` reports `drifted` ("rpmdb reports 2.4.0 but ANOLISA state records 2.5.0") → `repair` reconciles state to 2.4.0 → `uninstall` (no `--remove-system-package`) drops state but leaves `copilot-shell` installed → re-adopt → `forget` drops state with no package operation → re-adopt → out-of-band `rpm -e` → `status` reports `missing`. Every step matched the fake-world test's assertions.

Closes #963

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/mvp_lifecycle.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/mvp_lifecycle.rs
@@ -1,0 +1,286 @@
+//! End-to-end MVP lifecycle coverage (#963).
+//!
+//! Drives a single image-preinstalled system RPM through the whole flow —
+//! adopt → update → drift status → uninstall — on **one** state file and
+//! **one** evolving fake rpmdb world. Where the per-command unit tests
+//! each seed their own hand-written start state, these feed *each command's real
+//! output* into the next, so they catch the cross-command regressions those
+//! tests assume away: a state-contract drift between what `adopt` writes and
+//! what `update`/`status`/`uninstall` read, and rpmdb-evolution effects that a
+//! static per-command fake cannot model.
+//!
+//! Scope (#963) → coverage:
+//! - ① adopt under `default_backend = raw` — [`lifecycle_adopt_update_drift_then_uninstall`] (install step)
+//! - ② update delegates to RPM + refreshes state — same test (update step)
+//! - ④ rpm-observed uninstall — same test (uninstall step); `forget` is covered
+//!   by `super::forget::tests::forget_drops_object_and_records_operation`
+//! - ⑤ drift status — same test (drift step); the `Missing` counterpart is
+//!   covered by `super::status::tests::probe_rpm_drift_detects_missing`
+//! - ③ user-mode raw shadows system RPM — covered by
+//!   `super::update::tests::user_raw_override_does_not_touch_system_rpm`; not
+//!   re-done here because `common::resolve_layout` binds user mode to the real
+//!   `$HOME`, so a command-layer user-mode test cannot be isolated without
+//!   polluting the home dir or mutating process-global `XDG_*` (which would race
+//!   other tests).
+
+use std::cell::{Cell, RefCell};
+
+use anolisa_core::state::{InstalledState, ObjectKind, Ownership};
+use anolisa_platform::fs_layout::FsLayout;
+use anolisa_platform::pkg_query::{PackageInfo, PackageQuery, PackageQueryError, PackageVersion};
+use anolisa_platform::pkg_transaction::{PackageTransaction, PackageTransactionError};
+
+use crate::commands::common;
+use crate::context::{CliContext, InstallMode};
+
+use super::install::{InstallArgs, InstallOutcome, RpmExec, handle_one_with_exec};
+use super::status::{RpmDrift, probe_rpm_drift};
+use super::uninstall::{UninstallArgs, handle_with_deps};
+use super::update::update_component_with_deps;
+
+/// An in-memory rpmdb + dnf that **mutates as commands act on it**, so one fake
+/// backs the whole flow: it answers [`PackageQuery`] (adopt probe, status drift)
+/// and runs [`PackageTransaction`] (update/remove).
+///
+/// `installed` is the live rpmdb view. A successful [`update`](PackageTransaction::update)
+/// advances it to `upgrade_to` (dnf moving the version forward); the
+/// `simulate_*` helpers model out-of-band `dnf`/`rpm` changes that ANOLISA state
+/// does not know about, which is exactly what the drift/missing probes must
+/// catch.
+struct RpmWorld {
+    package: String,
+    installed: RefCell<Option<PackageInfo>>,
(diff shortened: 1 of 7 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/mvp_lifecycle.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/// Adopt the preinstalled RPM, update it (rpmdb advances, state refreshes),
+/// observe drift after an out-of-band dnf upgrade, then uninstall (state
+/// dropped, system RPM left). Each step consumes the previous step's *real*
+/// persisted state and the same evolving rpmdb world.
+#[test]
+fn lifecycle_adopt_update_drift_then_uninstall() {
+    let (_tmp, ctx) = system_ctx_with_raw_repo();
+    let pkg = "anolisa-copilot-shell";
```
kongche-jbw: [P2] This E2E still uses the default component→RPM naming path (`copilot-shell` → `anolisa-copilot-shell`), while the real MVP flow described in the PR is `component = cosh` and `RPM = copilot-shell` with `--package copilot-shell`. That means a future regression where update/status/uninstall re-derive `anolisa-{component}` instead of consuming the adopted `rpm_metadata.package_name` could slip through. Consider making this test use `component = "cosh"`, `package = "copilot-shell"`, and `install_args("cosh").package = Some("copilot-shell".to_string())` so the cross-command contract actually proves package identity is carried from adopt into later commands.

ikunkun-sys: Good catch — adopted. The test now uses `component = "cosh"`, `package = "copilot-shell"` (no shared `anolisa-` prefix) and `install_args("cosh", Some("copilot-shell"))`. The `RpmWorld` answers only for `copilot-shell`, so update/status/uninstall now pass solely by consuming the adopted `rpm_metadata.package_name`; a regression that re-derives `anolisa-{component}` would miss the world and fail its internal package assertion. Also added an explicit `assert_eq!(meta.package_name, "copilot-shell")` right after adopt to pin the identity. This mirrors the real `cosh` ↔ `copilot-shell` mismatch the PR's real-host test exercised. Pushed as e03dee1.

kongche-jbw (approved):
lgtm

PR #1030, opened 2026-06-21
feat(cosh): show authenticated models in /model dialog

### LINKED ISSUE #1021: [cosh] feat: support quick switching between Qwen models in cosh
## Component
cosh

## Problem Statement
When using cosh with a Qwen/OpenAI-compatible provider, I cannot quickly switch between different models such as qwen-3.5 and qwen-3.6 from a selectable list.

The current model switch experience depends on models configured in settings.modelProviders, or on a single model saved through /auth. If the model I want is not already configured, I need to go back to the provider configuration page, delete the current model name, and manually type the new one.

This is slower than qwen-code, where users can switch models more directly. It also makes an in-progress task feel risky because changing the model through manual reconfiguration can interrupt the workflow and make the user feel like the previous context or progress was lost.

## Proposed Solution
Add a faster model switching experience for cosh.

For Qwen/OpenAI-compatible providers, cosh should allow users to select from multiple available or recently used models instead of only editing a single model text field. For example, users should be able to quickly switch between qwen-3.5, qwen-3.6, and other configured Qwen models from /model or the provider configuration flow.

The model switch should keep the current session context when possible and clearly indicate that only the active model changed.

## Alternatives Considered
Users can manually edit settings.modelProviders or re-enter the model name in /auth, but this is not discoverable or convenient for normal model switching.

## Additional Context
Current code inspection shows that /model does not intentionally create a new session. The main user-facing issue is the model selection experience: only registered models or one saved /auth model are available for quick selection, so switching between multiple Qwen models requires manual configuration.

kongche-jbw:
## Description

The `/model` dialog previously only showed `/auth`-configured models
as a fallback when the `modelProviders` registry was empty. This meant
users with registered models could never see or switch to their
`/auth`-verified models.

This PR merges both sources: registered models appear first, followed
by the current `/auth` model (labeled "Current /auth model") and any
previously authenticated models (labeled "Verified via /auth"), with
deduplication. A new `openaiModels` / `aliyunModels` array setting
tracks the MRU list of authenticated models (capped at 10). Selecting
a registered model from `/model` no longer overwrites the `/auth`
model setting.

## Related Issue

closes #1021

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [ ] `anolisa` (anolisa-cli)
- [ ] Multiple / Project-wide

## Checklist

## Testing

```bash
./node_modules/.bin/vitest run \
  packages/cli/src/ui/models/availableModels.test.ts \
  packages/cli/src/ui/components/ModelDialog.test.tsx \
  packages/cli/src/ui/auth/useAuth.test.ts
# 39 passed

### DIFF
--- a/src/copilot-shell/packages/cli/src/config/settingsSchema.ts
+++ b/src/copilot-shell/packages/cli/src/config/settingsSchema.ts
@@ -1091,6 +1091,16 @@ const SETTINGS_SCHEMA = {
             description: 'Last used model name for OpenAI authentication.',
             showInDialog: false,
           },
+          openaiModels: {
+            type: 'array',
+            label: 'OpenAI Models',
+            category: 'Security',
+            requiresRestart: false,
+            default: [] as string[],
+            description:
+              'Recently authenticated model names for OpenAI compatible authentication.',
+            showInDialog: false,
+          },
           aliyunModel: {
             type: 'string',
             label: 'Aliyun Model',
@@ -1098,6 +1108,16 @@ const SETTINGS_SCHEMA = {
             requiresRestart: true,
             default: undefined as string | undefined,
             description: 'Last used model name for Aliyun authentication.',
+            showInDialog: false,
+          },
+          aliyunModels: {
+            type: 'array',
+            label: 'Aliyun Models',
+            category: 'Security',
+            requiresRestart: false,
+            default: [] as string[],
+            description:
+              'Recently authenticated model names for Aliyun authentication.',
             showInDialog: false,
           },
         },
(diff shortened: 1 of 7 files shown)


samchu-zsl (approved):
LGTM, thx!

PR #1031, opened 2026-06-21
feat(skill): preflight model endpoint before gateway startup

### LINKED ISSUE #1022: [skill] feat: preflight OpenClaw model endpoint configuration before gateway startup
## Component

skill

## Problem Statement

The `install-openclaw` skill can complete config writing and start the OpenClaw Gateway even when the resolved API Key, BaseUrl, and Model ID combination is invalid.

From a user's perspective, the failure is discovered too late. The visible symptom can be a vague Gateway or agent-side error such as `gateway disconnected`, instead of a clear explanation that the API key is invalid, the endpoint does not match the selected billing plan, or the model is unavailable for that endpoint.

This makes one-shot non-interactive installation unreliable, especially for Coding Plan, Token Plan, and custom compatible endpoints.

Current implementation already supports `--base-url` and has default endpoints for pay-as-you-go, Coding Plan, and Token Plan, but the skill documentation does not make the custom endpoint path obvious enough. Small agents may still assume the setup is tied to the standard DashScope endpoint.

## Proposed Solution

Add a pre-flight validation step to `install_openclaw.py` before writing config and starting Gateway.

After resolving billing, provider id, BaseUrl, API key, and primary model, the script should make a lightweight API call using the same endpoint configuration that will be written into OpenClaw config.

If validation fails, stop before Gateway startup and print an actionable diagnosis, for example:

- API key rejected or expired.
- BaseUrl does not match the selected billing plan.
- Model ID is not available for the selected plan or endpoint.
- Network or DNS error reaching the endpoint.
- Endpoint returned an unexpected API format.

Suggested flags:

- `--skip-preflight` for offline/debug cases.
- `--preflight-timeout` to bound the lightweight check.

Also update `SKILL.md` to explicitly document:

- `--base-url` is supported.
- Coding Plan and Token Plan use dedicated endpoints by default.
- Custom compatible endpoints should be passed with `--base-url`.
- Gateway startup should happen only after pre-flight succeeds, unless `--skip-preflight` is explicitly passed.

## Alternatives Considered

Relying on `openclaw gateway health` is not enough because it only verifies Gateway health. It does not prove that the configured model endpoint can authenticate and serve the selected model.

Relying on `openclaw agent --message "hello"` after startup also detects the issue too late and may surface as a vague Gateway or fallback error.

## Additional Context

Current code references:

- `src/os-skills/ai/install-openclaw/scripts/install_openclaw.py` already supports `--base-url`.
- `dependency_precheck()` checks local dependencies only.
- `validate_api_key_for_billing()` only performs a basic Coding Plan key-shape check.
- `start_gateway()` starts/restarts Gateway and waits for `openclaw gateway health`, but does not validate model authentication or model availability.
- `SKILL.md` lists normal parameters but does not clearly document `--base-url` usage.

kongche-jbw:
## Description

Add a pre-flight validation step to `install_openclaw.py` that checks the
resolved API key, Base URL, and model ID against the live endpoint before
writing config or starting Gateway. On failure the script exits with an
actionable diagnosis (auth, model not found, rate limit, network) instead
of letting the user discover a vague gateway error after startup.

Also updates SKILL.md to document `--base-url` usage, custom endpoint
examples, and the new `--skip-preflight` / `--preflight-timeout` flags.

## Related Issue

closes #1022

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [x] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [ ] `anolisa` (anolisa-cli)
- [ ] Multiple / Project-wide

## Checklist

- [x] For `skill`: Skill directory structure is valid and shell scripts pass syntax check

## Testing

Tested manually with real Alibaba Cloud keys against 6 scenarios:

1. Coding Plan + valid sk-sp key → preflight passes
2. PAYG + valid sk key → preflight passes
3. Coding Plan + invalid key → preflight catches HTTP 401
4. Coding Plan + non-existent model → preflight catches HTTP 400
5. `--skip-preflight` + invalid key → preflight skipped, exit 0
6. PAYG key on coding endpoint → key format validation rejects before preflight

## Additional Notes

Only `anthropic-messages` provider API is supported for preflight;
other provider APIs are skipped with a log message.
Uses stdlib `urllib.request` only — no new dependencies.

### DIFF
--- a/src/os-skills/ai/install-openclaw/SKILL.md
+++ b/src/os-skills/ai/install-openclaw/SKILL.md
@@ -5,7 +5,7 @@ description: Install and configure OpenClaw non-interactively with Alibaba Cloud
 
 # OpenClaw Non-Interactive Setup
 
-Use this skill to turn a user's one-sentence request into a complete local OpenClaw setup. The normal path is one script command: install Node.js/npm/OpenClaw when needed, write Alibaba Cloud Model Studio config, install/restart the local gateway service, and verify `openclaw gateway health`.
+Use this skill to turn a user's one-sentence request into a complete local OpenClaw setup. The normal path is one script command: resolve the Alibaba Cloud Model Studio plan, validate the API key/Base URL/model combination with the official `anthropic-messages` endpoint shape, install Node.js/npm/OpenClaw when needed, write config, install/restart the local gateway service, and verify `openclaw gateway health`.
 
 Do not use `openclaw onboard` unless the user explicitly asks for interactive setup. Do not configure DingTalk unless the user provides DingTalk credentials or asks for DingTalk access.
 
@@ -29,7 +29,7 @@ Default to `--billing payg --region china` when the user does not specify billin
 - Coding Plan: https://bailian.console.aliyun.com/cn-beijing/?tab=model#/efm/coding_plan
 - Token Plan: https://bailian.console.aliyun.com/?tab=plan#/efm/subscription/overview
 
-Use `--region singapore` only for pay-as-you-go Singapore. For deeper billing details, read `references/aliyun-model-studio-openclaw.md`.
+Use `--region singapore` only for pay-as-you-go Singapore. If the user gives a custom Alibaba Cloud compatible endpoint, pass it through with `--base-url`; otherwise use the billing-specific defaults from the official OpenClaw guide. For deeper billing details, read `references/aliyun-model-studio-openclaw.md`.
 
 ## Execute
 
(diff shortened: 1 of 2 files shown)


Ziqi002:
Precheck and dry run maintain their original semantic installation dependency checks, with the expectation of adding installation dependency and model configuration checks

image
image

kongche-jbw:
Updated the `dry-run` behavior in the amended commit.

The intended split is now:

- `--precheck-only` remains a local dependency check. It does not require an API key and does not validate the model endpoint. This keeps the original “check whether the machine can install/run OpenClaw” workflow intact.
- `--dry-run` now previews the full install flow without making local changes, but still runs the model endpoint pre-flight check unless `--skip-preflight` is passed. This makes `dry-run` useful for validating the resolved billing/baseUrl/model/API-key combination before writing config or starting Gateway.
- `--skip-preflight` is still available for offline/local-flow previews, but it is now the explicit opt-out rather than something implied by `--dry-run`.

I also adjusted config writing so `--dry-run` prints the target config path and validated sections, but does not write `openclaw.json`.

Ziqi002 (approved):
LGTM

PR #1035, opened 2026-06-22
fix(anolisa): infer install-mode from euid when flag is omitted

### LINKED ISSUE #996: [anolisa] sudo anolisa update/adopt resolves user-scope layout: install_mode is flag-only, not euid-aware
## Summary

Privilege gating and install-scope resolution use two different signals that
can disagree. Real runs are gated on **euid** (`privilege::is_root()`), but the
**install scope** (which state store and filesystem layout a command reads and
writes) is derived purely from the `--install-mode` flag, which defaults to
`user`. As a result `sudo anolisa update <component>` passes the root gate yet
resolves the *user*-scope layout, operating on the wrong state store instead of
the system-scope component the user clearly intended.

This was surfaced while reviewing the RPM-backed update / delegated install work
(#959) but is **cross-cutting**: the merged `rpm-observed` adopt path (#958) is
also a system-scope action reached through the same resolution.

## Repro

```
# A component installed/adopted in system scope (/usr/local, rpm-managed/observed)
sudo anolisa update mycomponent
```

## Expected vs actual

- **Expected:** running as root operates on the system-scope state where the
  rpm-managed / rpm-observed component lives.
- **Actual:** `install_mode` stays `user` (the flag default), so the command
  resolves the user-scope layout and looks up state under the user home — not
  the system store. To hit system scope the caller must additionally pass
  `--install-mode system`, which is easy to miss and not implied by `sudo`.
  (Under `sudo`, the user-branch home is whatever `EnvService::detect().home`
  returns — typically root's `$HOME` — so the resolved path is not even the
  invoking user's state.)

## Root cause

- `--install-mode` defaults to `user`:
  `crates/anolisa-cli/src/commands.rs:43`
- The context copies the flag verbatim, with no euid awareness:
  `crates/anolisa-cli/src/context.rs:56`
- Scope resolution branches solely on `install_mode`:
  `crates/anolisa-cli/src/commands/common.rs:31` (User -> `FsLayout::user(home)`,
  System -> `FsLayout::system(prefix)`)
- The update privilege gate, by contrast, keys on euid:
  `crates/anolisa-cli/src/commands/tier1/update.rs:160`
  (`privilege::is_root()`).

So the privilege check and the scope selection are decided by independent
inputs that can contradict each other.

## Affected commands

- `anolisa update <component>` (RPM update + delegated install, #959)
- `anolisa install` adopt / `rpm-observed` path (#958, merged)
- Any other system-scope action that gates on root but resolves scope from the
  flag.

## Proposed direction (for discussion)

Make scope resolution euid-aware rather than flag-only, e.g. when no explicit
`--install-mode` is given, default to `system` if running as root (and keep
`user` otherwise), or at minimum warn/error when root runs a system-scope
component under a resolved `user` layout. The exact policy (auto-switch vs.
explicit-only with a clear diagnostic) should be agreed before implementing,
since it changes the default behavior of every scope-sensitive command.

## Scope note

Deliberately left out of #959 because it is broader than the update path and
also affects the already-merged adopt path; filing separately so it can be
fixed once across all scope-sensitive commands.

ikunkun-sys:
## Description

Makes `--install-mode` scope resolution euid-aware so that `sudo anolisa update <component>` automatically resolves to system scope without requiring an explicit `--install-mode=system` flag (#996).

Previously `--install-mode` defaulted to `user` unconditionally via `default_value_t`. The privilege gate (`privilege::is_root()`) and scope resolution (`resolve_layout`) were decided by independent inputs — euid and the CLI flag respectively — that could contradict each other: root passed the privilege check but still resolved the user-scope layout, operating on the wrong state store.

Behavior after this change:
- `--install-mode` omitted + root → system scope (inferred).
- `--install-mode` omitted + non-root → user scope (inferred, same as before).
- `--install-mode=system` explicit → system scope (respected unconditionally).
- `--install-mode=user` explicit + root → user scope (respected, stderr warning emitted).

## Design Note

The `Cli.install_mode` field type changes from `InstallMode` (with `default_value_t = User`) to `Option<InstallMode>`. This lets clap distinguish "omitted" from "explicitly set to user". A new pure function `resolve_install_mode(explicit, is_root)` in `context.rs` encapsulates the inference logic and is called by `CliContext::from_cli` with `privilege::is_root()`. `CliContext.install_mode` remains `InstallMode` (never `Option`), so all 40+ downstream callsites that read `ctx.install_mode` — including `resolve_layout`, privilege gates, and the osbase system-only validation — are unchanged.

## Ship Note

- The osbase sandbox install gate (`ctx.install_mode == System`) now passes automatically under `sudo` instead of requiring the explicit flag — this is a user-facing improvement, not a regression.
- The stderr warning for root + explicit `--install-mode=user` is suppressed by `--quiet`.

## Test

- Unit (context): four-case matrix for `resolve_install_mode` — omitted/root → System, omitted/non-root → User, explicit User + root → User, explicit System + non-root → System.
- Unit (commands): clap parse tests — omitted yields `None`, explicit `--install-mode=system` yields `Some(System)`.
- Full gate green: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check --workspace --all-targets`, `cargo test --workspace`.

Closes #996

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands.rs
@@ -40,9 +40,10 @@ pub struct Cli {
     #[command(subcommand)]
     pub command: Commands,
 
-    /// Install scope: user (~/.local) or system (/usr/local)
-    #[arg(long, global = true, value_enum, default_value_t = InstallMode::User)]
-    pub install_mode: InstallMode,
+    /// Install scope: user (~/.local) or system (/usr/local).
+    /// When omitted, defaults to system if running as root, user otherwise.
+    #[arg(long, global = true, value_enum)]
+    pub install_mode: Option<InstallMode>,
 
     /// Custom install prefix (system-mode only)
     #[arg(long, global = true, value_name = "PATH")]
@@ -248,6 +249,8 @@ fn has_dot_segment(path: &Path) -> bool {
 
 #[cfg(test)]
 mod tests {
+    use clap::FromArgMatches as _;
+
     use super::*;
 
     fn ctx_with_prefix(prefix: PathBuf) -> CliContext {
@@ -321,5 +324,23 @@ mod tests {
             help.contains("ls"),
             "visible alias `ls` should appear in help output"
         );
+    }
+
+    #[test]
+    fn install_mode_is_none_when_omitted() {
+        let cmd = build_cli();
+        let matches = cmd.try_get_matches_from(["anolisa", "status"]).unwrap();
+        let cli = Cli::from_arg_matches(&matches).unwrap();
+        assert_eq!(cli.install_mode, None);
+    }
+
+    #[test]
+    fn install_mode_is_some_when_explicitly_set() {
+        let cmd = build_cli();
+        let matches = cmd
+            .try_get_matches_from(["anolisa", "--install-mode=system", "status"])
+            .unwrap();
+        let cli = Cli::from_arg_matches(&matches).unwrap();
+        assert_eq!(cli.install_mode, Some(InstallMode::System));
     }
 }
--- a/src/anolisa/crates/anolisa-cli/src/context.rs
+++ b/src/anolisa/crates/anolisa-cli/src/context.rs
@@ -6,9 +6,13 @@
 //! every command handler. Handlers must not re-parse globals from the args
 //! struct; instead they read from the shared context so that semantics stay
 //! consistent across surfaces.
+//!
+//! When `--install-mode` is omitted, the effective scope is inferred from
+//! the process's effective UID: root defaults to system, non-root to user.
 
 use std::path::PathBuf;
 
+use anolisa_platform::privilege;
 use clap::ValueEnum;
 
 /// Where ANOLISA installs files: user-mode (`file-hierarchy(7)` under `$HOME`)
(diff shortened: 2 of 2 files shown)


src/anolisa/crates/anolisa-cli/src/commands.rs
```diff
@@ -39,9 +39,10 @@ pub struct Cli {
     #[command(subcommand)]
     pub command: Commands,
 
-    /// Install scope: user (~/.local) or system (/usr/local)
-    #[arg(long, global = true, value_enum, default_value_t = InstallMode::User)]
-    pub install_mode: InstallMode,
+    /// Install scope: user (~/.local) or system (/usr/local).
+    /// When omitted, defaults to system if running as root, user otherwise.
```
kongche-jbw, line 44: 这里已经把 omitted `--install-mode` 的语义改成 euid-aware；建议顺手更新 `crates/anolisa-cli/src/commands/osbase.rs` 里仍写着 “global --install-mode is `user`” 的旧注释，否则后续维护者会以为默认值仍然固定为 user。

kongche-jbw (dismissed):
lgtm

src/anolisa/crates/anolisa-cli/src/commands.rs
ikunkun-sys: 已在 cd8ba74 中更新：将 osbase.rs 中描述 `--install-mode` 默认为 `user` 的旧注释和错误提示改为反映 euid-aware 推断的新语义。

kongche-jbw (approved):

PR #1044, opened 2026-06-22
feat(anolisa): implement raw-backend component update

### LINKED ISSUE #1037: [anolisa] feat(anolisa): implement raw-backed update for raw-managed components
## Goal
Let `anolisa update <component>` update **raw-managed** components, closing the
gap left by the RPM update path (#959), which only covers `rpm-observed` /
`rpm-managed`.

## Current State
- The RPM path is implemented (#959): `rpm-observed` / `rpm-managed` run
  `rpmdb query -> dnf repo query -> dnf update -> refresh state`.
- `raw-managed` is **not implemented**: `update_component_with_deps` returns
  `not_implemented` for `Ownership::RawManaged` (in the `update` command,
  `anolisa-cli`), with the hint *"raw component update is not implemented yet
  (update planner / distribution resolver pending); only RPM-backed components
  update today"*.
- `update all` is likewise `NOT_IMPLEMENTED`, blocked on the same raw
  distribution resolver.
- The `update` module doc records this boundary.

## Scope
- Implement the update flow for `Ownership::RawManaged`: query the raw
  index/distribution for candidate versions -> compare versions -> download the
  new artifact -> replace owned files -> refresh ANOLISA state (`version`
  field).
- **Do not switch backend**: update must follow the recorded
  ownership/`install_backend` (raw stays raw).
- Reuse the existing install-time artifact download and owned-file management
  (install_runner) rather than duplicating it.
- Extend the result wire shape: `ComponentUpdatePayload.backend` is currently a
  fixed `"rpm"`; it must be able to report `raw`.
- Unblock `update all` once the raw resolver lands.
- A user raw override updates only the user raw object, not the shadowed system
  RPM (mirroring the #959 semantics).

## Acceptance Criteria
- `anolisa update <raw-component>` updates a raw-managed component to a newer
  version from the raw index and refreshes state.
- "Already latest" is a clean no-op.
- After update, ownership/backend remain raw (no backend switch).
- A dry-run previews available candidates without applying.
- `update all` no longer returns `NOT_IMPLEMENTED` for raw-managed objects.
- The `not_implemented` branch for `Ownership::RawManaged` in the `update`
  command is replaced by the implementation.

ikunkun-sys:
## Description

Implements `anolisa update <component>` for raw-managed components (#1037). The `Ownership::RawManaged` branch previously returned `NOT_IMPLEMENTED`; it now resolves the latest version from the raw distribution index and replaces the owned files. Because `InstallRunner::install_files` refuses to overwrite (`O_CREAT|O_EXCL`), the apply step backs up and removes the old owned files, installs the new artifact, rewrites the manifest snapshot, and refreshes state — all wrapped in a `Transaction` that rolls back to the prior version on any failure.

Behavior:
- Resolves the highest published version (no explicit `--version`); already-latest is a clean no-op.
- Version comparison is semver-aware (tolerates a leading `v`); a repo that offers only an older release is refused rather than silently downgraded, and a version that cannot be ordered (non-semver, differing strings) is refused too rather than optimistically treated as an upgrade.
- Ownership / `install_backend` are preserved; version, distribution source, owned files, services, status, health, and the operation pointer move to the new version.
- A component installed with `--package` updates against that same package, not one re-derived from the component name.
- `update all` stays `NOT_IMPLEMENTED` (tracked separately).

## Design Note

To avoid duplicating install, the raw resolver is reused rather than copied: `resolve_raw`, `prepare_raw_execution`, `write_installed_component_manifest`, and the `RawResolution`/`PreparedInstall`/`ResolveInputs` types become `pub(crate)`, plus a new `resolve_raw_inputs_for_component` that rebuilds resolve inputs from recorded state (update has no CLI args). `CliError::with_command` re-stamps the command label on errors surfaced by those install-tagged helpers, and the backup primitive `prepare_backup` is exposed from anolisa-core.

Resolve + download run before the install lock (so the global lock is never held across network I/O); the destructive phase then re-validates under the lock — it re-reads the owned-file list from freshly loaded state and aborts if the component drifted from the planned version, so a concurrent update/uninstall is never clobbered by a stale plan. Version ordering goes through a four-state `version_relation` (older / same / newer / indeterminate) so the downgrade guard stays effective for non-semver versions instead of defaulting them to "upgrade".

`InstalledObject` gains a `raw_package` field, recorded at raw install time from the resolved package; update prefers it over repo.toml derivation so a component installed with `--package` re-fetches the same package (falling back to derivation when absent, e.g. pre-existing state).

The dry-run candidate list reuses `resolve`'s own selection predicate via a new `DistributionIndex::matching_versions`, so the preview agrees with what an update actually picks (same channel / libc / pkg_base / install_mode filtering and semver ordering). Transaction operation ids embed the operation verb (`op-update-…`, `op-uninstall-…`) so audit tooling can group by operation.

## Ship Note

- Explicit `--version` / downgrade and `update all` are out of scope.
- On the success path the per-operation backup tree is pruned (mirroring uninstall); a failed-and-rolled-back update retains it and writes a `RolledBack` record to the central log, with any rollback-restore failures surfaced to stderr and the audit log rather than silently swallowed.
- Known deferred items: a brand-new file declared by the next version that already exists on disk (and is not an ANOLISA-owned record) aborts the update via `DestExists`; and folding `raw_update_rollback` / the four-phase transaction into a shared anolisa-core primitive is left as follow-up.

## Test

- Unit (raw update): upgrade success (files/sha256 refreshed, version bumped, ownership preserved, operation recorded, backups pruned, `op-update-` id); already-latest no-op; dry-run touches nothing; install-failure rollback restores old files + version; downgrade refused; non-semver version refused; under-lock abort when the recorded version drifted (a concurrent update); status/service state reset after a clean update; `--package` reused on update (artifact published only under a non-default key, with the no-recorded-package path failing to resolve); raw dispatch never runs dnf; `version_relation` classification.
- Unit (install): a raw install records `raw_package`.

Closes #1037


src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
```diff
@@ -222,6 +253,628 @@ pub(crate) fn update_component_with_deps(
...
+    let _lock = InstallLock::acquire(&layout.lock_file).map_err(|err| CliError::Runtime {
+        command: command.to_string(),
+        reason: format!("failed to acquire install lock: {err}"),
+    })?;
+    let mut state = common::load_installed_state(ctx, command)?;
+
+    // Re-validate under the lock: the component must still exist and still be
+    // raw-managed. A concurrent uninstall or backend change must not be
+    // overwritten by a stale update.
+    match state.find_object(ObjectKind::Component, component) {
```
kongche-jbw: Could we also revalidate the raw component snapshot after reloading state under the install lock? `handle_component_update()` snapshots `from_version` and `old_files` before the lock, then this path only checks that the object still exists and is still `raw-managed`. If another install/update/repair changes the same component while this command is resolving or downloading, the stale plan can still remove the pre-lock file set and persist its resolved version over the newer state. At minimum, it seems safer to compare the locked object version and owned-file path set with the pre-lock snapshot and ask the user to retry if they changed.

kongche-jbw:
Thanks for the follow-up fixes. I rechecked the latest head and the raw update TOCTOU concern looks addressed now: the apply path reloads state under the install lock, validates the planned version, and reads the owned-file list from the locked state instead of carrying a pre-lock file snapshot.

One small PR-description cleanup: the Ship Note still lists the install-time `--package` override as a deferred limitation, but the latest commits now record `InstalledObject::raw_package` and reuse it during raw update. Could you update that note to remove the stale limitation, or rephrase it along these lines?

> Raw install now records `raw_package`, so raw update preserves install-time `--package` selection. Older state without `raw_package` falls back to repo.toml package derivation.

No new blocking code issue from my side after the latest changes.

src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
ikunkun-sys: Done

ikunkun-sys:
> Thanks for the follow-up fixes. I rechecked the latest head and the raw update TOCTOU concern looks addressed now: the apply path reloads state under the install lock, validates the planned version, and reads the owned-file list from the locked state instead of carrying a pre-lock file snapshot.
> 
> One small PR-description cleanup: the Ship Note still lists the install-time `--package` override as a deferred limitation, but the latest commits now record `InstalledObject::raw_package` and reuse it during raw update. Could you update that note to remove the stale limitation, or rephrase it along these lines?
> 
> > Raw install now records `raw_package`, so raw update preserves install-time `--package` selection. Older state without `raw_package` falls back to repo.toml package derivation.
> 
> No new blocking code issue from my side after the latest changes.

PR描述已清理

kongche-jbw (approved):
lgtm

PR #1045, opened 2026-06-22
fix(sight): canonicalize uprobe path for container support

jfeng18:
## Summary

`ssl_libs_from_maps` prepends `/proc/{pid}/root` to library paths from `/proc/{pid}/maps`. On non-container hosts this symlink resolves to `/` and libbpf handles it fine — **A/B testing on ECS confirmed both paths attach successfully**.

However, in container scenarios `/proc/{pid}/root` points to an overlay rootfs. The kernel's `kern_path` (used by `perf_event_open`) can resolve these cross-namespace paths, but libbpf's userspace `elf_open` may fail because the overlay mount is not directly accessible from the host mount namespace.

This PR adds `std::fs::canonicalize()` before passing the path to uprobe APIs, resolving the symlink to the real host-visible path. This handles both cases:
- Non-container: `/proc/{pid}/root/usr/lib/libssl.so` → `/usr/lib/libssl.so` (no behavior change)
- Container: `/proc/{pid}/root/usr/lib/libssl.so` → `/var/lib/containerd/.../usr/lib/libssl.so` (resolves overlay)

Falls back to the original path if `canonicalize` fails (e.g. process already exited).

## Correction from original description

The original PR described the root cause as "libbpf cannot resolve `/proc/{pid}/root` paths on ECS". **This was incorrect.** A/B testing showed the old code (without canonicalize) attaches to claude BoringSSL successfully on non-container ECS. The actual reason claude genai capture was failing was a missing `*claude*` entry in the cmdline allow-list in `config.json` — a config issue, not a code issue. The canonicalize change is a defensive improvement for container deployments, not a fix for the originally reported symptom.

## Verification

- **ECS A/B test (non-container)**: OLD binary (no canonicalize) and NEW binary (with canonicalize) both attach to claude BoringSSL successfully. No behavior difference on non-container hosts.
- **Unit test**: `uprobe_path_canonicalize_resolves_real_path` covers symlink resolution for both scenarios.
- Container scenario: not yet verified (no container environment available).

## Changed files

- `src/agentsight/src/probes/sslsniff.rs`: add `std::fs::canonicalize` before uprobe attach, keep original path for `find_boringssl_offsets` (reads file content, not affected)

### DIFF
--- a/src/agentsight/src/probes/sslsniff.rs
+++ b/src/agentsight/src/probes/sslsniff.rs
@@ -326,30 +326,44 @@ impl SslSniff {
 
             log::debug!("[attach_process] pid={pid}: attaching {kind:?} → {path}");
 
+            // uprobe attach requires host filesystem paths, not /proc/{pid}/root/... paths.
+            // canonicalize resolves /proc/{pid}/root symlink to the real host path,
+            // handling both container (overlay rootfs) and non-container (/ symlink) cases.
+            // Falls back to original path if the process has already exited.
+            let uprobe_path = std::fs::canonicalize(&path)
+                .map(|p| p.to_string_lossy().into_owned())
+                .unwrap_or_else(|_| path.clone());
+
             let result = match kind {
                 // Use pid=-1 for global attach (all processes), avoiding per-process duplicate attaches
-                SslLibKind::OpenSsl => attach_openssl(&mut self.skel, &path, -1),
-                SslLibKind::GnuTls => attach_gnutls(&mut self.skel, &path, -1),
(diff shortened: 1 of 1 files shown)


chengshuyi (approved):
LGTM. canonicalize 处理 /proc/{pid}/root 路径是容器 overlay 场景的正确做法，find_boringssl_offsets 与 uprobe attach 的路径分离逻辑清晰，fallback 处理合理。一个小建议：可在 canonicalize 后补一行 debug 日志（路径变化时），方便容器环境调试，不 block 合入。

PR #1052, opened 2026-06-23
fix(anolisa): isolate user-layout tests from host environment

### LINKED ISSUE #1050: Tests can mutate real user state/env directories when user-mode layout and process env are used
## Summary

Some tests can accidentally read from or write to the developer's real user environment, especially when user-mode layout resolution is involved.

The main pattern is that `FsLayout::user(home)` still honors process-level `XDG_*` overrides. In tests, passing a temp `home` is therefore not enough to isolate filesystem writes if the developer or CI environment has `XDG_STATE_HOME`, `XDG_CONFIG_HOME`, etc. set.

There are also a few process-global env mutation cases (`OPENCLAW_*`, `PATH`) that can leak between tests or depend on the host environment.

## Examples found

- A `repair` command test used `InstallMode::User` with a temp prefix. User mode ignores `prefix`, so seeding state can write to the process home, e.g. `$HOME/.local/state/anolisa/installed.toml`.
- `adapter_manager` tests construct user layouts from temp homes, but `FsLayout::user(...)` can redirect state/config/cache through real `XDG_*` env vars.
- `adapter_manager` tests mutate `OPENCLAW_BIN`, `OPENCLAW_HOME`, and `FAKE_OPENCLAW_FAIL`; tests should start from a clean env contract and restore the original values.
- A `PATH` unit test mutates process-global `PATH`; it should test path composition without changing global env.

## Reproduction idea

Run tests under an isolated HOME/XDG root and inspect leftovers:

```bash
root="$(mktemp -d /tmp/anolisa-isolated.XXXXXX)"
mkdir -p "$root"/{home,xdg-data,xdg-config,xdg-state,xdg-cache,xdg-runtime,tmp}

env \
  HOME="$root/home" \
  XDG_DATA_HOME="$root/xdg-data" \
  XDG_CONFIG_HOME="$root/xdg-config" \
  XDG_STATE_HOME="$root/xdg-state" \
  XDG_CACHE_HOME="$root/xdg-cache" \
  XDG_RUNTIME_DIR="$root/xdg-runtime" \
  TMPDIR="$root/tmp" \
  cargo test --workspace

find "$root/home" "$root"/xdg-* -maxdepth 8 -mindepth 1 -print
```

Tests that intend to be fully isolated should not leave ANOLISA state/config/cache files under these fake HOME/XDG roots unless the test explicitly asserts that behavior.

## Expected behavior

Tests should not touch the real user environment or rely on host-level `XDG_*`, `PATH`, or framework-specific variables.

## Suggested fixes

- Use system-mode temp prefixes for command-layer tests that need to seed `installed.toml`.
- Provide/use a test-friendly layout constructor that accepts explicit XDG roots instead of reading process env.
- Guard process-global env mutations with a mutex, clear the relevant variables at test start, and restore them on drop.
- Avoid mutating `PATH` in unit tests by injecting the existing PATH value into the helper under test.

ikunkun-sys:
## Description

Fixes several test-isolation leaks where tests could depend on, mutate, or write through the developer's real process environment.

The main issue is user-mode layout resolution: `FsLayout::user(home)` intentionally honors process-level `XDG_*` overrides, so passing a temp `home` is not enough for tests. If `XDG_STATE_HOME` is set, tests can still write ANOLISA state under that real XDG root.

Behavior:
- User-mode test layouts that need isolation now use explicit override inputs instead of reading process `XDG_*`.
- The `repair` raw-component test uses a system-mode temp prefix, because user mode ignores `prefix`.
- `adapter_manager` tests start from a clean `OPENCLAW_*` contract and restore the previous env on drop.
- The `PATH` helper test no longer mutates process-global `PATH`.

## Design Note

`FsLayout::user_with_overrides(...)` is made public so cross-crate tests can construct user layouts without reading process-global XDG variables. Production `FsLayout::user(home)` is unchanged and still honors `XDG_DATA_HOME`, `XDG_CONFIG_HOME`, `XDG_STATE_HOME`, `XDG_CACHE_HOME`, and `XDG_RUNTIME_DIR`.

For OpenClaw adapter integration tests, env mutation is wrapped in a guard that serializes on `ENV_LOCK`, clears `OPENCLAW_BIN`, `OPENCLAW_HOME`, and `FAKE_OPENCLAW_FAIL` at test start, and restores the original values on drop. This prevents both test-to-test leakage and accidental dependence on a developer's real OpenClaw installation.

## Ship Note

- No production layout behavior changes are intended.
- System command paths such as `/run/anolisa`, `/var/log/anolisa`, and `/etc/anolisa` remain production behavior and are not exercised by these tests.
- The remaining absolute paths in tests are either pure path assertions or hostile-path rejection fixtures; they do not write to those real paths.

## Test

- Unit/workspace gate green: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check --workspace --all-targets`, `cargo test --workspace`, `cargo doc --workspace --no-deps`.
- Isolation run: `cargo test --workspace` with fresh isolated `HOME`, `XDG_DATA_HOME`, `XDG_CONFIG_HOME`, `XDG_STATE_HOME`, `XDG_CACHE_HOME`, `XDG_RUNTIME_DIR`, and `TMPDIR`.
- Post-run verification: inspected the isolated HOME/XDG roots and confirmed no ANOLISA state/config/cache files were left behind.

close #1050

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/repair.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/repair.rs
@@ -910,7 +910,9 @@ mod tests {
     #[test]
     fn repair_raw_component_is_not_implemented() {
         let tmp = tempfile::tempdir().expect("tmpdir");
-        let c = ctx(tmp.path().to_path_buf(), InstallMode::User, false);
+        // User mode ignores `prefix` and resolves from the process home, so
+        // this test uses system mode to keep the seeded state under `tmp`.
+        let c = ctx(tmp.path().to_path_buf(), InstallMode::System, false);
         seed(&c, raw_object("copilot-shell", "9.9.9"));
         let rpm = FakeQuery::new("anolisa-copilot-shell", None);
         let err = repair_with_query("copilot-shell", &c, &rpm)
--- a/src/anolisa/crates/anolisa-core/src/adapter/manager.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/manager.rs
@@ -1254,8 +1254,15 @@ fn run_capture(cmd: &FrameworkCommand) -> Result<CliOutput, AdapterError> {
 
 /// Build a `PATH` value with `prepend` dirs in front of the current one.
 fn prepend_path(prepend: &[PathBuf]) -> std::ffi::OsString {
+    prepend_path_with_existing(prepend, std::env::var_os("PATH"))
+}
+
+fn prepend_path_with_existing(
+    prepend: &[PathBuf],
+    existing: Option<std::ffi::OsString>,
+) -> std::ffi::OsString {
     let mut parts: Vec<PathBuf> = prepend.to_vec();
-    if let Some(existing) = std::env::var_os("PATH") {
+    if let Some(existing) = existing {
         parts.extend(std::env::split_paths(&existing));
     }
     // join_paths only fails if a component contains the path separator,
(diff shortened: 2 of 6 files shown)


kongche-jbw (approved):
lgtm thx

PR #1056, opened 2026-06-23
fix(sight): merge embedded defaults with user config

### LINKED ISSUE #1055: [sight] config.json silently replaces built-in agent rules, causing zero capture with no diagnostic
## Problem

When `/etc/agentsight/config.json` exists and loads successfully, the embedded defaults in `agentsight.json` (compiled via `include_str!`) are **completely discarded** — `config.json` rules fully replace them rather than extending them. This caused a real incident where two developers spent hours debugging zero `genai` capture for Claude, when the root cause was simply that `config.json` had only `cosh` rules and no `*claude*` entry.

**Root cause chain (verified from code):**

1. `AgentsightConfig::default()` initializes `cmdline_rules` as `Vec::new()` (`config.rs:545`)
2. `unified.rs:137-138` calls `config.load_from_file()` when `config.json` exists
3. `load_from_json` (`config.rs:710`) does `self.cmdline_rules.extend(rules_from_file)` — extending an **empty** Vec
4. The embedded defaults with `"claude*"` rule are never loaded when `config.json` succeeds
5. The fallback at `unified.rs:149` that loads `default_cmdline_rules()` only fires on config load **failure**
6. No log output at any level indicates that processes were scanned but not matched — the user gets `genai=0` with zero explanation

Additionally, `HealthChecker` (`health/checker.rs:80`) independently uses `default_cmdline_rules()`, ignoring `config.json`. This means the health dashboard can show Claude as "discovered" while the trace pipeline silently ignores it.

### 1. Merge defaults instead of replacing (~15 lines in `unified.rs`)

Before calling `config.load_from_file()`, pre-seed `config.cmdline_rules` and `config.https_rules` with the embedded defaults. Since `load_from_json` uses `extend()`, user config rules will be appended after defaults. Use user-rules-first ordering so user patterns win on first-match.

This also makes the error-path fallback at L149 redundant (defaults already seeded), but keep it for safety.

### 2. Add zero-match WARN diagnostic (~10 lines in `scanner.rs`)

After `scan()` completes, if zero agents were discovered and matchers are non-empty, emit a single `log::warn!`:

```
WARN: Scanned 412 /proc entries, 0 matched any allow rule — verify config.json cmdline.allow
```

Key details:
- Must be `warn!` not `info!` — default log level is `warn` (`config.rs:68`), so `info!` is invisible
- No cmdline content in warn-level logs (avoid leaking API keys/tokens)
- Add `is_denied()` check to `scan()` loop for consistency with runtime `handle_procmon_event` path

### 3. Add deny-rule check to initial scan (~3 lines in `scanner.rs`)

The runtime path (`handle_procmon_event` L794) checks deny rules before matching, but `scan()` does not. Add `is_denied()` check before `try_match_process()` for consistency.

## Phase 2 (follow-up PR)

- Expand `agentsight.json` cmdline.allow: add `codex`, `cursor`, `aider`, `windsurf` patterns
- Fix Claude pattern: current `["claude*"]` only matches argv[0], but Claude Code runs as `node`; needs `["*node*", "*claude*"]`
- Expand `agentsight.json` https rules: add `api.anthropic.com`, `api.openai.com`, `*.zjapi.com`, `generativelanguage.googleapis.com`
- Optional: add `"merge_defaults": false` config field for deployments that intentionally want minimal monitoring

## Rejected: Default allow-all + deny-list

Rejected for five reasons: (1) BPF-side blocker — `sslsniff.bpf.c is_pid_traced()` hard-gates on `traced_processes` hash map (max 1024 entries), making user-space allow-all useless without rewriting BPF programs. (2) Breaking backward compatibility. (3) Privacy regression — GenAI pipeline persists full request/response bodies. (4) Unmaintainable deny-list. (5) Hot-path performance.

## Key files

- `config.rs:545` — `Vec::new()` initialization
- `config.rs:710` — `extend()` that operates on empty Vec
- `unified.rs:131-152` — config load + fallback logic
- `scanner.rs:106-135` — `scan()` with silent skip
- `scanner.rs:262-269` — `find_match()` returns None silently
- `agentsight.json` — embedded defaults (thin coverage)
- `health/checker.rs:80` — HealthChecker uses defaults independently (inconsistency)

jfeng18:
## Summary

When `/etc/agentsight/config.json` exists, it silently **replaces** the embedded default agent rules (`agentsight.json`) instead of extending them. This caused a real incident where a config with only `cosh` rules and no `*claude*` entry led to hours of debugging zero `genai` capture.

This PR makes user config **extend** the embedded defaults, and adds a zero-match diagnostic.

### 1. Pre-seed defaults before config load (~8 lines, `unified.rs`)

Before `load_from_file()`, pre-seed `cmdline_rules` and `https_rules` with embedded defaults. `load_from_json`'s `extend()` appends user rules after defaults. First-match semantics in `AgentScanner` mean defaults apply transparently.

### 2. New `default_rules()` function (~6 lines, `config.rs`)

Returns both cmdline and https rules from embedded JSON. Existing `default_cmdline_rules()` unchanged (used by HealthChecker).

### 3. scan() deny check + diagnostic (~15 lines, `scanner.rs`)

- Added `is_denied()` check in scan loop (was missing; runtime `handle_procmon_event` has it)
- Added `warn!` when 0 agents found with non-empty matchers (default log level is warn, so this is visible)
- Eliminated double `/proc/{pid}/cmdline` read by passing pre-read cmdline to `try_match_process`

## Verification

- `cargo +1.89.0 fmt --check`: pass
- `cargo +1.89.0 clippy -D warnings`: pass  
- `cargo +1.89.0 test --lib`: 699 passed, 0 failed
- **ECS E2E**: old config (cosh-only, no claude) + new binary loaded **21 cmdline rules** (16 defaults + 5 user) instead of old 5. Merge confirmed working.
- **Adversarial review**: 10-agent workflow, 3 dimensions (correctness/backward-compat/security), 7 findings → 5 confirmed (all low, addressed) + 2 refuted

## Test plan

- [x] `test_default_rules_returns_both_cmdline_and_https`: embedded defaults contain Claude + https rules
- [x] `test_load_extends_defaults_not_replaces`: pre-seed + load_from_json preserves both defaults and user rules (discriminating)
- [x] `test_scan_zero_match_with_matchers`: scan with impossible pattern returns empty cleanly
- [x] `test_is_denied_blocks_allow_match`: deny overrides allow
- [x] ECS E2E with old config (no claude entry) verified merge

Fixes #1055

### DIFF
--- a/src/agentsight/src/config.rs
+++ b/src/agentsight/src/config.rs
@@ -366,6 +366,13 @@ pub fn default_cmdline_rules() -> Vec<CmdlineRule> {
     rules
 }
 
+/// Load both default cmdline and https rules from the embedded config.
+pub fn default_rules() -> (Vec<CmdlineRule>, Vec<HttpsRule>) {
+    let (cmdline, https, _) =
+        parse_json_rules(DEFAULT_AGENTS_JSON).expect("embedded DEFAULT_AGENTS_JSON is valid");
+    (cmdline, https)
+}
+
 // ==================== Chrome Trace Export ====================
 
 /// Check if chrome trace export is enabled (set once at startup)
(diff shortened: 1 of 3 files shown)


jfeng18:
Closing in favor of #1079. Shuyi confirmed that replace semantics is intentional design — users are expected to include all needed agent rules in their config file. #1079 documents this behavior in AGENTS.md, PITFALLS.md, and code comments, which is the right approach.

The scan() zero-match warning from this PR could still be useful as a separate diagnostic improvement, but the core config merge change is no longer needed.

PR #1057, opened 2026-06-23
feat(cosh): add SLS JSONL session telemetry with expanded metrics

samchu-zsl:
## Description

Add local JSONL-based session telemetry for SLS agent collection. At session end, `logSessionSummary` writes a comprehensive 32-field record to `/var/log/anolisa/sls/ops/cosh.jsonl`. Fields cover component identification, session configuration, audit decisions, tool call statistics with error classification, file/sandbox metrics, token usage, API stats, and environment info. Uses open→write→close pattern to support logrotate.

## Related Issue

no-issue: New telemetry capability per Agentic OS data collection specification

## Type of Change

- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)

## Checklist

## Testing

```
npm run test --workspace=packages/core   → 179 files, 3765 passed
npm run test --workspace=packages/cli    → 219 files, 3454 passed
```

## Additional Notes

Key files:
- `packages/core/src/telemetry/sls-logger.ts` — new JSONL append writer
- `packages/core/src/telemetry/loggers.ts` — `logSessionSummary` with 32 fields
- `packages/core/src/telemetry/uiTelemetry.ts` — `toolErrorCounts` metrics
- `packages/core/src/core/coreToolScheduler.ts` — await duration tracking

### DIFF
--- a/src/copilot-shell/packages/cli/src/nonInteractive/io/BaseJsonOutputAdapter.test.ts
+++ b/src/copilot-shell/packages/cli/src/nonInteractive/io/BaseJsonOutputAdapter.test.ts
@@ -1197,6 +1197,15 @@ describe('BaseJsonOutputAdapter', () => {
           totalRuns: 0,
           totalBlocked: 0,
         },
+        toolErrorCounts: {
+          modelError: 0,
+          executionError: 0,
+          denied: 0,
+        },
+        awaitApproval: {
+          totalDurationMs: 0,
+          count: 0,
+        },
       };
       const options: ResultOptions = {
         isError: false,
--- a/src/copilot-shell/packages/cli/src/nonInteractive/io/JsonOutputAdapter.test.ts
+++ b/src/copilot-shell/packages/cli/src/nonInteractive/io/JsonOutputAdapter.test.ts
@@ -584,6 +584,15 @@ describe('JsonOutputAdapter', () => {
           totalRuns: 0,
           totalBlocked: 0,
         },
+        toolErrorCounts: {
+          modelError: 0,
+          executionError: 0,
+          denied: 0,
+        },
+        awaitApproval: {
+          totalDurationMs: 0,
+          count: 0,
+        },
       };
 
       adapter.emitResult({
--- a/src/copilot-shell/packages/cli/src/nonInteractiveCli.test.ts
+++ b/src/copilot-shell/packages/cli/src/nonInteractiveCli.test.ts
@@ -138,6 +138,7 @@ describe('runNonInteractive', () => {
       getFolderTrust: vi.fn().mockReturnValue(false),
       getIncludePartialMessages: vi.fn().mockReturnValue(false),
       getSessionId: vi.fn().mockReturnValue('test-session-id'),
+      getAuthType: vi.fn().mockReturnValue('dashscope'),
       getModel: vi.fn(() => currentModel),
       setModel: vi.fn(async (model: string) => {
         currentModel = model;
--- a/src/copilot-shell/packages/cli/src/nonInteractiveCli.ts
+++ b/src/copilot-shell/packages/cli/src/nonInteractiveCli.ts
@@ -18,6 +18,7 @@ import {
   InputFormat,
   uiTelemetryService,
   parseAndFormatApiError,
+  logSessionSummary,
 } from '@copilot-shell/core';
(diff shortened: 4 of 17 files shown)


samchu-zsl:
Manual testing completed successfully.
image

kongche-jbw:
Thanks for adding the session telemetry. I found two issues that may affect the accuracy of the collected data:

1. The normal non-interactive path appears to emit the session summary twice. `runNonInteractive` writes it in its `finally` block, then `gemini.tsx` calls `runExitCleanup()`, which can reach `Config.shutdown()` -> `GeminiClient.shutdown()` and write the same session summary again. This would duplicate one session record and inflate session-level metrics. Please either centralize the summary emission in one shutdown path, or make `logSessionSummary` idempotent per process/session id.

2. `session.avg_await_duration_seconds` currently uses the tool call `startTime` that is set when the call enters validation. When the tool later enters `awaiting_approval`, the recorded duration can include validation, hook, and scheduling time before the user is actually waiting on an approval prompt. Please record a separate timestamp when transitioning into `awaiting_approval`, and compute the approval wait duration from that point.

samchu-zsl:
Thanks for the thorough review! Both issues are fixed in the latest force-push (`b557292`):

**1. Duplicate session summary emission**

Added an idempotent guard (`sessionSummaryLogged` flag) at the top of `logSessionSummary()`. Regardless of how many exit paths invoke it, only the first call writes to SLS — subsequent calls are no-ops.

**2. Inaccurate `avg_await_duration_seconds`**

Introduced `awaitStartTime` on `WaitingToolCall`. The timestamp is captured when the tool call *first transitions* to `awaiting_approval` (via `setStatusInternal`), and `handleConfirmationResponse` now computes duration from `awaitStartTime` instead of `startTime`. This excludes validation, hook, and scheduling time from the metric.

kongche-jbw:
Thanks for the update. The previous two issues look addressed in `b557292`, but I found one more session-boundary case around `/clear`:

`clearCommand` calls `config.startNewSession()` and then `uiTelemetryService.reset()`, while `GeminiClient.resetChat()` only fires the session-end hook and does not emit the SLS summary. This means the metrics for the session before `/clear` can be reset without ever being written to `cosh.jsonl`.

Also, the new `sessionSummaryLogged` guard is process-wide. If we fix `/clear` by flushing the current session before resetting metrics, a process-wide boolean would then prevent the later session in the same process from being emitted. It would be safer to make the idempotency guard keyed by `session.id` instead of a single boolean, and flush the current session summary before switching/resetting session state.

kongche-jbw:
Thanks, the `/clear` path now looks fixed: it flushes before resetting metrics, and the idempotency guard is keyed by `session.id`.

I found one more similar session-switching path in `/resume`: `useResumeCommand` calls `config.startNewSession(sessionId, sessionData)` and then reinitializes the client. During `GeminiClient.initialize()`, resumed telemetry is replayed from the selected session history, which resets/rebuilds the telemetry metrics. Since the current session is not flushed before that switch, the current session summary can still be lost when a user resumes another session.

Could you also call `logSessionSummary(config)` before the resume path switches session state, similar to the `/clear` fix? That should keep all in-process session transitions consistent.

samchu-zsl:
Good catch on the `/clear` session boundary. Fixed in `d519de3`:

1. **Flush before reset** — `clearCommand` now calls `logSessionSummary(config)` before `startNewSession()` + `uiTelemetryService.reset()`, so the outgoing session's metrics are persisted to `cosh.jsonl` before being cleared.

2. **Per-session idempotency** — Replaced the process-wide boolean guard with a `Set<string>` keyed by `session.id`. Each session can be emitted exactly once, and new sessions created via `/clear` are not blocked from writing their summary on exit.

samchu-zsl:
Fixed in `fd75760`. The `/resume` path now calls `logSessionSummary(config)` before `config.startNewSession()`, consistent with the `/clear` fix. All in-process session transitions (clear, resume) now flush the outgoing session summary before switching state.

kongche-jbw:
非阻塞建议：clearCommand.test.ts 和 useResumeCommand.test.ts 目前补了 logSessionSummary mock，没有断言它在 startNewSession / reset / initialize 前调用。
可以考虑补一下调用顺序测试，避免后续重构又把 flush 放到 session 切换之后。

kongche-jbw (approved):

Issue #1070, opened 2026-06-23
[anolisa] feat(install): declarative service/capability/hook operations with raw–RPM backend parity


ikunkun-sys:
## Component

anolisa

## Problem Statement

#870 extends `anolisa install/uninstall` to all distributed components, but explicitly defers two things to follow-up issues: **adapter install** (#998 / #1061) and **service enable/disable (systemd unit activation)**. In practice, several components are *not actually usable* after a file-only install — they need install-time operations beyond laying files:

- **agentsight (sight)**: needs file capabilities `cap_bpf,cap_perfmon` applied to the binary via `setcap`, plus its `agentsight.service` enabled and started. Without the caps, eBPF load fails for non-root users; without the service it never runs.
- **ws-ckpt (ckpt)**: a btrfs COW daemon. On **uninstall it must first `ws-ckpt recover --all` while the daemon is still up**, otherwise it leaves half-checkpointed btrfs subvolumes and corrupts workspace data; after files are removed it must tear down loop devices and the state dir.
- **agent-memory (memory)**: a per-user background service managed via `systemctl --user` (template unit `anolisa-memory@<user>.service`).

The **RPM backend** already performs these via `.spec` scriptlets (`%post` / `%preun` / `%postun`, file caps). The **raw (tarball) backend** has no equivalent, so the *same component reaches a different system state* depending on which backend installed it. There is also no contract surface for a component to declare "this unit must be enabled", "this binary needs these caps", or "run this step at this lifecycle phase" — today it is hand-coded per component and per backend.

## Proposed Solution

Let a component **declare**, in its `component.toml` contract, the three classes of *install-related* extra operations, and have each backend honor the **same single declaration**:

- `[[component.services]]` — `{ unit, scope = system|user, enable, start, instance? }`
- `[[component.capabilities]]` — `{ path, caps = [...], optional }` (file caps via `setcap`)
- `[[component.hooks]]` — `{ phase, script, timeout_secs, strict? }` for component-specific imperative steps that declarative file/service/cap entries cannot express (e.g. ws-ckpt's pre-uninstall recover)

Execution model:

[...]

PR #1080, opened 2026-06-23
feat(anolisa): execute declared install operations on the raw backend

ikunkun-sys:
## Summary

Implement the **raw (tarball) backend** half of declarative install operations
from #1070: a component declares `[[component.services]]`,
`[[component.capabilities]]`, and `[[component.hooks]]` in its `component.toml`
contract, and anolisa's raw install/uninstall executors honor that single
declaration.

## Background / Motivation

#1070 notes that several components are not usable after a file-only install:
agentsight needs `cap_bpf,cap_perfmon` on its binary and its
`agentsight.service` enabled/started; ws-ckpt must `recover` while its daemon is
still up *before* uninstall removes files. The RPM backend already performs
these via `.spec` scriptlets, so the *same component reached a different system
state* depending on which backend installed it. This PR gives the raw backend
parity for those install-related operations, driven by one declarative contract.

## What changed

- **Contract parsing**: the `[component.layout]` minimal-schema parser now reads
  sibling `[[component.services]]` / `[[component.capabilities]]` /
  `[[component.hooks]]` declarations.
- **Capabilities executor** (`setcap`): applies declared Linux file caps on raw
  install/update; install keeps `optional=false` abort/rollback semantics, while
  raw update applies caps only after the new state is durable and reports any
  failure as a committed warning because xattrs are not transaction-journaled.
- **Service activation**: enables and starts declared systemd units on raw
  install; stops then disables them on uninstall.
- **Lifecycle hooks**: runs the contract's standalone hook scripts in the fixed
  order - install: `pre_install -> lay files -> setcap -> post_install ->
  service enable/start -> post_enable`; uninstall: `pre_uninstall ->
  service stop/disable -> remove files -> post_uninstall`. Hook scripts are
  shipped as ordinary layout files; no shell is inlined into the contract.
  On a fresh raw install, an artifact-local `pre_install` script is not on disk
  yet and is expected to skip as Missing; that phase is mainly useful when a
  prior version already laid the script.
- **Failure semantics**: file/symlink -> transactional rollback; install
  `setcap` optional -> warn, required -> abort/rollback; service activation ->
  best-effort warn; hook `strict=false` -> warn-and-continue, `strict=true` ->
  abort and roll back laid files and the manifest snapshot. Post-service hard
  failures best-effort stop/disable units activated by the failed install before
  rolling back files; raw update applies capability/service side effects only
  after the new state has been persisted, so those failures are reported as
  committed warnings instead of triggering an incomplete rollback.
- **Backend gating (no double-execution)**: these executors run **only for
  raw-managed** components; rpm-managed / rpm-observed teardown returns early
  and is left to `.spec` scriptlets.
- **Mode gating**: `setcap` and `scope=system` services run only in system mode;
  user mode / containers / non-Linux skip silently (treated as "not supported").

## Not addressed in this PR (follow-ups to #1070)

This PR intentionally covers only the raw backend, so #1070 stays open for:

- **`scope=user` service activation** via `systemctl --user` (needed by
  agent-memory); service enable/start currently handles system scope only.
- **`{unitdir}` / `{userunitdir}` layout template variables.**

(agent-sec-core stays RPM-only by design, per #1070.)

## Test

- Unit (core): capability executor (apply / optional-warn / required-abort /
  unsupported-skip); service activate and deactivate (enable+start, stop+disable,
  not-supported skip); hook resolution with phase filtering.
- Unit (cli): install-side hook phase classification and invalid-placeholder
  rejection.
- E2E (cli, raw backend): post_install hook runs; a strict post_install failure
  rolls back installed files and the manifest snapshot; pre_install is silently
  skipped as Missing on fresh install (script not yet laid); a contract-declared
  pre_uninstall hook runs on uninstall.
- Full gate green: `cargo fmt --all -- --check`,
  `cargo clippy --workspace --all-targets -- -D warnings`,
  `cargo test --workspace`, `cargo doc --workspace --no-deps`.

Refs #1070

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
@@ -21,12 +21,12 @@
 //! then pick an action by `(backend, rpmdb hit, install mode)`. `npm` remains
 //! NOT_IMPLEMENTED.
 //!
-//! Deliberately out of scope for this milestone: execution-policy gating,
-//! pre/post hooks, health checks, and service start/enable. Installed
-//! services are recorded in state with `enabled: false`.
+//! Deliberately out of scope for this milestone: execution-policy gating and
+//! health checks.
 
 use clap::Parser;
 use serde::{Deserialize, Serialize};
+use std::collections::BTreeSet;
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::path::{Path, PathBuf};
@@ -44,8 +44,11 @@ use anolisa_core::state::{
     ObjectStatus, OperationRecord, OwnedFile, Ownership, RpmMetadata, ServiceRef,
 };
 use anolisa_core::{
-    ArtifactType, ComponentManifest, DistributionEntry, DistributionIndex, FileKind, ResolveQuery,
-    expand_layout_placeholders,
+    ArtifactType, CapabilityRequest, ComponentManifest, DistributionEntry, DistributionIndex,
+    FileKind, HookPhase, HookSpec, ResolveQuery, ServiceActivation, ServiceManager, ServiceRequest,
+    ServiceRunOutcome, apply_capabilities, apply_services, capability_for_install_mode,
+    deactivate_services, expand_layout_placeholders, resolve_manifest_hooks, run_hooks,
+    service_for_install_mode,
 };
 use anolisa_platform::fs_layout::FsLayout;
 use anolisa_platform::pkg_query::{PackageInfo, PackageQuery, PackageQueryError};
@@ -118,7 +121,8 @@ pub(crate) struct RawResolution {
 struct InstallPreview {
     resolution: RawResolution,
     files: Vec<ResolvedInstallFile>,
-    services: Vec<String>,
+    services: Vec<ServiceRequest>,
+    capabilities: Vec<CapabilityRequest>,
 }
 
 /// Execution input after the artifact has been verified and its install
@@ -130,7 +134,12 @@ pub(crate) struct PreparedInstall {
     pub(crate) resolution: RawResolution,
     pub(crate) artifact_path: PathBuf,
     pub(crate) files: Vec<ResolvedInstallFile>,
-    pub(crate) services: Vec<String>,
+    /// Declared service activations (unit + scope + enable/start), applied
+    /// after files land. Carried resolved with template instances expanded.
+    pub(crate) services: Vec<ServiceRequest>,
+    /// Linux file capabilities to apply after files land (raw, system mode
+    /// only). Carried resolved — path already layout-expanded and bounded.
+    pub(crate) capabilities: Vec<CapabilityRequest>,
     pub(crate) manifest_toml: String,
 }
 
@@ -168,6 +177,9 @@ struct InstallPlanPayload {
     artifact: ArtifactInfo,
     files: Vec<String>,
     services: Vec<String>,
+    /// Human-readable `path: cap,cap` lines for the capabilities install
+    /// would apply. Rendered for `--dry-run`; setcap is never run here.
+    capabilities: Vec<String>,
     dry_run: bool,
     warnings: Vec<String>,
 }
@@ -1913,10 +1925,11 @@ fn build_install_preview(
             resolution,
             files: Vec::new(),
             services: Vec::new(),
+            capabilities: Vec::new(),
         });
     };
 
-    let (files, services) = match resolve_manifest_contract(
+    let (files, services, capabilities) = match resolve_manifest_contract(
         &contract.manifest,
         layout,
         &resolution,
@@ -1929,7 +1942,7 @@ fn build_install_preview(
                 "local catalog manifest does not match resolved artifact; file and service details are unavailable: {}",
                 err.reason()
             ));
-            (Vec::new(), Vec::new())
+            (Vec::new(), Vec::new(), Vec::new())
         }
         Err(err) => return Err(err),
     };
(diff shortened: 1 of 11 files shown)


kongche-jbw:
## Review Notes

### Findings

1. **[P1] install enables/starts services before state is durable, so later failures can leave orphan systemd state**

`install.rs` runs `apply_services(... Start ...)` before the final `state.save()`. If a strict `post_enable` hook fails after services are started, or if `state.save()` fails later, the code rolls back files and the manifest snapshot but does not stop/disable the already enabled/started service.

Result: the command returns failure, files may be removed, and state has no component record, but systemd can still have an enabled/running unit.

Suggested fix: either move durable state save before service activation, or best-effort `deactivate_services()` on every post-service hard failure / state-save failure before rolling back files. Strict `post_enable` failure needs the same cleanup path.

2. **[P1] update restarts services before state is durable, so rollback can leave a new-version process running**

`update.rs` runs `apply_services(... Restart ...)` before state persistence. If state save fails and `raw_update_rollback()` restores the old files, the service may still be running the new version process.

This is more subtle than install: disk state returns to the old version, but runtime state may still be the restarted new binary.

Suggested fix: move service restart after state has been successfully persisted. Since restart is already best-effort, failures can be warnings after the update is committed. If restart must stay before persistence, rollback needs an explicit service cleanup/restart strategy, which is probably more complex.

3. **[P2] successful install output still says services are “recorded, not started”**

The PR now actually enables/starts services when supported, but the human output still renders:

```text
services (recorded, not started):
```

This is misleading. Dry-run can keep “would apply” style wording, but successful install output should say something like `services:` or `services (enabled/started when supported):`, with warnings showing unsupported/skipped cases.

4. **[P2] PR description overstates fresh-install `pre_install` behavior**

The PR description lists install order as `pre_install -> lay files`, but for a fresh raw install the hook script usually ships inside the artifact and is not on disk yet. The code handles this as `Missing`, which means `pre_install` is mostly not an executable fresh-install capability unless the script already exists from a prior version.

If this is intentional, please make the PR description / contract wording explicit: fresh raw install artifact-local `pre_install` scripts are expected to skip as missing; the phase is mainly useful for update or prior-version hooks.

### Confirmed Reasonable

- raw/rpm gating direction looks right: rpm-managed / rpm-observed uninstall goes through the rpm-delegating path, not the raw file-removal executor.
- capability/service/hook execution is scoped to the raw backend, avoiding double execution with RPM scriptlets.
- setcap and service executors have Linux/system/container gating, and user mode currently skips as documented.
- The PR does not appear to conflict with the already merged adapter contract/status work in #1066 / #1072.

### Summary

The direction is good and matches #1070, but I would not approve yet. The two P1 items should be fixed first: service activation/restart should not happen before the last rollback-capable persistence point unless rollback also cleans up service side effects.

kongche-jbw:
## 继续 Review

### 仍需关注

1. **[P1] raw update rollback 可能丢失旧文件上的 file capabilities**

这版已经把 service restart 移到 durable state save 之后，这个方向是对的。但 `update.rs` 现在仍然在 state 持久化前执行 capability apply。

问题是：一旦后续发生 rollback，`raw_update_rollback()` 会从 transaction backup 恢复旧文件；但当前 transaction backup/restore 只恢复文件字节或 symlink，不恢复 xattr。file capabilities 是 xattr，所以旧版本二进制如果之前带有 `setcap`，rollback 后可能只恢复了旧文件内容，却丢了旧文件的 capabilities。

结果会变成：state 仍然回到旧版本 installed 状态，但旧版本运行所需的 capability 可能已经不存在。

可能触发场景包括：
- required capability apply 后续又发生 rollback；
- state save 失败触发 rollback；
- 任何 capability 已应用、新文件已落盘、但最终未完成提交的 update 失败路径。

建议处理方式三选一：

- 更简单：把 update 的 capability apply 也移到 durable state save 之后，失败作为已提交更新后的 warning/degraded 处理，而不是 rollback 触发点；
- 更完整：transaction backup/restore 支持保存并恢复 capability xattr；
- 或者 rollback 完旧文件后，根据旧 manifest 重新 apply 旧版本声明的 capabilities，并把失败显式暴露。

至少需要补一个测试或在 PR limitation 里明确说明。当前 PR 描述里说 raw update 是 transactional rollback，但对新引入的 capability 状态并不完整。

### 已确认修复

- 之前提到的 install service orphan 问题已修：`post_enable` failure 和 `state.save()` failure 会调用 `rollback_activated_services()`，再回滚文件和 manifest。
- update service restart 现在已经挪到 state save 之后，避免了“磁盘回滚到旧版本但进程仍是新版本”的问题。
- 成功输出里的 `services (recorded, not started)` 已改成 `services (enabled/started when supported)`，用户语义对齐了。

### 额外说明

`pre_install` 对 fresh raw install 来说仍然基本会是 Missing，因为 hook script 通常也在当前 artifact 里，文件还没 lay down。这个可以接受，但 PR 描述里最好说清楚：fresh install 的 artifact-local `pre_install` 不是一个真正可执行的阶段，主要适用于已有旧版本文件的 update / prior-version hook 场景。

整体比上一版稳很多，但 capability rollback 这个点建议先收口后再 approve。

ikunkun-sys:
Addressed the raw update capability rollback risk in `dd80e601`.

Raw update now applies declared file capabilities only after the new state has been durably persisted, matching the post-commit service restart behavior. Since capability xattrs are not covered by the rollback journal, failures after commit are surfaced as committed warnings instead of triggering an incomplete rollback that could restore old file bytes without old xattrs.

I also added a focused unit test for converting required post-commit capability failures into warnings and updated the PR description to document the behavior.

kongche-jbw (approved):

PR #1071, opened 2026-06-23
feat(tokenless): add compression toggle with dry-run compare mode

shiloong:
## Description

Add a compression on/off toggle to tokenless, plus a dry-run compare mode that enables end-to-end A/B measurement of compression savings on the same task.

- New `TOKENLESS_COMPRESSION_ENABLED` config field (default `true`) with env override; priority: env > config.json > default. Empty env values are treated as unset, consistent with the existing `stats`/`sls` toggles.
- `CompressionMode` enum (`Active`/`DryRun`) persisted per stats record; the `stats` SQLite table gains a `mode` column (migrated via `pragma_table_info`), and legacy NULL/unknown rows fall back to `Active` so historical data stays readable.
- In `compress-schema`/`compress-response`/`compress-toon`, the predicted savings are always computed and recorded. When compression is disabled, the original (uncompressed) text is emitted so the LLM context stays uncompressed, while the predicted savings are still captured — enabling same-task A/B comparison.
- New `stats summary --compare BASELINE_SESSION TOKENLESS_SESSION [--json]` reports side-by-side token usage by operation, with a warning when a session's recorded mode does not match the expected baseline=DryRun / tokenless=Active.

## Related Issue

no-issue: standalone observability feature; no tracked issue. Adds a toggle + measurement tooling to quantify tokenless compression savings.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

## Scope

- [x] `tokenless` (tokenless)

## Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [x] My code follows the project's code style
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly
- [x] For `tokenless`: `cargo clippy -- -D warnings` and `cargo fmt --check` pass
- [x] Lock files are up to date (`package-lock.json` / `Cargo.lock`)

## Testing

Ran the tokenless pre-commit gate locally on the rebased branch:

```bash
cd src/tokenless
cargo fmt --all --check      # clean
cargo clippy --workspace -- -D warnings   # 0 warnings
cargo test --workspace      # 67 passed, 0 failed
```

New unit tests added: `CompressionMode` roundtrip / `from_db` legacy fallback, default-mode-active, env/file override precedence, empty-env-as-unset, and `records_by_session` filtering for the compare flow.

## Additional Notes

[...]

PR #1085, opened 2026-06-24
fix(tokenless): align compression mode serde/db form and dedup config load

shiloong:
## Description

Follow-up fixes for the compression-toggle feature merged in #1071:

- `CompressionMode::as_str()` now returns `"dry-run"` to match its `#[serde(rename_all = "kebab-case")]` serialization; `from_db()` accepts both `"dry-run"` and the legacy `"dryrun"` so historical rows stay readable. Removes a latent mismatch where a future JSON round-trip of `StatsRecord` would have deserialized `"dry-run"` back to `Active`.
- `record_compression_stats` now takes `&TokenlessConfig` instead of reloading config each call, eliminating a duplicate config file read per compression command (consistent with the existing slow-filesystem avoidance rationale). `#[allow(clippy::too_many_arguments)]` follows repo convention (used in anolisa-core/anolisa-cli); the config param is a real dependency, not a band-aid.
- Recorder tests return `(StatsRecorder, TempDir)` instead of `std::mem::forget(dir)`, keeping the temp dir alive via ownership so it is cleaned up on test exit.

## Related Issue

no-issue: small correctness/efficiency hardening of the compression-toggle stats path; no tracked issue.

## Type of Change

## Scope

- [x] `tokenless` (tokenless)

## Checklist

- [x] For `tokenless`: `cargo clippy -- -D warnings` and `cargo fmt --check` pass

## Testing

```bash
cd src/tokenless
cargo fmt --all --check        # clean
cargo clippy --workspace -- -D warnings   # 0 warnings
cargo test --workspace        # 67 passed, 0 failed
```

Updated `test_compression_mode_roundtrip` asserts the new `"dry-run"` string and the legacy `"dryrun"` back-compat path.

## Additional Notes

Backward compatible: existing DBs with `"dryrun"` rows continue to read as `DryRun`; only newly written rows use `"dry-run"`.

### DIFF
--- a/src/tokenless/crates/tokenless-cli/src/main.rs
+++ b/src/tokenless/crates/tokenless-cli/src/main.rs
@@ -333,7 +333,8 @@ fn run() -> Result<(), (String, i32)> {
                 after_compact.clone()
             };
 
-            let compression_on = TokenlessConfig::load().is_compression_enabled();
+            let config = TokenlessConfig::load();
+            let compression_on = config.is_compression_enabled();
             let mode = resolve_mode(compression_on, before_tokens, after_tokens);
             let emit_text = if compression_on {
                 output_text.clone()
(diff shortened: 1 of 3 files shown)


Forrest-ly (approved):
LGTM

PR #1089, opened 2026-06-24
feat(anolisa): place declarative systemd units via {unitdir}/{userunitdir}

ikunkun-sys:
## Description

The raw backend parses `[[component.services]]` and runs `systemctl enable/start <unit>`, but the unit file was laid under `{datadir}` (e.g. `/usr/local/share/anolisa/agentsight/agentsight.service`). systemd resolves units by name against its search path, so activation failed with `Unit file ... does not exist` and degraded to a warning. This exposes the systemd unit search directories as layout placeholders so contracts place units where systemd looks, and reloads the unit database before activation.

Behavior:
- `{unitdir}` in a layout target → the **system** unit dir: `/usr/local/lib/systemd/system` (system mode), `~/.config/systemd/user` (user mode).
- `{userunitdir}` → the **user** unit dir: `/usr/local/lib/systemd/user` in system mode (system-wide — every user's `systemctl --user` can find it, the `/usr/local` analogue of `/usr/lib/systemd/user`), `~/.config/systemd/user` in user mode.
- Snake-case aliases `{unit_dir}` / `{user_unit_dir}` are accepted, matching the existing `{lib_dir}`/`{etc_dir}` style.
- A unit installed this run is `systemctl daemon-reload`-ed once before enable/start, so a freshly-placed unit is loadable and `start` no longer fails with "not found".
- Unit targets are owned: uninstall removes them with the rest of the manifest.

## Design Note

`FsLayout::systemd_unit_dir` already existed; this adds a sibling `systemd_user_unit_dir` (new `fhs::SYSTEMD_USER_UNITS = /usr/local/lib/systemd/user`; user mode reuses `~/.config/systemd/user`, since a user install has no system-manager units). Both map through the same `expand_layout_placeholders` that already renders layout files, adapters, capabilities, and hook scripts, so the layout renderer picks up the new placeholders with no call-site change.

`path_safety::lexical_roots` gains both unit dirs. This is required in both modes, not just system: in user mode `~/.config/systemd/user` lives **outside** the anolisa `etc_dir` (`~/.config/anolisa`), and in system mode `/usr/local/lib/systemd/{system,user}` are outside the `/usr/local/share/anolisa` tree — without listing them the owned-path boundary check rejects unit targets as `External`. Adding them as owned roots is also what lets uninstall remove the placed unit.

`ServiceManager::daemon_reload` is a defaulted no-op overridden only by the systemd backend (`systemctl daemon-reload`); the not-supported and fake managers inherit the no-op. `apply_services` invokes it once, before the per-unit loop, only when the backend is supported and at least one **system-scope** unit will actually be enabled/started. User-scope units are skipped in the loop, so no `systemctl --user daemon-reload` is attempted. Reload failure is best-effort (collected as a warning), consistent with enable/start.

## Ship Note

- Placeholders resolve mode-aware via the layout. Placing a system-scope unit under `{unitdir}` in a *user*-mode install lands it in `~/.config/systemd/user` — harmless, since the service manager is `NotSupported` in user mode and never activates it.
- User-scope **auto**-activation is still unsupported: `apply_services` records a documented skip for `scope=user` units. `{userunitdir}` fixes unit *placement* only — the template unit lands where `systemctl --user enable <unit>` can find it and a future user-scope executor can drive it — but this PR does not enable/start user-scope services.
- `expand_layout_placeholders` still hard-fails on a genuinely unknown placeholder (`UnknownPlaceholder`), so a contract using `{unitdir}` requires a CLI that knows it. Gating that with a clean `min_anolisa_version` message is left to the release that bumps the version; this change keeps the workspace at 0.1.13.
- Uninstall-side `daemon-reload` (after the unit file is removed) is not added here; this PR targets the install-side activation failure.

## Test

- Unit (fs_layout): `systemd_user_unit_dir` resolves to `/usr/local/lib/systemd/user` (system mode, incl. custom prefix) and `~/.config/systemd/user` (user mode, incl. explicit XDG overrides).
- Unit (adapter): `{unitdir}`/`{unit_dir}` → the system unit path; `{userunitdir}`/`{user_unit_dir}` → system-wide vs per-user user-unit path.
- Unit (path_safety): `validate_owned_path` accepts targets under both unit dirs.
- Unit (service): existing `apply_services` ordering tests updated to expect the leading `daemon-reload`; new test that a failed reload warns but still enable/starts; a user-scope-only request triggers no reload (manager untouched); central-log audit emits the `service:daemon-reload` Info line.
- Real-host dry-run (system mode): component contracts declaring services resolve their unit to `/usr/local/lib/systemd/system/agentsight.service`, `/usr/local/lib/systemd/system/ws-ckpt.service`, and `/usr/local/lib/systemd/user/anolisa-memory@.service` — no expansion or owned-path rejection, capabilities and services planned with no warnings.

Refs #1070

### DIFF
--- a/src/anolisa/crates/anolisa-core/src/adapter.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter.rs
@@ -380,6 +380,13 @@ pub fn expand_layout_placeholders(
     replacements.insert("log_dir", &layout.log_dir);
     replacements.insert("cachedir", &layout.cache_dir);
     replacements.insert("cache_dir", &layout.cache_dir);
+    // systemd unit search dirs: `{unitdir}` for system-scope units,
+    // `{userunitdir}` for user-scope template units. Both are mode-aware
+    // via the layout (see `FsLayout::systemd_user_unit_dir`).
+    replacements.insert("unitdir", &layout.systemd_unit_dir);
+    replacements.insert("unit_dir", &layout.systemd_unit_dir);
+    replacements.insert("userunitdir", &layout.systemd_user_unit_dir);
+    replacements.insert("user_unit_dir", &layout.systemd_user_unit_dir);
 
     let mut result = template.to_string();
     let mut search_from = 0;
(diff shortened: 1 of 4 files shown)


kongche-jbw:
Reviewed against the current anolisa adapter/service progress. No blocking issue found.

This PR closes the install-side gap for declarative systemd unit placement: `{unitdir}` / `{userunitdir}` resolve into systemd unit search paths, those paths are included in owned roots, and `systemctl daemon-reload` runs before system-scope enable/start.

Known limitations look acceptable for this PR:
- user-scope activation is still skipped;
- uninstall-side daemon-reload is not included yet;
- custom-prefix unit dirs are suitable for staging/testing but are not visible to host systemd automatically.

I suggest tracking uninstall-side daemon-reload as a follow-up under the service lifecycle issue.

kongche-jbw (approved):

PR #1088, opened 2026-06-24
feat(tokenless): enable SLS recording by default and document usage

shiloong:
## Description

Flip the `sls_enabled` default from `false` to **`true`** so SLS JSONL logging is on out of the box, mirroring SQLite stats. tokenless user manuals previously documented SQLite stats but **never mentioned SLS** — this fills that gap.

- `TokenlessConfig::default()`: `sls_enabled` false → true; serde default changed from `#[serde(default)]` to `#[serde(default = "default_true")]`.
- Updated the affected config tests (`sls_enabled_default`, `load_sls_missing_file`, `sls_env_override_*`) to assert the new default.
- Added an **SLS Log Recording (JSONL)** section to both the zh and en user manuals: default-on note, field list, env vars (`TOKENLESS_SLS_ENABLED` / `TOKENLESS_SLS_PATH`), output path, production notes; plus a troubleshooting row and file-location entries.
- Added a cross-link note in `response-compression.md`.

Backward compatible: SLS can still be turned off via `TOKENLESS_SLS_ENABLED=0` or `sls_enabled: false` in `~/.tokenless/config.json`.

## Related Issue

no-issue: enable SLS recording by default + document its usage; no tracked issue.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

## Scope

- [x] `tokenless` (tokenless)

## Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [x] My code follows the project's code style
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly
- [x] For \`tokenless\`: \`cargo clippy -- -D warnings\` and \`cargo fmt --check\` pass

## Testing

```bash
cd src/tokenless
cargo fmt --all --check               # clean
cargo clippy --workspace -- -D warnings   # 0 warnings
cargo test --workspace                # 67 + 41 + 24 + 19 passed, 0 failed
```

Updated `sls_enabled_default` / `load_sls_missing_file` / `sls_env_override_*` tests assert the new default-on state.

PR #1092, opened 2026-06-24
fix(tokenless): make SLS writer append-only and skip when log file absent

shiloong:
## Description

Follow-up to #1088. The SLS JSONL file is owned and lifecycle-managed (created, rotated, removed) by the **anolisa SLS component**; tokenless must not manage it. This changes `SlsWriter::write()` accordingly:

- On each write, check whether the file exists first; **append only when it exists, silently skip otherwise** (treat as "SLS collection not active").
- Drop `create_dir_all` (no directory creation), drop `.create(true)` + `mode(0o600)` (no file creation), drop the 100 MiB size-warning block (rotation is the SLS component's job).
- Keep `O_NOFOLLOW` to refuse a swapped symlink at the final path component.

Documents the ownership / append-only contract in both user manuals and `response-compression.md`.

## Related Issue

no-issue: SLS log file lifecycle is owned by the anolisa SLS component; tokenless should only append, never create/delete.

## Type of Change

- [x] Refactoring (no functional change for the happy path; behavior change only when the file is absent — now skips instead of creating)

## Scope

- [x] \`tokenless\` (tokenless)

## Checklist

- [x] For \`tokenless\`: \`cargo clippy -- -D warnings\` and \`cargo fmt --check\` pass

## Testing

```bash
cd src/tokenless
cargo fmt --all --check               # clean
cargo clippy --workspace -- -D warnings   # 0 warnings
cargo test --workspace                # 67 + 41 + 24 + 19 passed, 0 failed
```

Updated writer tests pre-create the file (as the SLS component would) and a new `test_sls_writer_skips_when_file_missing` asserts tokenless never creates it.


Forrest-ly:
PR #1092 Review: fix(tokenless): make SLS writer append-only and skip when log file absent

  Scope: +64/-75 (net -11), 4 files. Changes SlsWriter::write() from create-on-demand to append-only-if-exists. Removes create_dir_all, .create(true), mode(0o600), and the 100 MiB size warning. Keeps O_NOFOLLOW. Updates docs (EN+ZH) and
  tests.

  ---
  High

  1. TOCTOU between path.exists() and opts.open() — sls.rs:252-267

  if !self.path.exists() {
      return;                 // ① file exists here
  }
  // ... serialize ...
  opts.open(&self.path)       // ② file may be gone (rotated/removed by SLS component)

  Between the existence check and the open, the SLS component can rotate or remove the file. The open will fail and hit the existing fail-silent if let Err(e) handler, so there's no crash — but the comment says O_NOFOLLOW "narrows the
  TOCTOU window" when the real TOCTOU is this check-then-open gap. The exists check is redundant: just remove it and let the open fail naturally (.append(true) without .create(true) will return NotFound for a missing file, hitting the same
  fail-silent path). This eliminates the race and simplifies the code.

  ---
  Medium

  2. Removing the 100 MiB size warning loses operational visibility — sls.rs:286-303 (removed)

  The old code emitted a one-time stderr warning when the file exceeded 100 MiB. The rationale for removal ("rotation is the SLS component's job") is valid, but if the SLS component's rotation is misconfigured or not running, the file
  grows unbounded with zero warning. Since tokenless is the writer, it's the only component that observes file growth in real time. A compromise: keep the warning but raise the threshold (e.g., 500 MiB) and reword it to say "SLS component
  may not be rotating."

  3. O_NOFOLLOW without .create(true) changes error behavior — sls.rs:261-267

  Previously, O_NOFOLLOW + .create(true) meant: if the file is a symlink, fail with ELOOP; otherwise create/open it. Now, without .create(true), if the file doesn't exist, the open fails with NotFound before O_NOFOLLOW is even evaluated.
  The O_NOFOLLOW flag is still correct for the attacker-swaps-file-to-symlink scenario, but the comment could note that O_NOFOLLOW only matters when the file exists (after the existence check passes).

  ---
  Low

  4. Tests pre-create with fs::write(&path, "") but use std::fs not the import — sls.rs:449, 466

  The test file has use std::fs::OpenOptions; but not use std::fs;. The fs::write calls work because of the existing use crate::{StatsRecord, VERSION}; scope, but they rely on the module-level use std::fs; that was just removed from the
  production code (line 22 of the diff). Check that the test module has its own use std::fs; — if it relies on a glob or prelude import, this compiles only by accident.

  5. Unused std::fs import removal may break other call sites — sls.rs:22

  The diff shows use std::fs; removed and replaced with use std::fs::OpenOptions;. If any other function in this file uses fs::metadata, fs::read_to_string, etc., this will fail to compile. The test suite passing suggests it's fine, but
  worth confirming no other fs:: usage exists in the file outside the removed blocks.

  ---
  Positive

  - Clean ownership model: tokenless is a log appender, not a log manager — correct separation of concerns
  - Docs updated in both EN and ZH with consistent messaging and practical examples (mkdir -p /tmp && touch)
  - New test_sls_writer_skips_when_file_missing directly asserts the core invariant
  - Existing tests updated to pre-create the file, modeling the real deployment topology
  - O_NOFOLLOW retained for symlink-swap hardening despite simplified flow
  - Net negative LOC (-11 lines) — the change removes complexity

  ---
  Summary: Well-scoped cleanup with correct ownership semantics. The main concern is the unnecessary TOCTOU (issue #1) — the existence check adds a race window that the fail-silent open already handles; removing it would be both simpler
  and safer. The size warning removal (issue #2) trades operational visibility for cleaner separation; consider keeping a high-threshold warning as a safety net.

shiloong:
Thanks @Forrest-ly for the thorough review. Responses per item:

**#1 (TOCTOU / `exists()` check) — keeping as-is.** The existence gate is intentional: it preserves the required "skip *silently* when the SLS collector hasn't created the file" contract. With it, a missing file returns immediately with no stderr noise; removing it would route the missing-file case through `opts.open()` and emit a one-time `write error` warning, which contradicts the silent-skip behavior this PR introduces. The check→open race is benign: if the SLS component rotates/removes the file between `exists()` and `open()`, the `open()` fails (NotFound) and falls through to the existing fail-silent handler — no data loss or corruption, just that one record is skipped. `O_NOFOLLOW` narrows the symlink-swap window between those two calls. Net: the explicit gate + silent skip is the behavior we want here.

**#2 (size warning) — declining.** This is a deliberate ownership boundary: the anolisa SLS component owns the file lifecycle (creation, rotation, removal), and tokenless is a pure appender that does not observe or manage file size/rotation. Re-adding a size monitor would re-couple tokenless to lifecycle concerns this PR is explicitly decoupling. Operational visibility for misconfigured rotation belongs to the SLS component, not the writer.

**#3 (O_NOFOLLOW comment)** — the existing comment already scopes it to "blocks a swap-to-symlink between the existence check and the open," i.e. the window after `exists()` passes. Leaving as-is.

**#4 (test `fs::write` import)** — confirmed fine: the test module has its own `use std::fs;` (sls.rs:287), so `fs::write` / `fs::read_to_string` resolve there directly — not relying on a glob/prelude.

**#5 (removed `use std::fs;`)** — confirmed: after removal, the only `fs` references in production code are `use std::fs::OpenOptions;` (line 14) and the fully-qualified `std::os::unix::fs::OpenOptionsExt;` (line 258); no bare `fs::` call sites remain. `cargo fmt --check` / `cargo clippy -D warnings` / `cargo test --workspace` all pass.

No code changes needed; happy to revisit if any of the above doesn't land.

Forrest-ly (approved):
LGTM

PR #1095, opened 2026-06-24
fix(skillfs): require precreated session metrics log

kongche-jbw:
## Summary

Require the SkillFS session metrics log file to be pre-created before writing.

SkillFS now only appends to the deployment-managed SLS JSONL file and no longer creates it implicitly. This keeps file ownership, permissions, and log rotation policy fully controlled by the logging/deployment setup.

## Changes

- Remove implicit `create(true)` from `SessionStatsWriter`.
- Open the session metrics log in append-only mode.
- Treat a missing log file as a non-fatal configuration error.
- Update writer tests to pre-create log files for success cases.
- Add a regression test proving missing log files return `NotFound` and are not created.

## Log Path

```text
/var/log/anolisa/sls/ops/skillfs.jsonl
```

This file must be created by deployment or log collection setup before SkillFS writes session metrics.

## Validation

```bash
cargo fmt
cargo test -p skillfs-fuse session_stats --lib
cargo check -p skillfs -p skillfs-fuse
```

Result:

- `session_stats`: 15 passed
- `cargo check`: passed

## Notes

- Missing log files do not affect mount success.
- Write failures remain best-effort and non-fatal.
- This PR keeps log rotation responsibility outside SkillFS.

### DIFF
--- a/src/skillfs/crates/skillfs-fuse/src/security/session_stats_writer.rs
+++ b/src/skillfs/crates/skillfs-fuse/src/security/session_stats_writer.rs
@@ -5,9 +5,11 @@
 //!
 //! Design contract:
 //!
+//! * The target file must already exist. Deployment owns file creation,
+//!   ownership, permissions, and logrotate policy. SkillFS only appends to
+//!   the pre-created file.
 //! * Every write opens the file, appends one JSONL line, and closes the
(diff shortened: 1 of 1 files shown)


yummypeng (approved):
LGTM

PR #1100, opened 2026-06-24
feat(anolisa): activate user-scope services and reload units on uninstall

ikunkun-sys:
## Description

Completes the two install-operations follow-ups tracked under #1070, on top of the declarative service/capability/hook executor (#1080) and unit placement (#1089):

1. **User-scope service activation** — `scope=user` units (e.g. agent-memory's `anolisa-memory@.service`) are now *activated* via `systemctl --user`, not merely placed. A user-mode install drives the caller's `systemctl --user enable/start`, and a raw `update` restarts them the same way; a system-mode install still only places the unit (root has no single target-user session) and leaves per-user `systemctl --user enable` to the user.
2. **Uninstall-side daemon-reload** — after a component's owned unit files are removed, the manager runs `daemon-reload` so systemd drops the now-deleted units (the mirror of the install-side reload from #1089). A user-mode uninstall issues `systemctl --user daemon-reload`. Requested by the #1089 review.

Behavior:
- `SystemdServiceManager` carries a scope; a user-scoped instance prefixes every `systemctl` call with `--user` and labels itself `systemd-user`, so central-log records and `ServiceOutcome.manager` (enable/start/restart/reload) match the `systemd-user` label install state already writes for user mode.
- New `user_service_for_install_mode` factory drives `systemctl --user` only in a user-mode install (Linux, no container); otherwise a quiet skip.
- Both **install** and **update** pick the scope-matched backend by the requests' scope, so a user-scope service is enabled/started on install and *restarted* on update via `systemctl --user` rather than left running on the replaced binary.
- `ServiceManager::handles_scope` routes each request to the backend that drives its scope; a request the chosen backend does not handle (a mixed-scope contract) is a documented skip, never mis-driven.
- The uninstall reload runs only when the component declared service units and the host drives systemd; it selects the user-scope factory in a user-mode uninstall (`systemctl --user daemon-reload`) and the system factory otherwise.

## Design Note

User-scope activation gets its **own** factory rather than making `for_install_mode` scope-aware, because `for_install_mode` is also the factory the **status/health** path uses to probe `kind = "systemd"` checks — and it relies on user mode being *unsupported* so a system-service health check degrades to `not_supported` instead of probing a system unit from a user-mode install. Flipping `for_install_mode("user")` to drive `systemctl --user` regressed that path. So `for_install_mode` stays the system-scope factory (user mode → unsupported), and user-scope work goes through `user_service_for_install_mode`, called by the install/update executors and the uninstall reload.

`apply_services` / `deactivate_services` keep their single-manager signatures; the install/update executors pick the system or user backend by the requests' scope (a contract's services are single-scope in practice). `handles_scope` — defaulted `true`, overridden by the systemd and fake backends — is the safety net that skips any off-scope request.

The uninstall reload runs after the file-removal loop and before `post_uninstall` hooks, and is best-effort (a failed reload only warns). It selects the user-scope factory when `install_mode == "user"` so the removed `~/.config/systemd/user` units get a `systemctl --user daemon-reload`, and the system factory otherwise.

## Ship Note

- **No production behavior change today**: the only user-scope component, agent-memory, declares `enable = false` / `start = false`, so nothing auto-activates. This adds the capability + tests and lands the unit where `systemctl --user enable` can find it (via `{userunitdir}`, #1089).
- **Per-unit user-scope stop/disable on uninstall is intentionally out of scope**: the persisted `ServiceRef` does not record a unit's scope, so uninstall drives the system manager for stop/disable. The post-removal *reload* is already scope-aware by install mode (a user-mode uninstall runs `systemctl --user daemon-reload`), but threading scope through state — so a system-mode uninstall can stop/disable per-unit user services — is a separate follow-up.
- The uninstall reload is a quiet skip on container / non-Linux / user-mode-without-session hosts.

## Test

- Unit (service): `user_service_for_install_mode` support matrix — drives `--user` only in user mode (labeled `systemd-user`), skips system mode (place-only) and containers; `for_install_mode` stays unsupported in user mode (keeps systemd health checks degrading). `apply_services` drives a user-scope unit when the backend handles its scope (reload → enable → start) and skips it under a system backend without touching it.
- Unit (lifecycle): a no-service uninstall never emits a `service:daemon-reload` record (the gate) and still removes owned files.
- Existing `apply_services` / `deactivate_services` and status-health tests unchanged and green.
- Full gate: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, `cargo doc --workspace --no-deps`. One pre-existing platform test (`systemd::tests::unimplemented_unit_operations_return_errors_instead_of_panicking`) fails on the dev host only because `agentsight.service` happens to be installed there; it fails identically on clean `main` and is unrelated to this change.

Refs #1070

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
@@ -46,9 +46,9 @@ use anolisa_core::state::{
 use anolisa_core::{
     ArtifactType, CapabilityRequest, ComponentManifest, DistributionEntry, DistributionIndex,
     FileKind, HookPhase, HookSpec, ResolveQuery, ServiceActivation, ServiceManager, ServiceRequest,
-    ServiceRunOutcome, apply_capabilities, apply_services, capability_for_install_mode,
-    deactivate_services, expand_layout_placeholders, resolve_manifest_hooks, run_hooks,
-    service_for_install_mode,
+    ServiceRunOutcome, ServiceScope, apply_capabilities, apply_services,
+    capability_for_install_mode, deactivate_services, expand_layout_placeholders,
+    resolve_manifest_hooks, run_hooks, service_for_install_mode, user_service_for_install_mode,
 };
 use anolisa_platform::fs_layout::FsLayout;
 use anolisa_platform::pkg_query::{PackageInfo, PackageQuery, PackageQueryError};
@@ -2681,12 +2681,23 @@ fn execute_raw(
     }
 
     // Capabilities done; bring declared services up (issue order: setcap →
-    // service enable/start). The manager gates itself to system + Linux +
-    // non-container; user-scope units and unsupported hosts are quiet skips.
-    // Activation is best-effort — a failed enable/start is a warning, not an
-    // abort: the component's files are installed and an operator can fix the
-    // unit out of band. Reuse the env + log opened for the capability step.
-    let service_manager = service_for_install_mode(ctx.install_mode.as_str(), &env);
+    // service enable/start). Activation is best-effort — a failed enable/start
+    // is a warning, not an abort: the component's files are installed and an
+    // operator can fix the unit out of band. Reuse the env + log opened for
+    // the capability step.
+    //
+    // A contract's services are single-scope in practice (a component is
+    // either a system daemon or a per-user service). Pick the matching
+    // backend: an all-user-scope set drives `systemctl --user` (and only in a
+    // user-mode install); otherwise the system backend. A request the chosen
+    // backend does not handle (a hypothetical mixed-scope contract) is skipped
+    // by `apply_services` via `handles_scope`, so this never mis-drives.
+    let service_manager: Box<dyn ServiceManager> =
+        if !services.is_empty() && services.iter().all(|s| s.scope == ServiceScope::User) {
+            user_service_for_install_mode(ctx.install_mode.as_str(), &env)
+        } else {
+            service_for_install_mode(ctx.install_mode.as_str(), &env)
+        };
     let service_run = apply_services(
         service_manager.as_ref(),
         &services,
--- a/src/anolisa/crates/anolisa-core/src/lifecycle.rs
+++ b/src/anolisa/crates/anolisa-core/src/lifecycle.rs
@@ -1381,6 +1381,57 @@ fn execute_uninstall_or_purge(
         ));
         combined.push(msg);
     }
+
+    // Step 8.5 — daemon-reload AFTER the owned unit files are gone, so the
+    // manager drops the now-deleted units from its database (the
+    // uninstall-side mirror of the install reload — without it systemd keeps
+    // a stale unit until the next manual reload). Best-effort: a failed
+    // reload only warns. Runs only when the component declared service units
+    // (whose files were just removed) on a host that drives systemd.
+    //
+    // A user-mode install places (and reloads) its units under the user
+    // manager, so the post-removal reload selects the user-scope factory in
+    // user mode — `for_install_mode("user")` is unsupported and would skip
+    // the `systemctl --user daemon-reload` the removed `~/.config/systemd/user`
+    // units need. (Per-unit user-scope stop/disable in a *system*-mode
+    // uninstall still waits on `ServiceRef` recording scope.)
+    if live_plan.components.iter().any(|c| !c.services.is_empty()) {
+        let env = EnvService::detect();
+        let manager = if install_mode == "user" {
+            service::user_service_for_install_mode(install_mode, &env)
+        } else {
+            service::for_install_mode(install_mode, &env)
+        };
+        if manager.supported() {
+            match manager.daemon_reload() {
+                Ok(_) => service::record_service_op(
+                    Some(&central),
+                    service::ServiceOp::DaemonReload,
+                    target_name,
+                    "",
+                    &operation_id,
+                    actor,
+                    install_mode,
+                    None,
+                ),
+                Err(err) => {
+                    let msg = format!("daemon-reload after unit removal failed: {err}");
+                    service::record_service_op(
+                        Some(&central),
+                        service::ServiceOp::DaemonReload,
+                        target_name,
+                        "",
+                        &operation_id,
+                        actor,
+                        install_mode,
+                        Some(&msg),
+                    );
+                    combined.push(msg);
+                }
+            }
+        }
+    }
+
     let warnings = combined;
 
     // Best-effort cleanup of backups on the success path.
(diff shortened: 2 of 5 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -2681,12 +2681,23 @@ fn execute_raw(
...
+    // operator can fix the unit out of band. Reuse the env + log opened for
+    // the capability step.
+    //
+    // A contract's services are single-scope in practice (a component is
+    // either a system daemon or a per-user service). Pick the matching
+    // backend: an all-user-scope set drives `systemctl --user` (and only in a
+    // user-mode install); otherwise the system backend. A request the chosen
+    // backend does not handle (a hypothetical mixed-scope contract) is skipped
+    // by `apply_services` via `handles_scope`, so this never mis-drives.
+    let service_manager: Box<dyn ServiceManager> =
```
kongche-jbw, line 2695: [P1] This scope-aware manager selection should also be reused by the raw update path. `update.rs` still calls `service_for_install_mode(ctx.install_mode, ...)`, which is intentionally unsupported in user mode, so a user-scope service can be enabled/started on install but will not be restarted after a raw update. That leaves the running user service on the old process after files are replaced.

src/anolisa/crates/anolisa-core/src/lifecycle.rs
```diff
@@ -1381,6 +1381,47 @@ fn execute_uninstall_or_purge(
...
+    // Step 8.5 — daemon-reload AFTER the owned unit files are gone, so the
+    // manager drops the now-deleted units from its database (the
+    // uninstall-side mirror of the install reload — without it systemd keeps
+    // a stale unit until the next manual reload). Best-effort: a failed
+    // reload only warns. Runs only when the component declared service units
+    // (whose files were just removed) on a host that drives systemd; the
+    // manager's scope follows install mode (`--user` in user mode).
+    if live_plan.components.iter().any(|c| !c.services.is_empty()) {
+        let env = EnvService::detect();
+        let manager = service::for_install_mode(install_mode, &env);
```
kongche-jbw: [P1] This does not match the comment above for user-mode uninstalls. `for_install_mode("user", ...)` returns the unsupported system-scope factory, so deleting `~/.config/systemd/user/*.service` will not be followed by `systemctl --user daemon-reload`. Even if stop/disable for user-scope units stays out of scope until `ServiceRef` records scope, the post-removal reload can still select the user-scope factory based on `install_mode == "user"`.

src/anolisa/crates/anolisa-core/src/service.rs
```diff
@@ -272,18 +322,41 @@ impl ServiceManager for NotSupportedServiceManager {
...
-    /// Build a manager that invokes `systemctl` from `PATH`.
+    /// Build a **system**-scope manager that invokes `systemctl` from
+    /// `PATH`.
     pub fn new() -> Self {
+        Self::with_scope(ServiceScope::System)
+    }
+
+    /// Build a manager bound to `scope`. A [`ServiceScope::User`] manager
+    /// prefixes every `systemctl` call with `--user`.
+    pub fn with_scope(scope: ServiceScope) -> Self {
```
kongche-jbw, line 341: [P2] A user-scoped `SystemdServiceManager` executes `systemctl --user`, but `manager()` / `ServiceOutcome.manager` still report `"systemd"`, while install state records `"systemd-user"` for user mode. This is not blocking behavior, but it will make central-log and state/JSON diagnostics diverge. Consider returning a stable `"systemd-user"` label for the user-scoped backend, or explicitly model scope separately from backend family.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
ikunkun-sys: Fixed in ebe67343. `update.rs` now selects the scope-matched backend exactly like install — `user_service_for_install_mode` for an all-user-scope contract, the system factory otherwise — so a user-scope service is restarted via `systemctl --user` after a raw update instead of being left running on the replaced binary.

src/anolisa/crates/anolisa-core/src/lifecycle.rs
ikunkun-sys: Fixed in ebe67343. The post-removal `daemon-reload` now selects the user-scope factory when `install_mode == "user"`, so the removed `~/.config/systemd/user` units get a `systemctl --user daemon-reload`. As you noted, per-unit user-scope stop/disable in a *system*-mode uninstall still depends on `ServiceRef` recording scope — that's the follow-up PR, which threads scope through state and routes both stop/disable and reload per unit.

src/anolisa/crates/anolisa-core/src/service.rs
ikunkun-sys: Fixed in ebe67343. A user-scoped `SystemdServiceManager` now reports `systemd-user` from `manager()`, so central-log records and `ServiceOutcome.manager` match the `systemd-user` label install state already writes for user mode.

kongche-jbw:
Follow-up on the PR description: the latest code now selects `user_service_for_install_mode` for user-mode uninstall reload, but the Design Note / Ship Note still say the uninstall reload reuses `for_install_mode(install_mode)` and describe it as system-scope only. Please sync the description so reviewers don't read it as intentionally skipping user-mode daemon-reload.

src/anolisa/crates/anolisa-core/src/service.rs
```diff
@@ -365,13 +438,23 @@ impl Default for SystemdServiceManager {
 
 impl ServiceManager for SystemdServiceManager {
     fn manager(&self) -> &str {
-        "systemd"
+        // Report the scope in the label so central-log and probe outcomes
```
kongche-jbw, line 441: [P2] This fixes the trait label, but the `ServiceOutcome.manager` constructors in `run_op()` and `daemon_reload()` still hardcode `"systemd"`. User-scope enable/start/restart/reload will therefore still surface `systemd` in JSON/log diagnostics. Please use `self.manager().to_string()` in those outcome constructors as well.

ikunkun-sys: Good catch — fixed in 14e2a4e7. `run_op()` (both the `Probe` early-return and the post-op outcome) and the `SystemdServiceManager` `daemon_reload()` override now build `ServiceOutcome.manager` from `self.manager().to_string()` instead of the `"systemd"` literal, so user-scope enable/start/restart/reload surface `systemd-user` in JSON/log diagnostics as well. (The trait-default `daemon_reload` already used `self.manager()`.)

kongche-jbw (approved):
lgtm

PR #1101, opened 2026-06-24
fix(cosh): skip SLS log write when file does not exist or is not writable

samchu-zsl:
## Summary

Add a pre-flight writability check to `appendSlsLog` so that SLS session telemetry is only written when the log file (`/var/log/anolisa/sls/ops/cosh.jsonl`) **already exists and is writable**.

## Changes

- `sls-logger.ts`: Added `isWritable()` guard that checks `fs.accessSync(path, W_OK)` on first call and caches the result. If the file does not exist, the directory is missing, or permissions are insufficient, all subsequent `appendSlsLog` calls are no-ops.

## Behavior

| Condition | Result |
|---|---|
| File exists and writable | Append log normally |
| File does not exist | No-op (no file creation) |
| Directory does not exist | No-op |
| Permission denied | No-op |

No exceptions are thrown in any case.

### DIFF
--- a/src/copilot-shell/packages/core/src/telemetry/sls-logger.ts
+++ b/src/copilot-shell/packages/core/src/telemetry/sls-logger.ts
@@ -14,13 +14,17 @@ const SLS_LOG_PATH = '/var/log/anolisa/sls/ops/cosh.jsonl';
 
 /**
  * Append a single JSON record as one line to the SLS log file.
- * Uses open→write→close pattern to support logrotate rename.
- * Silently fails if the file cannot be written (e.g., directory does not exist).
+ * Uses O_WRONLY | O_APPEND (without O_CREAT) so the call naturally fails
+ * when the file does not exist — this writer never creates the SLS file.
+ * Open→write→close on each call to support logrotate rename-by-path.
  */
 export function appendSlsLog(record: Record<string, unknown>): void {
   try {
     const line = JSON.stringify(record) + '\n';
-    const fd = fs.openSync(SLS_LOG_PATH, 'a');
+    const fd = fs.openSync(
+      SLS_LOG_PATH,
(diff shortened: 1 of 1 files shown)


kongche-jbw:
[P2] `writable=true` caching can still recreate the SLS file

The change intends to no-op unless `/var/log/anolisa/sls/ops/cosh.jsonl` already exists and is writable, which matches the Agentic OS telemetry contract. However, once `isWritable()` succeeds, `writable` is cached as `true` forever, and later writes call `fs.openSync(SLS_LOG_PATH, 'a')`. Node's `'a'` mode creates the file if it is missing.

So if the file existed at the first check but is later removed/recreated by registration, logrotate, or an operator, cosh can create `cosh.jsonl` itself on the next session summary. That breaks the PR's core guarantee that this writer never creates the SLS component file.

A simpler fix is to avoid the cached `accessSync` check and open the path on each write without `O_CREAT`, for example:

```ts
const fd = fs.openSync(
  SLS_LOG_PATH,
  fs.constants.O_APPEND | fs.constants.O_WRONLY,
);
```

If the file is missing or not writable, this naturally fails and the existing best-effort `catch` keeps the main process unaffected. It also keeps the current open/write/close behavior for logrotate by path.

samchu-zsl:
Good catch. Fixed in `2855d965` — removed the cached `isWritable()` guard entirely and switched to opening the file with explicit flags:

```ts
fs.openSync(SLS_LOG_PATH, fs.constants.O_WRONLY | fs.constants.O_APPEND)
```

Without `O_CREAT`, `openSync` naturally throws `ENOENT` when the file is missing, which the existing `catch` handles silently. This guarantees the writer never creates `cosh.jsonl` regardless of timing — no caching, no TOCTOU window.

kongche-jbw (approved):
lgtm thx

PR #999, opened 2026-06-17
feat(anolisa): extend adapter manifest contract

kongche-jbw:
### Ship Note

This PR adds the first-stage adapter manifest contract for ANOLISA components.

What changed:

- Extended `component.toml` adapter schema with display metadata,
  adapter type, trust level, bundle metadata, and compatibility metadata.
- Kept adapter manifests as publishing contracts only; no executable
  commands, scripts, argv, or generic DSL are introduced.
- Updated the runtime component template to use the standardized
  artifact-side adapter layout: `adapters/<framework>`.
- Preserved current install behavior: `anolisa install` only lays adapter
  resources under `{datadir}/adapters/{component}/{framework}/`.

Known limitations:

- This is a foundation step for agent integrations, not the full framework
  implementation.
- No new framework drivers are implemented in this PR.
- RPM package manifest extraction is not implemented here.
- No `adapter install` alias is added; current CLI remains
  `adapter enable/disable/status/scan`.

## Related Issue

refs #998

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional change)
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `sec-core` (agent-sec-core)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] Multiple / Project-wide

## Checklist

- [ ] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [x] My code follows the project's code style
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly
- [ ] For `cosh`: Lint passes, type check passes, and tests pass
- [ ] For `sec-core` (Rust): `cargo clippy -- -D warnings` and `cargo fmt --check` pass
- [ ] For `sec-core` (Python): Ruff format and pytest pass
- [ ] For `skill`: Skill directory structure is valid and shell scripts pass syntax check
- [ ] For `sight`: `cargo clippy -- -D warnings` and `cargo fmt --check` pass
[...]

PR #1102, opened 2026-06-24
feat(tokenless): expand RPM component contract (bundle.entry + hermes)

shiloong:
Extends the tokenless RPM component contract (from #999 / 260471f7) so it is fully self-describing and covers both framework adapters tokenless ships through the anolisa adapter-contract flow.

## Changes

1. **Explicit `bundle.entry` for openclaw.** Declare `[adapters.bundle] entry = "openclaw.plugin.json"` instead of relying on the OpenClaw driver's hard-coded default (`openclaw.rs` `read_bundle`). The contract now names the platform entry manifest itself and survives driver default changes. Value matches the file installed at the dest root (id = `tokenless`).

2. **Add hermes adapter.** Previously the contract declared only openclaw, so `enable`/`scan` via `resolve_component_contract` did not cover hermes even though tokenless ships a hermes plugin bundle. New entry mirrors the openclaw shape:
   ```toml
   [[adapters]]
   framework    = "hermes"
   adapter_type = "plugin"
   plugin_id    = "tokenless"
   source       = "adapters/hermes"
   dest         = "{datadir}/adapters/{component}/hermes/"

   [adapters.bundle]
   entry = "plugin.yaml"
   ```
   - `dest` → `/usr/share/anolisa/adapters/tokenless/hermes/`, which the spec already populates (`__init__.py`, `plugin.yaml`, `scripts/*`) and declares in `%files`, so `resolve_resource_root` finds a real, non-empty bundle.
   - `plugin_id = "tokenless"` matches the hermes identity used by the spec's `%postun` cleanup (`~/.hermes/plugins/tokenless`) and `plugin.yaml`'s `name`; declaring it lets the driver skip manifest parsing.
   - `bundle.entry = "plugin.yaml"` names the Hermes-native entry manifest (the driver's default fallback chain looks for `hermes.plugin.json` / `hermes.manifest.yaml`, neither of which tokenless ships).

Both `.in` template and generated artifact updated, mirroring the `manifest.json` template pattern in the Makefile.

## Verification

Built the tokenless RPM via `scripts/rpm-build.sh tokenless` and inspected the installed `component.toml`:

- `PASS: hermes adapter declared`
- `PASS: hermes bundle.entry` (`entry = "plugin.yaml"`)
- `PASS: openclaw bundle.entry retained` (`entry = "openclaw.plugin.json"`)
- `PASS: hermes plugin.yaml packaged` at dest root
- `PASS: hermes __init__.py packaged` at dest root

TOML parse-checked: 2 `[[adapters]]` elements, each with its own `[adapters.bundle].entry` correctly attached.

## Notes

- Contract still covers only `openclaw` + `hermes` via the new model; the remaining frameworks tokenless ships (cosh/qoder/claude-code/codex/qwencode) continue to use the legacy `manifest.json`/bundle path. Expanding those is intentionally out of scope here.

### DIFF
--- a/src/tokenless/.anolisa/component.toml
+++ b/src/tokenless/.anolisa/component.toml
@@ -14,3 +14,25 @@ adapter_type = "plugin"
 plugin_id = "tokenless"
 source = "adapters/openclaw"
 dest = "{datadir}/adapters/{component}/openclaw/"
+
+[adapters.bundle]
+# Entry-point manifest inside the bundle directory (the file the OpenClaw
+# driver reads to identify the plugin). Declared explicitly rather than
+# relying on the driver's "openclaw.plugin.json" default, so the contract is
+# self-describing and survives driver default changes.
+entry = "openclaw.plugin.json"
+
+[[adapters]]
+framework = "hermes"
+adapter_type = "plugin"
+plugin_id = "tokenless"
+source = "adapters/hermes"
+dest = "{datadir}/adapters/{component}/hermes/"
+
+[adapters.bundle]
+# Hermes-native plugin manifest at the bundle root. The driver scans this
+# file for the plugin id when plugin_id is not declared; we declare
+# plugin_id above so the driver need not parse it, but keep entry explicit
+# so the contract names the platform entry manifest (matches the file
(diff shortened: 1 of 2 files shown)


Forrest-ly (approved):
LGTM

PR #1105, opened 2026-06-24
fix(anolisa): scope adapter datadir to contract root

### LINKED ISSUE #1104: [anolisa] fix(anolisa): scope adapter datadir to component contract root
## Problem

`anolisa adapter enable` currently derives the effective `{datadir}` scope from the adapter resource root resolved from `dest`.

This breaks RPM-provided component contracts when the adapter bundle lives outside the ANOLISA datadir, for example:

```toml
[[adapters]]
framework = "openclaw"
plugin_id = "agent-sec"
dest = "/opt/agent-sec/openclaw-plugin/"

[adapters.openclaw]
skills = [
  { name = "code-scanner", source = "{datadir}/skills/code-scanner/" },
]
```

If the contract is installed at:

```text
/usr/share/anolisa/components/sec-core/component.toml
```

then `{datadir}` in `source` should resolve to:

```text
/usr/share/anolisa
```

However, with the current implementation it may resolve to the user/system default candidate first, for example:

```text
/usr/local/share/anolisa
```

That makes `adapter enable` fail while copying skills:

```text
error[EXECUTION_FAILED]: io error while accessing /usr/local/share/anolisa/skills/code-scanner/: source directory does not exist
```

## Expected Behavior

The component contract location should determine the effective datadir scope used for `{datadir}` expansion in related adapter fields.

For a manifest loaded from:

```text
/usr/share/anolisa/components/sec-core/component.toml
```

and an adapter with:

```toml
dest = "/opt/agent-sec/openclaw-plugin/"
source = "{datadir}/skills/code-scanner/"
```

ANOLISA should resolve:

```text
resource_root = /opt/agent-sec/openclaw-plugin/
skill source  = /usr/share/anolisa/skills/code-scanner/
```

`dest` should still support absolute paths, but an absolute `dest` should not cause `{datadir}` to be inferred from an unrelated default datadir candidate.

## Suggested Fix

When `load_visible_component_manifest()` finds a manifest under a visible root, keep the matched visible root / contract datadir scope and use it as the effective datadir for `{datadir}` expansion.

In other words:

- manifest loaded from `/usr/share/anolisa/components/...` -> `{datadir}` should use `/usr/share/anolisa`
- manifest loaded from `/usr/local/share/anolisa/components/...` -> `{datadir}` should use `/usr/local/share/anolisa`
- `dest = "/opt/..."` should remain valid as the adapter resource root, but should not override the contract datadir scope

## Acceptance Criteria

- A component contract installed under `/usr/share/anolisa/components/<component>/component.toml` can declare an adapter `dest = "/opt/..."` and `skills.source = "{datadir}/skills/..."`.
- `anolisa adapter enable <component> openclaw --dry-run` shows skill sources under `/usr/share/anolisa/skills/...`.
- Add a regression test for an absolute adapter `dest` with `{datadir}` skill sources, verifying that the skill source uses the manifest's matched datadir root instead of `/usr/local/share/anolisa`.

kongche-jbw:
## Description

Fixes adapter `{datadir}` expansion when a component contract declares an absolute adapter `dest`.

Previously, `adapter enable` could derive the effective datadir from the first scoped datadir candidate when `dest` did not reference `{datadir}`. For RPM-provided contracts under `/usr/share/anolisa/components/<component>/component.toml`, this could incorrectly expand skill sources to `/usr/local/share/anolisa/skills/...`.

This change keeps:

- `resource_root` from the absolute `dest`, for example `/opt/agent-sec/openclaw-plugin/`
- `{datadir}` expansion scoped to the datadir root that owns the component contract, for example `/usr/share/anolisa`

### Ship Note

What changed:
- Absolute adapter `dest` no longer changes the datadir scope used by `{datadir}` skill source expansion.
- Added a regression test for an RPM-style contract with `dest = "/opt/..."` and `source = "{datadir}/skills/..."`.

Known limitations:
- This does not change the component manifest schema.
- This does not change framework driver behavior or sec-core packaging layout.

## Related Issue

closes #1104

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

## Testing

```bash
cd src/anolisa
cargo fmt --all -- --check
cargo test -p anolisa-core --locked adapter --quiet
cargo test -p anolisa-cli --locked adapter --quiet
```

## Additional Notes

This unblocks RPM contracts that keep adapter bundles outside the ANOLISA datadir while still declaring skill sources under `{datadir}/skills/...`.

### DIFF
--- a/src/anolisa/crates/anolisa-core/src/adapter/contract.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/contract.rs
@@ -18,6 +18,15 @@ use anolisa_platform::fs_layout::FsLayout;
 
 use crate::manifest::{ComponentManifest, ManifestError};
 
+/// A resolved component contract plus the concrete file that supplied it.
+#[derive(Debug, Clone)]
+pub struct ResolvedComponentContract {
+    /// Parsed component manifest.
+    pub manifest: ComponentManifest,
+    /// Candidate path that won contract resolution.
+    pub path: PathBuf,
+}
+
 /// Errors from component contract resolution.
 #[derive(Debug, thiserror::Error)]
 pub enum ContractError {
@@ -84,8 +93,20 @@ pub fn resolve_component_contract(
     state_roots: &[PathBuf],
     datadir_roots: &[PathBuf],
 ) -> Result<ComponentManifest, ContractError> {
+    resolve_component_contract_with_source(component, state_roots, datadir_roots)
+        .map(|resolved| resolved.manifest)
+}
+
+/// Resolve the component contract and return the file path that supplied
+/// the manifest. Use this when callers need to keep layout placeholder
+/// expansion scoped to the actual contract source.
+pub fn resolve_component_contract_with_source(
+    component: &str,
+    state_roots: &[PathBuf],
+    datadir_roots: &[PathBuf],
+) -> Result<ResolvedComponentContract, ContractError> {
     let candidates = candidate_paths(component, state_roots, datadir_roots);
-    resolve_from_candidates(component, &candidates)
+    resolve_from_candidates_with_source(component, &candidates)
 }
 
 /// Try each candidate path in order and return the first valid manifest.
@@ -97,11 +118,25 @@ pub fn resolve_from_candidates(
     component: &str,
     candidates: &[PathBuf],
 ) -> Result<ComponentManifest, ContractError> {
+    resolve_from_candidates_with_source(component, candidates).map(|resolved| resolved.manifest)
+}
+
+/// Try each candidate path in order and return the first valid manifest
+/// with its source path.
+pub fn resolve_from_candidates_with_source(
+    component: &str,
+    candidates: &[PathBuf],
+) -> Result<ResolvedComponentContract, ContractError> {
     let mut searched = Vec::new();
 
     for path in candidates {
         match try_load_contract(path) {
-            TryLoad::Loaded(manifest) => return Ok(*manifest),
+            TryLoad::Loaded(manifest) => {
+                return Ok(ResolvedComponentContract {
+                    manifest: *manifest,
+                    path: path.clone(),
+                });
+            }
             TryLoad::NotFound => {
                 searched.push(path.clone());
             }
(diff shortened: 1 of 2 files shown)


ikunkun-sys:
I think this patch fixes the narrow RPM datadir-only case, but the current implementation still infers the contract datadir by re-scanning `scoped_datadir_roots` instead of using the actual contract source that was resolved.

The risk is that component contract resolution already supports state snapshots, and snapshots have higher priority than datadir contracts:

```text
{state_dir}/component-manifests/<component>/component.toml
{datadir}/components/<component>/component.toml
```

`install` / RPM adopt paths can also snapshot a datadir contract into state. In that normal snapshot-first path, `manifest_datadir_root()` may find a datadir contract that exists on disk but was not the contract actually loaded. If an earlier datadir root contains a stale or unrelated same-component contract, the absolute `dest` can still resolve correctly to `/opt/...`, but `effective_datadir` may be selected from that other datadir. That would make `{datadir}` skill sources expand to the wrong root.

I suggest carrying contract provenance as part of the contract resolve result instead of re-scanning datadirs inside `resolve_resource_root()`. For a datadir contract hit, the matched datadir root can become `effective_datadir`; for a state snapshot, install/adopt can record the original datadir root when writing the snapshot, and adapter enable can later read that provenance.

ikunkun-sys:
Thanks for the update. This is a clear improvement: direct datadir hits now carry the resolved source path, and the snapshot path now handles the stale-local-contract case by matching snapshot content against scoped datadir contracts.

One remaining limitation is that snapshot provenance is still inferred from the current datadir files by content matching. If identical same-component contracts exist under multiple datadirs, the first matching root still wins, even though it may not be the original root that was snapshotted.

Should we track explicit snapshot provenance as a follow-up? For example, install/adopt could persist the original datadir root when writing the snapshot, and adapter enable could use that recorded source instead of inferring it from current filesystem state. If the intended contract is that identical datadir contracts are interchangeable, it may be worth documenting that assumption explicitly.

ikunkun-sys (approved):

PR #1106, opened 2026-06-24
feat(anolisa): tear down user-scope services on uninstall

### LINKED ISSUE #1070: [anolisa] feat(install): declarative service/capability/hook operations with raw–RPM backend parity

ikunkun-sys:
## Description

Completes user-scope service lifecycle under #1070, on top of #1100 (user-scope activation + a scope-aware uninstall reload by install mode). This PR persists each unit's scope and routes uninstall teardown per scope: a **user-mode** uninstall stops/disables/reloads user-scope services through `systemctl --user`, while a **system-mode** uninstall keeps user-scope units place-only — symmetric with install, where root has no single target user to activate them for.

Three parts:
1. **`%u` instance resolution** — a template service with `instance = "%u"` (e.g. agent-memory's `anolisa-memory@.service`) now resolves `%u` to the caller's login name in a user-mode install → `anolisa-memory@<user>.service`. `%u` is a systemd specifier that systemd does *not* expand in a command-line instance name, so anolisa resolves it itself. A system-mode install is place-only and keeps the bare template for per-user `systemctl --user enable`.
2. **Scope persisted in state** — `ServiceRef` gains a `scope` field (`#[serde(default)]` → `system`, so pre-existing `installed.toml` loads unchanged). Install and update record the unit's scope, `anolisa restart` routes by it (an all-user-scope component restarts via `systemctl --user`), and the persisted `manager` label is derived from scope (`systemd` / `systemd-user`) so it never disagrees with `scope`.
3. **Per-scope uninstall teardown** — the uninstall executor partitions removed units by scope: system units stop/disable + `daemon-reload` via `systemctl`, user units via `systemctl --user`. The user factory is only supported in a user-mode uninstall, so a user-scope unit removed in a system-mode uninstall is a place-only skip (mirroring install).

## Design Note

`ServiceRef.scope` uses `#[serde(default)]` (→ `ServiceScope::System`), so no migration step is needed — older state files deserialize as system-scope, correct for everything installed before this change. The resolved unit name (`anolisa-memory@<user>.service` in user mode, the bare template in system mode) is what gets persisted, so uninstall reconstructs the same name from state.

`%u` resolution is keyed on install mode, not just the presence of the specifier: a user-mode install activates the unit as the caller, so it bakes in the login name; a system-mode install only places the template (root has no single target user), so it leaves `%u` un-resolved for per-user enable. The username is detected once (via `EnvService`) and only when a `%u` instance needs it.

`ServiceAction` carries the scope from `ServiceRef` into the lifecycle plan; `execute_plan` partitions removed units by scope and drives each through its matching factory (`for_install_mode` for system, `user_service_for_install_mode` for user), for both stop/disable and the post-removal `daemon-reload`. The user factory is unsupported outside a user-mode install, so system-mode user-scope teardown is a recorded skip rather than a mis-drive through the system manager. `deactivate_services` keeps its single-manager signature — the executor calls it once per non-empty scope.

The persisted `manager` label and the running `SystemdServiceManager::manager()` both come from a single `ServiceScope::manager_label()`, so state and live diagnostics can't diverge.

## Ship Note

- **Completes the user-scope teardown deferred in #1100**: this PR threads scope through state so a user-mode uninstall stops/disables/reloads user-scope services via `systemctl --user`, and a system-mode uninstall cleanly skips them (place-only) instead of mis-driving them through the system manager. System-scope units are unaffected.
- **No production behavior change today**: agent-memory still declares `enable = false` / `start = false`, so nothing auto-activates; this completes the plumbing + tests so a future `enable = true` user-scope service works end-to-end (place → activate → tear down).
- Per-unit teardown is a quiet skip where the relevant manager is `NotSupported` (container / non-Linux / a user-scope unit in a system-mode uninstall).

## Test

- Unit (install): `resolve_service_instance` — `%u` → caller when a name is available, `None` (place-only) when not; a literal instance passes through. `resolve_manifest_services` resolves `%u` in user mode and keeps the bare template in system mode.
- Unit (state/update): a `ServiceRef` written without a `scope` field deserializes as `system`; an explicit `scope = "user"` round-trips; `raw_update_service_refs` labels `manager` from each unit's scope (`systemd` / `systemd-user`).
- Unit (lifecycle): `plan_services` carries scope into `ServiceAction`; a user-scope service in a system-mode uninstall is place-only (recorded stop/disable skip, no `service:daemon-reload`) while owned files are still removed.
- Full gate: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, `cargo doc --workspace --no-deps`. One pre-existing platform test (`systemd::tests::unimplemented_unit_operations_return_errors_instead_of_panicking`) fails on the dev host only because `agentsight.service` is installed there; it fails identically on clean `main` and is unrelated to this change.

Closes #1070

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
@@ -2248,7 +2248,7 @@ fn resolve_manifest_contract(
         &resolution.component,
     )?);
 
-    let services = resolve_manifest_services(manifest, &resolution.component)?;
+    let services = resolve_manifest_services(manifest, &resolution.component, mode)?;
     let capabilities = resolve_manifest_capabilities(manifest, layout, &resolution.component)?;
 
     Ok((files, services, capabilities))
@@ -2265,7 +2265,26 @@ fn resolve_manifest_contract(
 fn resolve_manifest_services(
     manifest: &ComponentManifest,
     component: &str,
+    mode: &str,
 ) -> Result<Vec<ServiceRequest>, CliError> {
+    // The `%u` instance specifier resolves to the caller's login name, but
+    // only in a user-mode install where the unit is activated as that user.
+    // A system-mode install merely *places* a user-scope template for later
+    // per-user `systemctl --user enable`, so it leaves `%u` un-resolved
+    // (the bare template) rather than baking in root's name. Detect the user
+    // at most once, and only when a `%u` instance actually needs it.
+    let caller = if mode == "user"
+        && manifest
+            .install
+            .services
+            .iter()
+            .any(|s| s.instance.as_deref().is_some_and(|i| i.contains("%u")))
+    {
+        Some(anolisa_env::EnvService::detect().user)
+    } else {
+        None
+    };
+
     let mut requests = Vec::with_capacity(manifest.install.services.len());
     for spec in &manifest.install.services {
         if spec.unit.trim().is_empty() {
@@ -2279,7 +2298,12 @@ fn resolve_manifest_services(
         // Template unit (`name@.service`) + instance → `name@<instance>.service`.
         let unit = match &spec.instance {
             Some(instance) if spec.unit.contains("@.") => {
-                spec.unit.replacen("@.", &format!("@{instance}."), 1)
+                match resolve_service_instance(instance, caller.as_deref()) {
+                    Some(resolved) => spec.unit.replacen("@.", &format!("@{resolved}."), 1),
+                    // `%u` with no resolved user (system-mode place-only):
+                    // keep the bare template; per-user enable instantiates it.
+                    None => spec.unit.clone(),
+                }
             }
             _ => spec.unit.clone(),
         };
@@ -2291,6 +2315,21 @@ fn resolve_manifest_services(
         });
     }
     Ok(requests)
+}
+
+/// Resolve a systemd template instance, expanding the `%u` specifier to the
+/// caller's login name.
+///
+/// `%u` is a systemd specifier that systemd does *not* expand in the instance
+/// portion of a command-line unit name, so anolisa resolves it itself. Returns
+/// `None` when the instance uses `%u` but no caller name is available (a
+/// system-mode install that only places the template) — the caller then keeps
+/// the bare template. A literal instance is returned verbatim in every mode.
+fn resolve_service_instance(instance: &str, caller: Option<&str>) -> Option<String> {
+    if !instance.contains("%u") {
+        return Some(instance.to_string());
+    }
+    caller.map(|user| instance.replace("%u", user))
 }
 
 /// Render the manifest's `[install.files]` against the layout: expand
@@ -2767,11 +2806,6 @@ fn execute_raw(
         .collect();
     installed_paths.push(manifest_path.display().to_string());
 
-    let service_manager_label = match ctx.install_mode {
-        crate::context::InstallMode::System => "systemd",
-        crate::context::InstallMode::User => "systemd-user",
-    };
-
     // Migrate away legacy capability rows on this state write; surfaced
     // in the result warnings and audited in the central log below. A
     // state-save failure rolls the prune back with the rest of the write.
@@ -2816,10 +2850,14 @@ fn execute_raw(
             .iter()
             .map(|svc| ServiceRef {
                 name: svc.unit.clone(),
-                manager: service_manager_label.to_string(),
+                // Label follows the unit's scope, not install mode: a
+                // place-only user-scope unit in a system install is still
+                // `systemd-user`, keeping `manager` consistent with `scope`.
+                manager: svc.scope.manager_label().to_string(),
                 restartable: true,
                 // Reflect what the executor actually enabled this run.
                 enabled: service_run.enabled_units.contains(&svc.unit),
+                scope: svc.scope,
             })
             .collect(),
         health: Vec::new(),
@@ -4403,7 +4441,7 @@ sha256 = "{sha}"
     fn resolve_manifest_services_carries_spec_and_expands_instance() {
         let toml = service_manifest("anolisa-memory@.service", true, false, Some("alice"));
         let manifest = ComponentManifest::from_toml_str(&toml).expect("parse manifest");
-        let reqs = resolve_manifest_services(&manifest, "agentsight").expect("resolve");
+        let reqs = resolve_manifest_services(&manifest, "agentsight", "system").expect("resolve");
         assert_eq!(reqs.len(), 1);
         assert_eq!(reqs[0].unit, "anolisa-memory@alice.service");
         assert!(reqs[0].enable);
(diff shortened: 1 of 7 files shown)


src/anolisa/crates/anolisa-core/src/state.rs
```diff
@@ -231,6 +232,12 @@ pub struct ServiceRef {
     /// Desired enabled-on-boot state when a manager supports it.
     #[serde(default)]
     pub enabled: bool,
+    /// Manager scope: `system` units are driven by `systemctl`, `user`
```
kongche-jbw, line 235: [P1] Persisting `ServiceRef.scope` also needs `anolisa restart` to consume it. `restart.rs` still builds one manager with `service_for_install_mode(install_mode, ...)`; that factory is intentionally unsupported in user mode, so a user-mode install can enable/start a user service but `anolisa restart <component>` will still report `not_supported` instead of `systemctl --user restart`. Please route restart per scope or pick the user-scope factory when all restartable units are user-scoped.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -2820,6 +2859,7 @@ fn execute_raw(
                 restartable: true,
                 // Reflect what the executor actually enabled this run.
                 enabled: service_run.enabled_units.contains(&svc.unit),
+                scope: svc.scope,
```
kongche-jbw, line 2860: [P2] `scope` is persisted from the service request, but `manager` just above is still derived from `ctx.install_mode`. In a system-mode place-only install of a user-scope unit this can write `manager = "systemd"` with `scope = "user"`, which makes state/diagnostics inconsistent. Now that scope is the authoritative routing field, consider deriving the manager label from `svc.scope` (`systemd` vs `systemd-user`) in both install and update.

kongche-jbw:
One PR-description mismatch: the top-level Description says this routes user-scope teardown through `systemctl --user` “even in a system-mode uninstall”, but the code and test define system-mode user-scope uninstall as place-only skip (`uninstall_user_scope_service_in_system_mode_is_place_only`). Please update the description to say user-mode uninstall tears down user-scope services, while system-mode uninstall still skips user-scope teardown/place-only.

src/anolisa/crates/anolisa-core/src/state.rs
ikunkun-sys: Good catch — fixed in b64513d7. `anolisa restart` now picks the scope-matched factory: when every restartable unit is user-scope it restarts through `systemctl --user` (`user_service_for_install_mode`), otherwise the system factory — so a user-scope service is actually restarted instead of being reported `not_supported`. This mirrors install/update's single-scope selection.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
ikunkun-sys: Fixed in b64513d7. `ServiceRef.manager` is now derived from the unit's `scope` (`systemd` vs `systemd-user`) in both install and update, via a new `ServiceScope::manager_label()` that `SystemdServiceManager::manager()` also delegates to — so the persisted label is a single source of truth and can't disagree with `scope`. A place-only user-scope unit in a system-mode install is now recorded as `systemd-user`.

ikunkun-sys:
Good catch — description corrected. It now states user-mode uninstall tears down user-scope services via `systemctl --user`, while system-mode uninstall keeps them place-only (symmetric with install, where root has no single target user). The misleading "even in a system-mode uninstall" line and the matching Ship Note bullet are both fixed; the per-scope teardown bullet (#3) already described the place-only skip and is unchanged.

src/anolisa/crates/anolisa-cli/src/commands/tier1/restart.rs
```diff
@@ -91,7 +95,21 @@ pub fn handle(args: RestartArgs, ctx: &CliContext) -> Result<(), CliError> {
     }
 
     let env = EnvService::detect();
-    let manager = service_factory(install_mode, &env);
+    // Route by the units' scope (single-scope per component in practice):
+    // an all-user-scope set restarts through `systemctl --user`, otherwise
+    // the system manager. Mirrors install/update's scope selection so a
+    // user-scope service is actually restarted instead of reported
+    // `not_supported` by the system-only factory.
+    let all_user_scope = comp
```
kongche-jbw: [P2] This still picks a single manager for the whole restart operation. For a mixed-scope component, `all_user_scope` is false, so user-scope restartable units are passed to the system manager and become `systemctl restart ...` calls. Uninstall already partitions by scope; restart should either partition similarly or explicitly skip off-scope units instead of mis-driving them. If single-scope services are an intentional constraint, please document/enforce it near manifest validation or here.

ikunkun-sys: Agreed — switched from the single-manager heuristic to full per-scope partitioning, matching uninstall (`ac00faa0`). Restart now builds both managers once and routes **each unit** through the one for its own `scope`:

```rust
let manager: &dyn ServiceManager = match u.scope {
    ServiceScope::System => sys_manager.as_ref(),
    ServiceScope::User   => user_manager.as_ref(),
};
if !manager.supported() { /* per-unit not_supported skip */ continue; }
manager.restart_service(&u.unit)
```

So a mixed-scope component never mis-drives: a user unit is never handed to the system `systemctl`, and an off-scope unit whose manager is `NotSupported` here (a user unit in a system-mode restart, container, non-Linux) becomes a per-unit `not_supported` result rather than a wrong-namespace call. The top-level summary fields now describe the set actually present — `supported` is true if any unit's manager can drive it, and the `manager` label combines the distinct namespaces (`systemd`, `systemd-user`, or `systemd+systemd-user`), so a single label never hides that two namespaces were driven.

The mis-drive avoidance reduces to the per-unit `match u.scope` plus the factories already covered in core tests (`user_service_for_install_mode` returns `NotSupported` outside a user-mode install; `for_install_mode` likewise outside system mode). Restart's actual `systemctl`-driving paths stay un-unit-tested by design — there's no manager-injection seam and they shell out — so I left the test surface as-is rather than mock it; happy to add an injection seam + a mixed-scope routing test if you'd prefer it locked down here.

kongche-jbw (approved):

PR #1126, opened 2026-06-25
fix(anolisa): persist adapter contract provenance

### LINKED ISSUE #1119: [anolisa] Persist adapter contract snapshot provenance for datadir expansion
## Problem

PR #1105 fixed the narrow RPM datadir case by keeping `{datadir}` expansion scoped to the contract root. One limitation remains: snapshot provenance can still be inferred from current datadir files by content matching.

If identical same-component contracts exist under multiple datadirs, the first matching root can win even if it was not the original root used when the snapshot was written. That can make `{datadir}` expansion depend on the current filesystem layout instead of the actual contract source.

## Goal

Persist explicit adapter contract snapshot provenance so adapter enable/disable/status can use the original contract source instead of inferring it from current datadir files.

## Proposed behavior

- When install/adopt writes a component contract snapshot, persist provenance such as the original datadir root or source contract path.
- When resolving adapter resources from a snapshot, prefer recorded provenance over re-scanning scoped datadir roots.
- If provenance is missing because the snapshot was created by an older version, keep the current inference fallback and surface a warning/debug note where appropriate.
- Document whether identical datadir contracts are considered interchangeable.

## Acceptance criteria

- Snapshot-backed adapter resolution uses explicit recorded provenance when available.
- Identical contracts under multiple datadirs do not cause `{datadir}` expansion to pick the wrong root.
- Backward compatibility is preserved for old snapshots without provenance.
- Tests cover direct datadir contract, state snapshot with provenance, old snapshot fallback, and duplicate identical datadir contracts.

## Related

Follow-up from #1105.
Related to #1104.

kongche-jbw:
## Description

Persists provenance for component contract snapshots so adapter `{datadir}`
expansion can use the original contract source.

This is a follow-up to #1105. Snapshot-backed adapter resolution now prefers
the recorded datadir root before falling back to content matching for older
snapshots.

### Ship Note

What changed:
- `install` / `adopt` snapshot writes best-effort `provenance.toml`.
- Adapter resolution uses provenance for snapshot-backed contracts.
- Old snapshots without provenance keep the content-match fallback.

Known limitations:
- Only `source_kind = "datadir"` is supported for now.
- Provenance write failure is a warning and does not fail install/adopt.

## Related Issue

closes #1119

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

## Testing

```bash
cd src/anolisa
cargo fmt --all -- --check
cargo test -p anolisa-core --locked -- adapter
cargo test -p anolisa-cli --locked -- install
cargo test -p anolisa-cli --locked -- adapter
```

## Additional Notes

Follow-up to #1105. The fallback remains for snapshots created before
provenance existed.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/common.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/common.rs
@@ -59,12 +59,24 @@ pub fn installed_component_manifest_path(
     component: &str,
     command: &str,
 ) -> Result<PathBuf, CliError> {
+    Ok(
+        installed_component_manifest_dir(layout, component, command)?
+            .join(INSTALLED_COMPONENT_MANIFEST_FILE),
+    )
+}
+
+/// Directory for the component manifest saved as part of an installed
+/// component's local state.
+pub fn installed_component_manifest_dir(
+    layout: &FsLayout,
+    component: &str,
+    command: &str,
+) -> Result<PathBuf, CliError> {
     validate_component_path_segment(component, command)?;
     Ok(layout
         .state_dir
         .join(INSTALLED_COMPONENT_MANIFESTS_SUBDIR)
-        .join(component)
-        .join(INSTALLED_COMPONENT_MANIFEST_FILE))
+        .join(component))
 }
 
 fn validate_component_path_segment(component: &str, command: &str) -> Result<(), CliError> {
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/forget.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/forget.rs
@@ -5,9 +5,10 @@
 //! `missing` case from `anolisa status`), or whenever the operator wants ANOLISA
 //! to stop tracking a component, `forget` removes the `installed.toml` object
 //! and records the operation. It performs **no** package operation — no
-//! `dnf remove`, no `rpm -e` — and deletes no files. An observed/managed RPM
-//! stays installed in rpmdb; a raw component's owned files stay on disk (use
-//! `anolisa uninstall` to remove those).
+//! `dnf remove`, no `rpm -e` — and leaves package/component files on disk. It
+//! only removes ANOLISA-owned state metadata for the forgotten component. An
+//! observed/managed RPM stays installed in rpmdb; a raw component's owned files
+//! stay on disk (use `anolisa uninstall` to remove those).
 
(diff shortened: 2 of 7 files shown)


ikunkun-sys:
[P2] Please clean up snapshot/provenance state sidecars when the component is removed

One lifecycle gap remains and I think it should be handled in this PR rather than as a follow-up.

The RPM/delegated install/adopt path writes a state-side contract snapshot under `state_dir/component-manifests/<component>/component.toml`, and this PR adds `provenance.toml` next to it. These files are ANOLISA-owned state metadata, but they are not part of `InstalledObject.files` for rpm/adopt records, and uninstall/forget currently drop the component state record without deleting `component-manifests/<component>`. As a result, after removing a component, stale `component.toml` / `provenance.toml` can remain in state.

The stale files do not make the component visible by themselves, but they can affect a future same-name component if it is added back to state and the new datadir snapshot is missing or fails to write. Adapter resolution searches state snapshots before datadir contracts, so it may read the old manifest and, with this PR, the old recorded datadir root. Since this change introduces another state sidecar that participates in adapter resolution, the sidecar should share the component state lifecycle.

Please make uninstall, and likely forget as well since it also drops component state, remove the component's snapshot directory or otherwise explicitly invalidate both `component.toml` and `provenance.toml` when the component state record is dropped.

ikunkun-sys:
Thanks, I re-reviewed the updated PR head (`17a61f84`).

The lifecycle gap I raised is addressed now: RPM-backed uninstall removes the component manifest snapshot directory, `forget` does the same when dropping component state, and the dry-run cases keep the snapshot directory intact. The added tests cover both the real-run cleanup and dry-run non-mutation paths. CI and PR Lint are green on the current head.

One small non-blocking thing I noticed: the `forget` module/function comments still say it "deletes no files" / "no file deletion", but `forget` now intentionally removes ANOLISA's component-manifest state sidecar. That is just stale wording, not a behavioral blocker.

No blocking issues from my side after this update.

ikunkun-sys (approved):

PR #1134, opened 2026-06-25
fix(skillfs): gate trusted skill metadata view

### LINKED ISSUE #1115: [skillfs] SkillFS: restrict .skill-meta access to trusted processes and expose live metadata for trusted access
## Background

Skill Ledger uses `<skill>/.skill-meta/` as the trusted metadata root for scan manifests, signed versions, snapshots, activation decisions, and related security state.

SkillFS exposes the agent-facing skill view according to activation state. A skill may be served from live source, from a signed snapshot, or be hidden/blocked from normal agent use. Regardless of that public view, Skill Ledger still needs to read and maintain the live metadata under `<source>/<skill>/.skill-meta/`.

SkillFS currently has a `trusted-writer` mechanism, but it only covers write-side access to `.skill-meta/**`. It does not cover read-side visibility, and it does not guarantee that trusted access through `<mount>/<skill>/.skill-meta/**` resolves to the live metadata root when the public skill view is snapshot-backed, hidden, or blocked.

## Requirement 1: `.skill-meta` should be readable/writable only by trusted processes

Please extend the current trusted process model so that `.skill-meta` is protected for both reads and writes.

Expected behavior:

- Trusted processes, such as Skill Ledger / agent-sec daemon, can list, stat, read, create, update, and delete `<skill>/.skill-meta/**`.
- Non-trusted processes cannot see `.skill-meta` in directory listings.
- Non-trusted processes cannot read `.skill-meta/**` through direct path access, `stat`, `open/read`, `access`, or similar operations.
- Non-trusted processes also remain unable to mutate `.skill-meta/**`.
- Normal agent processes should not be able to inspect ledger metadata via `ls -a`, direct `cat <skill>/.skill-meta/...`, direct `stat`, or any equivalent access path.

## Requirement 2: trusted `.skill-meta` access should always reach live metadata

When SkillFS exposes a snapshot as the public skill view, normal skill files should continue to come from that activated snapshot. However, trusted access to `<mount>/<skill>/.skill-meta/**` should resolve to the live metadata root under `<source>/<skill>/.skill-meta/**`.

This should hold for all activation states:

- If the public view is source/current, trusted `.skill-meta` access reaches live metadata.
- If the public view is snapshot/fallback, trusted `.skill-meta` access still reaches live metadata rather than the snapshot tree.
- If the skill is hidden or blocked from normal agent use, trusted `.skill-meta` access is still available.

In short:

- For normal agents, `<mount>/<skill>/...` remains the activation-controlled runtime skill view.
- For trusted Skill Ledger processes, `<mount>/<skill>/.skill-meta/...` always refers to the live metadata root.
- This should not expose `.skill-meta` to normal agents or otherwise change normal agent visibility of skill content.

## Acceptance Criteria

- A non-trusted process cannot list, stat, read, or write `<mount>/<skill>/.skill-meta/**`.
- A trusted process can list, stat, read, and write `<mount>/<skill>/.skill-meta/**`.
- When the public skill view is snapshot-backed, normal files still come from the snapshot, while trusted `.skill-meta` access reaches live source metadata.
- When a skill is hidden or blocked for normal agents, the trusted process can still access `<mount>/<skill>/.skill-meta/**`.
- Normal agent-facing skill behavior remains unchanged except that `.skill-meta` becomes fully hidden and inaccessible to non-trusted processes.
- The feature enables Skill Ledger / agent-sec daemon to inspect and maintain `.skill-meta` reliably in SkillFS in-place mount deployments.

kongche-jbw:
## Summary

This PR tightens `.skill-meta` visibility and access through the SkillFS
mount path.

- Hide `.skill-meta` from untrusted directory listings and exact-path probes.
- Route trusted `.skill-meta/**` access to the live source directory.
- Preserve hidden/fallback public skill views for ordinary skill files.
- Keep mutating `.skill-meta` opens on the existing policy/audit path.

## What changed

Trusted metadata access now uses the existing trusted-writer identity gate.
When the caller is trusted, exact-path `.skill-meta/**` lookup/getattr/access/open
and directory listing use the live source tree, even if the public skill view is
hidden or fallback-mapped to a snapshot.

Untrusted callers receive `ENOENT` for `.skill-meta/**` exact-path access and do
not see `.skill-meta` in directory listings. Staging and pending install listings
also filter `.skill-meta`.

## Validation

- `cargo +1.86.0 fmt --all --check`
- `cargo +1.86.0 clippy -p skillfs-fuse --tests -- -D warnings`
- `cargo +1.86.0 test -p skillfs-fuse --test trusted_skill_meta_view_tests`
- `cargo +1.86.0 test -p skillfs-fuse --test install_staging_tests`
- `cargo +1.86.0 test -p skillfs-fuse --test posix_xattr_tests`

## Related Issue

Closes #1115

### DIFF
--- a/src/skillfs/crates/skillfs-fuse/src/fs/callbacks/dir.rs
+++ b/src/skillfs/crates/skillfs-fuse/src/fs/callbacks/dir.rs
@@ -9,12 +9,15 @@ use super::super::SkillFs;
 use super::super::read_resolution::ReadResolution;
 use crate::attr::dir_entry_file_type;
 use crate::path::{PathType, is_skill_discover_path, parse_path};
-use crate::security::{inbox::INBOX_DIR_NAME, lifecycle::is_reserved_lifecycle_name};
+use crate::security::{
+    SKILL_META_DIR, inbox::INBOX_DIR_NAME, lifecycle::is_reserved_lifecycle_name,
+};
 use crate::sys::errno;
 
 impl SkillFs {
     pub(in crate::fs) fn readdir_dynamic(
         &mut self,
+        req: &Request,
         ino: u64,
         offset: i64,
         mut reply: ReplyDirectory,
@@ -130,6 +133,7 @@ impl SkillFs {
                     if self.is_staging_skill_root(&skill_name)
                         || self.is_pending_install(&skill_name)
                     {
+                        let show_meta = self.should_show_skill_meta_in_listing(&skill_name, req);
                         let parent_ino = self.skills_dir_ino();
                         let mut entries: Vec<(u64, FileType, String)> = vec![
                             (ino, FileType::Directory, ".".to_string()),
@@ -139,6 +143,9 @@ impl SkillFs {
                         if let Ok(dir_iter) = std::fs::read_dir(&phys_dir) {
                             for entry in dir_iter.flatten() {
                                 let name = entry.file_name().to_string_lossy().to_string();
+                                if name == SKILL_META_DIR && !show_meta {
+                                    continue;
+                                }
                                 let kind = dir_entry_file_type(&entry);
                                 let entry_path = format!("{}/{}", path, name);
                                 let entry_ino = self.inodes.readdir_ino(&entry_path);
@@ -167,6 +174,8 @@ impl SkillFs {
                         entries.push((md_ino, FileType::RegularFile, "SKILL.md".to_string()));
 
                         if skill_name != "skill-discover" {
+                            let show_meta =
+                                self.should_show_skill_meta_in_listing(&skill_name, req);
                             // D1.1: fallback skills enumerate from the
                             // snapshot tree; current and no-resolver skills
                             // enumerate from the live source.
@@ -177,6 +186,9 @@ impl SkillFs {
                                 for entry in dir_iter.flatten() {
                                     let name = entry.file_name().to_string_lossy().to_string();
                                     if name == "SKILL.md" {
+                                        continue;
+                                    }
+                                    if name == SKILL_META_DIR && !show_meta {
                                         continue;
                                     }
                                     let kind = dir_entry_file_type(&entry);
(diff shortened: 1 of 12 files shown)


src/skillfs/crates/skillfs-fuse/src/fs/callbacks/read.rs
```diff
@@ -252,11 +252,44 @@ impl SkillFs {
...
+        // denied; trusted read-only opens route to the live source
+        // directory. Mutating opens fall through to enforce_skill_meta
+        // so the existing audit/policy path is preserved.
+        let is_mutating_open = is_write || (flags & libc::O_TRUNC) != 0;
+        match self.is_trusted_skill_meta_access(&path_type, req) {
+            Some(false) => {
+                reply.error(libc::ENOENT);
+                return;
+            }
+            Some(true) if !is_mutating_open => {
```
yummypeng, line 265: trusted read-only open 分支只处理了 Passthrough（用 skill_physical_dir 路由到 live source），如果通过 inbox 路径访问 .skill-meta（InboxPassthrough），该分支的 if let 不匹配就会 fall through，走了后面的 mutation 检查路径而不是预期的 live
  source 路由。不过 inbox 路径本来就直接映射到物理 source，所以实际影响可能不大，但代码逻辑上不完整。

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/dir.rs
```diff
@@ -139,6 +142,9 @@ impl SkillFs {
                         if let Ok(dir_iter) = std::fs::read_dir(&phys_dir) {
                             for entry in dir_iter.flatten() {
                                 let name = entry.file_name().to_string_lossy().to_string();
+                                if name == SKILL_META_DIR {
```
yummypeng: 在 staging 和 pending install 的 SkillDir 列表中，.skill-meta 被无条件过滤，即使是 trusted 进程也看不到。这与 Passthrough 子目录中 trusted 进程可以 readdir .skill-meta/** 的行为不一致。 如果是有意为之那么在注释里说明一下。

kongche-jbw: 已改成行为一致的实现。
现在 staging / pending / normal SkillDir 的 parent listing 都统一走
should_show_skill_meta_in_listing(skill_name, req)：
- trusted caller：可以在 SkillDir listing 中看到 .skill-meta
- untrusted caller：仍然过滤 .skill-meta
同时 exact-path 的 .skill-meta/** traversal 仍然走 live source，hidden/fallback
普通文件视图没有被放宽。

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/read.rs
kongche-jbw: thx for review。
trusted read-only open fast path 已补齐 InboxPassthrough 分支。现在 Passthrough 走 skill_physical_dir，InboxPassthrough 走 inbox_skill_dir；mutating open 仍然 fall through 到 enforce_skill_meta，保留原有 policy/audit 路径。

yummypeng (approved):
LGTM

PR #1137, opened 2026-06-25
feat(sight): add container_id to AgentsightLLMData

jfeng18:
## Summary

Extract container ID from `/proc/{pid}/cgroup` at LLM callback time and expose it via the FFI `AgentsightLLMData` struct. Supports Docker (cgroup v1/v2), containerd, and Kubernetes cgroup layouts. Returns NULL for non-container processes.

Partial implementation of #148 (Phase 0: cgroup-based container_id extraction only, no Docker API dependency).

## Changes

| File | What changed |
|------|-------------|
| `container.rs` | NEW: `extract_container_id(pid)` + `parse_container_id_from_cgroup(content)` pure function + 8 tests |
| `ffi.rs` | `AgentsightLLMData` gains `container_id: *const c_char`, `build_llm_data()` calls extraction |
| `lib.rs` | `pub mod container;` |

## Design decisions

- **No new dependencies**: pure `str` parsing, no regex, no `bollard`
- **FFI-only for now**: `LLMCall` struct unchanged, container_id resolved at FFI boundary
- **Graceful degradation**: non-container processes get NULL, /proc read failure returns NULL with debug log

## Test plan

- [x] 880 tests pass (10 new container parsing tests)
- [x] Covers: Docker v1, Docker v2, containerd, k8s, non-container, empty, multiline, short-hex-rejection
- [ ] ECS E2E: non-container returns NULL (ECS is bare metal, not in container)

Ref #148

### DIFF
--- a/src/agentsight/AGENTS.md
+++ b/src/agentsight/AGENTS.md
@@ -89,6 +89,7 @@ eBPF Probes → Event → Parser → ParsedMessage → Aggregator → Aggregated
 | **Tokenizer** | `src/tokenizer/` | LLM Token 计数 | `LlmTokenizer`, `MultiModelTokenizer` |
 | **ATIF** | `src/atif/` | 轨迹格式导出 | `AtifDocument`, `convert_trace_to_atif` |
 | **Server** | `src/server/` | HTTP API + 嵌入式前端 | `AppState`, `run_server` |
+| **Container** | `src/container.rs` | 容器 ID 提取（/proc/pid/cgroup） | `extract_container_id`, `parse_container_id_from_cgroup` |
 | **Config** | `src/config.rs` | 统一配置 | `AgentsightConfig` |
 | **Unified** | `src/unified.rs` | 主编排器 | `AgentSight` |
 
--- a/src/agentsight/src/FFI_AGENTS.md
+++ b/src/agentsight/src/FFI_AGENTS.md
@@ -8,3 +8,4 @@
 4. 错误通过 thread-local `LAST_ERROR` 返回，调用方用 `agentsight_last_error()` 读取
 5. 指针参数必须在函数入口做 null check，返回错误码而非 UB
 6. 修改函数签名后必须确认 `build.rs` drift guard 通过（注意：drift guard 只检查函数名，不检查签名）
+7. `AgentsightLLMData` 字段变更（新增/删除/重排）会改变 C ABI layout — cbindgen 自动拾取 `#[repr(C)]` struct 的所有字段，但变更后必须确认下游 C 消费方已同步更新
--- a/src/agentsight/src/bin/cli/audit.rs
+++ b/src/agentsight/src/bin/cli/audit.rs
@@ -222,6 +222,7 @@ mod tests {
                 args: Some(args.into()),
                 exit_code: None,
             },
+            session_id: None,
         }
     }
 
--- a/src/agentsight/src/container.rs
+++ b/src/agentsight/src/container.rs
@@ -1,0 +1,170 @@
+//! Container ID extraction from `/proc/{pid}/cgroup`.
+//!
+//! Supports Docker (cgroup v1 & v2), containerd, and Kubernetes cgroup
+//! layouts.  Returns `None` for non-container processes.
+
+/// Read `/proc/{pid}/cgroup` and extract the container ID.
+///
+/// Returns `None` when the process is not running inside a container or
+/// when the cgroup file cannot be read.
+pub fn extract_container_id(pid: u32) -> Option<String> {
+    let path = format!("/proc/{pid}/cgroup");
+    match std::fs::read_to_string(&path) {
+        Ok(content) => parse_container_id_from_cgroup(&content),
+        Err(e) => {
+            log::debug!("failed to read {path}: {e}");
+            None
+        }
+    }
+}
+
+/// Pure function: extract a 64-char hex container ID from raw cgroup
+/// file content.
+///
+/// Recognised layouts (checked in order):
+///
+/// 1. Docker cgroup v1 — `.../docker/<64hex>`
+/// 2. Docker cgroup v2 — `docker-<64hex>.scope`
+/// 3. Kubernetes       — `/kubepods/.../<64hex>`
+/// 4. containerd       — last path segment is exactly 64 hex chars
+pub fn parse_container_id_from_cgroup(content: &str) -> Option<String> {
+    for line in content.lines() {
+        // The third colon-separated field is the cgroup path.
+        let cgroup_path = match line.splitn(3, ':').nth(2) {
+            Some(p) => p,
+            None => continue,
+        };
+
+        if let Some(id) = try_extract_from_path(cgroup_path) {
+            return Some(id);
+        }
+    }
+    None
+}
+
+/// Try to extract a container ID from a single cgroup path string.
+fn try_extract_from_path(path: &str) -> Option<String> {
+    // 1. Docker cgroup v1: .../docker/<64hex>
+    if let Some(pos) = path.find("/docker/") {
+        let candidate = &path[pos + "/docker/".len()..];
+        // Strip any trailing path segments
+        let candidate = candidate.split('/').next().unwrap_or(candidate);
+        if is_64_hex(candidate) {
+            return Some(candidate.to_string());
+        }
+    }
+
+    // 2. Docker cgroup v2: docker-<64hex>.scope
+    for segment in path.rsplit('/') {
+        if let Some(rest) = segment.strip_prefix("docker-") {
+            if let Some(hex) = rest.strip_suffix(".scope") {
+                if is_64_hex(hex) {
+                    return Some(hex.to_string());
+                }
+            }
+        }
+    }
+
+    // 3. Kubernetes: /kubepods/.../<64hex>
+    if path.contains("/kubepods") {
+        // The container ID is the last 64-hex segment.
+        if let Some(segment) = path.rsplit('/').next() {
+            if is_64_hex(segment) {
+                return Some(segment.to_string());
+            }
+        }
+    }
+
+    // 4. containerd / generic: last path segment is exactly 64 hex chars.
+    if let Some(segment) = path.rsplit('/').next() {
+        if is_64_hex(segment) {
+            return Some(segment.to_string());
+        }
+    }
+
+    None
+}
+
+/// Returns `true` when `s` is exactly 64 lowercase-hex characters.
+fn is_64_hex(s: &str) -> bool {
+    s.len() == 64 && s.chars().all(|c| c.is_ascii_hexdigit())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn docker_cgroup_v1() {
+        let content =
+            "12:devices:/docker/a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2\n";
+        let id = parse_container_id_from_cgroup(content).unwrap();
+        assert_eq!(
+            id,
+            "a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2"
+        );
+    }
+
+    #[test]
+    fn docker_cgroup_v2() {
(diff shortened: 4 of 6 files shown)


chengshuyi:
# Code Review: feat(sight): add container_id to AgentsightLLMData

Thanks @jfeng18 for this PR! The cgroup parsing logic is clean and well-tested. Below are findings across the AgentSight review dimensions.

---

## Summary

| Dimension | Status | Findings |
|-----------|--------|----------|
| Hard Rules Compliance | ⚠️ | 1 (style consistency) |
| eBPF Safety | ✅ | N/A |
| FFI Boundary | ⚠️ | 2 (cbindgen sync, panic safety annotation) |
| Footprint Ladder | ⚠️ | 1 (new module rationale missing) |
| Pipeline Test Coverage | ⚠️ | 1 (edge cases missing) |
| Documentation Sync | ❌ | 1 (AGENTS.md / FFI_AGENTS.md not updated) |

**Overall**: +177/-0 across 3 files. Code quality is solid — parsing logic is clear, tests cover major cgroup layouts. Main gaps are in **FFI contract synchronization** and **documentation updates**.

---

## Findings

### 1. [FFI Boundary] `src/agentsight/src/ffi.rs:135` — cbindgen.toml not updated

`AgentsightLLMData` gains a new `container_id: *const c_char` field, but the diff doesn't include changes to `cbindgen.toml`. If this struct is exposed via cbindgen-generated C headers, the new field may be missing from the output, causing ABI mismatch or compile errors on the C side.

**Suggestion**: Verify that `cbindgen` correctly picks up the new field in the generated header. If manual declaration in `after_includes` is required, please add it. Either way, confirm the generated `.h` file is correct and mention this in the PR description.

### 2. [FFI Boundary] `src/agentsight/src/ffi.rs:276` — panic safety at FFI boundary

`extract_container_id()` is called inside `build_llm_data()`, which constructs an FFI struct. The current implementation is safe (pure string parsing, no panic paths), but there's no defensive annotation guaranteeing this invariant holds as the function evolves.

**Suggestion**: Either wrap the call in `std::panic::catch_unwind`, or add a doc comment on `extract_container_id` explicitly stating `// SAFETY: This function and all callees are guaranteed no-panic.` to prevent future regressions from introducing UB across the FFI boundary.

### 3. [Hard Rules] `src/agentsight/src/container.rs:49,75,85` — inconsistent defensive style

Line 49 uses `unwrap_or(candidate)` defensively after `split('/').next()`, while lines 75 and 85 use bare `.next()` on `rsplit('/')` without fallback. Both are logically correct (`rsplit` always yields at least one element), but the inconsistency makes the code harder to audit.

**Suggestion**: Unify the style — either add `unwrap_or("")` / `unwrap_or(path)` everywhere with a brief comment, or remove the defensive wrapper on line 49 and add a comment explaining why `next()` is safe in all three locations.

### 4. [Footprint Ladder] `src/agentsight/src/container.rs` — new module without rationale

A new 170-line public module is introduced. The PR description notes "no new dependencies" but doesn't explain why this couldn't be implemented as internal functions within an existing module (e.g., `ffi.rs` or a utils module). Per the Footprint Ladder policy, new files at Level 2-3 should justify why extending existing modules (Level 1-2) is insufficient.

**Suggestion**: Add a one-liner in the module doc comment or PR description explaining the choice, e.g., _"Standalone module to accommodate planned cgroup v2 unified hierarchy support and container runtime detection."_ If this is expected to remain a single-purpose utility, consider inlining into `ffi.rs`.

### 5. [Documentation Sync] Multiple docs not updated

New public module `container` and new FFI field `container_id` require documentation updates:

- **`AGENTS.md`** — Module Map should include `container` entry
- **`src/FFI_AGENTS.md`** — `AgentsightLLMData` struct field list needs `container_id` added
- **`docs/ARCHITECTURE.md`** (if applicable) — LLM data collection architecture should mention container_id source and limitations

**Suggestion**: Update at minimum `AGENTS.md` Module Map and `src/FFI_AGENTS.md` struct definition before merge.

### 6. [Test Coverage] `src/agentsight/src/container.rs` — missing edge case tests

The 8 unit tests cover major layouts well. Missing edge cases:

- **overlay2 false positive**: `/docker/overlay2/<long-hex>` — overlay2 layer IDs are also long hex strings but are NOT container IDs. Does the current parser incorrectly match these?
- **Uppercase hex**: Some runtimes may produce uppercase hex. Current `is_64_hex` only accepts lowercase via `is_ascii_hexdigit()` (which does accept both A-F and a-f — please verify and add a test to lock this behavior).
- **Single-line no trailing newline**: `/proc/{pid}/cgroup` content without trailing `\n` — `lines()` handles this correctly, but a test would prevent regressions.

**Suggestion**: Add test cases for overlay2 path and uppercase hex to document expected behavior explicitly.

---

## Minor Nits (non-blocking)

- `container.rs:10` — `format!("/proc/{pid}/cgroup")` could use `concat!` or a const template if performance becomes a concern in hot paths. Not urgent.
- Consider adding `#[inline]` to `is_64_hex` since it's called in a tight loop during parsing.

---

## Verdict

Code is well-structured and the approach (pure parsing, no Docker API dependency) is sound. Please address findings #1 (cbindgen sync) and #5 (doc updates) before merge — these are process-critical for FFI correctness and maintainability. Other findings are recommendations that can be addressed in follow-up.

chengshuyi:
## Follow-up: Add LRU cache for container_id extraction

Two issues with the current per-call `/proc/{pid}/cgroup` read:

1. **Performance**: Every LLM callback triggers a file read + string parse. For long-lived Agent processes making frequent calls, this is redundant work.
2. **Correctness after process exit**: Once the target process exits, `/proc/{pid}/cgroup` disappears and `extract_container_id` returns `None`. If downstream consumers need to correlate historical LLM calls with their originating container (e.g., for post-hoc analysis), the container_id is permanently lost.

### Proposed approach

Add a process-level LRU cache keyed by PID, with TTL to handle PID reuse safely:

```rust
use std::collections::HashMap;
use std::sync::Mutex;
use std::time::{Duration, Instant};

const CACHE_CAPACITY: usize = 256;
const CACHE_TTL: Duration = Duration::from_secs(60);

struct CacheEntry {
    container_id: Option<String>,
    inserted_at: Instant,
}

static CONTAINER_ID_CACHE: Mutex<Option<HashMap<u32, CacheEntry>>> =
    Mutex::new(None);

pub fn extract_container_id_cached(pid: u32) -> Option<String> {
    // 1. Check cache
    {
        let guard = CONTAINER_ID_CACHE.lock().unwrap();
        if let Some(ref map) = *guard {
            if let Some(entry) = map.get(&pid) {
                if entry.inserted_at.elapsed() < CACHE_TTL {
                    return entry.container_id.clone();
                }
            }
        }
    }

    // 2. Cache miss or expired — read from /proc
    let result = extract_container_id(pid);

    // 3. Insert into cache (evict oldest if at capacity)
    {
        let mut guard = CONTAINER_ID_CACHE.lock().unwrap();
        let map = guard.get_or_insert_with(|| HashMap::with_capacity(CACHE_CAPACITY));

        // Simple eviction: if at capacity, remove the oldest entry
        if map.len() >= CACHE_CAPACITY {
            if let Some(oldest_key) = map
                .iter()
                .min_by_key(|(_, v)| v.inserted_at)
                .map(|(k, _)| *k)
            {
                map.remove(&oldest_key);
            }
        }

        map.insert(
            pid,
            CacheEntry {
                container_id: result.clone(),
                inserted_at: Instant::now(),
            },
        );
    }

    result
}
```

### Design notes

- **TTL (60s)**: Balances freshness vs. PID reuse risk. Container lifetimes are typically >> 60s; PID reuse within 60s is rare on systems with high PID space. Adjust based on observed workload.
- **Capacity (256)**: Upper-bounds memory. Each entry is ~100 bytes, so max ~25KB.
- **Mutex contention**: LLM callbacks are not expected to be extremely high-frequency. If profiling shows lock contention, switch to `parking_lot::RwLock` or a concurrent cache crate (`moka`).
- **Zero-dependency option above**: Uses only `std`. If the team prefers a battle-tested implementation, `moka` (async-aware, concurrent) or `lru` (simple sync LRU) are good alternatives but add a dependency.
- **FFI integration**: Replace `crate::container::extract_container_id(call.pid as u32)` with `extract_container_id_cached(...)` in `build_llm_data()`. The rest of the FFI struct construction stays unchanged.

### Trade-offs acknowledged

| Concern | Mitigation |
|---------|-----------|
| PID reuse returns stale container_id | TTL ensures entries expire; 60s window is short relative to typical container lifetime |
| Memory growth | Hard cap at 256 entries with LRU eviction |
| Lock contention under high concurrency | Start with `std::sync::Mutex`; upgrade path to concurrent cache if needed |
| Added complexity vs. Phase 0 simplicity | Cache is opt-in via separate function; original `extract_container_id` remains available for callers that don't want caching |

Suggest implementing this as a follow-up commit in this PR or a separate PR, depending on urgency. Happy to help with the implementation if needed.

chengshuyi:
@jfeng18 Added code review comments and a follow-up suggestion for LRU caching. Please take a look when you get a chance. Thanks!

jfeng18:
Thanks for the thorough review! All 6 findings addressed in the latest push:

**#1 cbindgen**: Verified — cbindgen auto-picks up `#[repr(C)]` struct fields. No manual `after_includes` change needed. Added a note to FFI_AGENTS.md (#7) about ABI layout changes.

**#2 panic safety**: Added doc comment on `extract_container_id` explicitly stating no-panic guarantee.

**#3 unwrap_or style**: Unified — all `split/rsplit().next()` calls now have `unwrap_or("")` or a comment explaining why `next()` is safe.

**#4 module rationale**: Added to module doc: Footprint Ladder Level 3, justified by planned expansion (cgroup v2 unified hierarchy, container runtime detection).

**#5 docs**: Updated AGENTS.md (Module Map) and FFI_AGENTS.md.

**#6 edge cases**: Added 3 tests — overlay2 (not a false positive: "overlay2" is 8 chars, not 64), uppercase hex (accepted by `is_ascii_hexdigit`), no trailing newline. Fixed doc comment from "lowercase-hex" to "case-insensitive".

---

Re: **LRU cache follow-up** — agree the concern is valid but suggest deferring to a separate PR. At LLM callback time the process is alive (it just completed an API call), so `/proc/{pid}/cgroup` is always readable. Per-call file read cost is negligible for LLM-frequency events (~tens/min). Happy to add caching if profiling shows it matters.

chengshuyi:
### Suggestion: Use a well-maintained LRU crate instead of hand-rolled implementation

The current cache uses `HashMap` + linear-scan eviction to simulate LRU. While correct and zero-dependency, this is a well-solved problem in the Rust ecosystem. Hand-rolling common data structures carries maintenance risk (edge cases, concurrency bugs, performance regressions) that public crates have already absorbed through community testing.

**Recommended options:**

| Crate | Pros | Cons |
|-------|------|------|
| [`lru`](https://crates.io/crates/lru) | Simple sync API, O(1) eviction via intrusive linked list, widely used (~2k downloads/day), no async overhead | No built-in TTL |
| [`moka`](https://crates.io/crates/moka) | Built-in TTL/TTI, concurrent (lock-free reads), production-proven at scale | Heavier dependency, async-first design (sync API available via `moka::sync::Cache`) |
| [`mini-moka`](https://crates.io/crates/mini-moka) | Lightweight subset of moka with TTL support, minimal deps | Smaller community |

For this use case (sync, small capacity, TTL needed), I'd lean toward **`moka::sync::Cache`** or **`mini-moka`** since TTL is a core requirement here. The `lru` crate would need an external TTL wrapper, which defeats the purpose of using a library.

This aligns with the project's general practice of leveraging ecosystem crates for standard algorithms rather than maintaining custom implementations. Happy to help evaluate which crate fits best if needed.

jfeng18:
Good call — switched to `lru::LruCache` (already in Cargo.toml, used by aggregator/response_map/id_resolver). Removed the hand-rolled HashMap + linear-scan eviction. O(1) eviction, 8 fewer lines, consistent with the rest of the codebase. TTL check stays on `get()`.

jfeng18:
Thanks for the crate recommendations! After evaluating, decided to stay with lru::LruCache for this case:

- Already the project standard (used in response_map, http2, http/aggregator, id_resolver) — consistency matters
- Zero incremental dependency cost (already in Cargo.toml)
- TTL is a 3-line elapsed() check, not a wrapper — maintainable as-is

moka would add ~18 transitive deps for a 256-entry cache called tens of times per minute — the concurrency optimizations would not pay off here.

Happy to revisit if we hit a use case that genuinely needs built-in TTL/TTI at scale.

chengshuyi (approved):
LGTM after 3 rounds of review. All findings addressed, LRU cache implemented with project-standard crate.

PR #1140, opened 2026-06-25
feat(anolisa): add raw-backend runtime dependency preflight

### LINKED ISSUE #1117: [anolisa-cli] Resolve missing component dependencies with official-source first, system-package fallback
## Problem

`anolisa install <component>` needs a consistent strategy when a component requires runtime dependencies that are not present on the host, such as language runtimes or helper binaries. Today this can become component-specific behavior, which makes install results hard to predict across distributions.

## Goal

Add a dependency resolution flow that prefers the component official dependency source first, then falls back to OS system packages when that is appropriate for the target environment.

## Proposed behavior

Resolution order:

1. Use the component official dependency source or bundled/declared dependency channel when available.
2. If the official source is unavailable or not applicable, detect the host package manager and resolve through system packages.
3. If neither path is available, return a clear actionable error with the missing dependency and suggested manual command.

The flow should work with dry-run output and should not silently escalate privileges.

## Acceptance criteria

- `anolisa install <component> --dry-run` shows missing dependency checks and the selected resolution path.
- The resolver prefers official dependency sources over system package fallback.
- System package fallback is explicit and records what package was installed or required.
- Human and JSON output both expose unresolved dependencies and remediation hints.
- Tests cover official-source success, system-package fallback, unsupported OS/package manager, and dry-run behavior.

## Related

Part of the anolisa-cli install lifecycle closure.

ikunkun-sys:
## Description

The raw install backend is an HTTPS GET + untar with no package manager, so RPM `Requires` never runs on a raw install. A missing runtime dependency surfaces only as a service that silently dies after a "successful" install — e.g. `install ws-ckpt --backend raw` on a host without `btrfs-progs` reports success, enables/starts the unit, then the service fails at runtime with `wrong fs type` and no dependency warning anywhere.

Add a preflight that probes declared dependencies while the install lock is held but before any filesystem mutation, and fails fast with per-dependency remediation instead of installing a dead component.

Behavior:
- Three dependency kinds via a `DependencyKind` enum, each with its own missing-dep policy:
  - `system-package` → remediation `<dnf|apt> install <name>` (name picked per host package family); **reported, not auto-installed**.
  - `platform-capability` → kernel-capability check (`btf` / `cap_bpf` / `btrfs` + `min_kernel`); **never installable**, fails fast.
  - `language-runtime` → manual install hint (node; **no vendoring yet**).
- Probes run under the install lock, before the first filesystem mutation; every miss is aggregated and reported in one shot.
- `install --dry-run` renders a read-only preflight section and never fails.
- The RPM backend never reaches the resolver (dnf owns its `Requires`), so a dependency is never resolved twice.

## Design Note

- **Schema lives on the artifact contract** (`.anolisa/component.toml`), as a structured `[[component.dependencies]]` array (`runtime_deps`), distinct from the flat `[dependencies]` spec. The contract is what the raw backend actually executes (it already drives layout / capabilities / services / hooks) and travels inside the tarball; the bundled `manifests/runtime/*.toml` are catalog/planning artifacts and schema-sensitive, so they are left untouched.
- **platform-capability is a first-class kind, not a degenerate package.** agentsight's BTF / kernel-5.8 and ws-ckpt's btrfs kernel support cannot be installed by any package manager (`kernel-headers` is a package; "kernel built with BTF/btrfs" is not). A two-way split would misclassify these into the package bucket and fail; the third kind makes "check-only, never installable" explicit.
- **Resolver (anolisa-core) never mutates.** It probes each dependency through an injectable `CommandRunner` (fake runner → unit-testable without a real host), maps each kind to its policy, and aggregates every miss. `execute_raw` calls it before the first mutation and refuses on any miss; `--dry-run` runs the same probe read-only.

## Ship Note

- **Check + report only.** Auto-installing system packages (`dnf`/`apt`) and language-runtime vendoring are both deferred to later phases — this PR only turns a silent runtime death into an actionable pre-install error.
- doctor/status's missing post-hoc runtime-dep check (doctor is `NOT_IMPLEMENTED`, status only checks file integrity) is an orthogonal gap, tracked separately and out of scope here.
- No general cross-distro package-name database: only the rpm/deb names actually used by current components are mapped. Target package managers are alinux (dnf) and Ubuntu (apt); anything else is reported unsupported rather than guessed.
- The btrfs platform-capability check reads `/proc/filesystems` — it verifies btrfs is **currently exposed by the running kernel**, not that the module could be loaded on demand. On a modular kernel where btrfs is not yet loaded this conservatively reports the capability as missing (a deliberate false-negative: the preflight would rather over-report a miss than let a real one through). A stronger probe (`modprobe -n btrfs` / module availability) is deferred.

## Test

- **Unit (manifest):** `[[component.dependencies]]` parsing for all three kinds (kebab-case tags), covering `version` / `probe` / `source` / `packages` / `check` / `min_kernel`.
- **Unit (resolver):** per-kind probing via a fake `CommandRunner` — system-package remediation picks the rpm vs deb name per host family; platform-capability honors `min_kernel` + `btf` / `cap_bpf` / `btrfs`; language-runtime emits the manual hint; version-constraint verdicts; multiple misses aggregated in one pass; resolver never mutates.
- **Unit (install/CLI):** `execute_raw` refuses before the first filesystem mutation; `--dry-run` renders the preflight section, never fails, and writes nothing; the RPM backend bypasses the resolver entirely.
- **Full gate green:** `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, `cargo doc --workspace --no-deps`.
- **Real-host:** on a host missing the declared deps, `install <comp> --backend raw` now fails fast listing the misses (previously reported success, then the service died); re-running once the deps are present proceeds; `--dry-run` shows the preflight without touching anything.

close #1117

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
@@ -44,9 +44,10 @@ use anolisa_core::state::{
     ObjectStatus, OperationRecord, OwnedFile, Ownership, RpmMetadata, ServiceRef,
 };
 use anolisa_core::{
-    ArtifactType, CapabilityRequest, ComponentManifest, DistributionEntry, DistributionIndex,
-    FileKind, HookPhase, HookSpec, ResolveQuery, ServiceActivation, ServiceManager, ServiceRequest,
-    ServiceRunOutcome, ServiceScope, apply_capabilities, apply_services,
+    ArtifactType, CapabilityRequest, ComponentManifest, DependencyKind, DependencyResolution,
+    DependencyResolver, DependencyStatus, DistributionEntry, DistributionIndex, FileKind,
+    HookPhase, HookSpec, ResolveQuery, ResolverEnv, ServiceActivation, ServiceManager,
+    ServiceRequest, ServiceRunOutcome, ServiceScope, apply_capabilities, apply_services,
     capability_for_install_mode, deactivate_services, expand_layout_placeholders,
     resolve_manifest_hooks, run_hooks, service_for_install_mode, user_service_for_install_mode,
 };
@@ -123,6 +124,58 @@ struct InstallPreview {
     files: Vec<ResolvedInstallFile>,
     services: Vec<ServiceRequest>,
     capabilities: Vec<CapabilityRequest>,
+    /// Runtime-dependency preflight outcomes. Empty when the artifact was not
+    /// downloaded (file/service details unavailable) or the component declares
+    /// none.
+    dependencies: Vec<DependencyResolution>,
+}
+
+/// Project detected host facts onto the slice the dependency resolver needs.
+fn resolver_env_from_facts(facts: &anolisa_env::EnvFacts) -> ResolverEnv {
+    ResolverEnv {
+        kernel: facts.kernel.clone(),
+        // `os_id` (raw `/etc/os-release` ID) maps to the coarse rpm/deb family;
+        // the legacy `EnvFacts::pkg_base` is Anolis-specific and unsuitable here.
+        pkg_base: facts
+            .os_id
+            .as_deref()
+            .and_then(anolisa_env::pkg_base_from_id),
+        btf: facts.btf,
+        cap_bpf: facts.cap_bpf,
+    }
+}
+
+/// Runtime-dependency preflight shared by the fresh-install (`execute_raw`) and
+/// update (`execute_raw_update`) paths. Probes every declared dependency
+/// through the system resolver and returns the satisfied plan's (soft)
+/// warnings, or an error listing every miss so the caller can refuse **before
+/// touching the host**. Empty `runtime_deps` is a no-op. The RPM backend never
+/// calls this — dnf owns its `Requires`, so a dependency is never resolved
+/// twice. Pure probe: never mutates.
+pub(crate) fn run_runtime_preflight(
+    manifest: &ComponentManifest,
+    env: &anolisa_env::EnvFacts,
+    command: &str,
+) -> Result<Vec<String>, CliError> {
+    if manifest.runtime_deps.is_empty() {
+        return Ok(Vec::new());
+    }
+    let plan = DependencyResolver::system()
+        .resolve(&manifest.runtime_deps, &resolver_env_from_facts(env))
+        .map_err(|err| CliError::Runtime {
+            command: command.to_string(),
+            reason: format!("invalid runtime dependency declaration: {err}"),
+        })?;
+    if !plan.is_satisfied() {
+        return Err(CliError::Runtime {
+            command: command.to_string(),
+            reason: format!(
+                "missing runtime dependencies; no files were changed:\n  {}",
+                plan.unsatisfied_lines().join("\n  ")
+            ),
+        });
+    }
+    Ok(plan.warnings)
 }
 
 /// Execution input after the artifact has been verified and its install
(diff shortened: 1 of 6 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -2621,6 +2740,33 @@ fn execute_raw(
             command: command.to_string(),
             reason: format!("failed to parse component manifest for hook resolution: {err}"),
         })?;
+
+    // Runtime-dependency preflight — probe declared dependencies while the lock
```
kongche-jbw, line 2777: [P1] This preflight currently only runs through the fresh raw install path. Raw update reuses `prepare_raw_execution()` and then enters `execute_raw_update()`, but never calls `DependencyResolver`, so a new artifact can add `[[component.dependencies]]` and still replace files/start services on a host that misses the new dependency. Please extract this into a shared helper and call it from both `execute_raw()` and `execute_raw_update()` before any filesystem mutation.

src/anolisa/crates/anolisa-core/src/resolver.rs
```diff
@@ -0,0 +1,855 @@
...
+        Some(false) => CheckResult::Unsupported {
+            reason: format!("{label} is not available on this host"),
+        },
+        None => CheckResult::Unsupported {
+            reason: format!("{label} could not be determined on this host"),
+        },
+    }
+}
+
+/// Whether the running kernel supports btrfs, read from `/proc/filesystems`.
```
kongche-jbw, line 376: [P2] This checks only `/proc/filesystems`, which can false-negative on modular kernels where btrfs exists but has not been loaded yet. If this is intentionally “currently exposed by the running kernel” rather than “available to be loaded”, please call that out in the Ship Note. A later stronger check could also probe `modprobe -n btrfs` or module availability.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
ikunkun-sys: Fixed — folded into the preflight commit (now `b1fa42d5`). Extracted the preflight into a shared `run_runtime_preflight` helper, now called from both `execute_raw()` and `execute_raw_update()` before the first filesystem mutation (and before the update transaction opens). Added a regression test: a v2 artifact declaring an unsatisfiable dependency is refused with the old binary and recorded version left untouched.

src/anolisa/crates/anolisa-core/src/resolver.rs
ikunkun-sys: Intentional — this checks "currently exposed by the running kernel", and the conservative false-negative is by design: the preflight would rather over-report a miss than let a real one through. Called it out in the Ship Note now; a stronger `modprobe -n btrfs` / module-availability probe is deferred as a follow-up.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
kongche-jbw: Confirmed, this addresses the P1 concern. The shared `run_runtime_preflight` call in the raw update path runs before the transaction and before Phase 1 backup/remove, and the regression test covers the old-file/state-not-touched behavior.

src/anolisa/crates/anolisa-core/src/resolver.rs
kongche-jbw: Thanks, that explanation works for this PR. Treating btrfs as a conservative currently-exposed kernel capability is acceptable as long as the Ship Note calls out the possible modular-kernel false-negative and leaves stronger module probing as follow-up.

kongche-jbw (approved):

PR #1157, opened 2026-06-26
fix(anolisa): use bare component name for rpm package default

### LINKED ISSUE #1150: [anolisa] rpm backend delegated install prepends "anolisa-" prefix, producing a wrong package name
## Summary

When installing a component through the **rpm backend**, `anolisa` delegates a
`dnf install` using a package name that is hardcoded with an `anolisa-` prefix
(e.g. `anolisa-tokenless`). The real RPM package name has **no** such prefix
(it is just `tokenless`), so the delegated install targets a non-existent
package.

## Steps to reproduce

```bash
sudo anolisa --install-mode system install --backend rpm tokenless
```

(any component that is not yet present as a system RPM, has no `--package`
override, no `[backends.rpm].package` in its manifest, and no `package_map`
entry in `repo.toml`)

## Expected vs. actual

- **Expected:** `dnf install tokenless`
- **Actual:** `dnf install anolisa-tokenless`

## Root cause

The rpm-backend package-name resolution lives in `rpm_package_candidates`
(`crates/anolisa-cli/src/commands/tier1/install.rs`). Its precedence chain is:

1. CLI `--package`
2. component manifest `[backends.rpm].package`
3. `repo.toml` `package_map`
4. RPM virtual-provides reverse lookup `anolisa-component(<name>)`
5. **default fallback: `anolisa-<component>`**

When the first four levels miss, it falls through to the default:

```rust
// rpm_package_candidates
Ok(vec![format!("anolisa-{component}")])
```

This resolved name then flows through `RpmSituation::Absent { package }` →
`execute_delegated_install` → `exec.txn.install(package)` (i.e. `dnf install`).

## Why the default is wrong

The `anolisa-` prefix assumption does not match the actual packaging
convention:

- **Every component's RPM spec `Name:` is the bare component name**, with no
  prefix: `tokenless`, `copilot-shell`, `agentsight`, `agent-memory`,
  `os-skills`, `ws-ckpt`, `skillfs`, `agent-sec-core`.
- No spec declares `Provides: anolisa-component(<name>)`, so level 4 always
  returns empty and resolution necessarily reaches the wrong level 5.
- Component `component.toml` files do not declare `[backends.rpm].package`
  either, so level 2 also misses.
- **By contrast, the raw backend** (`RepoConfig::package_name` in
  `crates/anolisa-cli/src/repo_config.rs`) defaults to `component.to_string()`
  — the bare name, no prefix. The two backends disagree, and the rpm side is
  the one that contradicts the real spec names.

## Affected locations

The same `anolisa-<name>` assumption appears in three places:

- `rpm_package_candidates` default fallback — `Ok(vec![format!("anolisa-{component}")])`
- `probe_rpm_situation` defensive fallback — `.unwrap_or_else(|| format!("anolisa-{component}"))`
- adopt path — `package_name: format!("anolisa-{name}")` in
  `crates/anolisa-cli/src/commands/tier1/adopt.rs`

The doc comments on these functions also document `default anolisa-<name>` as
the spec, so they need updating too.

## Suggested fix

Change the rpm-backend default package name from `anolisa-<component>` to the
bare `<component>`, aligning it with both the raw backend and the actual spec
`Name:` fields. Update the affected fallbacks, the related doc comments, and
the tests that assert `anolisa-...` names.

(If instead the intended convention is that RPM package names *should* carry an
`anolisa-` prefix, then the fix belongs in the 8 `*.spec.in` files and the
published artifacts, not in the CLI — but that contradicts what is currently
shipped.)

ikunkun-sys:
## Description

Installing a component through the **rpm backend** delegated a `dnf install` whose package name was hardcoded with an `anolisa-` prefix (e.g. `anolisa-tokenless`). The real RPM package name carries no such prefix — every component spec's `Name:` is the bare component name (`tokenless`, `copilot-shell`, ...) — so the delegated install targeted a non-existent package.

The rpm package-name resolution chain is `--package` > manifest `[backends.rpm].package` > `package_map` > virtual-provides reverse lookup > default. Only the final default was wrong: it produced `anolisa-<component>` instead of the bare `<component>`. This fixes the default in both `rpm_package_candidates` and the `probe_rpm_situation` fallback.

## Design Note

- **Align with the raw backend and the real spec names.** The raw backend's `RepoConfig::package_name` already defaults to the bare component name; the rpm path now matches it, and both match the published spec `Name:` fields.
- **Virtual-provides contract untouched.** The `anolisa-component(<name>)` capability namespace is unrelated to the package name and is preserved, so a future packaging-side `Provides:` still short-circuits the default.
- **Test fixtures de-prefixed.** RPM package-name fixtures across the suite used the misleading `anolisa-<component>` form; they now use the bare name so test data matches reality. Repo names (`anolisa-release`) and the virtual-provides capability are left intact.

## Test

- **Full gate green:** `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check --workspace --all-targets`, `cargo test --workspace`, `cargo doc --workspace --no-deps`.
- **Focused:** `cargo test -p anolisa-cli commands::tier1::install` — delegated install / adopt / probe now resolve the bare `<component>` name.
- **Not run:** real-host `dnf install` verification against a live rpmdb.

close #1150

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/adopt.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/adopt.rs
@@ -246,7 +246,7 @@ mod tests {
             install_backend: Some(if is_rpm { "rpm" } else { "raw" }.to_string()),
             ownership: Some(ownership),
             rpm_metadata: is_rpm.then(|| RpmMetadata {
-                package_name: format!("anolisa-{name}"),
+                package_name: name.to_string(),
                 evr: Some("1.0.0-1.al8".to_string()),
                 arch: Some("x86_64".to_string()),
                 source_repo: Some("@System".to_string()),
@@ -296,10 +296,10 @@ mod tests {
         seed(&c, Vec::new());
         let q = FakeQuery {
             installed: vec![(
-                "anolisa-copilot-shell".to_string(),
-                pkg_info("anolisa-copilot-shell", "2.2.0", Some("1.al8"), "x86_64"),
+                "copilot-shell".to_string(),
+                pkg_info("copilot-shell", "2.2.0", Some("1.al8"), "x86_64"),
             )],
-            origins: vec![("anolisa-copilot-shell".to_string(), "@System".to_string())],
+            origins: vec![("copilot-shell".to_string(), "@System".to_string())],
             ..Default::default()
         };
         adopt_with_query("copilot-shell", None, &c, &q).expect("adopt ok");
@@ -311,7 +311,7 @@ mod tests {
         assert_eq!(obj.effective_ownership(), Ownership::RpmObserved);
         assert_eq!(obj.status, ObjectStatus::Adopted);
         let meta = obj.rpm_metadata.as_ref().expect("rpm metadata");
-        assert_eq!(meta.package_name, "anolisa-copilot-shell");
+        assert_eq!(meta.package_name, "copilot-shell");
         assert_eq!(meta.evr.as_deref(), Some("2.2.0-1.al8"));
         assert!(
             after
@@ -338,8 +338,8 @@ mod tests {
         );
         let q = FakeQuery {
             installed: vec![(
-                "anolisa-copilot-shell".to_string(),
-                pkg_info("anolisa-copilot-shell", "2.0.0", Some("1.al8"), "x86_64"),
+                "copilot-shell".to_string(),
+                pkg_info("copilot-shell", "2.0.0", Some("1.al8"), "x86_64"),
             )],
             ..Default::default()
         };
@@ -373,7 +373,7 @@ mod tests {
                 ObjectStatus::Adopted,
             )],
         );
-        // Existing observed package is `anolisa-copilot-shell`; the user pins a
+        // Existing observed package is `copilot-shell`; the user pins a
         // different installed package via --package.
         let q = FakeQuery {
             installed: vec![(
@@ -387,7 +387,7 @@ mod tests {
 
         assert_eq!(err.code(), "INVALID_ARGUMENT");
         assert!(
-            err.reason().contains("anolisa-copilot-shell")
+            err.reason().contains("copilot-shell")
                 && err.reason().contains("anolisa-other")
                 && err.reason().contains("forget"),
             "refusal must name both packages and point at forget: {}",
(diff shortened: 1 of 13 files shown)


kongche-jbw:
Review result: no blocking issue found. The package-name resolution change looks reasonable and the core precedence chain still holds (`--package` > manifest `[backends.rpm].package` > repo `package_map` > virtual provides > default bare component name).

Two non-blocking consistency fixes are recommended in this PR:

1. `src/anolisa/crates/anolisa-core/src/manifest.rs`

The doc comment for `ManifestBackends` still says the fallback is `default anolisa-<component>`. It should be updated to say the fallback is the bare component name, otherwise future readers may keep the old mental model.

2. `src/anolisa/templates/installed-state.toml`

The RPM example still uses:

```toml
package_name = "anolisa-copilot-shell"
```

It should be updated to:

```toml
package_name = "copilot-shell"
```

This is template/documentation consistency rather than a runtime blocker, but since this PR is specifically changing the RPM default-name semantics, it is worth fixing here.

Validation note: I ran the tier-1 focused tests locally on the PR branch:

```bash
cargo test -p anolisa-cli --locked commands::tier1 --quiet
```

Result: `289 passed`.

ikunkun-sys:
> Review result: no blocking issue found. The package-name resolution change looks reasonable and the core precedence chain still holds (`--package` > manifest `[backends.rpm].package` > repo `package_map` > virtual provides > default bare component name).
> 
> Two non-blocking consistency fixes are recommended in this PR:
> 
> 1. `src/anolisa/crates/anolisa-core/src/manifest.rs`
> 
> The doc comment for `ManifestBackends` still says the fallback is `default anolisa-<component>`. It should be updated to say the fallback is the bare component name, otherwise future readers may keep the old mental model.
> 
> 2. `src/anolisa/templates/installed-state.toml`
> 
> The RPM example still uses:
> 
> ```toml
> package_name = "anolisa-copilot-shell"
> ```
> 
> It should be updated to:
> 
> ```toml
> package_name = "copilot-shell"
> ```
> 
> This is template/documentation consistency rather than a runtime blocker, but since this PR is specifically changing the RPM default-name semantics, it is worth fixing here.
> 
> Validation note: I ran the tier-1 focused tests locally on the PR branch:
> 
> ```shell
> cargo test -p anolisa-cli --locked commands::tier1 --quiet
> ```
> 
> Result: `289 passed`.

Changed.

kongche-jbw (approved):

kongche-jbw:
lgtm thx

PR #1158, opened 2026-06-26
feat(anolisa): support skill-only adapter bundles

kongche-jbw:
### What changed

- Added `adapter_type = "skill_bundle"` for adapters that only deliver
  declared skills.
- OpenClaw and Hermes skip plugin registration / enable flows for
  `skill_bundle` adapters.
- Skill delivery still uses the existing adapter manager IO boundary,
  receipts, and disable cleanup path.
- Updated manifest/template docs and adapter tests.

### Known limitations

- `skill_bundle` currently means skill delivery only. It does not run
  framework config, service setup, or plugin registration.
- Adapter status still follows the existing receipt-oriented model and
  does not add a dedicated skill health summary.

## Description

This adds the minimal adapter framework support needed for components that
only need to copy declared skills into agent frameworks. It is intended to
support os-skills style component contracts without forcing a plugin manifest
or plugin registration step.

## Related Issue

no-issue: split framework support for os-skills component contract

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

## Testing

```bash
cargo fmt --all -- --check
cargo test -p anolisa-core --locked adapter --quiet
cargo test -p anolisa-cli --locked adapter --quiet
```

## Additional Notes

This PR is intentionally separated from the os-skills contract packaging PR.


ikunkun-sys (commented):
Thanks for the change. I found two issues that look worth fixing before merge:

1. `skill_bundle` receipts will report `unknown` status immediately after a successful enable.

For both OpenClaw and Hermes, `prepare_enable` now creates a receipt with `plugin_id = None` for `adapter_type = "skill_bundle"`, which is correct for skipping plugin registration. But `status()` still treats plugin registration as the core health signal. When the receipt has no plugin id, it emits `PluginRegistered = Unknown` with `"receipt has no plugin id"`, and `summarize(...)` turns that into an overall `unknown` summary. So a successfully enabled skill-only adapter will never look healthy through `anolisa adapter status`.

Relevant spots:
- `src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs`: `status()` around the no-plugin-id branch and `summarize(...)`
- `src/anolisa/crates/anolisa-core/src/adapter/hermes.rs`: same pattern

I think the fix needs an explicit way to identify skill-only receipts during status, then skip/replace the plugin-registration condition for that adapter type. That likely means persisting `adapter_type` or equivalent typed capability in the claim rather than relying on the enable-time `DriverCtx`, since manager status/disable currently rebuilds `DriverCtx` with `adapter_type: None`.

2. OpenClaw `skill_bundle` config entries are silently ignored.

The manager still parses `declared_config` and passes it into the driver. For OpenClaw plugin adapters, the driver records config resources and runs `openclaw config set`. For `skill_bundle`, this PR skips both the claim resources and the apply step. That means a manifest can declare config, enable successfully, and get no indication that the config was not applied.

Relevant spots:
- `src/anolisa/crates/anolisa-core/src/adapter/manager.rs`: config is still collected for all adapter types
- `src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs`: `prepare_enable` and `apply_enable` skip config when `ctx.is_skill_bundle()`

The PR description says `skill_bundle` does not run framework config, so I would fail fast during enable/plan when config is present for a skill bundle, rather than silently dropping it.

I did not run the local test suite; this is based on the PR diff/head inspection. The added tests cover plan/claim behavior, but they do not cover the post-enable status path where the first issue shows up.

ikunkun-sys (approved):

PR #1173, opened 2026-06-29
fix(anolisa): delegate rpm-owned self update to dnf

ikunkun-sys:
## Description

Fixes `anolisa update self` for RPM-packaged CLI installs. The command still checks the release manifest for a newer CLI version, but the apply step now detects whether the running executable is owned by an installed RPM via `rpm -q --whatprovides <current_exe>`.

Behavior:
- RPM-owned current executable -> delegates to `dnf update -y <package>` instead of replacing the binary path directly.
- RPM-owned current executable without root -> refuses before running dnf and prints a sudo hint.
- Non-RPM install or host without rpm tooling -> keeps the existing verified tar.gz binary replacement path.
- Ambiguous RPM ownership -> refuses rather than choosing an arbitrary package.

## Design Note

The ownership decision lives in the CLI update layer rather than in `anolisa_core::self_update`: core continues to own manifest/artifact validation and verified binary replacement, while the CLI chooses the correct writer for the current executable.

The implementation reuses the existing `PackageQuery` / `PackageTransaction` seams and adds a small `SelfUpdateOps` test seam so unit tests can prove the RPM branch never calls the binary replacement path. Output now also tracks the apply mode: binary updates keep the existing HTTPS-only signature warning, while RPM updates report the dnf package that was updated.

## Ship Note

- `update self --dry-run` remains a manifest/artifact availability check and does not probe rpm ownership or run dnf.
- The RPM path relies on the system package manager and repo signature policy; it does not consume the release-manifest artifact bytes for replacement.
- If rpm ownership probing hits a hard rpm query failure, self-update refuses rather than risking rpmdb drift.

## Test

- `cargo test -p anolisa-cli commands::tier1::update::tests`
- `cargo check --workspace --all-targets`

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
@@ -27,7 +27,7 @@
 //! `update all` remains `NOT_IMPLEMENTED`.
 
 use std::io::Read;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::time::Duration;
 
 use chrono::{SecondsFormat, Utc};
@@ -1573,6 +1573,9 @@ fn fetch_and_write_repo_config(url: &str, dest: &Path) -> Result<(), String> {
 pub(in crate::commands) fn handle_self_update(ctx: &CliContext) -> Result<(), CliError> {
     let url = self_update::update_url();
     let current_version = env!("CARGO_PKG_VERSION");
+    let ops = SystemSelfUpdateOps;
+    let query = RpmPackageQuery::system();
+    let txn = RpmTransaction::system();
 
     let progress_cb: Option<ProgressFn> = if !ctx.json && !ctx.quiet {
         Some(Box::new(move |downloaded: u64, total: Option<u64>| {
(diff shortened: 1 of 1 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
```diff
@@ -1619,18 +1627,205 @@ pub(in crate::commands) fn handle_self_update(ctx: &CliContext) -> Result<(), Cl
...
+    let apply_mode = if let Some(package) = rpm_owner_for_current_exe(query, &current_exe)? {
+        if !is_root {
+            return Err(CliError::Runtime {
+                command: "update self".to_string(),
+                reason: format!(
+                    "updating RPM-owned anolisa package '{package}' requires root privileges; re-run with sudo: `sudo anolisa update self`"
+                ),
+            });
+        }
+        txn.update(&package)
```
kongche-jbw, line 1768: [P1] 这里执行完 `dnf update` 后就直接把结果报成 `UpdateAvailable { to: manifest.version }`，但 `PackageTransaction::update` 把 dnf 的 no-op / already-latest 也视为成功。如果 dnf repo 还没同步 release manifest 里的 CLI 版本，命令会成功但 RPM 包没有变化，human/JSON 仍会显示已经更新到 manifest version。建议 RPM self-update 像组件 RPM update 一样在 dnf 后重新 `query_installed(package)`，对比前后 EVR；没变就报告 no-op / already latest，不要声称更新到 manifest version，并补一个 dnf no-op 的回归测试。

src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
```diff
@@ -1582,29 +1585,34 @@ pub(in crate::commands) fn handle_self_update(ctx: &CliContext) -> Result<(), Cl
...
 
-    let outcome = result.map_err(|e| CliError::Runtime {
-        command: "update self".to_string(),
-        reason: e.to_string(),
-    })?;
+    let run = result?;
 
     if ctx.json {
-        return render_json_outcome(&outcome, ctx.dry_run);
+        return render_json_outcome(&run.outcome, ctx.dry_run);
```
kongche-jbw, line 1607: [P2] 这里 JSON 输出只传了 `run.outcome`，所以机器调用方拿不到这次实际的 apply mode。PR 描述里说 output tracks apply mode，human 输出也能看到 `applied through dnf package ...`，但 JSON 仍只有版本和 updated 布尔值，无法区分 binary replacement 还是 RPM/dnf delegate。建议让 JSON payload 接收 `run.apply_mode`，增加 `apply_mode`，RPM 分支再带上 package 字段。

PR #1194, opened 2026-06-29
fix(anolisa): separate osbase helper command scope

ikunkun-sys:
## Description

Separates `osbase` helper-backed mutations from install-mode system mutations:

- Adds a dedicated `HelperGatedSystemOperation` command scope for `osbase` mutations.
- Keeps `osbase` outside the `--install-mode system/user` install-layout semantics.
- Preserves non-root `anolisa osbase ...` access to `osbase_preflight()` so the existing system-helper handshake path remains reachable.
- Keeps explicit `--install-mode user osbase ...` invalid because `osbase` has no user-mode install-layout behavior.

## Design Note

`SystemOnlyMutation` now models commands that are actually tied to ANOLISA system install-mode semantics, such as system-scope package/state mutations.

`osbase` is a system operation, but its capability gate lives in the handler: `osbase_preflight()` first attempts the privileged system-helper socket and only falls back to direct root execution. Modeling it as `HelperGatedSystemOperation` makes that distinction explicit and avoids treating non-root omitted `--install-mode` as a user-layout request for `osbase`.

Read-only `osbase` commands remain `ReadOnly`; mutation commands still reach the same handler and helper IPC code paths as before.

## Ship Note

- `anolisa osbase ...` remains usable by non-root users when system-helper is installed and reachable.
- `sudo anolisa osbase ...` still works through the direct root fallback path when no helper is used.
- `anolisa --install-mode user osbase ...` remains invalid because `osbase` does not support user install mode.
- No helper IPC protocol, socket path, or handshake behavior changes are included in this PR.

## Test

- Unit (dispatcher): helper-gated `osbase` mutation reaches handler when `--install-mode` is omitted under non-root.
- Unit (dispatcher): helper-gated `osbase` mutation reaches handler with explicit `--install-mode system` under non-root.
- Unit (dispatcher): helper-gated `osbase` mutation rejects explicit `--install-mode user`.
- Helper/IPC: `cargo test -p anolisa-core system_helper::tests::`, `cargo test -p anolisa-platform ipc::tests::`.
- Full gate green: `cargo fmt --all -- --check`, `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, `git diff --check`.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands.rs
@@ -226,19 +226,9 @@ pub fn dispatch(cli: Cli, ctx: &CliContext) -> Result<(), CliError> {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum CommandScope {
     ReadOnly,
-    ModeScopedMutation {
-        dry_run_without_root: bool,
-    },
-    SystemOnlyMutation {
-        dry_run_without_root: bool,
-        privilege_gate: PrivilegeGate,
-    },
-}
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum PrivilegeGate {
-    Dispatcher,
-    Handler,
+    ModeScopedMutation { dry_run_without_root: bool },
+    SystemOnlyMutation { dry_run_without_root: bool },
+    HelperGatedSystemOperation,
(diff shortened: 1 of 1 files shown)


kongche-jbw (approved):
lgtm

PR #1210, opened 2026-06-30
fix(memory): unblock vector/hybrid search + reserved-path/doc fixes

shiloong:
## Description

Fixes the vector/hybrid search path in `agent-memory` 0.2.0, which was completely non-functional whenever an embedding provider was configured, plus three related issues found during full functional verification against a real DashScope (OpenAI-compatible) embedding endpoint.

**Root cause of the primary bug (double failure):**
1. The index worker runs on a dedicated `std::thread` where `Handle::try_current()` returns `None`, so the embedding branch was skipped and `files_vec` stayed empty — no vectors were ever indexed.
2. `memory_search mode=vector|hybrid` called `Handle::block_on` from a tokio worker thread, panicking with `Cannot start a runtime from within a runtime`.

Existing CI only covered the no-provider BM25 fallback, so the runtime path escaped detection.

**Commits in this PR:**
- `fix(memory): unblock vector/hybrid search on tokio runtime` — capture the runtime handle at spawn time and thread it through to the worker; use `block_in_place` in the search path. Adds a regression test with a mock `EmbeddingProvider`.
- `fix(memory): stop leaking .git internals via memory_get_context` — `memory_get_context` only filtered `.anolisa/`, surfacing `.git/logs/HEAD` as agent context. Extract `is_under_git` to `safe_fs` as the single shared predicate.
- `docs(memory): align tool parameter tables with actual signatures` — `memory_timeline` was missing required `session_id`; the sovereignty/import-export table had no parameter columns. Rewritten against `mcp_server/tools.rs` (zh + en).
- `refactor(memory): resolve embedding dimensions from first response` — `OpenAiEmbedding` hardcoded 1536 for unknown models (DashScope `text-embedding-v3` is 1024). Store dimensionality in an `AtomicUsize` seeded with the estimate, overwrite from the first real response.
- `fix(memory): backfill vectors for preexisting files in full_scan` — `full_scan` (startup + inotify-overflow recovery) only built BM25, never vectors, so pre-existing files were invisible to vector search until modified. Adds a `paths_without_vec` query + a backfill pass; centralises the `block_on` strategy in an `embed_sync` helper used by both `flush` and `full_scan`.

## Related Issue

no-issue: vector/hybrid search was non-functional when an embedding provider was configured; the defects were discovered and root-caused during full functional verification of the installed 0.2.0 build, and no tracking issue exists yet.

## Type of Change

- [ ] New feature (non-breaking change that adds new functionality)
- [x] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [x] `memory` (agent-memory)
- [ ] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

- [x] For `memory` (Linux only): `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test` pass

## Testing

Branch rebased onto latest `alibaba/main`; pre-push CI all green:

- `cargo fmt --check` — pass
- `cargo clippy --all-targets -- -D warnings` — 0 warnings
- `cargo test` — 161 unit + all integration suites pass (tier_b now 23, +2 new regression tests)

End-to-end verification against a real DashScope `text-embedding-v3` (1024-dim) endpoint via a stdio JSON-RPC driver:

- `memory_search mode=vector` returns the semantic match (`decisions/2026-06/db-pick.md`, score 0.805) for a Chinese query with no keyword overlap — previously panicked.
- `memory_search mode=hybrid` returns the same hit (RRF score 0.33) — previously panicked.
- 46/46 functional cases pass, 0 panics.
- `memory_get_context` output no longer contains `.git/` entries (was leaking `.git/logs/HEAD`).
- Startup log now reads `dims=… estimated; resolves on first embed`; true 1024-dim is locked in after the first embed.

New regression tests in `tests/tier_b_test.rs`:
- `vector_and_hybrid_search_with_provider_hit_without_panic` — mock provider + multi-thread runtime; asserts vectors land in the store and search returns hits without panicking.
- `get_context_skips_git_dir` — enables git, asserts `.git/` is absent from context while real notes are present.
- `full_scan_backfills_vectors_for_preexisting_files` — writes files with no provider, reopens with a mock provider, asserts vector search finds the preexisting files with no new modify events.

## Additional Notes

- No LLM is required for vector search; only an embedding model (`/v1/embeddings` endpoint). agent-memory itself makes zero LLM calls.
- The startup `full_scan` now computes embeddings for pre-existing files synchronously, which adds HTTP latency proportional to file count at startup. This is bounded by the embedding provider's timeout and is the intended trade-off for correctness (vector search must see pre-existing memory). Files already embedding-persisted in `files_vec` are skipped via the `paths_without_vec` query.
- Commits are `Signed-off-by: Shile Zhang <shile.zhang@linux.alibaba.com>`.

### DIFF
--- a/src/agent-memory/docs/user_manual.md
+++ b/src/agent-memory/docs/user_manual.md
@@ -155,7 +155,7 @@ All tools are invoked via MCP `tools/call` with JSON object arguments. Errors re
 | `memory_observe` | `content` | `hint`, `type` | `observed at notes/observed/<ulid>.md` |
 | `memory_get_context` | — | `max_tokens` (default 2048) | markdown preview of recently modified files; each entry has `suspicious` |
 | `memory_sessions` | — | `limit` (default 10) | historical session list |
-| `memory_timeline` | — | — | cross-session timeline |
+| `memory_timeline` | `session_id` | `limit` (default 50) | tool-call timeline for a specific session |
 | `mem_index_refresh` | — | — | force-rebuild the FTS5 index |
 
 ### Tier C — governance & versioning (7)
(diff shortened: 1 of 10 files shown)


Forrest-ly (approved):
LGTM

PR #1219, opened 2026-06-30
feat(sight): json output + graceful cli errors

jfeng18:
## Summary

Give the machine-facing `discover` / `token` subcommands a structured `--json`
contract and graceful, fail-loud error handling — the surface agents and
scripts consume.

## Problem

1. `discover` had no JSON output — agents couldn't parse discovered/known agents.
2. `token --data-file <missing>` silently created an empty DB and returned
   all-zeros (misleading success).
3. `TokenStore::new()` panicked on a bad DB instead of failing gracefully.

## Fix

| file | change |
|------|--------|
| `bin/cli/discover.rs` | `--json` for both `scan` and `--list-known` |
| `bin/cli/token.rs` | missing `--data-file` → exit 1 (`check_data_file`); unreadable DB → graceful error; `--json` output |
| `storage/sqlite/token.rs` (lib) | `TokenStore::new` returns `Result` instead of panicking |
| `storage/mod.rs` | add `check_data_file` (fail early on a bad `--data-file`) |
| `bin/cli/mod.rs` | `print_json` centralizes `--json` serialization |
| `discovery/agent.rs` | `AgentInfo` / `DiscoveredAgent` derive `Serialize` |

## Testing

- `tests/cli_contract.rs`: 8 integration tests run the built binary and assert
  the discover/token JSON + exit-code contract end-to-end (discriminating).
- Incremental diff-cover **98%**; `cargo test` (1084 lib + 8 contract),
  `clippy --all-targets -D warnings`, `fmt --check` all green.

## Scope / not changed

- **The `audit` exit-code change is intentionally out of scope.** `audit` has no
  `--data-file` and its query-error arm only fires on a corrupt DB, so it is not
  testable under the 80% diff-cover gate; it stays at `main`'s behavior and can
  land as a standalone change if a user needs it.
- `summary` behavior and the default (no `--data-file`) path are unchanged.

### DIFF
--- a/src/agentsight/src/bin/cli/discover.rs
+++ b/src/agentsight/src/bin/cli/discover.rs
@@ -20,6 +20,10 @@ pub struct DiscoverCommand {
     /// Path to JSON configuration file
     #[structopt(short, long, default_value = "/etc/agentsight/config.json")]
     pub config: String,
+
+    /// Output as JSON
+    #[structopt(long)]
+    pub json: bool,
 }
 
 impl DiscoverCommand {
@@ -74,6 +78,12 @@ impl DiscoverCommand {
         let mut scanner = AgentScanner::from_rules(&rules, &[]);
         let running_agents = scanner.scan();
 
+        if self.json {
+            let infos: Vec<_> = matchers.iter().map(|m| m.info().clone()).collect();
+            super::print_json(&infos);
+            return;
+        }
+
         println!("已知 AI Agent（共 {} 条规则）:", matchers.len());
         println!("{}", "=".repeat(60));
         println!();
@@ -112,6 +122,11 @@ impl DiscoverCommand {
         let mut scanner = AgentScanner::from_rules(&rules, &[]);
         let agents = scanner.scan();
 
+        if self.json {
+            super::print_json(&agents);
+            return;
+        }
+
         if agents.is_empty() {
             println!("未发现正在运行的 AI Agent。");
             println!();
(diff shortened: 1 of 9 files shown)


jfeng18:
Rebased onto latest main. Merged main's --config flag and richer list_known output with the PR's --json support. Fixed TokenStore::new/with_table callers for the Result return type. Updated cli_contract test to match main's Chinese output. ECS gates: fmt ✓ / clippy ✓ / 1787 passed, 0 failed ✓

viron-xiao:
Reviewed against main — LGTM, approving from my side.

Verified:
- Call-site coverage for the `TokenStore::new` signature change is complete (unified.rs production paths, CLI token.rs, sqlite/token.rs tests — all updated in this diff).
- Backward compatibility: default text output unchanged; `token --json` stays equivalent to the previous pretty-printed output.
- The one intentional behavior change (`token --data-file <missing>` now exits 1 with stderr instead of silently creating a store and exiting 0) is a contract improvement — the old silent-success behavior was the worse failure mode — and the 8 cli_contract tests lock the new contract.

One nit before merge: add a CHANGELOG line for the `token --data-file` exit-code change, since external scripts parsing the old silent-success behavior will observe a different exit code.

viron-xiao (approved):
LGTM — approving.

Verified: full call-site coverage for the `TokenStore::new` signature change (unified.rs, CLI token.rs, sqlite/token.rs tests all updated in this diff), backward-compatible default output, and the one intentional behavior change (`token --data-file <missing>` → exit 1) is an improvement over the old silent-success contract and is locked by the cli_contract tests.

Nit (non-blocking): a CHANGELOG line for the exit-code change would help script consumers.

PR #1231, opened 2026-07-01
fix(anolisa): use repo.toml rpm source

ikunkun-sys:
## Description

Fixes RPM-backed `anolisa install` so `[backends.rpm]` from `repo.toml`
is used for both RPM candidate queries and delegated DNF installs. The
RPM backend now injects the configured source as a transient DNF repo
with system repos disabled, preventing installs from resolving older
packages from host-enabled repositories.

## Design Note

The platform layer now owns `DnfRepoSource`, a small command-line
adapter for `--repofrompath`, `--enablerepo`, and per-repo `gpgcheck`.
`RpmPackageQuery` and `RpmTransaction` both accept that source, so
resolution and install use the same repository contract. The CLI only
renders the RPM source when an RPM path is selected or already recorded,
so raw-only installs do not fail on unrelated RPM variables.

## Ship Note

No persistent DNF repo files are written. DNF may still report installed
packages as `@System` when ANOLISA refreshes rpmdb origin metadata after
a transient-repo transaction.

## Test

- `cargo test --locked`
- On Alibaba Cloud Linux 4, installed `os-skills` with real
  `anolisa install --backend rpm os-skills`; it installed
  `os-skills-0.6.0-1.alnx4` from repo.toml's agentic-os source while
  host-enabled repos only offered older `0.0.x` builds.
- On the same host, installed Node.js 22 with real `dnf`, then installed
  `cosh` with real `anolisa install --backend raw cosh`; status reported
  raw-managed `cosh` v2.6.0 with integrity checks passing.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
@@ -56,6 +56,7 @@ use anolisa_platform::pkg_query::{PackageInfo, PackageQuery, PackageQueryError};
 use anolisa_platform::pkg_transaction::{PackageTransaction, PackageTransactionError};
 use anolisa_platform::privilege;
 use anolisa_platform::rpm_query::RpmPackageQuery;
+use anolisa_platform::rpm_repo::DnfRepoSource;
 use anolisa_platform::rpm_transaction::RpmTransaction;
 use chrono::{SecondsFormat, Utc};
 
@@ -73,6 +74,7 @@ use crate::resolution::{
 use crate::response::{CliError, render_json, render_json_with_status};
 
 const COMMAND: &str = "install";
+const ANOLISA_RPM_REPO_ID: &str = "anolisa-configured";
 
 #[derive(Debug, Parser)]
 // `--version` here means the *component* version (the `cargo install`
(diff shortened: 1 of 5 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/install.rs
```diff
@@ -528,6 +557,51 @@ pub(crate) fn handle_one_with_exec(
     )
 }
 
+fn configured_rpm_repo_source(
+    repo_config: &RepoConfig,
+    env: &anolisa_env::EnvFacts,
+) -> Result<Option<DnfRepoSource>, CliError> {
+    let Some(backend) = repo_config.backends.get("rpm") else {
+        return Ok(None);
```
kongche-jbw, line 565: When `[backends.rpm]` is absent, returning `Ok(None)` lets explicit RPM installs build `RpmTransaction::system()`. If the package is absent, the delegated install then runs bare host `dnf install`, so it can still resolve from host-enabled repos. Adopt may not need repo config, but delegated RPM installs should fail clearly when the RPM backend is not configured, or require a `DnfRepoSource` before reaching install.

ikunkun-sys: Addressed in eb62ef58. Delegated RPM installs now call a configured-RPM-backend gate before reaching `dnf install`, so a raw-only `repo.toml` returns the existing `BackendNotConfigured` error and never invokes the transaction runner.

I also added `delegated_install_requires_configured_rpm_backend` to cover this path, and moved delegated-install success/failure tests onto a fixture that includes `[backends.rpm]`.

Verified locally:
- `cargo test --locked`

kongche-jbw (approved):
lgtm thx

PR #1239, opened 2026-07-01
fix(anolisa): include adopted RPM components in list

ikunkun-sys:
## Description

Fixes `anolisa list` / `anolisa ls` so RPM components recorded through `adopt` are shown instead of being treated as `not_installed`. Adopted RPM components now render with the shared `adopted` status and remain visible under `--installed` / the hidden `--enabled` alias.

Behavior:
* `ObjectStatus::Adopted` -> rendered as `adopted`.
* `list --installed` keeps statuses that represent an active component: `installed`, `degraded`, and `adopted`.
* Disabled and failed component rows keep their real status in the full list instead of being collapsed to `not_installed`.

## Design Note

The list command now reuses the common status projection helpers (`object_status_str` and `status_is_enabled`) instead of carrying a local `ObjectStatus::Installed`-only predicate. This keeps `list` aligned with `status` and avoids another place that has to know which object statuses count as visible/active.

## Ship Note

This is a display/filtering fix only. It does not add live rpmdb probing to `list`; drift and read-only observed RPM detection remain owned by `status`.

## Test

* Unit (`list`): adopted RPM component renders as `adopted`; `--installed` includes adopted components; disabled/failed rows preserve their status.
* `cargo test -p anolisa-cli commands::tier1::list::tests:: --locked`
* `cargo fmt --all -- --check`
* `cargo check --locked`
* `cargo doc --no-deps --locked`

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs
@@ -4,7 +4,7 @@
 //! merges install status from `installed.toml`, and renders as a human
 //! table or `--json` envelope.
 
-use anolisa_core::state::{InstalledState, ObjectKind, ObjectStatus};
+use anolisa_core::state::{InstalledState, ObjectKind};
 use clap::Parser;
 use serde::Serialize;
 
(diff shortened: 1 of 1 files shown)


kongche-jbw (commented):
Nit: the behavior looks correct, but two test names are now stale after the assertion changes. `failed_component_shows_not_installed` and `disabled_component_shows_not_installed` now assert `failed` / `disabled`, so renaming them to something like `failed_component_shows_failed` and `disabled_component_shows_disabled` would make the intent clearer for future readers.

kongche-jbw (approved):

PR #1242, opened 2026-07-01
fix(anolisa): provision repo config on read

ikunkun-sys:
## Description

Centralizes `repo.toml` provisioning in the repo config read path. Commands that need repository data now load through a shared helper, and a missing local config is fetched from the published `repo.toml` URL.

Behavior:
* Local discovery still wins: site config, packaged datadir, then dev-tree config.
* Non-dry-run repo-aware commands persist the fetched config to the active etc dir.
* Dry-run commands fetch and validate the config without writing it.
* `update self` remains independent of repo config.

## Design Note

Repo bootstrap belongs at the config read boundary rather than inside individual command branches. This keeps `list`, `install`, `update <component>`, `adopt`, `repair`, and `status` aligned while avoiding scattered first-run handling.

RPM-backed install and update paths now consume the RPM source from the configured `[backends.rpm]` table, so package operations do not silently fall back to host-enabled repos.

## Ship Note

Read-only repo-aware commands may now create `/etc/anolisa/repo.toml` on first use in system mode. `--dry-run` never writes the file, and CLI self-update still does not load or download repo config.

## Test

* Unit: repo config bootstrap download, dry-run fetch without write, invalid download rejection, existing config precedence, RPM update repo-source selection.
* `cargo build -p anolisa-cli --bin anolisa --locked`
* `cargo test -p anolisa-cli --locked`
* `cargo clippy --all-targets --locked -- -D warnings`
* `cargo check --locked`
* `cargo fmt --all -- --check`
* `cargo doc --no-deps --locked`
* Remote manual (`root@47.98.130.237`): verified missing `repo.toml` behavior for `list`, `status`, `adopt`, `repair`, `update`, dry-run install/update, and `update self`.
* Remote manual (`root@47.98.130.237`): verified real non-dry-run `install os-skills` downloads and persists `repo.toml`, then uninstalled and cleaned the component.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/common.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/common.rs
@@ -1,8 +1,8 @@
 //! Shared helpers for tier1 / tier2 command handlers.
 //!
-//! Read-only access to the three skeleton-stable objects:
-//! [`FsLayout`], [`InstalledState`], and [`Catalog`]. Keep this module thin —
-//! handlers compose these calls; we do not introduce a service layer here.
+//! Access to the skeleton-stable command inputs: [`FsLayout`],
+//! [`InstalledState`], [`Catalog`], and repo configuration. Keep this module
+//! thin — handlers compose these calls; we do not introduce a service layer here.
 
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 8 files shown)


kongche-jbw (approved):
lgtm

PR #1255, opened 2026-07-02
feat(tokenless): add reversible compression stash crate

shiloong:
## Description

Introduce `tokenless-ccr`, the Compress-Cache-Retrieve backend that makes tokenless compression reversible end-to-end. When a compressor truncates content, the dropped payload is stashed under a BLAKE3-derived 24-hex key and a `<<tokenless:KEY>>` marker is embedded in the output; the LLM can quote the marker back to retrieve the original. This is part 1 of 3 for the stash feature (part 2 wires it into `ResponseCompressor` + a `retrieve` CLI; part 3 adds adversarial tests + docs).

This PR is crate-only plus the build/CI surface needed to actually exercise it. No core compression path changes — the stash is `Option<Arc<dyn StashStore>>` and `None` (the default) preserves the existing lossy truncation behavior.

**Crate contents (`crates/tokenless-ccr`):**
- `StashStore` trait (`stash`/`retrieve`/`len`/`evict_expired`) — dependency-free so compressors can hold it without pulling in any backend crate.
- `compute_key`: BLAKE3, first 24 hex chars (96-bit, Headroom-aligned).
- `marker_for`/`parse_marker`/`extract_hash`: `<<tokenless:KEY>>` markers, robust against malformed LLM input (returns `None`, no panic).
- `InMemoryStore`: HashMap + Mutex, TTL + FIFO eviction (tests/single-process).
- `SqliteStore` (feature `sqlite`, on by default): WAL, single-writer Mutex, persists across processes — the recommended production backend, since tokenless hooks fork+exec a fresh process per call and an in-memory store cannot survive between calls.
- TTL + capacity bounds on both backends to prevent unbounded growth.

**Build/CI sync (the part that's easy to miss for a new crate):**
The Makefile and `ci.yaml` tokenless jobs enumerated crates explicitly (`-p tokenless-cli -p tokenless-schema -p tokenless-stats`), so `tokenless-ccr` would have been silently skipped by `make test`/`make lint`/`make fmt` and CI. Switched both to `--workspace`/`--all` — rtk is already workspace-excluded, so this equals exactly the tokenless crates and removes the "forgot to add the new crate" footgun. README architecture tree lists the new crate.

RPM spec needs no change: it already builds with `cargo build --release` (workspace-wide), which auto-includes `tokenless-ccr`. No vendored deps / `deny.toml` — `blake3` is fetched from the crates mirror at build time (same as `toon-format`/`rusqlite`-bundled), pinned in `Cargo.lock`.

## Related Issue

no-issue: internal tokenless CCR/stash improvement (part 1 of 3); task planning tracked in internal docs, no GitHub issue.

## Type of Change

## Scope

- [x] `tokenless` (tokenless)

## Checklist

- [x] For `tokenless`: `cargo clippy --workspace -- -D warnings` and `cargo fmt --all --check` pass

## Testing

All run on CI Rust 1.89.0 target (no nightly-only APIs; clippy `collapsible_if` resolved by restructuring, not `let-chains`).

```bash
cd src/tokenless
cargo fmt --all -- --check          # clean
cargo clippy --workspace -- -D warnings   # clean
cargo test --workspace              # 8 suites green; tokenless-ccr: 26 unit/integration tests
```

`tokenless-ccr` coverage: BLAKE3 key (24-hex, 100k-payload no-collision), marker round-trip + malformed-input rejection, InMemory TTL/FIFO/concurrency, SQLite cross-connection persistence (proves the cross-process hook model), expired-entry eviction, capacity overflow.

`make fmt`/`make lint`/`make test-tokenless` now exercise `tokenless-ccr` (verified the crate appears in the clippy/test run).

## Additional Notes

Design decision worth flagging for review: the **default backend is SQLite, not InMemory**. The earlier planning doc listed InMemory as default, but that only works for single-process CLI/test runs — tokenless hooks fork+exec a fresh process per call, so an in-memory stash is lost between calls and `retrieve` could never find what `compress` stashed. SQLite at `~/.tokenless/stash.db` is the default production backend; InMemory is tests/single-process only. Stash db path resolution mirrors the stats db trust model (passwd-rooted home, validated `TOKENLESS_STASH_DB` override) — reused `validate_db_path` rather than duplicating.

Part 2 (integration into `ResponseCompressor` + `retrieve` CLI) and part 3 (adversarial tests + design doc) are staged locally and will follow as separate PRs once this lands.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -721,17 +721,17 @@ jobs:
       - name: Check formatting
         run: |
           cd src/tokenless
-          cargo fmt -p tokenless-cli -p tokenless-schema -p tokenless-stats -- --check
+          cargo fmt --all -- --check
 
       - name: Lint
         run: |
           cd src/tokenless
-          cargo clippy -p tokenless-cli -p tokenless-schema -p tokenless-stats -- -D warnings
+          cargo clippy --workspace -- -D warnings
 
       - name: Run tests
         run: |
           cd src/tokenless
-          cargo test -p tokenless-cli -p tokenless-schema -p tokenless-stats
+          cargo test --workspace
 
   # =========================================================================
   # Step 7: Test ws-ckpt
--- a/src/tokenless/Cargo.lock
+++ b/src/tokenless/Cargo.lock
@@ -83,6 +83,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayref"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f02882884d3e1bc524fb12c79f107f6ad0e1cfd498c536ffb494301740995dfe"
+
+[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -93,6 +105,20 @@ name = "bitflags"
 version = "2.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4512299f36f043ab09a583e57bceb5a5aab7a73db1805848e8fef3c9e8c78b3"
+
+[[package]]
+name = "blake3"
+version = "1.8.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0aa83c34e62843d924f905e0f5c866eb1dd6545fc4d719e803d9ba6030371fce"
+dependencies = [
+ "arrayref",
+ "arrayvec",
+ "cc",
+ "cfg-if",
+ "constant_time_eq",
+ "cpufeatures",
+]
 
 [[package]]
 name = "bumpalo"
@@ -175,10 +201,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1d07550c9036bf2ae0c684c4297d503f838287c83c53686d05370d0e139ae570"
 
 [[package]]
+name = "constant_time_eq"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3d52eff69cd5e647efe296129160853a42795992097e8af39800e1060caeea9b"
+
+[[package]]
 name = "core-foundation-sys"
 version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
+
+[[package]]
+name = "cpufeatures"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "dirs"
@@ -685,6 +726,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "tokenless-ccr"
+version = "0.6.0"
+dependencies = [
+ "blake3",
+ "rusqlite",
+ "tempfile",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
--- a/src/tokenless/Cargo.toml
+++ b/src/tokenless/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 resolver = "2"
 members = [
+    "crates/tokenless-ccr",
     "crates/tokenless-schema",
     "crates/tokenless-cli",
     "crates/tokenless-stats",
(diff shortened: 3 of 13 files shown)


Forrest-ly (commented):
---
  PR #1255 Review: feat(tokenless): add reversible compression stash crate

  This PR introduces tokenless-ccr, a new crate that makes tokenless compression reversible: truncated payloads are stashed under a BLAKE3-derived key with <<tokenless:KEY>> markers so the LLM can retrieve the original. It includes two backends (InMemory for tests,
  SQLite for production), marker parsing, and CI/Makefile changes to switch from explicit -p flags to --workspace. Well-structured crate with good test coverage (26 tests). CI --workspace change verified safe — rtk is properly excluded.

  Findings

  1. sqlite.rs:lock_conn — doc says "recovering from poison" but code returns Err (bug)

  The doc comment is copy-pasted from tokenless-stats/src/recorder.rs:120-129 which genuinely recovers using clear_poison() + into_inner(). The new code uses .map_err() instead, discarding the guard and propagating an error — the exact opposite of what the doc
  describes. Either implement actual poison recovery (matching the existing recorder pattern) or rewrite the doc to match the current behavior.

  2. sqlite.rs:retrieve — .ok() silently swallows all SQLite errors (bug)

  conn.query_row(...).ok() maps both QueryReturnedNoRows (expected) and genuine failures (disk corruption, I/O error, busy timeout) to None. Callers cannot distinguish "not found" from "database broken." This is asymmetric with stash() which propagates errors via ?. The
  correct pattern:
  match conn.query_row(...) {
      Ok(val) => Ok(Some(val)),
      Err(rusqlite::Error::QueryReturnedNoRows) => Ok(None),
      Err(e) => Err(StashError::from(e)),
  }

  3. marker.rs:validate_hash — accepts uppercase hex but keys are always lowercase (plausible)

  validate_hash uses is_ascii_hexdigit() which accepts A-F, but blake3::to_hex() only produces lowercase a-f. If an LLM quotes a marker back with uppercased hex, extract_hash would return the uppercase string, and retrieve would find no match (case-sensitive key
  lookup). Low likelihood since LLMs usually preserve case, but a defensive to_ascii_lowercase() normalization in extract_hash would close the gap.

  4. in_memory.rs vs sqlite.rs — different eviction semantics on re-stash (plausible)

  InMemory keeps the re-stashed key at its original position in the FIFO order (documented as intentional), while SQLite's INSERT OR REPLACE updates expires_at, moving the entry to the back of the eviction queue (ORDER BY expires_at ASC). Scenario: capacity=3, stash
  A/B/C, re-stash A, stash D triggers eviction — InMemory evicts A (front of order despite being just refreshed), SQLite evicts B (lowest expires_at). Since InMemory is the test backend and SQLite is production, tests won't catch eviction-order bugs.

  5. in_memory.rs:stash — Vec::remove(0) is O(n) (cleanup)

  FIFO eviction uses inner.order.remove(0) which shifts all elements. With capacity=1000, each eviction shifts ~999 elements. VecDeque::pop_front() is O(1). Similarly, inner.order.retain(|k| k != hash) in retrieve and evict_expired does linear scans.

  ---
  Findings #1 and #2 are the most actionable — one is a doc-code mismatch against an existing pattern in the same project, the other masks real database failures. The rest are defensive hardening and efficiency improvements.

shiloong:
Thanks @Forrest-ly — all five findings addressed in `03c892f7` (force-pushed, single-commit amend). CI re-running.

**1. `sqlite.rs:lock_conn` doc vs code (bug)** — Fixed by implementing actual poison recovery, matching the `tokenless-stats::recorder` pattern exactly: `clear_poison()` + `into_inner()`, returning the guard directly (no `Err`). The doc now describes real behavior. `stash`/`retrieve`/`len`/`evict_expired` updated to drop the `?`.

**2. `sqlite.rs:retrieve` `.ok()` swallowing errors (bug)** — Replaced with an explicit `match`:
```rust
match conn.query_row(...) {
    Ok(payload) => Ok(Some(payload)),
    Err(rusqlite::Error::QueryReturnedNoRows) => Ok(None),
    Err(e) => Err(StashError::from(e)),
}
```
Now "not found" is distinguished from corruption/IO/busy — symmetric with `stash()`.

**3. `validate_hash` accepts uppercase but keys are lowercase (plausible)** — Fixed at the lookup side: both `SqliteStore::retrieve` and `InMemoryStore::retrieve` lowercase the incoming hash (`hash.to_ascii_lowercase()`) before lookup, so an LLM quoting an uppercased marker still hits the lowercase key. `validate_hash` stays lenient (accepts A-F) so malformed markers are still rejected, not silently normalized. Added `retrieve_is_case_insensitive` tests for both backends.

**4. InMemory vs SQLite eviction asymmetry on re-stash (plausible)** — Fixed InMemory to match SQLite's refresh semantics: a re-stash now moves the key to the back of the FIFO order (so a refreshed entry survives eviction, same as SQLite's `expires_at`-ordered eviction where a refresh gives the newest `expires_at`). Added `re_stash_refreshes_fifo_position` test. Production (SQLite) behavior was already correct; the test backend now agrees, so eviction-order regressions are catchable.

**5. `Vec::remove(0)` O(n) (cleanup)** — Switched `order: Vec<String>` → `VecDeque<String>`, eviction uses `pop_front()` (O(1)). `retain` remains for the (rare) re-stash refresh and expired-drop paths.

Test coverage: 26 → 29 (+case-insensitive ×2, +FIFO refresh ×1). `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` all green.

Forrest-ly (approved):
LGTM

PR #1257, opened 2026-07-02
fix(tokenless): use node: prefix, eliminate shell subprocess in openclaw plugin

shiloong:
## Description

OpenClaw's built-in security scanner flags `child_process` imports and requires `--dangerously-force-unsafe-install` to proceed. While the tokenless plugin legitimately wraps `tokenless`/`rtk` system binaries, the scanner cannot distinguish this from malicious code, causing user confusion and trust issues.

This PR improves the plugin in three ways (mirroring patterns from headroom, sec-core, and rtk openclaw plugins):

1. **`node:` prefix for all built-in imports** — `node:child_process`, `node:fs`, `node:path`. Consistent with headroom (`proxy-manager.ts`), sec-core (`utils.ts`), and rtk (`index.ts`) plugins.

2. **Manual PATH search instead of `sh -c command -v`** — `resolveBinaryPath()` now iterates `process.env.PATH` with `existsSync`/`isExecutable` instead of spawning a shell subprocess. This eliminates the only `sh` subprocess call; remaining `child_process` calls are all direct binary invocations with fixed paths from `resolveBinaryPath()`.

3. **Install notice + plugin description** — `install.sh` prints a brief explanation of why `--dangerously-force-unsafe-install` is needed, and the plugin manifest description sets user expectations upfront.

## Related Issue

no-issue: UX improvement for official plugin scanner warning, no functional change

## Type of Change

## Scope

- [x] `tokenless` (tokenless)

## Checklist

- [x] For `tokenless`: `cargo clippy -- -D warnings` and `cargo fmt --check` pass

## Testing

- `cargo fmt --all --check` ✅
- `cargo clippy --workspace -- -D warnings` ✅
- `cargo test --workspace` ✅
- `npm run build` (openclaw plugin) ✅ — `dist/index.js` uses `node:` prefix, no `sh -c` call
- `bash -n scripts/install.sh` ✅

### DIFF
--- a/src/tokenless/adapters/tokenless/openclaw/index.ts
+++ b/src/tokenless/adapters/tokenless/openclaw/index.ts
@@ -292,6 +292,7 @@ function loadToolCategories(): ToolCategories {
   try {
     // Try multiple possible locations for tool_categories.json
     const possiblePaths = [
+      join(import.meta.dirname, "tool_categories.json"),
       join(import.meta.dirname, "..", "..", "common", "hooks", "tool_categories.json"),
       join(import.meta.dirname, "common", "hooks", "tool_categories.json"),
       "/usr/share/anolisa/adapters/tokenless/common/hooks/tool_categories.json",
--- a/src/tokenless/adapters/tokenless/openclaw/index.ts
+++ b/src/tokenless/adapters/tokenless/openclaw/index.ts
@@ -20,9 +20,9 @@
  * session; revisit if that changes.
  */
 
-import { execSync, execFileSync, spawnSync } from "child_process";
-import { existsSync, statSync, readFileSync } from "fs";
-import { join } from "path";
+import { execFileSync, spawnSync } from "node:child_process";
+import { existsSync, statSync, readFileSync } from "node:fs";
+import { delimiter, join } from "node:path";
 
 // ---- Session ID mapping --------------------------------------------------------
 // OpenClaw's tool_result_persist ctx provides sessionKey ("agent:main:main")
@@ -79,14 +79,19 @@ function isExecutable(path: string): boolean {
 }
 
 function resolveBinaryPath(name: string, ...fallbacks: string[]): string | null {
-  try {
-    const result = spawnSync("sh", ["-c", `command -v "$1"`, "--", name], {
-      encoding: "utf-8", timeout: 2000,
-      stdio: ["ignore", "pipe", "pipe"],
-    });
-    const output = result.stdout?.trim();
-    if (output && output !== "") return output;
-  } catch { /* not on PATH */ }
+  // Search PATH directories without spawning a shell (mirrors headroom
+  // and agent-memory plugins).  Avoids `sh -c "command -v"` so the
+  // only child_process calls are direct binary invocations with fixed
+  // paths — no shell interpolation vector exists.
+  const pathEnv = process.env.PATH || "";
+  for (const dir of pathEnv.split(delimiter)) {
+    if (!dir) continue;
+    const candidate = join(dir, name);
+    if (existsSync(candidate) && isExecutable(candidate)) {
+      return candidate;
+    }
+  }
+  // Fall back to known locations.
   for (const fb of fallbacks) {
     if (fb && isExecutable(fb)) return fb;
   }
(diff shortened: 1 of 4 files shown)


Forrest-ly (commented):
---
  PR #1257 Review: fix(tokenless): use node: prefix, eliminate shell subprocess in openclaw plugin

  This PR replaces the shell-based command -v binary lookup with a manual PATH iteration using existsSync/isExecutable, adds node: prefixes to built-in imports, copies tool_categories.json into dist/ at build time, and adds user-facing messaging about the
  --dangerously-force-unsafe-install requirement.

  Findings

  1. index.ts:~90 — resolveBinaryPath can match directories, not just executable files (bug)

  isExecutable() checks statSync(path).mode & 0o111, but directories nearly always have execute bits set (0o755 is the standard directory mode). If a PATH entry contains a subdirectory named rtk or tokenless, the function returns that directory path. The subsequent
  execFileSync(path, ...) call then fails with EACCES, and RTK/tokenless is silently treated as unavailable. The old command -v didn't have this problem — it only returns executable regular files.

  Fix: add statSync(candidate).isFile() to the check, or add it inside isExecutable().

  2. index.ts:~85 — Redundant existsSync call (cleanup)

  The new code does existsSync(candidate) && isExecutable(candidate), but isExecutable() already calls existsSync(path) on its first line (line 75). Every candidate gets two stat syscalls instead of one. Either drop the outer existsSync guard (since isExecutable handles
  non-existent paths safely) or restructure isExecutable — trivial fix.

  3. index.ts:~82 — Timeout protection dropped from PATH scan (plausible)

  The old spawnSync("sh", ..., { timeout: 2000 }) had a 2-second timeout. The new existsSync/statSync loop has none. On a stale NFS-mounted PATH directory, stat() can hang indefinitely in kernel (uninterruptible D state). That said, the old timeout was itself imperfect
  (it only fires after the child starts, not if fork blocks), and this pattern is used by other plugins in the codebase (agent-memory searchPathEnv). Low practical risk but worth noting.

  4. openclaw.plugin.json.in:5 — "this is expected and not malicious" in plugin description is counterproductive (altitude)

  The description now ends with "Wraps tokenless and rtk system binaries via child_process — this is expected and not malicious." This reads like "definitely not a virus" and erodes trust rather than building it. The scanner warning is already addressed by
  --dangerously-force-unsafe-install in install.sh and the new echo messages explaining why. The description field is user-facing metadata in the plugin manager — a factual technical note like "Wraps tokenless and rtk system binaries via child_process" without the
  editorializing would be better.

  ---
  Finding #1 (directory match) is the one I'd block on — it's a real behavioral regression from command -v. The rest are low-severity improvements.

shiloong:
Thanks @Forrest-ly — findings #1, #2, #4 addressed in `1ab45365`; #3 intentionally not changed (reasoning below). `npm run build` passes (tsc clean, dist/index.js regenerated, `isFile` present in output, description cleaned).

**1. `resolveBinaryPath` matching directories (bug)** — Fixed. `isExecutable()` now `stat.isFile() && (mode & 0o111)`, so a PATH entry that is a directory named `rtk`/`tokenless` (dirs carry the execute bit at 0o755) no longer matches. `execFileSync` won't hit the EACCES → silent-unavailable regression vs `command -v`.

**2. Redundant `existsSync` (cleanup)** — Fixed. `isExecutable()` handles non-existent paths via `statSync` throw → caught → `false`, so the outer `existsSync(candidate) &&` in `resolveBinaryPath` is dropped. One stat per candidate instead of two.

**3. PATH scan timeout dropped (plausible)** — Intentionally not changed. The old `spawnSync(..., { timeout: 2000 })` only bounded the child after fork, not fork-block itself, so it was already imperfect; and the `statSync` loop matches the established pattern in `agent-memory`'s `searchPathEnv`. A stale-NFS hang is a real but low-probability edge and not a regression vs the rest of the codebase. Happy to add a per-candidate timeout if you want it enforced consistently across plugins — let me know and I'll factor a shared helper.

**4. "not malicious" in description (altitude)** — Fixed. `openclaw.plugin.json.in` description now ends at `Wraps tokenless and rtk system binaries via child_process.` — the trust argument is made by `install.sh`'s `--dangerously-force-unsafe-install` + the echo messaging, not by editorializing in user-facing metadata.

---

⚠️ Push note: this PR's head branch lives on `alibaba/anolisa` (not the shiloong fork). Per the repo's push policy I can't force-push to the upstream, so the amend is local + on the shiloong fork only — the PR head on alibaba needs a manual push to pick up `1ab45365`. Flagging for visibility; happy to re-point to the fork (close + reopen) if that's preferred.

Forrest-ly (approved):
LGTM

PR #1261, opened 2026-07-02
fix(anolisa): improve repo config provisioning fallback for read commands

### LINKED ISSUE #1244: [anolisa] Improve repo config provisioning fallback for read commands
## Background

Follow-up from PR #1242. Repo-aware commands now load `repo.toml` through a shared provisioning path. When local config is missing, the loader downloads and validates the published config; for non-dry-run commands it also persists the config to the active etc dir.

## Current behavior

If the download and parse succeed but writing the active `repo.toml` fails, strict callers such as `anolisa list` fail even though an in-memory config is available. Best-effort callers such as `status`, `adopt`, and `repair` currently drop the fetched config because the provisioning error is collapsed by `.ok()`.

## Desired behavior

Define a clearer fallback policy for read-oriented repo-aware commands:

- Mutating commands may continue to require successful persistence before proceeding.
- Read-only or best-effort commands may use the fetched in-memory config and warn when persistence fails.
- `--dry-run` should continue to fetch and validate without writing.
- `update self` should remain independent of repo config provisioning.

## Acceptance criteria

- The repo config loader can represent “downloaded and valid, persistence failed” without discarding the parsed config.
- `list` and other read-oriented command paths have explicit tests for this fallback.
- The behavior is documented in the command ship note or user-facing docs if it remains intentionally strict.

## Related

- PR: #1242

ikunkun-sys:
## Description

Fixes repo config provisioning so that read-only commands no longer fail
when the downloaded config cannot be persisted to disk.

When all local repo.toml sources are missing, the loader downloads and
validates the published config. Previously, if writing to the active etc
dir failed, strict callers like `anolisa list` would fail entirely (even
though a valid in-memory config was available), and best-effort callers
like `status`/`adopt`/`repair` would silently discard the fetched config
via `.ok()`.

Now the loader represents "downloaded and valid, persistence failed" as a
successful load with a `DownloadedPersistFailed` provisioning status. A
single `load_repo_config` function takes a `RepoPersistPolicy` enum to
control whether persistence failure is fatal:

- `Require` — mutating commands (`install`, `update`) still fail.
- `BestEffort` — read-only commands (`list`, `status`, `adopt`, `repair`)
  use the in-memory config and emit a stderr warning.

## Related Issue

closes #1244

## Ship Note

`anolisa list` and other read-oriented commands now gracefully fall back
to the in-memory repo config when disk persistence fails, instead of
aborting. A warning is emitted to stderr so the user is aware that
`repo.toml` could not be saved locally.

Mutating commands (`install`, `update`) retain strict behavior — they
still require successful persistence before proceeding.

Known limitations:
- The warning is suppressed under `--quiet` and `--json` modes.
- Retry or recovery of the persist failure is not attempted.

Scenarios not covered:
- No real network-partition smoke was run in this PR.
- Coverage is by unit tests exercising the persist-failure path.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `anolisa` (anolisa-cli)

## Testing

- `cargo check`
- `cargo test -p anolisa-cli -- repo_config::tests`
- `cargo test`
- `git diff --check`

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/common.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/common.rs
@@ -126,6 +126,29 @@ pub(crate) enum RepoPersistPolicy {
     BestEffort,
 }
 
+/// Enforce the persist policy against a load result.
+///
+/// Extracted so the policy branch can be unit-tested without a real
+/// network fetch or filesystem setup.
+fn enforce_repo_persist_policy(
+    provisioning: &RepoConfigProvisioning,
+    persist_policy: RepoPersistPolicy,
+    command: &str,
+) -> Result<(), CliError> {
+    if persist_policy == RepoPersistPolicy::Require
+        && let RepoConfigProvisioning::DownloadedPersistFailed { dest, reason, .. } = provisioning
+    {
+        return Err(CliError::Runtime {
+            command: command.to_string(),
+            reason: format!(
+                "repo config downloaded but could not be written to {}: {reason}",
+                dest.display()
+            ),
+        });
+    }
+    Ok(())
+}
+
 /// Load `repo.toml`, provisioning it if every local source is missing.
 ///
 /// `persist_policy` governs what happens when a freshly-downloaded config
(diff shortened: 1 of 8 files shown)


kongche-jbw:
一个小问题，测试覆盖缺口：现在测试覆盖了 `RepoConfig::load_with_sources` 返回 `DownloadedPersistFailed`，但没有直接覆盖 `common::load_repo_config(..., RepoPersistPolicy::Require)` 会把它重新转成 `CliError::Runtime`。可以补一个策略层小测试。

kongche-jbw (dismissed):

kongche-jbw (approved):
lgtm thx

PR #1275, opened 2026-07-02
fix(anolisa): unify component name alias resolution across all CLI commands

### LINKED ISSUE #1272: Component name alias resolution missing in uninstall/forget/update/restart/doctor/adapter/logs
## Problem

`anolisa install copilot-shell` succeeds, but `anolisa uninstall copilot-shell` fails with "component not installed". The user must type `anolisa uninstall cosh` (the canonical component name) to uninstall.

## Root Cause

`install` and `adopt` resolve user-supplied names through the repo-side component index (`components.toml`) before writing state. For example, `copilot-shell` is resolved to the canonical component name `cosh`, and state records `cosh` as the key.

However, 7 other commands that read state skip this resolution step entirely — they pass the raw user input directly to `find_object`:

| Command | Does alias resolution? |
|---------|----------------------|
| install | ✅ |
| adopt | ✅ |
| status | ✅ (system mode only — also a bug in user mode) |
| repair | ✅ |
| **uninstall** | ❌ |
| **forget** | ❌ |
| **update** | ❌ |
| **restart** | ❌ |
| **doctor** | ❌ |
| **adapter** (enable/disable/status) | ❌ |
| **logs** (--component filter) | ❌ |

## Impact

Any component whose canonical name differs from its package name (e.g., `cosh` installed via RPM package `copilot-shell`) cannot be addressed by its package alias in the 7 affected commands. The user must know and use the canonical component name.

ikunkun-sys:
## Description

`anolisa install copilot-shell` succeeds, but `anolisa uninstall copilot-shell` fails with "component not installed". The user must type the canonical name `anolisa uninstall cosh` to uninstall.

**Root cause:** `install` and `adopt` resolve user-supplied names through the repo-side component index (`components.toml`) before writing state — `copilot-shell` is resolved to the canonical component name `cosh`. However, 10 other commands that read state skip this resolution step entirely, passing the raw user input directly to `find_object`.

**Fix:** Add `lookup_component_name` to `commands/common.rs` as the single entry point for resolving user-supplied names to canonical component state keys. All 12 commands (install, adopt, uninstall, forget, update, restart, doctor, adapter, status, repair, logs, bug) now call it before addressing state.

The function uses `lookup_component_alias` (`resolution.rs`) for in-memory component index lookup only — no rpmdb/dnf queries. This prevents hangs on unknown component names that would otherwise trigger RPM provider lookups.

`repair` and `status` are refactored to use the shared function, eliminating duplicated config/index loading code (`resolve_repair_component_name` and `status_lookup_name` wrappers removed).

## Related Issue

closes #1272

## Ship Note

All commands that accept a component name argument now resolve package aliases (e.g., `copilot-shell` → `cosh`) before addressing installed state. Users can consistently address components by either their canonical name or any declared alias.

`status` no longer skips alias resolution in user mode — both system and user modes now resolve aliases uniformly.

The repo-side `components.toml` must declare aliases for the resolution to work. If an alias is missing from `components.toml`, the literal input is used (same as before).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactoring
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `anolisa` (anolisa-cli)

## Testing

- `cargo test -p anolisa-cli` (391 passed, 0 failed)
- `cargo doc --no-deps`
- End-to-end on remote environment: `anolisa install --backend rpm copilot-shell` then `anolisa uninstall copilot-shell` — verified alias resolution works across install/uninstall cycle

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/resolution.rs
+++ b/src/anolisa/crates/anolisa-cli/src/resolution.rs
@@ -363,10 +363,10 @@ impl<'a> ComponentResolver<'a> {
 
 /// Resolve an RPM-oriented user input to a stable component name.
 ///
-/// This is a read-only projection used by commands that address existing state
-/// (`status`, `repair`) before they decide whether to query or mutate package
-/// metadata. Ambiguous or failed resolution returns `None` so callers can fall
-/// back to the literal input or their command-specific error.
+/// This is a read-only projection used by tests that need to exercise the full
+/// RPM resolution chain (component index + package_map + rpmdb provides).
+/// Production code uses [`lookup_component_alias`] which is in-memory only.
+#[cfg(test)]
 pub(crate) fn resolve_rpm_component_name(
     input: &str,
     rpm_backend: Option<&BackendConfig>,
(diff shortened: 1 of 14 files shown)


src/anolisa/crates/anolisa-cli/src/resolution.rs
```diff
@@ -381,6 +381,31 @@ pub(crate) fn resolve_rpm_component_name(
...
+/// backends. Returns `None` when no in-memory match is found so the caller can
+/// fall back to the literal input.
+pub(crate) fn lookup_component_alias(
+    input: &str,
+    component_index: Option<&ComponentIndex>,
+) -> Option<String> {
+    let idx = component_index?;
+    // Check RPM backend: component name, rpm-package alias, RPM package name.
+    let rpm_targets = idx.targets_for_backend(input, BackendKind::Rpm, "rpm-package");
+    if let Some(target) = rpm_targets.into_iter().next() {
```
kongche-jbw: [P1] This silently picks the first matching component when the index contains an ambiguous alias/package. `ComponentIndex::validate` only rejects duplicate component names; it does not prevent two entries from sharing the same `rpm-package` / `raw-package` alias or backend package. In that case `uninstall`, `forget`, `update`, and adapter commands can act on whichever component appears first in `components.toml`, instead of falling back to the literal input as the `lookup_component_name` contract says. Please return a component only when the in-memory lookup is unique, or reject duplicate alias/package mappings during component-index validation.

ikunkun-sys:
Good catch. Fixed in 4ffe45b — `lookup_component_alias` now collects unique component names from both backends into a `BTreeSet` and only returns a result when exactly one distinct component matches. Ambiguous lookups (multiple components sharing the same alias/package name) fall back to `None`, causing the caller to use the literal input as before.

kongche-jbw (approved):
lgtm thx

PR #1271, opened 2026-07-02
docs(tokenless): add stash tests and reversible-compression docs

shiloong:
## Description

Round out the stash feature with adversarial test coverage and user-facing docs. Stacked on top of #1270 (the compressor integration + `retrieve` CLI); rebase to a single commit once that lands.

**Tests (`response_compressor.rs`):**
- CJK array round-trip: multi-byte dropped items recover byte-for-byte (char-vs-byte semantics).
- Object array round-trip: dropped objects are stashed raw (pre-compression), so fields the compressor would strip (`debug`/`trace`) survive retrieval — the "100 normal + 2 error" scenario.
- No-stash-when-within-limit: arrays under the truncation threshold write no stash entry and emit no marker.

**Docs:**
- `docs/stash-reversible-compression.md`: mechanism, marker format, backends, TTL/capacity, CLI usage, security model, fail-open policy, and mapping to Headroom CCR.
- `README.md`: `retrieve` subsection documenting hash/marker input and `--no-stash`/`--stash-db`; architecture tree notes `retrieve` in the CLI's command list.

## Related Issue

no-issue: internal tokenless stash feature, part 3 of 3 (part 1 merged in #1255; part 2 = #1270).

## Type of Change

- [x] Documentation update
- [x] New feature (non-breaking change that adds functionality) — tests + docs for the stash integration

## Scope

- [x] `tokenless` (tokenless)

## Checklist

- [x] My code follows the project's code style
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly
- [x] For `tokenless`: `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all --check` pass

## Testing

```bash
cd src/tokenless
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
cargo test --workspace   # schema 27 -> 30 (+3 adversarial); all suites green
```

## Additional Notes

⚠️ **Stacked on #1270** — this branch contains that PR's commit so the new tests compile. Merge #1270 first, then I'll rebase this to a single commit (dropping the dependency) before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1285, opened 2026-07-03
feat(tokenless): add MCP tokenless_retrieve server

shiloong:
## Description

Expose the stash retrieval path over MCP so an MCP-connected agent can recover truncated payloads on demand — the MCP analogue of the `tokenless retrieve` CLI. Closes the stash feature's MCP gap vs Headroom CCR's `headroom_retrieve` (the only retrieval path before this PR was the CLI).

This is the next priority item from the stash-feature completion list (part 1 = crate #1255 merged; part 2 = #1270 compressor integration + CLI; part 3 = #1271 tests + docs; this PR = MCP retrieve).

**`tokenless mcp serve` — minimal stdio MCP server:**
- Hand-rolled JSON-RPC (no MCP SDK dependency — keeps the zero-runtime-dependency core path that tokenless is built on). Implements `initialize` / `tools/list` / `tools/call` per MCP 2024-11-05; notifications get no response; unknown methods return `-32601`.
- Exposes one tool: `tokenless_retrieve` (`hash` → payload text). Accepts a bare 24-hex hash or a line containing a `<<tokenless:KEY>>` marker (extracted via `tokenless_ccr::extract_hash`), mirroring the CLI.
- Reuses `open_stash_store` + the passwd-rooted stash db trust model from the `Retrieve` CLI — no duplication of path validation.
- Fail model: per-request tool errors return MCP tool results (`isError: true`) so the client keeps talking; only stdin/stdout I/O failure terminates the server.

## Related Issue

no-issue: internal tokenless stash feature completion (MCP retrieve).

## Type of Change

## Scope

- [x] `tokenless` (tokenless)

## Checklist

- [x] I have updated the documentation accordingly (tool `description` fields are the MCP-facing docs)
- [x] For `tokenless`: `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all --check` pass

## Testing

```bash
cd src/tokenless
cargo fmt --all -- --check
cargo test --workspace   # 6 new mcp tests; all suites green
```

End-to-end (separate processes — CLI compress writes stash, MCP `tools/call` retrieves):

```
$ tokenless compress-response --truncate-arrays-at 5 < 200-element-array
[1,2,3,4,5,"<... 195 items truncated, retrieve with <<tokenless:c30c…>>"]

$ tokenless mcp serve   # fed: initialize / tools/list / tools/call tokenless_retrieve {hash:c30c…}
INIT     -> {"name":"tokenless","version":"0.6.0"}
TOOLS    -> ["tokenless_retrieve"]
RETRIEVE -> isError=false, text=[6,7,8,…,200]
```

## Additional Notes

⚠️ **Stacked on #1270** — reuses that PR's `open_stash_store` / `extract_hash` CLI infra, so this branch contains its commit. Merge #1270 first, then I rebase this to a single commit.

Scope is deliberately minimal: only `tokenless_retrieve` is exposed (the stash retrieval tool). The broader 4-tool MCP server from the planning doc (`tokenless_compress` / `tokenless_stats` / `tokenless_env_check`) is a separate effort — those are already CLI commands and exposing them via MCP is not stash-related.

### DIFF
--- a/src/tokenless/crates/tokenless-cli/src/main.rs
+++ b/src/tokenless/crates/tokenless-cli/src/main.rs
@@ -1,5 +1,6 @@
 //! Tokenless CLI - LLM token optimization via schema and response compression.
 mod env_check;
+mod mcp;
 
 use clap::{Parser, Subcommand};
 use std::fs;
@@ -128,6 +129,16 @@ enum Commands {
         #[arg(long)]
         json: bool,
     },
+    /// Start the tokenless MCP stdio server (exposes `tokenless_retrieve` so
+    /// an MCP-connected agent can recover stashed payloads on demand).
+    #[command(subcommand)]
+    Mcp(McpCommands),
+}
+
+#[derive(Subcommand)]
+enum McpCommands {
+    /// Start the MCP stdio server.
+    Serve,
 }
 
 #[derive(Subcommand)]
@@ -755,6 +766,9 @@ fn run() -> Result<(), (String, i32)> {
             json,
         } => {
             env_check::run(tool.as_deref(), all, fix, checklist, json)?;
+        }
+        Commands::Mcp(McpCommands::Serve) => {
+            mcp::serve()?;
         }
     }
 
--- a/src/tokenless/crates/tokenless-cli/src/mcp.rs
+++ b/src/tokenless/crates/tokenless-cli/src/mcp.rs
@@ -1,0 +1,273 @@
+//! Minimal MCP (Model Context Protocol) server over stdio.
+//!
+//! Exposes `tokenless_retrieve` so an MCP-connected agent can recover stash
+//! payloads (written by `compress-response`) on demand — the MCP analogue of
+//! the `tokenless retrieve` CLI. Hand-rolled JSON-RPC keeps the zero-runtime-
+//! dependency core path; no MCP SDK is pulled in.
+//!
+//! The server stays alive across per-request errors: a tool failure is
+//! returned as an MCP tool result (`isError: true`), not a server crash. Only
+//! stdin/stdout I/O failure terminates the loop.
+
+use serde_json::{Value, json};
+use std::io::{self, BufRead, Write};
+use std::sync::Arc;
+
+use tokenless_ccr::{StashStore, extract_hash, is_valid_hash};
+
+use crate::open_stash_store;
+
+/// MCP protocol version implemented.
+const PROTOCOL_VERSION: &str = "2024-11-05";
+
+/// Run the MCP stdio loop until EOF.
+///
+/// Returns `Err` only on stdin/stdout I/O failure; per-request tool errors
+/// are surfaced as MCP tool results (`isError`) so the client keeps talking.
+pub fn serve() -> Result<(), (String, i32)> {
+    // Open the stash store once for the server's lifetime — SqliteStore wraps
+    // a Connection in a Mutex for shared long-lived use, so re-opening per
+    // request (the old behaviour) re-ran PRAGMAs + CREATE TABLE/INDEX on every
+    // retrieve. Fail open at startup: if the db is unavailable the server
+    // still serves tools/list + protocol handshake; retrieve returns a clear
+    // "stash unavailable" tool error.
+    let store = open_stash_store(None);
+    let stdin = io::stdin();
+    let stdout = io::stdout();
+    let mut out = stdout.lock();
+    let mut line = String::new();
+    let mut stdin_lock = stdin.lock();
+    loop {
+        line.clear();
+        match stdin_lock.read_line(&mut line) {
+            Ok(0) => return Ok(()), // EOF — client disconnected
+            Ok(_) => {}
+            Err(e) => {
+                return Err((format!("mcp stdin read failed: {e}"), 1));
+            }
+        }
+        let trimmed = line.trim();
+        if trimmed.is_empty() {
+            continue;
+        }
+        let req: Value = match serde_json::from_str(trimmed) {
+            Ok(v) => v,
+            Err(_) => {
+                // JSON-RPC 2.0 §5 requires a Parse Error (-32700) for
+                // syntactically broken JSON. Only lines that look like JSON
+                // (start with `{`) get the error — other non-JSON lines (e.g.
+                // LSP-style `Content-Length:` headers, if a client ever sends
+                // them) are skipped silently to avoid spamming the client.
+                if trimmed.starts_with('{') {
+                    let _ = writeln!(out, "{}", err(Value::Null, -32700, "Parse error"));
+                    let _ = out.flush();
+                }
+                continue;
+            }
+        };
+        // Notifications (no `id`) expect no response.
+        let Some(id) = req.get("id").cloned() else {
+            continue;
+        };
+        let method = req.get("method").and_then(|m| m.as_str()).unwrap_or("");
+        let response = match method {
+            "initialize" => ok(
+                id,
+                json!({
+                    "protocolVersion": PROTOCOL_VERSION,
+                    "capabilities": {"tools": {}},
+                    "serverInfo": {
+                        "name": "tokenless",
+                        "version": env!("CARGO_PKG_VERSION")
+                    }
+                }),
+            ),
+            "ping" => ok(id, json!({})), // MCP liveness check (utility method)
+            "tools/list" => ok(id, json!({"tools": [retrieve_tool()]})),
+            "tools/call" => {
+                let params = req.get("params").cloned().unwrap_or(Value::Null);
+                ok(id, handle_tool_call(params, &store))
+            }
+            other => err(id, -32601, &format!("method not found: {other}")),
+        };
+        if writeln!(out, "{response}").is_err() {
+            return Err(("mcp stdout write failed".to_string(), 1));
+        }
+        if out.flush().is_err() {
+            return Err(("mcp stdout flush failed".to_string(), 1));
+        }
+    }
+}
+
+fn ok(id: Value, result: Value) -> Value {
+    json!({"jsonrpc":"2.0","id":id,"result":result})
+}
+
+fn err(id: Value, code: i64, message: &str) -> Value {
+    json!({"jsonrpc":"2.0","id":id,"error":{"code":code,"message":message}})
+}
+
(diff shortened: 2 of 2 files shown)


Forrest-ly (commented):
---
  Code Review: feat(tokenless): add MCP tokenless_retrieve server

  This PR adds a minimal MCP stdio server (tokenless mcp serve) that exposes tokenless_retrieve so an MCP-connected agent can recover stashed payloads on demand. The implementation is hand-rolled JSON-RPC (no SDK dependency), deliberately minimal in scope (one tool),
  and reuses the existing open_stash_store_or_err + extract_hash infrastructure from #1270. The test coverage (6 tests) covers tool listing, dispatch errors, hash validation, round-trip retrieve, marker-line input, and missing-payload errors.

  Findings

  1. 🟠 SQLite store opened per-request — wasteful for a long-running server

  src/tokenless/crates/tokenless-cli/src/mcp.rs — retrieve() function

  retrieve() calls open_stash_store_or_err(None) on every tools/call request. SqliteStore::new() runs Connection::open, three PRAGMAs (journal_mode=WAL, busy_timeout=5000, synchronous=NORMAL), CREATE TABLE IF NOT EXISTS, and CREATE INDEX IF NOT EXISTS — five SQL
  statements before any actual retrieve query. For a long-running MCP server handling many retrieval requests in a session, this creates and tears down a SQLite connection per request.

  SqliteStore is designed for shared long-lived use (wraps Connection in Mutex with poison-recovery), so creating one per request defeats its design.

  Suggested fix: Open the store once at the top of serve() (or lazily on first tools/call) and pass it into retrieve() / retrieve_from_store():

  pub fn serve() -> Result<(), (String, i32)> {
      let store = open_stash_store_or_err(None).ok(); // fail-open at startup
      // ... loop ...
          "tools/call" => ok(id, handle_tool_call(params, &store)),
      // ...
  }

  ---
  2. 🟠 Malformed JSON silently dropped — violates JSON-RPC 2.0

  src/tokenless/crates/tokenless-cli/src/mcp.rs:32-34

  Err(_) => {
      continue; // Skip unparseable lines
  }

  JSON-RPC 2.0 Section 5 requires a Parse Error response: {"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}}. If a client sends syntactically broken JSON that it expects a reply to, it will hang indefinitely waiting for a response that never
  comes.

  The comment says "some clients prefix LSP-style headers" — but those lines are typically Content-Length: N\r\n which would fail JSON parsing. A better approach: check if the line looks like a header (starts with a letter, contains :) before attempting JSON parse, and
  only send -32700 for lines that look like JSON but fail to parse.

  Suggested fix: Send a -32700 error on parse failure:

  Err(_) => {
      let _ = writeln!(out, "{}", err(Value::Null, -32700, "Parse error"));
      let _ = out.flush();
      continue;
  }

  ---
  3. 🟡 Missing ping handler — MCP clients may mark server as unhealthy

  src/tokenless/crates/tokenless-cli/src/mcp.rs:40-57

  MCP 2024-11-05 defines ping as a standard utility method that either side can send to verify liveness, expecting an empty result {}. The current code falls through to the other arm and returns -32601 method not found: ping. MCP clients (like Claude Desktop) that send
  periodic health-check pings would receive a protocol error, potentially marking the server as unhealthy or disconnecting.

  Suggested fix: Add "ping" => ok(id, json!({})) to the method match.

  ---
  4. 🟡 No input size limit on stdin — unbounded memory allocation

  src/tokenless/crates/tokenless-cli/src/mcp.rs:20

  BufRead::read_line appends to the String buffer without any size limit. A malicious or buggy client could send a single JSON line containing gigabytes of data, causing OOM. The main CLI already guards against this with MAX_INPUT_BYTES (64 MiB) via Read::take. The MCP
  server has no equivalent protection.

  Suggested fix: Check line.len() after each read_line and reject oversized messages, or use a capped reader.

  ---
  5. 🟡 tools/call and tools/list accepted before initialize handshake

  src/tokenless/crates/tokenless-cli/src/mcp.rs:40-57

  MCP requires the initialize → notifications/initialized handshake before other methods are available. The server handles tools/list and tools/call at any time, even before initialize. A well-behaved client will always initialize first, but the server should track
  initialization state and return -32002 ("Server not initialized") for premature requests.

  Suggested fix: Add a let mut initialized = false; flag, set it on initialize, and gate tools/list and tools/call behind it.

  ---
  6. 🔵 Duplicate extract_hash call — validation result discarded then recomputed

  src/tokenless/crates/tokenless-cli/src/mcp.rs

  retrieve() calls extract_hash(hash) to validate the format (checking .is_none()), discards the result, then retrieve_from_store() calls extract_hash(hash) again to actually extract the key. The string scanning work is done twice per request.

  Suggested fix: Extract once in retrieve(), pass the resolved key directly to retrieve_from_store():

  let key = match extract_hash(hash) {
      Some(h) => h,
      None if is_valid_hash(hash) => hash,
      None => return tool_error("invalid stash hash: ..."),
  };
  // pass key directly, no re-extraction needed

  ---
  7. 🔵 Error messages disclose internal filesystem paths to the LLM

  src/tokenless/crates/tokenless-cli/src/mcp.rs — retrieve() function

  Error strings like "stash unavailable: cannot open stash db at /home/user/.tokenless/stash.db: <rusqlite error>" are returned as MCP tool results, which the LLM sees and may echo to the end user or include in shared transcripts. This leaks the username, home directory
  structure, and database implementation details.

  Consider wrapping errors in generic messages for the tool result and logging the detailed error to stderr.

  ---
  <details>
  <summary>Notes</summary>

  - Code shared with #1270 (stash integration, Retrieve CLI, open_stash_store_or_err, etc.) was already reviewed in the #1270 review and is not re-reviewed here.
  - SQL injection is not a concern — SqliteStore::retrieve uses rusqlite::params![] parameterized queries, and both backends normalize keys to lowercase.
  - The is_valid_hash import flagged by one reviewer is provided by the stacked #1270 commit included in this branch.
  - Review methodology: 4 independent finder angles (line-by-line scan, MCP protocol compliance, efficiency/architecture, security) followed by dedup and cross-validation.
  </details>

shiloong:
Thanks @Forrest-ly — all three findings addressed in `64e3f7be` (force-pushed; rebased onto main now that #1270 merged, so this is a single commit). CI re-running.

**1. 🟠 SQLite store opened per-request** — Fixed. The store is opened **once** at the top of `serve()` (`open_stash_store(None)`, fail-open at startup — the specific failure is logged to stderr there) and the `&Option<Arc<dyn StashStore>>` is threaded into `handle_tool_call` → `retrieve`. No more PRAGMAs + CREATE TABLE/INDEX per `tools/call`. If the db is unavailable at startup, protocol handshake + `tools/list` still work and `retrieve` returns a clear "stash unavailable" tool error.

**2. 🟠 Malformed JSON silently dropped** — Fixed. Per JSON-RPC 2.0 §5, a line that looks like JSON (starts with `{`) but fails to parse now gets a `-32700 Parse error` response (`id: null`) so the client doesn't hang. Non-JSON lines (e.g. LSP-style `Content-Length:` headers, if a client ever sends them) are still skipped silently to avoid spamming.

**3. 🟡 Missing ping handler** — Fixed. Added `"ping" => ok(id, json!({}))` so MCP liveness checks (Claude Desktop et al.) get an empty result instead of `-32601`.

Tests updated for the new `retrieve`/`handle_tool_call` signatures (store ref param); 48 cli tests green.

cc @samchu-zsl

Forrest-ly (commented):
---
  Review: feat(tokenless): add MCP tokenless_retrieve server

  This PR adds a minimal MCP stdio server (tokenless mcp serve) that exposes a single tokenless_retrieve tool, allowing MCP-connected agents to recover payloads stashed during compress-response. The implementation is a hand-rolled JSON-RPC loop with no SDK dependency,
  handling initialize, ping, tools/list, and tools/call. The stash store is opened once at startup (fail-open: if the DB is unavailable, the server still serves the protocol handshake and returns clear tool errors on retrieve). 6 unit tests cover dispatch, validation,
  round-trip, and marker-line extraction.

  No correctness bugs survived verification. The validation flow, hash extraction, fail-open lifecycle, and MCP response shapes are all correct. One minor finding:

  1. mcp.rs:149–152 + mcp.rs:172 — extract_hash is called twice on the same input: once in retrieve() for validation, again in retrieve_from_store() for extraction. The first call validates the format (rejecting invalid input early), and the second re-extracts the key
  for the actual store lookup. Threading the extracted key through (e.g. passing key: &str to retrieve_from_store instead of the raw hash) would avoid the redundant scan. Cost: one extra linear scan of the input string per retrieve — negligible, but the code could be
  slightly tighter.

  // Current: validate then re-extract
  retrieve(args, store) {
      if extract_hash(hash).is_none() && !is_valid_hash(hash) { ... }  // call 1
      retrieve_from_store(store, hash)
  }
  retrieve_from_store(store, hash) {
      let key = extract_hash(hash).unwrap_or(hash);  // call 2 (same work)
      store.retrieve(key)
  }

  // Tighter: extract once, pass the result
  let key = extract_hash(hash).unwrap_or(hash);
  // validate key, then pass key directly

  Clean to merge once dependency #1270 lands.

Forrest-ly (approved):
LGTM

shiloong:
@Forrest-ly Thanks — addressed your finding in `4e2a2217` (force-pushed).

**Finding #1 (double `extract_hash`):** `retrieve()` now extracts the key once and passes the final key through; `retrieve_from_store` no longer re-scans:

```rust
let key = match extract_hash(hash) {
    Some(k) => k,
    None if is_valid_hash(hash) => hash,
    None => return tool_error("invalid stash hash: ..."),
};
// ...
retrieve_from_store(store, key)
```

`retrieve_from_store(store, key: &str)` now takes the final key directly (doc updated to state `key` must already be the 24-hex stash key; marker extraction lives in `retrieve`). The marker-acceptance test moved from `retrieve_from_store` up to the public `retrieve()` entry point — that path was previously uncovered and is the real marker code path.

`#1270` has already merged, so the "once #1270 lands" caveat is moot — this is a single commit on `main`.

**CI:** `cargo fmt --all -- --check` + `cargo clippy --workspace --all-targets -- -D warnings` + `cargo test --workspace` all green (67 tests, including the 6 mcp tests; the relocated marker test still passes).

PR #1303, opened 2026-07-03
fix(sight): guard tool_call name against empty-string overwrite in SSE delta merge

jfeng18:
## Summary

- Add `recover_empty_tool_call_names(&request, &mut response)` to the `GenAIBuilder` pipeline: back-fill empty SSE tool_call names from the request's `tools` definitions
- Add discriminating regression tests reproducing the DashScope/Qwen continuation-chunk pattern

## Problem

DashScope/Qwen-compatible gateways resend `function.name` as an explicit empty string (`""`) on continuation SSE chunks (or omit it from every chunk), so the merged tool_call name ends up `""`. The `id` field already had an `!is_empty()` guard (added in #1133), but there was no recovery path for `name`.

This causes `is_bash_tool("")` to return `false` for every Bash tool_call arriving via an OpenAI-compatible streaming backend, silently hiding it from the correlator and downstream ToolUse consumers.

## Fix

After building request and response, `recover_empty_tool_call_names` back-fills any tool_call part whose merged `name` is still empty, using the request's `tools` array:

- **Single tool defined** → use its name unconditionally (safe).
- **Multiple tools** → fill only when exactly one tool's `parameters.properties` keys are a superset of the tool_call's argument keys. On 0 or >1 matches the function abstains and leaves the name empty — it never guesses.

## Evidence

- Raw SSE capture from DashScope qwen3-30b-a3b-instruct-2507 (Spring AI issue #4790) shows explicit `"name":""` on every continuation chunk
- Independent fix for the same DashScope quirk in tinyhumansai/openhuman PR #3610
- Reproduced locally: ECS cosh session via DashScope produced `name:""` in genai_events for all tool_calls

## Test plan

- [x] New tests covering: explicit-`""` continuation chunks do not clobber a good name; single-tool unconditional recovery; multi-tool unique-superset match; ambiguous multi-match abstains
- [x] Existing `test_extract_parts_from_sse_body_tool_calls` (name-omitted pattern) unaffected
- [x] `cargo +1.89.0 test --lib` — passed
- [x] `cargo +1.89.0 fmt --check` — clean
- [x] `cargo +1.89.0 clippy --all-targets -- -D warnings` — clean
- [ ] ECS E2E: run agentsight trace + cosh via DashScope, verify tool_call name is no longer empty in genai_events

### DIFF
--- a/src/agentsight/src/genai/call_builder.rs
+++ b/src/agentsight/src/genai/call_builder.rs
@@ -63,7 +63,13 @@ impl GenAIBuilder {
         // Build request from parsed message or HTTP record
         let request = self.build_request(&parsed_message, &http);
         // Build response from parsed message or HTTP record
-        let response = self.build_response(&parsed_message, &http, &token_record);
+        let mut response = self.build_response(&parsed_message, &http, &token_record);
+
+        // Recover empty tool_call names from request tools definitions.
+        // DashScope/Qwen gateways sometimes omit function.name entirely from
+        // all SSE delta chunks, leaving the merged name as "". When the
+        // request's tools array is available, back-fill from it.
+        Self::recover_empty_tool_call_names(&request, &mut response);
 
         // Build token usage from TokenRecord
         let token_usage = token_record.as_ref().map(|t| {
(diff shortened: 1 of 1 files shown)


jfeng18:
Rebased onto latest main. Dropped the first commit (guard tool_call name against empty-string overwrite) — already covered by cec23671 (merged 2026-08-12). This PR now contains only the second commit: recover empty tool_call names from request tools definitions when ALL SSE delta chunks omit the name.

ECS gates: fmt ✓ / clippy -D warnings ✓ / 1753 passed, 0 failed ✓

viron-xiao:
The code change itself is correct and appropriately conservative — every failure mode (ambiguous multi-tool match, non-object arguments mid-stream, no tools defined) falls back to the pre-fix behavior, the single-tool unconditional backfill is a near-certain inference, and the tests cover the key branches. No code concerns.

Requesting changes for one item:

**The PR description does not match the diff.** It claims this mirrors the existing `id` empty-string guard onto `name` in `extract_parts_from_sse_body`, and names a test `test_extract_parts_from_sse_body_tool_calls_empty_name_does_not_overwrite` — neither exists in this diff. That guard already landed on main (the DashScope `name:""` repeat case in openai_parse.rs). What this PR actually adds is the complementary recovery: name omitted from all chunks → merged name is empty → backfill from the request tool definitions.

Please rewrite the description (including the Evidence / Test plan sections) to describe the actual change — as written it would mislead future audits and anyone reviewing against the description.

jfeng18:
Hi @viron-xiao — gentle nudge for a review when you have a moment (following your approval on #1219).

What this does: back-fills empty SSE tool_call names from the request's `tools` definitions — single tool defined fills unconditionally; with multiple tools it fills only when exactly one tool's `parameters.properties` is a superset of the call's argument keys, and abstains otherwise. The matching is conservative by construction (never guesses), and CI is green (20/20) with discriminating regression tests.

Note: the PR body now reflects the actual implementation — it previously described a simpler empty-string guard and was corrected today. Thanks!

viron-xiao (approved):
Approving — the description has been rewritten to match the actual implementation, and the code was already correct in the first review.

The recovery logic is appropriately conservative: single-tool unconditional backfill, multi-tool superset-match only, abstain on ambiguity. All failure modes fall back to the pre-fix behavior (empty name stays empty), so this cannot introduce wrong names. Tests cover the key branches (single/multi/ambiguous/no-tools).

CI green (20/20).

PR #1309, opened 2026-07-03
feat(anolisa): list command local-aware summary with bounded rpmdb probe

ikunkun-sys:
## Description

Adds local-aware summary columns to `anolisa list`, based on `docs/cli_state_commands_proposed.md` §3. The list command now performs a bounded rpmdb probe so users can see mirror-preinstalled or admin-preinstalled ANOLISA RPM components even when `installed.toml` has no record.

Behavior:

- New columns: `LOCAL STATE`, `OWNERSHIP`, `ACTION` (replacing the old `SUMMARY`/`STATUS` table).
- Row states: `not_installed`, `observed`, `tracked`, `installed`, `drifted`, `missing`, `failed`, `degraded`, `disabled`.
- `observed` = rpmdb has the package but ANOLISA state has no record; ownership shows `rpm` (package manager owns files); action is `install`.
- Bounded rpmdb probe covers three stages: RPM backend package names → rpm-package aliases → `whatprovides` capability lookup for legacy-adoptable components.
- `list` stays read-only: no state writes, no dnf/rpm transactions, no auto-track.
- JSON output preserves the `status` field for backward compatibility and adds `local_state`, `ownership`, `action`, and optional `rpm_package`/`rpm_evr`/`rpm_arch`/`rpm_source_repo`.
- Table column alignment fixed for ANSI color codes (pad colored strings via `measure_text_width`).

## Design Note

The state projection logic lives in `state_view.rs`, separated from rendering (`render.rs`) and the handler (`list.rs`). `LocalProjection` is the intermediate struct that maps `InstalledState` + component index + rpmdb probe into user-facing row fields.

The OWNERSHIP column for `observed` state shows `rpm` rather than `none`: the package manager owns file transactions even though ANOLISA has not claimed ownership. This is a display-level concern — the core `Ownership` enum is unchanged (`None` in state means "ANOLISA has not claimed ownership").

The provides-based probe (`what_provides_installed`) returns `None` for zero or ambiguous (>1) providers so the list summary never picks an arbitrary package. Detailed resolution is left to `status`.

## Ship Note

- Source repo enrichment for observed RPMs is deferred: `query_installed` returns `origin: None`, so `rpm_source_repo` is `None` for observed state in JSON. The `status` command already populates this via `installed_origin()`.
- The `--installed` filter now uses `local_state != not_installed` instead of the old `status_is_enabled` check, so `observed` components appear with `--installed`.

## Test

- Unit (projection): all 9 local states covered; observed via direct package, alias, and provides; ambiguous providers → not_installed; command-missing → graceful degradation.
- Unit (rows): observed/tracked/installed/drifted/missing row assertions; `--installed` filter semantics; rpm query failure tolerance.
- Unit (output): JSON retains `status` + adds local fields; human header contains `LOCAL STATE`; no `--catalog`/`--tracked` args.
- Unit (compatibility): all pre-existing list tests pass unchanged (using `None` rpm_query).
- Full gate green: `cargo fmt --all -- --check`, `cargo clippy --all-targets --locked -- -D warnings`, `cargo test -p anolisa-cli` (408 passed).

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs
@@ -4,16 +4,26 @@
 //! merges install status from `installed.toml`, and renders as a human
 //! table or `--json` envelope.
 
-use anolisa_core::state::{InstalledState, ObjectKind};
+mod render;
+mod state_view;
+
+#[cfg(test)]
+mod tests;
+
+use anolisa_core::state::InstalledState;
+use anolisa_platform::pkg_query::PackageQuery;
+use anolisa_platform::rpm_query::RpmPackageQuery;
 use clap::Parser;
 use serde::Serialize;
 
-use crate::color::{Palette, pad_right};
 use crate::commands::common;
 use crate::commands::common::RepoPersistPolicy;
-use crate::context::CliContext;
+use crate::context::{CliContext, InstallMode};
 use crate::resolution::{ComponentIndex, ComponentIndexEntry, load_component_index};
 use crate::response::{CliError, render_json};
+
+use self::render::render_human;
+use self::state_view::{LocalProjection, project_component};
 
 const COMMAND: &str = "list";
 
@@ -33,6 +43,17 @@ pub struct Row {
     pub summary: String,
     pub backends: Vec<String>,
     pub status: String,
+    pub local_state: String,
+    pub ownership: String,
+    pub action: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rpm_package: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rpm_evr: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rpm_arch: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rpm_source_repo: Option<String>,
 }
 
 #[derive(Serialize)]
@@ -55,7 +76,16 @@ pub fn handle(args: ListArgs, ctx: &CliContext) -> Result<(), CliError> {
         })?;
 
     let state = common::load_installed_state(ctx, COMMAND)?;
-    let rows = build_rows(&index, &args, &state);
+    let rpm_query = match ctx.install_mode {
+        InstallMode::System => Some(RpmPackageQuery::system()),
+        InstallMode::User => None,
+    };
+    let rows = build_rows(
+        &index,
+        &args,
+        &state,
+        rpm_query.as_ref().map(|query| query as &dyn PackageQuery),
+    );
 
     if ctx.json {
         return render_json(COMMAND, ListPayload { components: rows });
@@ -67,21 +97,30 @@ pub fn handle(args: ListArgs, ctx: &CliContext) -> Result<(), CliError> {
     Ok(())
 }
 
-fn build_rows(index: &ComponentIndex, args: &ListArgs, state: &InstalledState) -> Vec<Row> {
+fn build_rows(
+    index: &ComponentIndex,
+    args: &ListArgs,
+    state: &InstalledState,
+    rpm_query: Option<&dyn PackageQuery>,
+) -> Vec<Row> {
     index
         .components
         .iter()
-        .map(|entry| entry_to_row(entry, state))
-        .filter(|row| !args.installed || common::status_is_enabled(&row.status))
+        .filter_map(|entry| {
+            let projection = project_component(entry, state, rpm_query);
+            if args.installed && !projection.local_state.matches_installed_filter() {
+                return None;
+            }
+            Some(entry_to_row(entry, projection))
+        })
         .collect()
 }
 
-fn entry_to_row(entry: &ComponentIndexEntry, state: &InstalledState) -> Row {
+fn entry_to_row(entry: &ComponentIndexEntry, projection: LocalProjection) -> Row {
     let backends: Vec<String> = entry.backends.iter().map(|b| b.kind.clone()).collect();
-    let status = state
-        .find_object(ObjectKind::Component, &entry.name)
-        .map(|obj| common::object_status_str(obj.status))
-        .unwrap_or("not_installed");
+    let local_state = projection.local_state.label().to_string();
+    let ownership = projection.ownership_label().to_string();
+    let action = projection.action_label().to_string();
     Row {
         name: entry.name.clone(),
         display_name: entry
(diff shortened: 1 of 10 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs
```diff
@@ -67,21 +89,30 @@ pub fn handle(args: ListArgs, ctx: &CliContext) -> Result<(), CliError> {
...
+    state: &InstalledState,
+    rpm_query: Option<&dyn PackageQuery>,
+) -> Vec<Row> {
     index
         .components
         .iter()
-        .map(|entry| entry_to_row(entry, state))
-        .filter(|row| !args.installed || common::status_is_enabled(&row.status))
+        .map(|entry| entry_to_row(entry, state, rpm_query))
+        .filter(|row| !args.installed || row.local_state != LocalState::NotInstalled.label())
```
kongche-jbw: [P2] `--installed` / hidden alias `--enabled` now includes non-installed states because this filter only excludes `not_installed`. That lets `failed`, `disabled`, `drifted`, and `missing` rows show up under `anolisa list --installed`; for example a tracked RPM removed with `rpm -e` would still be listed as installed. This regresses the existing `status_is_enabled` semantics and the help text's “currently installed” promise.

src/anolisa/crates/anolisa-cli/src/commands/tier1/list/state_view.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+fn project_tracked_object(
+    object: &InstalledObject,
+    rpm_query: Option<&dyn PackageQuery>,
+) -> LocalProjection {
+    let ownership = object.effective_ownership();
+    let base_state = tracked_state_without_rpm_drift(object.status, ownership);
+    let local_state = match base_state {
+        LocalState::Installed => rpm_drift_state(object, rpm_query).unwrap_or(base_state),
```
kongche-jbw: [P2] Drift probing only runs when the base state is `Installed`, so adopted `rpm-observed` rows mapped to `Tracked` never surface rpmdb drift or missing packages. If a user runs `dnf update` or `rpm -e` after adoption, `anolisa status` reports `drifted` / `missing`, but `anolisa list` keeps saying `tracked`. Please run the RPM drift probe for clean RPM-backed tracked rows too, while still preserving failed/degraded/disabled states.

kongche-jbw (commented):
Reviewed the list CLI changes; left two inline comments.

src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs
ikunkun-sys: Addressed in `79b9c724`. `--installed` now filters on `LocalState::matches_installed_filter()` before rendering rows, so `failed`, `disabled`, `drifted`, `missing`, and `not_installed` are excluded while the currently-present states (`observed`, `tracked`, `installed`, `degraded`) remain visible. Added a regression test covering those included/excluded states.

src/anolisa/crates/anolisa-cli/src/commands/tier1/list/state_view.rs
ikunkun-sys: Addressed in `79b9c724`. Clean `Tracked` projections now run through the same RPM drift probe as `Installed`, so adopted `rpm-observed` rows surface `drifted` and `missing` when rpmdb changes. The match still preserves `failed`, `degraded`, and `disabled` states without probing over them, with focused regression coverage in `projection_tracked.rs`.

src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs
```diff
@@ -55,7 +76,8 @@ pub fn handle(args: ListArgs, ctx: &CliContext) -> Result<(), CliError> {
         })?;
 
     let state = common::load_installed_state(ctx, COMMAND)?;
-    let rows = build_rows(&index, &args, &state);
+    let rpm_query = RpmPackageQuery::system();
```
kongche-jbw: [P2] This unconditionally probes the system rpmdb, so user-mode `anolisa list --installed` can report system RPMs as `observed`. That disagrees with the rest of the CLI: `install` only auto-detects/adopts system RPMs in system mode, and `status` also gates observed RPM reporting on `InstallMode::System`. Please only pass the RPM query for system-mode list output, or otherwise keep user-scope list from surfacing system RPM state it cannot act on.

src/anolisa/crates/anolisa-cli/src/commands/tier1/list/state_view.rs
```diff
@@ -0,0 +1,278 @@
...
+    LocalProjection {
+        backend: Some("rpm".to_string()),
+        local_state: LocalState::Observed,
+        ownership: None,
+        action: ListAction::Install,
+        status: "not_installed".to_string(),
+        rpm_package: Some(info.name),
+        rpm_evr: Some(info.version.to_string()),
+        rpm_arch: Some(info.arch),
+        rpm_source_repo: info.origin,
```
kongche-jbw, line 281: [P3] For the real RPM backend this field will be `None` on the observed path because `query_installed()` uses `rpm -q`, whose parser sets `PackageInfo::origin` to `None`; `status` calls `installed_origin()` separately for this exact metadata. If `rpm_source_repo` is part of the new list JSON contract, list needs to fetch it explicitly, otherwise the field is misleading/dead for the main implementation.

kongche-jbw (commented):
One additional non-inline issue: this PR changes user-visible `anolisa list` human output and JSON fields (`local_state`, `ownership`, `action`, RPM metadata) but does not update `src/anolisa/CHANGELOG.md`. The repo rules require a changelog entry for user-perceivable behavior changes.

src/anolisa/crates/anolisa-cli/src/commands/tier1/list.rs
ikunkun-sys: Addressed in `dc874e0b`. `anolisa list` now only constructs and passes the RPM query in `InstallMode::System`; user-mode list builds rows with `None`, so untracked system RPMs are not surfaced as `observed`. Added regression coverage for the no-query/user-scope row path.

src/anolisa/crates/anolisa-cli/src/commands/tier1/list/state_view.rs
ikunkun-sys: Addressed in `dc874e0b`. Observed RPM projection now enriches `PackageInfo::origin` with `installed_origin()` when `query_installed()` returns no origin, treating origin lookup failures as optional metadata. Added regression coverage for an installed RPM whose `query_installed()` origin is `None`.

kongche-jbw (approved):
lgtm

PR #1329, opened 2026-07-04
fix(skillfs): prevent FUSE test mount leaks

### LINKED ISSUE #1328: [skillfs] bug: FUSE test mounts are leaked after cargo test
## Component

skillfs

## What happened?

**Bug description**

`cargo test --workspace` in the `test-skillfs` job leaves FUSE mounts behind. The tests only assert on process output, so they pass even when the mount is never torn down — the leaked mounts and their directories accumulate under `target/tmp/skillfs-nontmp-*` as dead FUSE endpoints, and any later access to those paths fails with `ENOTCONN`.

**Root cause**

In `crates/skillfs-cli/tests/cli_startup_tests.rs`, tests that pass the startup gate (e.g. `non_tmp_transport_paths_pass_gate`) actually establish a FUSE mount at `non_tmp_dir()` (= `CARGO_TARGET_TMPDIR` = `target/tmp/`), then terminate the process with `child.kill()` (**SIGKILL**). SkillFS only unmounts via `fusermount3 -u` on **SIGTERM/SIGINT** (`main.rs:2140`); SIGKILL cannot be caught, so the mount survives and `TempDir::drop` cannot remove the still-mounted directory — both the mount and the directory leak. The test code has no unmount/Drop fallback.

**Steps to reproduce**

1. On a host with fuse3: `cd src/skillfs && cargo test --workspace`, run a few times.
2. `grep skillfs-nontmp /proc/mounts` — one leaked mount accumulates per run.
3. `ls target/tmp/skillfs-nontmp-*` reports `Transport endpoint is not connected`.

**Expected behavior**

No FUSE mounts or mount directories remain after the test run.

kongche-jbw:
## Description

The tests terminate the foreground mount process with `child.kill()` (SIGKILL, which SkillFS cannot catch), so the FUSE mount is never unmounted and leaks under `target/tmp`.

This PR:
- Adds `stop_mount_child`: sends **SIGTERM** so SkillFS takes its normal unmount path, waits a bounded time for graceful exit, then force-unmounts as a fallback (`fusermount3 -u` / `-z` / `umount -l`) and finally SIGKILLs to reap the child. Replaces all 4 `child.kill()` call sites; `run_briefly` now takes the mountpoint so teardown always targets the mounted path.
- Adds a bounded, non-panicking, non-blocking `force_unmount_path` on `MountHandle`, invoked from both `Drop` and the `unmount()` error path so every teardown route is covered.

## Related Issue

closes #1328

## Type of Change

- [x] CI/CD or build changes

## Scope

- [x] `skillfs` (SkillFS)

## Checklist

- [x] I have added tests that prove my fix is effective (reuses existing tests; the `target/tmp` mount count no longer grows after a run)
- [x] For `skillfs`: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` pass

## Testing

- Verified with the CI-pinned **1.86.0** toolchain: `fmt --check` / `clippy -D warnings` / build all clean.
- `cargo +1.86.0 test --test cli_startup_tests` — 47 passed; `grep -c skillfs /proc/mounts` is unchanged before/after (it would increase by 1 before the fix).
- Re-ran the previously-leaking `non_tmp_transport_paths_pass_gate` in isolation — no new `target/tmp` mount appears.

## Additional Notes

`MountHandle::Drop` does not help the SIGKILL case directly (a SIGKILLed process never runs Drop); it is a defensive safeguard for library consumers and in-crate tests. The actual fix for the leak is the SIGTERM change on the test side.

### DIFF
--- a/src/skillfs/crates/skillfs-fuse/src/lib.rs
+++ b/src/skillfs/crates/skillfs-fuse/src/lib.rs
@@ -121,11 +121,18 @@ impl MountHandle {
                     // lazy). Best-effort force cleanup so we never leave the
                     // mountpoint dangling, but still surface the original
                     // failure to the caller for explicit `unmount()`.
+                    //
+                    // Detach the session thread first so a later `Drop` sees
+                    // `session == None` and skips a second `unmount_inner`
+                    // (which would run `force_unmount_path` again — up to a
+                    // full 5s each on a genuinely stuck mount).
                     let stderr = String::from_utf8_lossy(&output.stderr);
+                    self.session.take();
                     Self::force_unmount_path(&self.mountpoint);
                     return Err(FuseError::UnmountFailed(stderr.to_string()));
                 }
                 Err(e) => {
+                    self.session.take();
                     Self::force_unmount_path(&self.mountpoint);
                     return Err(FuseError::IoError(e));
                 }
@@ -360,16 +367,19 @@ mod tests {
     }
 
     #[test]
-    fn unmount_inner_skips_join_when_fusermount_fails() {
+    fn unmount_inner_detaches_session_when_fusermount_fails() {
         let mut handle = MountHandle {
             mountpoint: PathBuf::from("/nonexistent/mount/point"),
             session: Some(std::thread::spawn(|| {})),
         };
         let result = handle.unmount_inner();
         assert!(result.is_err(), "bogus mountpoint must produce an error");
+        // The error path detaches the session (takes it without joining) so a
+        // subsequent `Drop` sees `session == None` and does not run a second
+        // `unmount_inner` / `force_unmount_path` pass.
         assert!(
-            handle.session.is_some(),
-            "session must remain when fusermount3 fails (join skipped)"
+            handle.session.is_none(),
+            "session must be detached when fusermount3 fails so Drop skips re-unmount"
         );
     }
 
--- a/src/skillfs/crates/skillfs-fuse/src/lib.rs
+++ b/src/skillfs/crates/skillfs-fuse/src/lib.rs
@@ -117,10 +117,16 @@ impl MountHandle {
                     info!("unmount successful");
                 }
                 Ok(output) => {
+                    // A plain `fusermount3 -u` can fail transiently (busy /
+                    // lazy). Best-effort force cleanup so we never leave the
+                    // mountpoint dangling, but still surface the original
+                    // failure to the caller for explicit `unmount()`.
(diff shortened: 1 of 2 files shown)


src/skillfs/crates/skillfs-fuse/src/lib.rs
```diff
@@ -117,10 +117,16 @@ impl MountHandle {
                     info!("unmount successful");
                 }
                 Ok(output) => {
+                    // A plain `fusermount3 -u` can fail transiently (busy /
+                    // lazy). Best-effort force cleanup so we never leave the
+                    // mountpoint dangling, but still surface the original
+                    // failure to the caller for explicit `unmount()`.
                     let stderr = String::from_utf8_lossy(&output.stderr);
+                    Self::force_unmount_path(&self.mountpoint);
                     return Err(FuseError::UnmountFailed(stderr.to_string()));
```
yummypeng, line 132: **失败路径会重复触发 `force_unmount_path`，卡死时最多阻塞 ~15s**

这两个错误分支返回 `Err` 前都没有 `self.session.take()`，所以 `session` 仍是 `Some`。当调用方用显式 `unmount()`（消费 `self`）且失败时：

1. `unmount_inner` 错误分支 → `force_unmount_path` 第 1 次 → 返回 `Err`
2. `self` 析构 → `Drop` 里 `session.is_some()` 仍为 true → 再次 `unmount_inner` → `force_unmount_path` 第 2 次
3. `Drop` 尾部 → `force_unmount_path` 第 3 次

瞬时失败时第 2、3 次会因 `path_is_mounted() == false` 立即返回，代价很小；但挂载**真正卡死**时，每次都会跑完整的 50×100ms 循环，总计 ~15s 阻塞。因为这发生在 `Drop` 里，drop 该 handle 的线程会被静默卡住 —— 在频繁挂载/卸载的路径上容易变成难查的卡顿。

建议在两个错误分支返回前也 `self.session.take()`，分离线程，让 `Drop` 跳过第二次 `unmount_inner`：

```rust
Ok(output) => {
    let stderr = String::from_utf8_lossy(&output.stderr);
    self.session.take();
    Self::force_unmount_path(&self.mountpoint);
    return Err(FuseError::UnmountFailed(stderr.to_string()));
}
```

这样卡死场景从 3×5s 降到 2×5s。

yummypeng (approved):
LGTM — 问题 2 已按建议修复（错误分支 detach session，force_unmount_path 由 3 次降到 2 次），且单测契约同步更新。

PR #1331, opened 2026-07-04
feat(anolisa): add cosh, codex, claude-code adapter drivers

kongche-jbw:
## Description

Adds first-class ANOLISA adapter support for cosh (copilot-shell), Codex, and
Claude Code, implemented entirely as built-in Rust `FrameworkDriver`s — no
delegation to component-side install/uninstall shell scripts. Extends the
receipt security boundary with two new claim-resource kinds (`Symlink`,
`FrameworkMarketplace`), two controlled IO ops (`write_file`, `create_symlink`),
and the `extension` adapter type. Cosh uses a filesystem extension model with an
ownership marker so it never destroys user-installed extensions; Codex/Claude
Code drive their official marketplace CLIs and keep the receipt as
`cleanup_failed` when deregistration cannot complete.

## Related Issue

no-issue: internal adapter roadmap (design in ANOLISA-design/docs/anolisa/cli)

## Type of Change

## Scope

- [x] `tokenless` (tokenless)
- [x] `anolisa` (anolisa-cli)

## Checklist

- [x] For `anolisa`: `cargo clippy --all-targets --locked -- -D warnings`,
      `cargo fmt --all --check`, and `cargo test --locked` pass

## Testing

- `cargo test --workspace --locked` — all 11 test binaries pass
- `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean
- `cargo doc --workspace --no-deps` — clean
- New unit tests per driver + `tests/adapter_frameworks.rs` integration suite
  (fake codex/claude CLIs asserting exact argv; cosh copy/remove-only-own-dir;
  dry-run writes nothing; forged symlink target / resource_root rejected;
  CLI-removal-failure and ownership-marker-missing keep the receipt)

### DIFF
--- a/src/anolisa/crates/anolisa-core/src/adapter.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter.rs
@@ -21,12 +21,16 @@ use anolisa_platform::fs_layout::FsLayout;
 use crate::manifest::AdapterSpec;
 
 pub mod claim;
+pub mod claude_code;
+pub mod codex;
 pub mod contract;
+pub mod cosh;
 pub mod driver;
 pub mod hermes;
 pub mod manager;
 pub mod openclaw;
 pub mod registry;
+mod util;
 
 // ---------------------------------------------------------------------------
 // Error
@@ -93,9 +97,9 @@ pub enum AdapterError {
     },
 
     /// The manifest declares an `adapter_type` that is not yet supported
-    /// by any built-in driver (e.g. `extension`, `service`).
+    /// by any built-in driver (e.g. `service`).
     #[error(
-        "adapter type '{adapter_type}' for {component}/{framework} is not supported; only 'plugin' and 'skill_bundle' are implemented"
+        "adapter type '{adapter_type}' for {component}/{framework} is not supported; only 'plugin', 'skill_bundle', and 'extension' are implemented"
     )]
     UnsupportedAdapterType {
         /// Component whose adapter was requested.
--- a/src/anolisa/crates/anolisa-core/src/adapter/claim.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/claim.rs
@@ -120,11 +120,46 @@ impl AdapterClaim {
         layout: &FsLayout,
         allowed_external_roots: &[PathBuf],
     ) -> Result<(), ClaimValidationError> {
+        self.validate_with_owned_roots(layout, allowed_external_roots, &[])
+    }
+
+    /// Like [`Self::validate`], but additionally trusts `extra_owned_roots`
+    /// as ANOLISA-owned locations for symlink *targets*.
+    ///
+    /// A symlink target (e.g. Codex's plugin symlink pointing back at the
+    /// resource bundle) is normally validated against the primary layout's
+    /// owned roots. When the bundle is resolved from a packaged datadir that
+    /// differs from `layout.datadir` (registered via
+    /// [`push_primary_datadir_root`](super::manager::AdapterManager::push_primary_datadir_root)),
+    /// that target lives outside the layout roots yet is still legitimately
+    /// ANOLISA-owned. The Manager passes its trusted datadir roots here so
+    /// such targets validate.
+    ///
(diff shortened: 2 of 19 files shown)


ikunkun-sys:
I found three adapter correctness issues worth addressing before merge:

1. **Codex enable rejects packaged datadir resources.** The Codex receipt records the symlink target as the resolved `bundle.resource_root` (`src/anolisa/crates/anolisa-core/src/adapter/codex.rs:186`), but `ClaimResourceKind::Symlink` validates that target with `validate_owned_path(layout, target)` (`src/anolisa/crates/anolisa-core/src/adapter/claim.rs:200`). That only accepts the active `FsLayout` owned roots. If the component contract/resource bundle is resolved from an additional packaged datadir such as `/usr/share/anolisa` via `push_primary_datadir_root`, enable fails before writing anything. I reproduced this locally with a regression test that stages the Codex resource under an extra packaged datadir; it fails with `ClaimValidation(OwnedPath { id: "codex_symlink", source: External { ... } })`. The trusted target boundary probably needs to include the resolved resource root/effective datadir set, not only the primary layout roots.

2. **Claude Code marketplace ownership looks ambiguous for multiple components.** The driver uses one shared marketplace name, `anolisa` (`src/anolisa/crates/anolisa-core/src/adapter/claude_code.rs:43`), but each enable registers the current component's `resource_root` and skips registration when that marketplace already exists (`src/anolisa/crates/anolisa-core/src/adapter/claude_code.rs:237`). Disable then always removes the shared `anolisa` marketplace (`src/anolisa/crates/anolisa-core/src/adapter/claude_code.rs:382`). If multiple Claude Code adapters can be enabled, the second component may never register its source, and disabling one component can remove the marketplace used by another. If the intended invariant is "only one Claude Code adapter component at a time", this should be enforced or documented; otherwise the marketplace should be component-scoped or have shared ownership/ref-count handling.

3. **`adapter_type = "extension"` is accepted globally.** `is_supported_adapter_type` now allows `extension` for every framework (`src/anolisa/crates/anolisa-core/src/adapter/manager.rs:1817`), but the drivers do not validate framework/type compatibility. That means a manifest can declare `framework = "openclaw"` with `adapter_type = "extension"` and still proceed down OpenClaw's plugin path, while `framework = "cosh"` with `adapter_type = "plugin"` is treated as a filesystem extension. The type gate should probably be driver-specific, or the manager should explicitly reject unsupported framework/type pairs.

Verification I ran:

- `cargo test -p anolisa-core --test adapter_frameworks` passes on this PR head.
- Existing packaged-datadir manager regression test passes.
- A temporary Codex packaged-datadir regression test fails as described in item 1.

ikunkun-sys:
Re-reviewed latest head `586bb57f53d9d1c9b35cde6224c438f4cdd121fa`.

No remaining blocking findings from my side. The three issues I raised earlier look addressed:
- Codex packaged-datadir bundles now validate symlink targets against Manager-supplied trusted datadir roots.
- Claude Code marketplace ownership is now component-scoped (`anolisa-<component>`) and bundle validation rejects the old shared `anolisa` name.
- `adapter_type` is now checked against the per-framework supported type matrix, so wrong-but-known pairs fail closed.

Verified on a clean PR snapshot:
- `cargo test -p anolisa-core --test adapter_frameworks`
- `cargo test -p anolisa-core adapter::`
- `cargo check -p anolisa-core`

Only a non-blocking docs nit remains: the `ClaimResourceKind::Symlink` rustdoc still says both link and target are validated against external roots, while the implementation now validates the target against owned roots / extra trusted datadir roots.

kongche-jbw:
fresh [claim.rs (line 326)](/Users/kongche/codex/anolisa/anolisa_cli/src/anolisa/crates/anolisa-core/src/adapter/claim.rs:326)  ClaimResourceKind::Symlink

shiloong:
---

### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

---

shiloong (commented):
LGTM, approve with comments

ikunkun-sys (approved):

PR #1337, opened 2026-07-04
fix(ktuner): shmall target uses runtime page size

jfeng18:
## What

`eval_shmall` hardcoded a 4096-byte page size when converting RAM to pages. `kernel.shmall` is measured in **pages**, and page size is architecture-dependent (4 KiB on x86_64, up to 64 KiB on arm64). On 64K-page kernels the recommended value was ~16x too high.

## Fix

- Read the page size at runtime via `sysconf(_SC_PAGESIZE)` (fallback 4096).
- Extract a pure `shmall_target_pages(mem_gb, page_size)` that guards `page_size == 0`.
- Replace the previous vacuous `if let` test with a discriminating one: asserts the 4K result equals the 64K result x16, and `page_size 0 -> 0`.

## Testing

- Only unit-tested: 1.89.0 `cargo fmt --check` / `clippy --all-targets -- -D warnings` / `cargo test` all pass (195 tests).
- Mutation-tested: reverting to hardcoded 4096 fails the scaling assert; removing the zero-guard panics; flipping `/2` fails — the test is discriminating.
- Not E2E-verified on a real 64K-page arm64 kernel (dev host is x86_64/4K); the page-size scaling is covered by the pure-function test injecting 65536.

### DIFF
--- a/src/ktuner/src/rules/mod.rs
+++ b/src/ktuner/src/rules/mod.rs
@@ -4934,13 +4934,37 @@ fn eval_pipe_max_size(_info: &SystemInfo, recs: &mut Vec<Recommendation>) -> usi
     1
 }
 
+/// Query the running kernel's page size (bytes). `shmall` is measured in pages,
+/// and the page size is architecture/kernel dependent — 4 KiB on x86_64 but up
+/// to 64 KiB on arm64 — so it MUST be read at runtime, never hardcoded.
+/// Falls back to 4096 if `sysconf` fails (a non-positive return).
+fn current_page_size() -> u64 {
+    // SAFETY: sysconf(_SC_PAGESIZE) is a pure, side-effect-free query.
+    let sz = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
+    if sz > 0 {
+        sz as u64
+    } else {
+        4096
+    }
+}
+
+/// Recommended `kernel.shmall` (in pages): half of physical RAM, converted to
+/// pages using the given page size. Returns 0 for a zero page size (guards the
+/// division). Kept pure so it can be tested across page sizes.
+fn shmall_target_pages(memory_total_gb: u64, page_size: u64) -> u64 {
+    if page_size == 0 {
+        return 0;
+    }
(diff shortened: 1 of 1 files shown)


kongche-jbw (commented):
Review baseline: `87bfbf725d8f8a2aa3b991593ffea5c487f55366...f1b53414fa390363709e7e935a00ea52460eaebb`
**[P2] 回归测试没有覆盖 `eval_shmall` 的生产调用链**

`src/ktuner/src/rules/mod.rs:7443` 的测试只直接调用新 helper；另一个测试只检查
本机 `sysconf` 值。因此，把 `eval_shmall` 第 4967 行恢复成原先硬编码 4096 的
计算，整套新增测试仍会通过，64 KiB 主机上的 16 倍错误不受回归保护。

Possible direction: 让 evaluator 的页大小或读取依赖可注入，并通过 `eval_shmall`
断言 65536 页大小下实际生成的 `recommended_value`。

jfeng18:
Fixed in d245541a. The evaluator now accepts injectable page-size and current-value readers, and the production call chain passes the real ones from `evaluate_with_workload`.

The new test `test_eval_shmall_recommends_using_injected_page_size` asserts the exact recommended value for both 4 KiB (33554432) and 64 KiB (2097152) pages through `eval_shmall` itself — reverting the production code to hardcoded 4096 makes the 64 KiB assertion fail.

Also covered: no recommendation when current >= target, and skip when the parameter file doesn't exist.

kongche-jbw (commented):
Review baseline: `628533bbb100ddf1f0e6db23a27115171f061367...d245541a67649c125e9f71c0efb4d9dd8fe35049`
**[P2] 更新 PR 中已过期的测试计数**

`src/ktuner/src/rules/mod.rs:7449` 将 1 个 `shmall` 测试替换为 4 个，
当前 head 上执行 `cargo test` 会运行 195 个测试，而 PR 的 Testing 仍写 192 个。
这使验证记录无法按描述复现。

Possible direction: 将 Testing 的数量更新为 195；无需修改代码或新增测试。

kongche-jbw (approved):

PR #1338, opened 2026-07-04
test(ktuner): pin rollback keep-previous invariant

jfeng18:
## What

Extract `merge_entries` from `merge_rollback` as a pure (no-I/O) fn with zero behavior change, and pin the rollback keep-previous invariant with a discriminating test: across repeated tune/fix/import runs the ORIGINAL pre-ktuner value must survive so rollback restores pristine state.

## Why discriminating

A merge that clobbered `previous` on re-tune would silently make rollback restore to a post-tune value instead of the pristine one. The test fails if `and_modify` also sets `previous` (mutation-verified).

## Testing
- Done (ktuner crate, toolchain 1.89): `cargo fmt --check`, `clippy --all-targets -- -D warnings`, `cargo test` (192 pass).

## Note
Slimmed from an earlier version: dropped the `is_metric_degraded` extraction and its test. That covered the benchmark degradation direction, but `verify_and_report` / `auto_rollback_on_degradation` have zero callers in the shipped CLI (`ktuner tune` verifies via static `eval.score()`), so testing them hardened dead code. Kept only the live rollback-merge coverage.

### DIFF
--- a/src/ktuner/src/tuner/mod.rs
+++ b/src/ktuner/src/tuner/mod.rs
@@ -292,18 +292,15 @@ fn save_rollback(recommendations: &[Recommendation]) -> Result<()> {
     }))
 }
 
-/// Merge `(param, previous, applied)` entries into the cumulative rollback
-/// record. For a param already recorded, keep the ORIGINAL `previous` (the true
+/// Merge `(param, previous, applied)` entries into a cumulative rollback record.
+/// For a param already recorded, keep the ORIGINAL `previous` (the true
 /// pre-ktuner value) so rollback always restores pristine state even across
 /// multiple tune/fix/import runs; only refresh `applied`. New params are added.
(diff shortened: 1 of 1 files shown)


kongche-jbw (commented):
Review baseline: `87bfbf725d8f8a2aa3b991593ffea5c487f55366...4f21d77131f415a713663dbccf6a672ff040d4bc`
未发现阻塞问题。

Remaining risks:
- 本地未执行需要 root 和真实 `/proc`、`/sys` 的 I/O 场景。
- 本地工具链为 Rust 1.97.1；PR CI 已在 Rust 1.89.0 通过 ktuner 检查。

Verification scope:
- 完整 diff、相邻 rollback 调用链与现有/新增测试。
- `cargo fmt --all --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test`（192 passed）

kongche-jbw (approved):

PR #1359, opened 2026-07-06
[cosh-ng] fix SSH remote command execution for cosh login shell

### LINKED ISSUE #1357: [cosh-ng] bug: after setting cosh-ng as the default shell, SSH command execution fails.
### Component

cosh-ng

### What happened?

After setting cosh-ng as the default login shell, commands cannot be executed via SSH. For example, `ssh <user>@<host> "echo hello"` only logs into the machine without executing the command or disconnecting.
Image

### Diagnostic Output

```shell

```

### Additional Context

_No response_

SunnyQjm:
## Summary

Fix SSH remote command execution when `cosh` is configured as the user's login shell.

## Root Cause

OpenSSH executes remote commands through the user's login shell as `shell -c <command>`. The installed `/usr/bin/cosh` launcher always prepended `raw cosh-core`, so `ssh host 'echo ok'` became:

```bash
cosh-shell raw cosh-core -c 'echo ok'
```

The raw subcommand path did not recognize this as a non-interactive shell command and entered interactive raw mode instead.

## Changes

- Teach `cosh-shell raw <adapter> -c <command>` to normalize back into the existing non-interactive passthrough path.
- Preserve shell options such as `--shell` while stripping the raw adapter segment for passthrough execution.
- Update the RPM `cosh` launcher so explicit non-interactive CLI entry points (`-c`, `--`, `--help`, `--version`, non-TTY stdin) go directly to `cosh-shell`.
- Add raw CLI regression coverage for command output and exit status preservation.

Fixes #1357

## Validation

```bash
cargo test -p cosh-shell raw_passthrough_args
cargo test -p cosh-shell --test raw_cli passthrough
cargo check -p cosh-shell
git diff --check
awk '/cat > %\{buildroot\}%\{_bindir\}\/cosh << '\''EOF'\''/{flag=1; next} flag && /^EOF$/{flag=0} flag{print}' src/cosh-ng/cosh-ng.spec.in | bash -n
```

Manual verification on an affected host after applying the launcher hotfix:

```bash
ssh root@47.110.58.189 'echo __SSH_REMOTE_COMMAND_OK__'
ssh root@47.110.58.189 'exit 37'
ssh root@47.110.58.189 'cosh --version'
ssh root@47.110.58.189 'cosh -- echo __COSH_DOUBLE_DASH_OK__'
```

### DIFF
--- a/src/cosh-ng/cosh-ng.spec.in
+++ b/src/cosh-ng/cosh-ng.spec.in
@@ -46,6 +46,16 @@ install -m 0755 target/release/cosh-cli %{buildroot}%{_bindir}/cosh-cli
 # Install launcher script
 cat > %{buildroot}%{_bindir}/cosh << 'EOF'
 #!/bin/bash
+case "${1:-}" in
+  -c|--|--help|--version)
+    exec %{_libexecdir_cosh}/cosh-shell "$@"
+    ;;
+esac
+
+if [ "$#" -eq 0 ] && [ ! -t 0 ]; then
+  exec %{_libexecdir_cosh}/cosh-shell "$@"
+fi
+
 exec %{_libexecdir_cosh}/cosh-shell raw cosh-core "$@"
 EOF
 chmod 0755 %{buildroot}%{_bindir}/cosh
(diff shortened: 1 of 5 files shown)


samchu-zsl (approved):
LGTM, thx!

PR #1364, opened 2026-07-06
fix(sight): make hf-hub dependency optional behind local-tokenizer feature

chengshuyi:
## Description

The `hf-hub` dependency currently points to a GitHub-hosted fork (`chengshuyi/hf-hub`). When
that URL is unreachable — CI without external GitHub access, internal network builds, or the
fork being temporarily unavailable — the entire Cargo resolve/fetch fails even though most
deployments never use the local HuggingFace tokenizer download (the runtime tokenizer feature
is opt-in via `features.tokenizer.enabled` in config).

This PR gates `hf-hub` behind a new `local-tokenizer` Cargo feature. Default builds no longer
depend on the fork, preventing gratuitous build failures and reducing binary size for
environments that rely solely on runtime token counting.

## Related Issue

no-issue: discovered during build failure investigation on internal CI

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `sight` (agentsight)

## Key Changes

- `Cargo.toml`: add `local-tokenizer = ["dep:hf-hub"]` feature; make `hf-hub` an optional git dependency; remove `[patch.crates-io]` section
- `src/tokenizer/multi_model.rs`: gate `hf_api` field, `get_hf_api()`, and `register_from_hf()` with `#[cfg(feature = "local-tokenizer")]`; add `HF_TOKEN` env var support; replace `.expect()` with `?` for safe error propagation
- `src/token_breakdown/cli.rs`: replace `.unwrap()` with `?` on `get_global_tokenizer()` to gracefully report missing tokenizer when feature disabled

## Checklist

- [ ] `cargo clippy --all-targets -- -D warnings` pass (11 pre-existing warnings in unrelated files; no new warnings introduced)
- [x] `cargo test` pass
- [x] I have updated the documentation accordingly (inline doc comments)

## Testing

1. Default build (`cargo check`) — succeeds without fetching hf-hub fork
2. Feature-enabled build (`cargo check --features local-tokenizer`) — compiles with HF download support
3. Runtime without feature — calling `get_global_tokenizer()` returns a descriptive error instead of panic
4. Runtime with feature + `HF_ENDPOINT` override — tokenizer downloads from configured mirror

### DIFF
--- a/src/agentsight/Cargo.lock
+++ b/src/agentsight/Cargo.lock
@@ -1260,7 +1260,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1339,7 +1339,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1732,7 +1732,8 @@ checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
 [[package]]
 name = "hf-hub"
 version = "0.5.0"
-source = "git+https://github.com/chengshuyi/hf-hub.git?rev=05f5134f40dbb5dc9f1617df1aaef628d49e951f#05f5134f40dbb5dc9f1617df1aaef628d49e951f"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aef3982638978efa195ff11b305f51f1f22f4f0a6cabee7af79b383ebee6a213"
 dependencies = [
  "dirs",
  "futures",
@@ -1903,7 +1904,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -3287,7 +3288,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3750,7 +3751,7 @@ dependencies = [
  "getrandom 0.4.2",
  "once_cell",
  "rustix 1.1.4",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


jfeng18:
Reviewed the diff and verified the dependency behavior locally (1.89.0). A couple of points that seem to cut against the PR's stated goal — flagging for your judgment:

**1. Default builds still require the fork to be reachable — `Cargo.toml:90`**
`optional = true` gates *compilation*, not *source loading*: cargo checks out every declared git dependency during resolution, before feature selection. Verified:
```
# default features, fork removed from ~/.cargo/git, --offline
cargo metadata --locked --offline
→ exit 101: can't checkout from 'https://github.com/chengshuyi/hf-hub.git': you are in the offline mode
```
Restoring the cache makes it pass. So an uncached build with the fork unreachable still fails at the resolution step — just earlier than before rather than at compile time.

**2. Reverts the earlier `cargo vendor` fix — `Cargo.toml:93`**
Removing `[patch.crates-io]` restores two `hf-hub 0.5.0` entries in `Cargo.lock` (registry + git fork). This is the two-source state that `fd3556c4` ("fix cargo vendor error") collapsed onto one source via the patch. `.github/actions/package-vendor` runs `cargo vendor --locked`, so it'd be worth confirming vendor still succeeds before merge.

**3. hf-hub isn't actually dropped from default builds — `Cargo.toml:59`**
`llm-tokenizer` depends on registry `hf-hub 0.5.0` unconditionally (`cargo tree -i hf-hub`, default features → `hf-hub ← llm-tokenizer ← agentsight`), so hf-hub + reqwest/native-tls/ureq still compile by default. The PR removes the dependency on the *fork*, but not on hf-hub itself.

**Minor**
- Default builds now return "download is disabled" for uncached models; the two runtime callers (`analyzer/unified.rs:752`, `unified.rs:1400`) swallow it at `debug`/skip, so token-count fallback silently no-ops on default builds — might be worth a CHANGELOG note.
- The `local-tokenizer`-enabled path (register_from_hf, HF_TOKEN/HF_ENDPOINT env) has no test and isn't covered by CI clippy/test (both run default features only).

The `unwrap`→`?` cleanups and the HF_TOKEN/HF_ENDPOINT env support read as clear improvements. The main thing I'd double-check is #1 — whether the resolution-stage fork dependency defeats the "unreachable fork shouldn't break builds" goal in your target environment.

jfeng18 (approved):
Approving — thanks for iterating on this. The registry-at-rest + temp-patch-for-RPM approach cleanly resolves the two blocking concerns from the earlier review:

- **Default builds no longer need the fork reachable.** `hf-hub = "0.5"` (registry) with no `[patch.crates-io]` at rest — verified `cargo metadata --locked --offline` (default) now succeeds even with the fork cache removed (was exit 101 before).
- **No more two-source `cargo vendor` state.** `Cargo.lock` has a single registry `hf-hub 0.5.0`, so `fd3556c4`'s duplicate-source error can't recur; the fork is injected only transiently by `rpm-build.sh` for the mirror fixes.

CI is green. Two non-blocking nits, feel free to fold into a follow-up:

- `rpm-build.sh` restores `Cargo.toml` via `truncate` but doesn't restore `Cargo.lock`. `cargo build --release` with the injected patch rewrites the lock to the fork, so a local RPM build leaves a patched (fork) `Cargo.lock` in the tree until the next unpatched build overwrites it. (Local-script only — the CI vendor path uses the committed registry lock, so no release impact.)
- The comment above `hf-hub = "0.5"` still reads "patched to use GitHub version below", but there's no `[patch.crates-io]` below it anymore (it lives in `rpm-build.sh` now) — slightly misleading for the next reader.

PR #1373, opened 2026-07-07
feat(benchmark): add terminal-bench agent adapter

yinan1995:
## Description

This PR introduces the `terminal-runner` package under `src/benchmark/` — a Harbor-based agent adapter for TerminalBench task evaluation.

The package supports two execution modes:

🔧 **Installed mode** — Harbor's built-in `openclaw` agent is auto-installed inside the container via nvm + npm on every trial. The agent runs entirely within the container, communicating through Harbor's native lifecycle.

🚀 **External mode** — OpenClaw runs on the host machine. The `OpenClawExternalAgent` adapter bridges commands to the container via `environment.exec()`, eliminating per-trial install overhead. The same pattern extends to any host-side agent CLI (Claude Code, Codex, etc.).

**Key components:**

- `external_agent/openclaw_external_agent.py` — `BaseAgent` implementation that spawns the OpenClaw subprocess, streams stdout/stderr with heartbeat & safety guards (total-timeout ⏱️, no-output timeout 🔇, stdout cap 📏), extracts bash commands from responses, executes them in-container, and feeds results back for the next iteration
- `scripts/setup.sh` — Automated setup that detects Python version (conda fallback for <3.12), creates venv, clones Harbor v0.15.0 (pinned for stability), installs via `pip install -e`, and clones the terminal-bench dataset via git-lfs
- Bilingual README (EN/CN) covering prerequisites, both modes' usage, environment variables, per-iteration workflow, and the auto-skill system

**Verification ✅**

Both modes were verified end-to-end against the `nginx-request-logging` task:

| Mode | Reward | Runtime |
|---|---|---|---|
| 🚀 External | ✅ 1.0 | 54s |
| 🔧 Installed | ✅ 1.0 | 1m50s |

Test environment: Alibaba Cloud Linux 4, Python 3.12.13, Docker 24.0.9, Node.js v22.22.0, openclaw CLI 2026.6.11.

### DIFF
--- a/src/benchmark/terminal-runner/README.md
+++ b/src/benchmark/terminal-runner/README.md
@@ -1,5 +1,7 @@
 # terminal-runner
 
+[中文版](README_zh.md)
+
 This guide explains how to test terminal-bench tasks via the [harbor](https://github.com/harbor-framework/harbor) platform, focusing on two implementation approaches — **Installed mode** and **External mode** — their differences, trade-offs, and usage.
 
 ---
@@ -68,7 +70,7 @@ terminal-runner/
 ├── .gitignore
 ├── LICENSE
 ├── README.md
-└── README_CN.md
+└── README_zh.md
 ```
 
 ### What's NOT here (cloned by `setup.sh`)
--- a/src/benchmark/terminal-runner/README.md
+++ b/src/benchmark/terminal-runner/README.md
@@ -90,8 +90,8 @@ terminal-runner/
 ### Install
 
 ```bash
-git clone https://github.com/<your-org>/terminal-runner.git
-cd terminal-runner
+git clone https://github.com/alibaba/anolisa.git
+cd anolisa/src/benchmark/terminal-runner
 ./scripts/setup.sh
 ```
 
--- a/src/benchmark/terminal-runner/README.md
+++ b/src/benchmark/terminal-runner/README.md
@@ -1,0 +1,205 @@
+# terminal-runner
+
+This guide explains how to test terminal-bench tasks via the [harbor](https://github.com/harbor-framework/harbor) platform, focusing on two implementation approaches — **Installed mode** and **External mode** — their differences, trade-offs, and usage.
+
+---
+
+## Two modes, two support paths
+
+[harbor](https://github.com/harbor-framework/harbor) runs agent evaluations inside Docker containers. It supports two ways to supply the agent:
+
+### Installed mode — harbor native
+
+Built into harbor. The agent implements an `install()` method — harbor calls it during `setup()` to automatically install the agent inside the container at runtime (e.g., Node.js + openclaw via nvm). The Docker image stays clean; installation happens on every trial.
+
+```bash
+harbor run --agent openclaw ...
+```
+
+- **Who supports it:** Harbor out of the box via `--agent <name>`.
+- **Cost:** Agent is re-installed every trial (adds setup time).
+
+### External mode — harbor's extensibility API
+
+Harbor provides `BaseAgent` + `--agent`, an interface that lets you plug in any agent running outside the container — a host CLI, a remote agent behind an HTTP service, anything. The external agent communicates with the container through `environment.exec()`.
+
+```bash
+harbor run --agent-import-path external_agent.openclaw_external_agent:OpenClawExternalAgent ...
+```
+
+- **Who supports it:** Harbor's `BaseAgent` API. Anyone can implement an external agent.
+- **Advantage:** Install the agent once on host, skip per-trial install overhead. No runtime inside the container.
+
+### This repo
+
+A **working demo** of external mode, using OpenClaw as the host-side agent:
+
+```
+OpenClaw (host LLM) <-> OpenClawExternalAgent (router) <-> Harbor Docker container
+```
+
+The same pattern works for Claude Code, Codex, or any agent CLI.
+
+---
+
+## Install OpenClaw on host
+
+External mode runs the agent CLI on the host. For this repo's `openclaw_external_agent`, that means the `openclaw` CLI on your host PATH. If you write your own `xxx_external_agent.py`, install the corresponding agent CLI instead.
+
+```bash
+# Requires Node.js >= 22.19 (Node 24 recommended)
+npm install -g openclaw
+openclaw --version
+```
+
+See [OpenClaw docs](https://github.com/openclaw/openclaw) for API key and model setup.
+
+---
+
+## What's inside
+
+```
+terminal-runner/
+├── external_agent/
+│   ├── __init__.py                  # Package marker (no imports — lazy loading)
+│   └── openclaw_external_agent.py   # External agent adapter (extends harbor BaseAgent)
+├── scripts/
+│   └── setup.sh                     # Clone harbor upstream + dataset, install harbor
+├── .gitignore
+├── LICENSE
+├── README.md
+└── README_CN.md
+```
+
+### What's NOT here (cloned by `setup.sh`)
+
+- `harbor/` — upstream harbor framework
+- `dataset/` — terminal-bench task dataset
+
+---
+
+## Quick start
+
+### Prerequisites
+
+- Python ≥ 3.12
+- Docker ≥ 24 with the **compose v2 plugin** (`docker compose version` must work)
+- Node.js ≥ 22.19 + `openclaw` CLI on host PATH (external mode only)
+- `git`, `git-lfs`
+
+### Install
+
+```bash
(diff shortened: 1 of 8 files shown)


casparant:
Thanks for this adapter — the implementation is solid: complete type hints, well-factored constants, thorough docstrings, and three layered subprocess guards (total-timeout / no-output / stdout-cap). Comments grouped by severity below; nothing blocking from my side.

> The 3 **Major** items are posted as inline review comments on the relevant lines.

## Minor / Nit

- `scripts/setup.sh` L22: comment says `HARBOR_REF (default: main)` but the actual default (L34) is `v0.15.0`.
- `run()` raises `RuntimeError` when max iterations is reached (L243 + L578-588). For a benchmark, "not finished within N iterations" is closer to a task failure than an agent crash, and may be recorded by harbor as an agent error. Default `OPENCLAW_MAX_ITERATIONS=0` (unlimited) doesn't trigger it, hence minor.
- No-output guard only fires while `total_stdout_bytes == 0` (L926). If the process emits a few bytes and then the API hangs, only the total-timeout (600s) catches it. Consider basing it on "time since last output" instead.

## Project integration / conventions

- **commitlint scope**: `benchmark` isn't in `scope-enum` in `.github/commitlint.config.json` (nor in the `validScopes` lists in `prelint.yml`). It only emits a warning (scope-enum is level 1; per prelint.yml the only hard error is an empty scope), so it won't block the merge — but adding `benchmark` would clear the CI warning.
- **Doc naming**: this PR uses `README_CN.md`, while the existing `src/benchmark/README-CN.md` uses a hyphen and the repo root uses `README_zh.md`. Worth aligning on one convention.
- **Onboarding**: `benchmark` isn't listed in the root `AGENTS.md` / `CHANGELOG.md`. Please confirm whether it should be, per `specs/component-onboarding.md`.
- **OpenClaw references**: the README links `github.com/openclaw/openclaw` and `npm install -g openclaw`. Please verify the repo and the npm package are publicly available so the open-source docs don't point to unavailable resources.

## Highlights

- Apache-2.0 headers on every file; all-English comments.
- Empty-apiKey providers are filtered out to avoid API hangs; all subprocess calls use exec form (no shell) — no injection surface; API keys arrive via `--agent-env` and never hit disk.
- `.gitignore` explicitly excludes `jobs/`, `.env`, `*.key`, `harbor/`, `dataset/`.
- Per-trial `~/.openclaw-<profile>` isolation for safe concurrency.

Overall: good to merge after the 3 Major items (posted inline; especially #1 and #2); Minor and convention items can be same-PR or follow-up.

src/benchmark/terminal-runner/external_agent/openclaw_external_agent.py
```diff
@@ -0,0 +1,1365 @@
...
+     c. Router executes commands via ``environment.exec()`` inside the container.
+     d. Results are fed back to OpenClaw for the next iteration.
+  3. On completion, ``AgentContext`` is populated with execution metadata.
+
+This is *not* an installed agent (``BaseInstalledAgent``).  It extends
+``BaseAgent`` directly and is loaded at runtime via Harbor's
+``--agent`` mechanism:
+
+    harbor run --agent external_agent.openclaw_external_agent:OpenClawExternalAgent \
+               -p dataset/my-task -m openai/my-model ...
```
casparant, line 34: **[Major] Wrong harbor flag in the module docstring.**

This says the external agent is loaded via Harbor's `--agent` mechanism and shows `harbor run --agent external_agent...:OpenClawExternalAgent`. But the README (L27/L126) and `scripts/setup.sh` (L154) use `--agent-import-path`, and `--agent <name>` is reserved for installed mode. Please switch this to `--agent-import-path` so the docstring matches actual usage.

(The `OPENCLAW_AGENT_ID ... --agent flag` note in the env table refers to OpenClaw's own `--agent` at L408 and is correct — a different flag.)

src/benchmark/terminal-runner/README.md
```diff
@@ -0,0 +1,205 @@
...
+- Docker ≥ 24 with the **compose v2 plugin** (`docker compose version` must work)
+- Node.js ≥ 22.19 + `openclaw` CLI on host PATH (external mode only)
+- `git`, `git-lfs`
+
+### Install
+
+```bash
+git clone https://github.com/<your-org>/terminal-runner.git
+cd terminal-runner
+./scripts/setup.sh
```
casparant, line 97: **[Major] Placeholder / misleading clone URL.**

`git clone https://github.com/<your-org>/terminal-runner.git` — this component ships as a subdirectory (`src/benchmark/terminal-runner/`), not a standalone repo, so this won't work as written. Please reword to "clone the anolisa repo and cd into this subdirectory", or mark it clearly as a template placeholder. The same applies to `README_CN.md` L92-95.

src/benchmark/terminal-runner/external_agent/openclaw_external_agent.py
```diff
@@ -0,0 +1,1365 @@
...
+
+        return final_result
+
+    # ==================================================================
+    # Session token collection
+    # ==================================================================
+
+    def _collect_session_tokens(self, profile_dir: str) -> dict[str, Any] | None:
+        """Read OpenClaw session JSONL and extract per-round token usage."""
+        sessions_dir = os.path.join(profile_dir, "agents", "main", "sessions")
```
casparant: **[Major] Hard-coded `main` conflicts with the configurable `OPENCLAW_AGENT_ID`.**

`agent_id` is read from `OPENCLAW_AGENT_ID` (L393, default `main`), but this session-token path — and the profile dir at L280-283 — hard-code `main`. If a user sets `OPENCLAW_AGENT_ID` to anything else, token reporting silently returns nothing. Either derive these paths from `agent_id`, or document that agent-id is fixed to `main`.

casparant (commented):
Inline comments for the 3 Major items from my earlier review.

src/benchmark/terminal-runner/external_agent/openclaw_external_agent.py
yinan1995: Fixed. The docstring now uses --agent-import-path to match the actual usage in README and setup.sh.

src/benchmark/terminal-runner/README.md
yinan1995: Fixed in both README.md and README_CN.md. The clone URL now points to the anolisa monorepo with cd anolisa/src/benchmark/terminal-runner to enter the correct subdirectory.

src/benchmark/terminal-runner/external_agent/openclaw_external_agent.py
yinan1995: Fixed. Profile and session paths now derive from agent_id (read from OPENCLAW_AGENT_ID) instead of hard-coding "main". E2E verified — reward 1.0, PASS.

casparant:
**[Doc convention] `README_CN.md` should be renamed to `README_zh.md`.**

Per [`specs/documentation-standard.md`](https://github.com/alibaba/anolisa/blob/main/specs/documentation-standard.md) §1 (Bilingual Naming Convention), the Chinese counterpart of a component-root doc must use the `_zh.md` suffix. `*_CN.md` is explicitly listed as a **prohibited** legacy pattern that "must migrate to `*_zh.md`" — it is not just an alternative style. So `src/benchmark/terminal-runner/README_CN.md` should be renamed to `README_zh.md`.

While here, §1 also requires a cross-reference header linking the two language versions, which both files currently lack:

- in `README.md`: `[中文版](README_zh.md)`
- in `README_zh.md`: `[English](README.md)`

Worth adding in the same change.

(Note: this is a repo-wide convention; a few other components still carry legacy `_CN.md` / `-CN.md` names and should be migrated separately — not this PR's responsibility.)

src/benchmark/terminal-runner/external_agent/openclaw_external_agent.py
```diff
@@ -0,0 +1,1365 @@
...
+
+        hybrid_config: dict[str, Any] = {
+            "models": raw_models,
+            "agents": agents_config,
+        }
+        with open(profile_config_path, "w") as f:
+            json.dump(hybrid_config, f, indent=2)
+
+        main_models_path = os.path.expanduser(
+            "~/.openclaw/agents/main/agent/models.json"
```
casparant: 这是不是还漏了个

yinan1995: Good catch — that was a leftover from Major #3. Fixed in the latest commit.

casparant (approved):
LGTM

PR #1349, opened 2026-07-06
feat(sight): add conversation grader

haoowa:
## Description

Adds a manual AgentSight conversation grader MVP for persisted agent-quality evaluation. The change introduces a rule-based grader over GenAI and interruption evidence, SQLite-backed idempotent evaluation runs, dashboard controls for running and viewing evaluations, and ATIF deeplinks for evidence navigation.

It also fixes capture and scoring issues found while validating the grader flow: manually interrupted streamed calls are persisted as pending evidence, successful assistant text is no longer misclassified as runtime interruption text, and Anthropic request-side `tool_result` blocks are preserved as structured tool-call responses for deterministic tool-failure grading.

## Related Issue

Closes #1304

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional change)
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `sec-core` (agent-sec-core)
- [ ] `skill` (os-skills)
- [x] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [ ] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [x] My code follows the project's code style
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly
- [ ] For `cosh`: Lint passes, type check passes, and tests pass
- [ ] For `cosh-ng`: `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` pass
- [ ] For `sec-core` (Rust): `cargo clippy -- -D warnings` and `cargo fmt --check` pass
- [ ] For `sec-core` (Python): Ruff format and pytest pass
- [ ] For `skill`: Skill directory structure is valid and shell scripts pass syntax check
- [x] For `sight`: `cargo clippy -- -D warnings` and `cargo fmt --check` pass
- [ ] For `tokenless`: `cargo clippy -- -D warnings` and `cargo fmt --check` pass
[...]

PR #1393, opened 2026-07-08
fix(sight): avoid false interruption signals

haoowa:
## Description

Avoid false interruption records when a successful assistant response mentions troubleshooting words such as timeout, auth, or rate limit. The detector now only scans structured provider error text and non-2xx response bodies for runtime error classifications.

This PR is split out of #1349 so the interruption classifier fix can be reviewed independently. It also includes the current-toolchain mechanical AgentSight clippy fixes required for this split PR to pass `-D warnings` independently.

## Related Issue

Refs #1304

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `skill` (os-skills)
- [x] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [ ] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

## Testing

Ran on the Linux AgentSight test host from an archive of this split branch:

- `cargo test -p agentsight interruption::detector --locked --quiet`
- `cargo clippy -q -p agentsight --all-targets --locked -- -D warnings`

## Additional Notes

Split from #1349 following review feedback to keep capture fixes and the grader feature reviewable in smaller units.

### DIFF
--- a/src/agentsight/src/interruption/detector.rs
+++ b/src/agentsight/src/interruption/detector.rs
@@ -101,10 +101,15 @@ impl InterruptionDetector {
             .and_then(|s| s.parse().ok())
             .unwrap_or(200);
 
-        // 修复：从 call.response.raw_body 读取响应体，而非 call.metadata（builder 不会写入 metadata）
         let error_text = call.error.as_deref().unwrap_or("");
         let response_body = call.response.raw_body.as_deref().unwrap_or("");
-        let combined_error = format!("{error_text} {response_body}").to_ascii_lowercase();
+        let structured_error = error_text.to_ascii_lowercase();
+        let response_error_body = if status_code >= 400 {
+            response_body
+        } else {
+            ""
+        };
+        let combined_error = format!("{error_text} {response_error_body}").to_ascii_lowercase();
 
         let is_context_overflow = combined_error.contains("context_length_exceeded")
             || combined_error.contains("maximum context length")
@@ -174,9 +179,9 @@ impl InterruptionDetector {
         // ── 3. NetworkTimeout (408/504 / timeout) ─────────────────────────────
         if status_code == 408
             || status_code == 504
-            || combined_error.contains("timeout")
-            || combined_error.contains("timed out")
-            || combined_error.contains("deadline exceeded")
+            || structured_error.contains("timeout")
+            || structured_error.contains("timed out")
+            || structured_error.contains("deadline exceeded")
         {
             let detail = serde_json::json!({
                 "model": call.model,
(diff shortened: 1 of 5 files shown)


jfeng18:
Thanks for splitting this out — the false-positive suppression goal makes sense. One concern on the recall side:

After this change `combined_error` only folds in the response body when `status_code >= 400` (detector.rs:107-112), and `call.error` is itself only populated for `status_code >= 400` (call_builder.rs:137). So for an HTTP 200 stream that commits the status line and *then* emits a mid-stream error envelope, both inputs to `combined_error` are empty — none of the AuthError/RateLimit/ServiceUnavailable/ContextOverflow keyword rules, nor the generic LlmError fallback (detector.rs:284), can fire.

Concretely: Anthropic Messages streaming returns 200 then `event: error` / `{"type":"error","error":{"type":"overloaded_error"}}`; OpenAI-compatible proxies stream `{"error":{...}}` chunks at 200. These land in `raw_body` and were classified before this PR (e.g. ServiceUnavailable). After it they are either reclassified as `SseTruncated` (when duration ≥ `sse_min_duration_ns`) or, for fast-fail (<1s) errors, produce no interruption event at all. Since agents almost always stream, this looks like the common interruption path.

(Reproduced pre/post against the SSE parse path.) Flagging in case the 200-status-with-error-body case is collateral rather than intended.

jfeng18:
Thanks — this closes the recall regression I raised in round 1. I verified by building and running this time, not just reading the diff:

- `cargo test --lib interruption::detector` → **42 passed / 0 failed**; both new tests (`test_detects_anthropic_sse_error_event_in_200_stream`, `test_detects_structured_sse_error_object_in_200_stream`) green.
- Mutation-checked the fix: neutralizing `detector.rs` `stream_error` extraction to `None` flips both tests (detected 0 vs expected 1), so they genuinely gate the fix rather than passing formally.

Also confirmed my main worry does **not** happen: a >1s 200-SSE error stream is not double-classified — rule 7 (`stream_error.is_some()`) emits one event and returns before the SseTruncated rule is reached, so exactly one event fires.

Two non-blocking notes for later (not for this PR):
1. Extraction is gated on `status_code < 400 && is_sse`, so a non-SSE HTTP 200 carrying a buffered (non-streamed) error envelope is still not scanned — same recall class, different transport. Probably a separate follow-up rather than widening this diff.
2. The `test_ignores_*` false-positive guards all run with `is_sse=false`, so they don't exercise the new SSE stream_error path; the behavior is correct today but not locked by a committed test.

LGTM on the recall fix.

jfeng18 (approved):
Recall regression from round 1 is closed and verified by build + tests + mutation (42 passed; neutralizing the fix flips the two new SSE tests). No blockers; the two notes above are non-blocking follow-ups. LGTM.

PR #1394, opened 2026-07-08
fix(sight): persist idle streams and tool results

haoowa:
## Description

Persist idle in-flight streamed calls so manually interrupted conversations remain visible as pending evidence instead of being silently dropped. This also preserves Anthropic `tool_result` blocks as structured input message parts so downstream scoring can distinguish actual tool failures from historical raw context.

This PR also adds regression coverage for oversized request-body eviction and includes the current-toolchain mechanical AgentSight clippy fixes required for this split PR to pass `-D warnings` independently.

## Related Issue

Refs #1304

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `skill` (os-skills)
- [x] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [ ] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

## Testing

Ran on the Linux AgentSight test host from an archive of this split branch:

- `cargo test -p agentsight aggregator --locked --quiet`
- `cargo test -p agentsight genai::call_builder --locked --quiet`
- `cargo clippy -q -p agentsight --all-targets --locked -- -D warnings`

## Additional Notes

Split from #1349 following review feedback to keep capture fixes separate from the grader API and dashboard work.

### DIFF
--- a/src/agentsight/src/aggregator/http/aggregator.rs
+++ b/src/agentsight/src/aggregator/http/aggregator.rs
@@ -85,6 +85,8 @@ pub struct HttpConnectionAggregator {
     last_appended_src_ptr: LruCache<ConnectionId, usize>,
     /// Last activity timestamp per connection, used for idle timeout eviction.
     last_activity: LruCache<ConnectionId, Instant>,
+    /// Connections already snapshotted after idle timeout.
+    idle_snapshotted: LruCache<ConnectionId, ()>,
     /// Maximum bytes buffered per connection (request body or compressed SSE).
     max_body_bytes: usize,
     /// Idle timeout before a connection state is forcibly dropped.
@@ -142,6 +144,7 @@ impl HttpConnectionAggregator {
             sse_continuation_buffers: LruCache::new(cap),
             last_appended_src_ptr: LruCache::new(cap),
             last_activity: LruCache::new(cap),
+            idle_snapshotted: LruCache::new(cap),
             max_body_bytes: max_body_bytes.max(1024),
             idle_timeout,
         }
@@ -203,6 +206,7 @@ impl HttpConnectionAggregator {
                 self.sse_continuation_buffers.pop(key);
                 self.last_appended_src_ptr.pop(key);
                 self.last_activity.pop(key);
+                self.idle_snapshotted.pop(key);
                 evicted_idle += 1;
             }
         }
@@ -232,6 +236,7 @@ impl HttpConnectionAggregator {
             self.sse_continuation_buffers.pop(key);
             self.last_appended_src_ptr.pop(key);
             self.last_activity.pop(key);
+            self.idle_snapshotted.pop(key);
         }
         if !oversized.is_empty() {
             log::warn!(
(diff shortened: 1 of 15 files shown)


jfeng18:
Re-checked this one closely since it's timing-bounded — it holds, with a scoping note.

**New in this PR:** `drain_and_persist_idle_connections` (unified.rs:1287) persists a fresh-`generate_id()` pending row for a *live* process once a stream idles past `idle_timeout` (60s, config.rs:61), and pops the connection from the aggregator. Since `complete_pending` keys on the real call_id, that orphan row is never resolved even if the stream resumes and completes successfully.

**Consequence:** `ProcMonEvent::Exit` carries no exit code (procmon.rs:47-53), so `handle_agent_crash_detection` (unified.rs:861) fires on every tracked-agent exit and sweeps `status='pending'` rows for the pid into `InterruptionType::AgentCrash` (Critical, types.rs:77), timestamped at exit. So a stream that idled >60s, resumed and completed fine, followed by a clean `/exit` within ~[t0+60s, t0+300s], yields a spurious Critical agent_crash on a *successful* session. After ~t0+300s the stale scanner flips the row to `interrupted` and it no longer trips the sweep, so the window is ~4 min.

**Attribution, to be fair:** the exit-without-exit-code gate and "pending row at clean exit → agent_crash" behavior already exist on `main`; this PR doesn't introduce those. What's new is a durable, unresolvable pending row for a *successful* idled-then-resumed stream, which previously left no row at all. Flagging since a false Critical on a clean exit is costly for a detection product; the deeper "Exit has no exit code" limitation looks pre-existing and probably a separate discussion.

jfeng18:
The idle-drain fix closes the finding I raised — a clean process exit no longer sweeps idle-drained rows into a spurious Critical `agent_crash`. I built and ran it rather than just reading:

- `cargo test --lib storage::sqlite::genai` → **51 passed / 0 failed**; `test_crash_sweep_ignores_idle_drain_pending`, `test_list_pending_for_pid(s)`, and `test_complete_pending_promotes_idle_snapshot_by_match_key` all green.
- Mutation-checked the gate: removing the `AND pending_origin != 'idle_drain'` filter from `list_pending_for_pid` flips `test_crash_sweep_ignores_idle_drain_pending` (2 rows vs expected 1), so the exclusion is genuinely what prevents the false crash.

Both halves of the original concern are addressed: the `PendingOrigin::IdleDrain` tag + its exclusion from the crash sweep stops the spurious `agent_crash`, and `pending_match_key` + the non-destructive snapshot let a resumed stream reconcile the row instead of it being an unresolvable orphan. Nice.

One thing blocking merge, though — CI is red on **Commit Message Lint** (not the tests; `Test agentsight` passes). From the job log:

```
✖  body's lines must not be longer than 100 characters [body-max-line-length]
✖  found 1 problems, 0 warnings
```

Both commits trip it — the bodies have lines of 146/359 chars (`preserve idle stream snapshots`) and 120 chars (`persist idle streams and tool results`). Rewrapping the two commit bodies at ≤100 chars and force-pushing should turn it green; happy to approve once CI is clean.

haoowa:
Additional note on the code changes made for the earlier #1394 review feedback.

The review concern was that persisting idle streams could make an active stream
look like a crash-drained pending call, and that a resumed stream might not be
able to reconnect to the idle snapshot later.

The PR now addresses that by:

- Marking idle-timeout persisted rows explicitly as `idle_drain` via
  `pending_origin`, separate from normal request capture and dead-PID drains.
- Adding a deterministic `pending_match_key` so a later completed stream can
  promote the previously persisted idle snapshot instead of leaving duplicate
  or stale pending evidence.
- Changing idle timeout handling from destructive drain to non-destructive
  snapshot, so `SseActive` state remains in memory and the stream can continue
  if more SSE chunks arrive.
- Excluding `idle_drain` rows from crash/dead-PID pending sweeps, so a live but
  idle stream is not reclassified as an agent-crash artifact.
- Adding SQLite schema migrations and regression tests for:
  - persisted idle origin and match key
  - promotion by match key
  - crash sweep ignoring idle snapshots
  - resumed SSE streams after idle snapshot

Validation was run on the Linux test host:

- `cargo test --quiet`
- `cargo clippy --all-targets -- -D warnings`

The latest force-push only wrapped commit message bodies for commitlint; it did
not change the code tree from the reviewed fix.

jfeng18 (approved):
Idle-drain finding closed and verified (build +  51 passed + mutation on the  filter). CI now green after the commit-body rewrap; code is byte-identical to the version I verified. LGTM.

PR #1409, opened 2026-07-09
fix(ci): prefer VPC mirror on ARC runners

kongche-jbw:
## Summary

- Prefer `mirrors.cloud.aliyuncs.com` for ARC/ECS Alinux jobs.
- Fall back to public `mirrors.aliyun.com` after two failed VPC mirror attempts.
- Pre-import the Alinux GPG key before `dnf install` to avoid late transaction failures.
- Keep GitHub-hosted manual runs on the public mirror.

## Validation

- `git diff --check`
- YAML parse check for updated workflows

### DIFF
--- a/.github/workflows/_rpm-build.yaml
+++ b/.github/workflows/_rpm-build.yaml
@@ -41,15 +41,54 @@ jobs:
     container:
       image: alibaba-cloud-linux-4-registry.cn-hangzhou.cr.aliyuncs.com/alinux4/alinux4:latest
     steps:
-      - name: Fix repo mirror
-        run: sed -i -e "s/cloud.aliyuncs/aliyun/g" /etc/yum.repos.d/*.repo
-
       - name: Install build dependencies
+        env:
+          USE_VPC_MIRROR: "true"
+        shell: bash
         run: |
+          set -euo pipefail
+
+          configure_repo_host() {
+            local host="$1"
+            sed -i \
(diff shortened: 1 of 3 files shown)


kongche-jbw:
Verified on the ARC runner that the workflow now uses the VPC mirror instead of the public mirror.

Evidence from the latest run:
- Job: https://github.com/alibaba/anolisa/actions/runs/28985058137/job/86012133929?pr=1409
- Runner: `anolisa-k8s-general-ci-x64-4vmvk-runner-gldw8`
- Log shows:
  ```text
  USE_VPC_MIRROR: true
  Trying VPC Alinux mirror (1/2)
  Importing Alinux GPG key from http://mirrors.cloud.aliyuncs.com/alinux/4/RPM-GPG-KEY-ALINUX-4
  ```
- No fallback log appeared:
  ```text
  VPC Alinux mirror failed; falling back to public mirror
  ```
- Repo metadata download also completed via the selected mirror with normal speed:
  ```text
  alinux4-os       331 MB/s | 11 MB | 00:00
  alinux4-updates   50 MB/s | 14 MB | 00:00
  ```

So ARC jobs are using the ECS/VPC mirror successfully, while GitHub-hosted manual runs still use the public mirror path.

kongche-jbw:
网络偶发导致问题，暂时关闭此PR

PR #1412, opened 2026-07-09
feat(anolisa): add scoped installed state view

ikunkun-sys:
## Description

Add a shared CLI `StateView` for scoped installed-state reads. User-mode read
views now load the current user state plus readable system state, while system
mode reads only system state. The `list` command uses this projection so a
normal user can discover system-provided components without implying ownership.

List rows now expose:

- `scope`: `user`, `system`, or `none`
- `active`: whether the row is the selected visible lifecycle record
- `mutable_by_current_invocation`: whether the current command may mutate it
- `state_path`: the physical `installed.toml` backing active rows

This is the first PR in the dual-scope installed-state plan. It intentionally
keeps mutation behavior unchanged: lifecycle writes still go only to the
current install mode's physical state file. `status`, `doctor`, adapter source
health, orphaned adapter receipts, and lifecycle wrong-scope guards are left to
later PRs.

## Related Issue

N/A

## Ship Note

`anolisa list --installed` in user mode can now show system-provided installed
components. System rows are marked as system scope and immutable for the user
invocation, while user rows remain user-scoped and mutable. Malformed writable
state remains fatal; malformed visible system state is surfaced as a warning for
read-only list output.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `anolisa` (anolisa-cli)

## Testing

- `cargo check`
- `cargo test -p anolisa-cli state_view` (6 tests)
- `cargo test -p anolisa-cli list` (40 tests)
- `cargo test -p anolisa-cli` (423 tests)
- `cargo doc --no-deps`
- Local CLI QA: temp user/system `installed.toml` files, verifying system mode
  does not read user state, user mode shows system rows as immutable, and
  malformed visible system state returns a warning.
- Remote QA on `Alinux 4`: root/system mode sees only the system row;
  `nobody`/user mode sees both the system row and the user row with correct
  scope, mutability, and `state_path` fields.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands.rs
@@ -6,6 +6,7 @@
 //!   / runtime / osbase). Each surface uses its own appropriate vocabulary.
 
 pub mod common;
+pub(crate) mod state_view;
 pub mod tier1;
 
 // Tier 2 surfaces
@@ -45,7 +46,7 @@ pub struct Cli {
     #[arg(long, global = true, value_enum)]
     pub install_mode: Option<InstallMode>,
 
-    /// Custom install prefix (system-mode only)
+    /// Custom system install prefix, including user-mode system visibility
     #[arg(long, global = true, value_name = "PATH")]
     pub prefix: Option<PathBuf>,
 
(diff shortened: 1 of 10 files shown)


kongche-jbw (approved):
lgtm

PR #1416, opened 2026-07-09
chore(ci): relax commit body line-length to warning

samchu-zsl:
## Description

The `config-conventional` preset inherited by `.github/commitlint.config.json` enforces `body-max-line-length` at 100 as a hard error, which blocked commits that legitimately need long single lines (pasted URLs, stack traces). This change adds an explicit override downgrading the rule to a warning (level 1) at the same 100-char limit, so contributors get nudged without being blocked. `header-max-length` and the scope rules remain hard errors.

## Related Issue

no-issue: minor CI lint config tweak

## Type / Scope

- [x] ci

## Testing

Config-only change. Verified `body-max-line-length` is now level 1 (warning) in `.github/commitlint.config.json`; `wagoid/commitlint-github-action` defaults to `failOnWarnings: false`, so warnings no longer fail the lint job.

### DIFF
--- a/.github/commitlint.config.json
+++ b/.github/commitlint.config.json
@@ -33,6 +33,11 @@
       2,
       "always",
       120
+    ],
+    "body-max-line-length": [
+      1,
+      "always",
+      100
     ]
   }
 }


casparant:
Accepting this change — the problem is real and the fix is verified: `@commitlint/config-conventional` ships `body-max-line-length` as a hard error, and `wagoid/commitlint-github-action@v6` defaults to `failOnWarnings: false`, so downgrading to level 1 unblocks long pasted URLs / stack traces while keeping the nudge. It also matches the prelint intent that only scope enforcement stays a hard error.

One suggestion (non-blocking): consider downgrading `footer-max-line-length` to a warning as well. The preset keeps it at level 2, and our commits rely heavily on trailers (`Fixes: <hash> ("<subject>")`, `Assisted-by`, `Signed-off-by`), so a long referenced subject or URL in the footer would still hard-fail the lint job for the same reason this PR addresses in the body.

casparant (approved):

Issue #1410, opened 2026-07-09
[anolisa] feat(anolisa): add RPM-only update check for image upgrades


kongche-jbw:
## Background

Operations need a lightweight way to tell users that an older ANOLISA image, such as a v0.5 image, can be upgraded to the current v1.0 toolchain. Today operators can run package-manager commands manually, for example `yum upgrade --disablerepo=* --enablerepo=alinux4-agentic-os`, but users do not get a login-time signal and new default components still require manual discovery.

This issue proposes a read-only RPM-focused check surface that can be used by MOTD and by users before running a real upgrade.

## Suggested command shape

```bash
anolisa update --check
anolisa update --check --json
anolisa update --check --motd
anolisa update --check --refresh
anolisa update --check --target image-v1.0
```

Suggested parameter semantics:

| Parameter | Suggested behavior |
|---|---|
| `--check` | Enter read-only update detection mode. It must not mutate rpmdb, ANOLISA state, repo config, adapter state, or component files. |
| `--json` | Emit a structured machine-readable report for automation and MOTD cache generation. |
| `--motd` | Emit short login-friendly text. It should be safe for `/etc/profile.d` / MOTD integration and avoid noisy stack-like errors. |
| `--refresh` | Ignore the local check cache and query the configured RPM repository again. |
| `--target <PROFILE>` | Select the target image/toolchain profile, defaulting to the repository-declared default profile. |
| `--install-mode system` | The first phase should support only system/RPM image scenarios. User mode should return an explicit unsupported result. |

## Suggested detection model

The check should report three classes of work:

1. CLI package update availability
   - Detect whether the running `anolisa` executable is owned by an RPM package.
   - Compare the installed RPM version with candidates from the configured `[backends.rpm]` repository.
   - Report `update`, `noop`, `unsupported`, or `error`.

2. Installed RPM component updates
   - Inspect `installed.toml` and rpmdb for `rpm-managed` / `rpm-observed` ANOLISA components.
   - Query available candidates from the configured RPM repository.
   - Report components that can be upgraded.

3. Missing default components for the target image profile
   - Load a repository-declared image/toolchain profile such as `image-v1.0`.
   - Detect profile components that are absent from rpmdb / ANOLISA state.
   - Report them as install candidates, not as already-installed updates.

## Suggested JSON shape

```json
{
  "target": "image-v1.0",
  "backend": "rpm",
  "upgrade_available": true,
  "cli": {
    "package": "anolisa",
[...]

PR #1418, opened 2026-07-09
feat(anolisa): add rpm image upgrade

### LINKED ISSUE #1411: [anolisa] feat(anolisa): add RPM image upgrade command
## Background

Operators need a one-command path to upgrade older ANOLISA images, such as v0.5 images, to the current v1.0 toolchain. Manual package-manager upgrades can update already-installed RPMs, but they do not provide an ANOLISA-level workflow for missing default components, state refresh, result reporting, or user guidance.

This issue proposes a first-phase RPM-only `anolisa upgrade` command that consumes the plan produced by the update-check work and applies it for system images.

Related detection issue: #1410.

## Suggested command shape

```bash
sudo anolisa upgrade
sudo anolisa upgrade --dry-run
sudo anolisa upgrade --json
sudo anolisa upgrade --target image-v1.0
sudo anolisa upgrade --assume-yes
```

Suggested parameter semantics:

| Parameter | Suggested behavior |
|---|---|
| `upgrade` | New top-level command for image/toolchain upgrade orchestration. It should not replace `update self` or `update <component>`. |
| `--dry-run` | Build and render the upgrade plan without running `dnf`, writing ANOLISA state, or touching adapter/framework state. |
| `--json` | Emit a structured plan/result for automation. |
| `--target <PROFILE>` | Select the target image/toolchain profile, defaulting to the repository-declared default profile. |
| `--assume-yes` / `-y` | Allow non-interactive execution once the plan is known. |
| `--install-mode system` | The first phase should support only system/RPM image scenarios. User mode should fail clearly. |

## Suggested execution model

The command should be an orchestration layer over RPM-backed operations, not a generic backend migration mechanism.

Suggested phases:

1. Preflight
   - Require root for real execution.
   - Require rpm/dnf tooling.
   - Load `repo.toml` and the configured `[backends.rpm]` repository.
   - Load the target image/toolchain profile.
   - Acquire the ANOLISA install lock before mutating state.

2. Plan
   - Reuse the planner behind `anolisa update --check`.
   - Compute:
     - `to_update`: installed RPM-backed components and CLI packages with newer repo candidates.
     - `to_install`: target-profile default components missing from the image.
     - `unsupported`: raw-managed or otherwise non-RPM components that must not be touched in phase one.
     - `errors`: ambiguous package mapping, missing package metadata, repo query failures, or rpmdb drift.

3. Apply RPM transactions
   - Update the CLI package when it is RPM-owned.
   - Update already-installed RPM-backed components.
   - Install missing default components from the target profile.
   - Delegate file transactions to dnf; ANOLISA should not directly modify RPM-owned files.

4. Refresh ANOLISA state
   - Re-read rpmdb after successful transactions.
   - Refresh versions, architecture, source repo, ownership, and install backend in ANOLISA state.
   - Record newly installed default components as RPM-managed or the agreed RPM ownership model.
   - Write central log entries and a last-upgrade report.

5. Post-upgrade reporting
   - Print a concise summary of updated, installed, skipped, unsupported, and failed items.
   - Make `anolisa status` sufficient for users to inspect final state.
   - Surface adapter/framework reconciliation gaps as warnings rather than silently claiming success.

## Suggested JSON result shape

```json
{
  "target": "image-v1.0",
  "backend": "rpm",
  "status": "ok",
  "dry_run": false,
  "updated": [
    {
      "name": "tokenless",
      "package": "tokenless",
      "from": "0.5.0-1.al4",
      "to": "1.0.0-1.al4"
    }
  ],
  "installed": [
    {
      "name": "agent-memory",
      "package": "agent-memory",
      "version": "1.0.0-1.al4"
    }
  ],
  "skipped": [
    {
      "name": "local-tool",
      "reason": "raw-managed components are out of scope for RPM image upgrade"
    }
  ],
  "warnings": []
}
```

## Adapter and environment reconciliation

The upgrade command should make the environment-adaptation boundary explicit.

Suggested first-phase behavior:

- Detect installed agent/framework environments and existing adapter claims.
- Preserve existing user/framework configuration.
- Report adapter reconciliation needs in the result.
- Do not silently rewrite framework configuration unless the existing adapter subsystem already owns the specific claim/resource.
- Treat automatic full adapter repair as a follow-up unless the implementation can reuse existing safe `adapter enable/disable/status` semantics.

## Explicit first-phase boundaries

- RPM/system image scenario only.
- Do not update raw-managed components.
- Do not migrate raw-managed components to RPM.
- Do not remove or overwrite raw-managed component files.
- Do not implement npm/raw upgrade in this issue.
- Do not make `update all` part of this first phase unless it becomes a thin alias to the same RPM-only planner with identical boundaries.
- Do not change bare `anolisa update` semantics.
- Do not perform automatic cascade adapter cleanup or broad framework rewrites as part of RPM upgrade.

For raw-managed components, the command should skip them with a clear reason such as `raw-managed components are out of scope for RPM image upgrade`.

## Acceptance criteria

- `sudo anolisa upgrade --dry-run` shows the RPM image upgrade plan without changing host state.
- `sudo anolisa upgrade` can update already-installed RPM-backed ANOLISA components to the configured repository candidates.
- Missing default components from the target profile can be installed from the RPM repository.
- The command refreshes ANOLISA state after RPM transactions so `anolisa status` shows the upgraded result.
- Non-root execution fails with an actionable `sudo anolisa upgrade` hint.
- raw-managed components are skipped and reported, not updated or migrated.
- Unit tests cover argument parsing, dry-run planning, root gating, RPM transaction ordering, state refresh, missing default installs, partial failure reporting, and raw boundary behavior.

kongche-jbw:
## Description

This PR adds the first-phase `anolisa upgrade` command for RPM/system image
upgrade scenarios. The command reuses `update --check` as the read-only planner,
then applies CLI RPM updates, installed RPM component updates, and resolved
target-profile default installs through dnf.

Raw-managed components remain explicitly out of scope and are skipped without
migration or mutation. Upgrade results are refreshed into ANOLISA state where
applicable, and durable operation/audit records reflect `ok`, `partial`, or
`failed` outcomes.

## Related Issue

closes #1411

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope
- [x] `anolisa` (anolisa-cli)

## Checklist

## Testing

- `cargo clippy -p anolisa-cli --all-targets --locked -- -D warnings`
- `cargo test -p anolisa-cli --locked`
- `cargo doc -p anolisa-cli --no-deps`

Covered parse behavior, dry-run/root gating, transaction ordering, missing
default installs, raw-managed skip boundaries, state drift handling, partial
failure reporting, and CLI-only upgrade audit records.

## Additional Notes

First-phase scope is limited to RPM/system image upgrades. Raw-managed, npm, and
raw backend upgrade or migration paths are intentionally out of scope.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands.rs
@@ -100,6 +100,8 @@ pub enum ComponentCommands {
     Restart(tier1::restart::RestartArgs),
     /// Update a component (`update <component>`), the CLI binary (`self`), or everything (`all`)
     Update(tier1::update::UpdateArgs),
+    /// Upgrade the RPM/system image to the target toolchain profile (system-only)
+    Upgrade(tier1::upgrade::UpgradeArgs),
     /// Reconcile a component's ANOLISA state with rpmdb after manual RPM changes
     Repair(tier1::repair::RepairArgs),
     /// Drop a component's ANOLISA state record without any package operation
@@ -208,6 +210,7 @@ pub fn dispatch(cli: Cli, ctx: &CliContext) -> Result<(), CliError> {
             ComponentCommands::Logs(args) => tier1::logs::handle(args, ctx),
             ComponentCommands::Restart(args) => tier1::restart::handle(args, ctx),
             ComponentCommands::Update(args) => tier1::update::handle(args, ctx),
+            ComponentCommands::Upgrade(args) => tier1::upgrade::handle(args, ctx),
             ComponentCommands::Repair(args) => tier1::repair::handle(args, ctx),
             ComponentCommands::Forget(args) => tier1::forget::handle(args, ctx),
             ComponentCommands::Adopt(args) => tier1::adopt::handle(args, ctx),
@@ -399,6 +402,11 @@ fn command_policy(command: &Commands) -> CommandPolicy {
                 CommandPolicy::new("update", CommandScope::ReadOnly)
             }
             ComponentCommands::Update(_) => mode_scoped("update", true),
+            // `upgrade` is a system-only RPM image mutation: real execution
+            // needs root. `--dry-run` waives only the root requirement, not the
+            // system-mode one, so an explicit system-mode dry-run can preview
+            // the plan without root (matching `repair`).
+            ComponentCommands::Upgrade(_) => system_only("upgrade", true),
             ComponentCommands::Repair(_) => system_only("repair", true),
             ComponentCommands::Forget(_) => mode_scoped("forget", true),
             ComponentCommands::Adopt(_) => system_only("adopt", false),
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1.rs
@@ -13,6 +13,7 @@ pub mod restart;
 pub mod status;
 pub mod uninstall;
 pub mod update;
+pub mod upgrade;
 
 // Cross-command end-to-end MVP lifecycle coverage (#963); test-only.
 #[cfg(test)]
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
@@ -71,7 +71,9 @@ use crate::context::CliContext;
 use crate::repo_config::{HostVars, RepoConfig};
 use crate::response::{self, CliError};
 
-mod check;
+// `pub(crate)` so `anolisa upgrade` (issue #1411) can reuse the read-only
+// planner (`check::compute_update_check_report`) instead of re-deriving it.
+pub(crate) mod check;
 
 /// Command label for JSON envelopes and error routing.
 const COMMAND: &str = "update";
@@ -342,7 +344,9 @@ fn resolve_update_target(
     }
 }
 
-fn rpm_repo_source_for_update(
+// `pub(crate)`: shared with `check` (read-only) and `upgrade` (issue #1411) so
+// all three resolve the configured ANOLISA RPM repository the same way.
+pub(crate) fn rpm_repo_source_for_update(
     repo_config: &RepoConfig,
     env: &anolisa_env::EnvFacts,
     command: &str,
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/update/check.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/update/check.rs
@@ -83,22 +83,26 @@ const BUILTIN_DEFAULT_TARGET_PROFILE: &str = include_str!(concat!(
 const PROFILES_SUBDIR: &str = "profiles";
 
 // Stable action vocabulary shared by JSON, cache, and human/MOTD rendering.
-const ACTION_UPDATE: &str = "update";
-const ACTION_NOOP: &str = "noop";
-const ACTION_INSTALL: &str = "install";
-const ACTION_UNSUPPORTED: &str = "unsupported";
-const ACTION_UNSUPPORTED_RPM: &str = "unsupported_in_rpm_upgrade";
-const ACTION_ERROR: &str = "error";
+// `pub(crate)` so the `upgrade` command (issue #1411) can classify a report's
+// items against the same vocabulary the check produces.
+pub(crate) const ACTION_UPDATE: &str = "update";
+pub(crate) const ACTION_NOOP: &str = "noop";
+pub(crate) const ACTION_INSTALL: &str = "install";
+pub(crate) const ACTION_UNSUPPORTED: &str = "unsupported";
+pub(crate) const ACTION_UNSUPPORTED_RPM: &str = "unsupported_in_rpm_upgrade";
+pub(crate) const ACTION_ERROR: &str = "error";
 
 /// Wire shape for `update --check` (`--json`) and the on-disk cache.
 ///
 /// Owned `String`s (rather than `&'static str`) so the same struct round-trips
-/// through the cache via `Deserialize`.
+/// through the cache via `Deserialize`. Exposed `pub(crate)` so `anolisa
+/// upgrade` (issue #1411) can consume the same planning output; only the fields
+/// the upgrade planner reads are widened past module scope.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-struct UpdateCheckReport {
+pub(crate) struct UpdateCheckReport {
     /// Target profile evaluated, when `--target` was given.
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    target: Option<String>,
+    pub(crate) target: Option<String>,
     /// Upgrade backend this report covers. Always `rpm` in the first version.
     backend: String,
     /// True when at least one component/CLI `update` was found. Deliberately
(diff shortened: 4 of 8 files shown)


ikunkun-sys:
Review findings:

1. `upgrade` mishandles installed-but-unadopted target defaults.

`update --check` intentionally treats a target default that is already present in rpmdb but absent from ANOLISA state as an `ACTION_UPDATE` row (`update/check.rs:646`, `update/check.rs:831`). `upgrade` then puts every `ACTION_UPDATE` component into `plan.updates` (`upgrade.rs:263`) and, after `dnf update` succeeds, `finalize_upgrade` requires a pre-existing state row (`upgrade.rs:820`). For baked-in or otherwise unadopted defaults, that row never existed, so the command mutates rpmdb and then reports failure with “component vanished from ANOLISA state”. The apply path needs to distinguish this observed-default update case and either record/adopt it after the transaction or plan it through a separate path that does not require existing state.

2. Missing-default installs bypass the existing RPM resolver and lose supported fallbacks.

For `ACTION_INSTALL`, the package is populated only by `index_rpm_packages()` from the component index (`update/check.rs:661`). If the repo lacks `components.toml` but the component can be resolved through existing supported mechanisms such as `[backends.rpm.package_map]` or available `Provides: anolisa-component(...)`, `component.package` stays `None`, and `upgrade` turns the install into a blocking plan error (`upgrade.rs:271`). The normal RPM resolver already supports these fallbacks (`resolution.rs:304` for package_map and `resolution.rs:333` for available provides, with coverage at `resolution.rs:1075`). `upgrade` should reuse that resolver for missing defaults instead of implementing an index-only subset.

I verified the current PR snapshot with `cargo test -p anolisa-cli upgrade --locked` and `cargo run -p anolisa-cli --locked -- upgrade --help`; both pass, so these are uncovered behavior gaps rather than compile/test failures.

kongche-jbw:
Thanks for the detailed review. Both gaps are fixed in 7e669642.

1. Installed-but-unadopted target defaults are now distinguished from normal
state-backed component updates. `update --check` marks target-default rows that
were absent from ANOLISA state, and `upgrade` uses that hint to record the
post-transaction package as `rpm-observed` / adopted when the state row is still
missing. Normal component updates still treat a missing state row as drift.

2. Missing-default package resolution now uses the shared RPM `ComponentResolver`
instead of the previous index-only lookup. This preserves the existing
`components.toml`, `[backends.rpm.package_map]`, and available
`Provides: anolisa-component(...)` fallback chain.

Added regression coverage for:
- package_map resolving a missing default package
- available Provides resolving a missing default package
- package_map identifying an installed-but-unadopted default without Provides
- observed default update recording rpm-observed state instead of failing

Validation:
- cargo fmt --all -- --check
- cargo clippy -p anolisa-cli --all-targets --locked -- -D warnings
- cargo test -p anolisa-cli --locked
- cargo doc -p anolisa-cli --no-deps

ikunkun-sys:
Review update for latest head `7e669642`:

The two previous findings are mostly addressed: missing-default package resolution now uses the shared RPM resolver, and observed defaults with an update candidate can be recorded as `rpm-observed`.

I found one remaining gap:

`upgrade` still drops installed-but-unadopted target defaults when they are already up to date.

`update --check` can now mark a target default that is present in rpmdb but absent from ANOLISA state with `absent_from_state = true`. However, if that package is already current, `check_present_default` returns `ACTION_NOOP` (`update/check.rs:921-929`), and `build_plan` discards all `ACTION_NOOP` rows (`upgrade.rs:293`). That means a baked-in/default RPM that is already up to date is never written into ANOLISA state as `rpm-observed`; `anolisa upgrade` exits successfully but does not complete the adopt/record part for that default.

The new test covers `ACTION_UPDATE + absent_from_state` (`upgrade/tests.rs:786`), but not `ACTION_NOOP + absent_from_state`. I think absent-from-state noops should still be planned through a no-dnf observed/adopt state refresh path, or otherwise surfaced as a result that records the default.

Verification on the latest PR snapshot:

- `cargo test -p anolisa-cli upgrade --locked` passed: 23 tests
- `cargo test -p anolisa-cli update_check --locked` passed: 43 tests
- `cargo run -p anolisa-cli --locked -- upgrade --help` rendered normally

kongche-jbw:
Fixed, thanks. The noop-but-absent target default is no longer dropped from
the upgrade plan.

If the package is already present in rpmdb and no update is available,
`anolisa upgrade` now records it as rpm-observed state without running a dnf
transaction. I also added a regression test:
`observed_default_noop_missing_from_state_is_recorded_without_dnf`.

All local checks pass.

ikunkun-sys:
Review findings for the latest head `5c0964fe1da040bb0e1d4e83be3f8acd687bdb93`:

1. **Block RPM mutations on stale ANOLISA state before running `dnf`.**

   `upgrade` currently runs privileged RPM transactions before acquiring the ANOLISA install lock and revalidating the current state:

   - `upgrade.rs:546` updates the CLI package.
   - `upgrade.rs:570` updates RPM-backed components.
   - `upgrade.rs:604` installs missing defaults.
   - `upgrade.rs:900` acquires the install lock only later in `finalize_upgrade`.

   This leaves a stale-plan window: if state changes between planning and execution, `dnf` may already have mutated the host before ANOLISA detects the conflict. The existing test around `upgrade/tests.rs:686` documents exactly this shape: a raw-managed component appears between plan and save, the state row is preserved, but the RPM install has already run.

   Please acquire the install lock across the transaction-critical section, reload/revalidate the state under that lock, and only run `dnf update/install` while the locked state still authorizes that mutation. This is especially important for the issue #1411 boundary that raw-managed components must not be updated, migrated, or overwritten.

2. **Refresh `rpm_metadata.source_repo` through `installed_origin()`, not `query_installed().origin`.**

   `new_observed_rpm_component` and `new_rpm_component` write `refreshed.origin` into state:

   - `upgrade.rs:1178`
   - `upgrade.rs:1221`

   But the `PackageQuery` contract says installed RPM queries typically cannot report the source repo, and callers that need `source_repo` should query it separately via `installed_origin()` (`pkg_query.rs:291-302`). The existing RPM install/adopt paths already follow that pattern (`install/rpm.rs:453` and `install/rpm.rs:817`).

   Please mirror the existing install/adopt behavior here: after a successful transaction, call `installed_origin(package)`, degrade origin lookup failures to a warning, and persist that value into `rpm_metadata.source_repo`.

3. **Do not report unresolved or ambiguous missing defaults as installable.**

   In `check_default_component`, when a target-profile default is genuinely absent but the shared RPM resolver cannot find a unique package, the check still returns an `ACTION_INSTALL` row with `package = None` (`update/check.rs:668-720`). `upgrade` later turns that into a blocking plan error.

   That means `update --check` can tell users or automation that a default component “can be installed” even though `upgrade` cannot apply it. This affects human output, JSON, and MOTD.

   Please classify unresolved or ambiguous missing-default package resolution as an item error (or a distinct blocked state) in the shared planner, so `update --check` and `upgrade` agree on what is actually actionable.

4. **Update MOTD guidance now that `anolisa upgrade` exists.**

   Related issue #1410 says the update-check MOTD should direct users to `sudo anolisa upgrade` once this command exists. The current MOTD still says:

   - `render.rs:62-65`: `Run: anolisa update --check for details`

   The test also still asserts that MOTD must not mention `anolisa upgrade` because it was not yet implemented:

   - `update/check/tests.rs:891-895`

   Please update the MOTD wording and test expectation so actionable login guidance points to `sudo anolisa upgrade` for applying the plan, while still allowing `anolisa update --check` to provide details if that remains desired.

kongche-jbw:
Thanks, fixed in the latest head.

Changes made:
- `upgrade` now holds the install lock across the state authorization and RPM
  transaction section. Component updates/installs are revalidated against the
  locked state before dnf runs, so stale raw-managed or conflicting state blocks
  the RPM mutation instead of being detected only after the fact.
- `source_repo` is now populated through `installed_origin()`, matching the
  install/adopt paths. Origin lookup failures degrade to warnings.
- Missing defaults whose RPM package cannot be resolved uniquely are now item
  errors in `update --check`, instead of installable rows with no package.
- MOTD now points users to `sudo anolisa upgrade` for applying the plan, while
  keeping `anolisa update --check` for details.

Added/updated regression coverage for stale-state pre-transaction blocking,
source_repo persistence/warnings, unresolved and ambiguous defaults, and MOTD
wording.

Validation:
- `cargo clippy -p anolisa-cli --all-targets --locked -- -D warnings`
- `cargo test -p anolisa-cli --locked`
- `cargo doc -p anolisa-cli --no-deps`

ikunkun-sys:
Found one remaining blocker in the updated PR: `source_repo` can be erased when `installed_origin()` fails during upgrade finalization.

In `upgrade.rs`, `installed_origin_or_warn()` returns `None` on lookup failure after adding a warning. That `None` is then written directly into existing RPM metadata during refresh:

- update refresh: `meta.source_repo = update.source_repo.clone();`
- matching install refresh: `meta.source_repo = install.source_repo.clone();`

This means an existing tracked component with a previously valid `rpm_metadata.source_repo` can lose that provenance just because the origin lookup failed in this run. `repair` already handles this more safely by only overwriting `source_repo` when a fresh repo value is available; `upgrade` should preserve the existing value on lookup failure and only set it when `Some(repo)` is returned. A regression test should cover an existing RPM-managed component with `source_repo = "@System"` where `installed_origin()` fails during upgrade/refresh; after the command, the old `source_repo` should still be present.

ikunkun-sys (approved):

PR #1429, opened 2026-07-10
fix(sight): auto-upgrade stale config via schema_version

Daydreamer-Li:
## Description

Add `schema_version` field to `agentsight.json` and implement `ensure_default_agents_config` to check the on-disk version at startup. When the version is missing or outdated (e.g. 0.6 configs), back up the old file (`.bak`) and overwrite with the current default. This fixes broken token collection after upgrading from 0.6 to 0.7 where the config structure changed incompatibly.

RPM uses `%config(noreplace)` so the package manager does not clobber user edits; the binary handles the upgrade itself.

### Key changes
- `agentsight.json`: add `schema_version: 2` at top level
- `config.rs`: add `CURRENT_SCHEMA_VERSION` constant, `extract_schema_version()` helper, rewrite `ensure_default_agents_config()` to check and upgrade stale configs
- `unified.rs`: always call `ensure_default_agents_config()` (not just when file is missing)
- `agentsight.spec.in`: install `agentsight.json` with `%config(noreplace)`
- `Makefile`: install/uninstall `agentsight.json`
- `rpm-build.sh`: include `agentsight.json` in tarball
- `AGENTS.md`: add schema_version bump rules and checklist
- `CHANGELOG.md`: add 0.7.2 entry

## Related Issue

closes #1427

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `sight` (agentsight)

## Checklist

## Testing

- `cargo test --lib config::tests` — 59 tests passed (including 7 new schema_version tests)
- `cargo clippy --lib --all-targets -- -D warnings` — clean
- Verified on remote machine (8.136.46.64): old 0.6 config without `schema_version` was auto-upgraded, backup created at `.json.bak`, token collection resumed

### DIFF
--- a/src/agentsight/AGENTS.md
+++ b/src/agentsight/AGENTS.md
@@ -30,6 +30,10 @@
 - 禁止修改 BPF 程序而不验证 kernel >= 5.8 兼容性
 - BPF 变更必须在真实内核上测试，仅编译通过不够
 
+### 配置文件
+- 修改 `agentsight.json` 结构（新增/删除/重命名字段、改变语义）时，必须同步 bump `schema_version` 并更新 `config.rs` 中的 `CURRENT_SCHEMA_VERSION` 常量
+- 纯新增可选字段且旧配置完全兼容时，不需要 bump `schema_version`
+
 ## 1. Quick Start
 
 ```bash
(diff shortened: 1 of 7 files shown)


jfeng18 (commented):
Thanks for tackling this @Daydreamer-Li — a `schema_version` field is a genuinely good addition, and shipping `agentsight.json` via the RPM closes a real gap (today the package installs no config at all). A few grounded findings before this lands. Two of them are data-loss concerns.

### 🔴 A (blocking) — the overwrite discards user customizations on upgrade

`ensure_default_agents_config` (config.rs) does, when `schema_version` is missing/older:
```rust
let backup = path.with_extension("json.bak");
std::fs::rename(path, &backup)?;          // move the user's config aside
std::fs::write(path, DEFAULT_AGENTS_JSON)?; // replace with the factory default
```
Every config deployed before this PR has **no** `schema_version`, so `extract_schema_version` returns `None`, `None < Some(2)` holds, and **every existing config is renamed to `.bak` and replaced with the factory default on the first upgrade**. Anything the operator customized then silently stops taking effect: custom `cmdline.allow` / `https` rules (e.g. a site-specific endpoint or agent), `server.auth`, `encryption.public_key`, `runtime_limits`, sqlite `batch` tuning. The `ensure_default_agents_config_upgrades_stale_config` test encodes this as intended (`{"cmdline":{"allow":[]}}` → backed up + overwritten).

This trades the reported bug (missing keys → collection off) for a same-class silent loss in a different dimension (user config → inactive). For a config whose whole purpose is user-owned allow-lists, overwrite is a lossy primitive.

### 🔴 B (blocking) — the backup is destroyed on the *second* schema bump

The backup path is fixed (`config.json.bak`). Upgrade N renames the user config → `.bak`. The next schema bump (N+1) renames the now-factory config → `.bak` again, which **overwrites the backup that held the user's original customizations** (`fs::rename` replaces the destination). A user who upgrades across two bumps without manually merging loses their original config entirely.

### 🟡 C (should-fix) — the installed config path isn't the one the daemon reads

The spec installs to `%{_sysconfdir}/anolisa/agentsight.json` → `/etc/anolisa/agentsight.json`, but the daemon reads `/etc/agentsight/config.json` (`bin/cli/mod.rs:25` `DEFAULT_CONFIG_PATH`, `bin/cli/trace.rs:26` default; `scripts/agentsight-start.sh` runs `agentsight trace` with no `-c`). So the RPM-shipped file and its `%config(noreplace)` protection sit at a path nothing loads, while the path the daemon actually uses is unmanaged by the package.

### Direction (your call)

`schema_version` + packaging are worth keeping. The piece I'd reconsider is overwrite → **merge**: use the embedded factory config as a base layer and overlay the user's file on top (deep-merge objects, let user scalars/arrays win). That fixes #1427 (absent `features` inherits factory-on) *and* preserves user customizations, which removes A and B. I went fairly deep on the merge semantics (explicit `false` must survive, array-replace vs union, FFI path must stay unmerged, the sibling `unified.rs` parse-error branch also silently disables collection) — happy to share notes or pair on it so we converge on one clean fix. One scope note: `token_consumption` is `false` even in the factory config, so neither approach re-enables it — worth a line in the description.

chengshuyi (approved):
LGTM. Note: Makefile install target is missing agentsight.json installation step (Finding 1 in review), can be addressed in a follow-up.

PR #1437, opened 2026-07-12
fix(anolisa): handle closed stdout safely

### LINKED ISSUE #1430: [anolisa] bug: CLI panics when stdout pipe closes early
### Component

anolisa

### What happened?

ANOLISA CLI commands panic when a downstream process closes stdout early. The smallest reproduction is:

```bash
set -o pipefail
target/debug/anolisa list --no-color | head -1
printf 'PIPESTATUS=%s\n' "${PIPESTATUS[*]}"
```

Observed output:

```text
NAME            BACKENDS    SCOPE    LOCAL STATE      OWNERSHIP      ACTION

thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
PIPESTATUS=101 0
```

The installed release binary exits with status 134 because the release profile uses `panic = "abort"`. The debug binary exits with status 101. The failure is timing-dependent because the producer can occasionally finish writing before `head` closes the pipe.

Expected behavior: an early-closing stdout consumer should not produce a panic or an internal-error message. A read-only command should stop producing output and exit successfully. A mutating command must not be killed midway through a transaction merely because its output consumer exited.

This is not specific to `list`. Repeated local runs reproduced the same BrokenPipe panic in:

| Command | Normal output lines | Reproductions |
| --- | ---: | ---: |
| `anolisa env` | 11 | 10/10 |
| `anolisa status` | 21 | 10/10 |
| `anolisa logs` | 51 | 10/10 |
| `anolisa register status` | 9 | 9/10 |
| `anolisa doctor` | 4 | 3/10 |
| `anolisa system status` | 5 | 2/10 |
| `anolisa adapter scan` | 5 | 1/10 |

A static scan currently finds direct `print!`/`println!` calls in 22 CLI source files. The shared JSON renderer also uses `println!`, so sufficiently large JSON output has the same latent failure mode.

### Root cause

`crates/anolisa-cli/src/commands/tier1/list/render.rs` prints the header and each row with separate `println!` calls. After `head` reads the header and closes its end of the pipe, the next row write returns `EPIPE`; Rust's stdout print machinery turns that I/O error into a panic.

`main` only handles `CliError` returned by command dispatch, so it cannot classify this panic as a graceful BrokenPipe termination.

### Suggested direction

Introduce a shared fallible stdout abstraction for both human and JSON renderers:

- Return and propagate `io::Result` from output operations instead of using infallible `print!`/`println!` macros.
- For read-only commands, treat `ErrorKind::BrokenPipe` as graceful early termination.
- For mutating commands, stop further stdout writes but complete or roll back the active transaction; preserve the operation's real exit status.
- Continue mapping other output failures to `EXECUTION_FAILED`.
- Keep progress and diagnostics on stderr.
- Add deterministic tests using a writer that returns BrokenPipe after the first line, plus a subprocess pipe regression test.

Restoring the default Unix SIGPIPE disposition would remove the panic globally, but it may terminate `install`, `uninstall`, `update`, or repair operations in the middle of a transaction, so it should not be applied globally without defining that behavior explicitly.

### Diagnostic Output

```shell
anolisa 0.2.2
rustc 1.88.0 (6b00bc388 2025-06-23)
Linux 5.10.134-007.ali5000.al8.x86_64 x86_64
commit c4d510238243838d7f150226916b3c569a126cff
```

### Additional Context

Relevant locations:

- `crates/anolisa-cli/src/commands/tier1/list/render.rs:75-94`
- `crates/anolisa-cli/src/response.rs:332-335`
- `crates/anolisa-cli/src/main.rs:16-23`
- `Cargo.toml:78-82`

boluochoufeng:
## Description

This PR routes CLI-owned human, JSON, help, and version output through a shared fallible stdout boundary. Early-closing consumers now stop further output without panicking or interrupting an active command, while non-BrokenPipe write failures remain visible as `EXECUTION_FAILED`.

Raw dependency provisioning sends dnf/apt stdout to stderr through direct FD duplication, keeping progress out of structured stdout without introducing a pipe-drain lifecycle. The captured `RpmTransaction` / `CommandRunner` contract remains unchanged.

## Related Issue

closes #1430

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `anolisa` (anolisa-cli)

## Checklist

## Testing

- `git diff --check`
- `cargo test --workspace --locked` (1349 passed, 0 failed)

Regression coverage includes deterministic BrokenPipe writers, pre-closed subprocess stdout, human and JSON output, Clap help/version, business-error exit-code preservation, `/dev/full`, package progress FD routing, and a self-terminating descendant retaining the progress FD.

## Additional Notes

The package-manager execution seam is tested with isolated child processes; no mutating dnf or apt transaction was run. `.codegraph` and local `.omo` planning/evidence files were not included in the commit.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/register.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/register.rs
@@ -360,9 +360,9 @@ fn format_source(source: &Option<anolisa_core::RegisterSource>) -> String {
 }
 
 fn prompt_yn(prompt: &str, default: bool) -> bool {
-    use std::io::{self, BufRead, Write};
+    use std::io::{self, BufRead};
     print!("{prompt}");
-    io::stdout().flush().ok();
+    crate::output::flush_stdout();
     let line = io::stdin()
         .lock()
         .lines()
--- a/src/anolisa/crates/anolisa-cli/src/main.rs
+++ b/src/anolisa/crates/anolisa-cli/src/main.rs
@@ -1,3 +1,6 @@
+#[macro_use]
+mod output;
+
 mod color;
 mod commands;
 mod context;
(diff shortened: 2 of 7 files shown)


kongche-jbw (changes requested):
P1：`src/anolisa/crates/anolisa-cli/tests/clap_output.rs:5` 中的 `Stdio` 只在 Linux 条件测试中使用，却无条件导入。在 macOS 上执行本 PR 声明的 `cargo clippy --workspace --all-targets --locked -- -D warnings` 会因 `unused import: Stdio` 失败（该工作区将 warning 视为错误），从而破坏支持的平台验证。请将该导入也放入 `#[cfg(target_os = "linux")]`，或将它与 Linux 专用测试放在同一条件编译范围内。

boluochoufeng:
已处理该 P1：将 `Stdio` 与 `/dev/full` 测试保持在相同的 `#[cfg(target_os = "linux")]` 条件下，避免 macOS all-target Clippy 的 unused import。修复提交为 `e48f8fa3`。\n\n验证：\n- `cargo fmt --all -- --check`\n- `cargo clippy --locked -p anolisa-cli --all-targets -- -D warnings`\n- `cargo test --locked -p anolisa-cli --test clap_output`（5 passed）\n\n本机尝试 `aarch64-apple-darwin` 交叉 Clippy 时，构建在 `ring` 的 C 编译阶段因缺少 Apple arm64 C 编译器而停止（宿主 `cc` 不支持 `-arch arm64`）；条件导入本身已按建议修正。

kongche-jbw:
lgtm thx

kongche-jbw (approved):

PR #1447, opened 2026-07-13
fix(anolisa): reject raw binary artifacts

### LINKED ISSUE #1446: [anolisa] bug: reject raw binary artifacts without embedded contracts
### Component

anolisa

### What happened?

The raw install path currently reuses the lightweight sidecar `meta.toml`
loader as the execution-time contract source for bare binary artifacts.
However, `meta.toml` is the lightweight copy intended for dry-run and
preflight. Archive execution instead reads the authoritative
`.anolisa/component.toml` embedded in the verified tar.gz artifact.

Because a bare binary cannot embed that component contract, accepting one
promotes preview metadata into an execution contract. Requiring a digest for
the sidecar would protect its bytes, but would preserve the incorrect metadata
responsibility. No currently published ANOLISA component uses the single-file
binary artifact model; published raw artifacts are tar.gz bundles.

### Steps to reproduce

1. Publish a raw distribution entry with `artifact_type = "binary"`, using
   either an HTTP(S) or `file://` URL.
2. Publish a valid sidecar `meta.toml` next to the binary.
3. Run `anolisa install <component>` through the raw backend.
4. Observe that the lightweight sidecar contract can drive the real
   installation.

### Expected behavior

- Reject every bare binary artifact before downloading the artifact or its
  sidecar, including `file://` entries.
- Tell publishers to use a tar.gz artifact containing
  `.anolisa/component.toml`.
- Keep `meta.toml` limited to dry-run/preflight; its existing optional
  `manifest_digest` verification remains optional.
- Keep tar.gz execution based exclusively on the manifest embedded in the
  digest-verified artifact.
- Remove the unused bare-binary branch from the raw install runner so its
  supported artifact list matches the published component model.

### Additional Context

Tests should cover early rejection of both local and remote binary URLs and
unchanged tar.gz behavior, including the existing destination-safety checks.

ikunkun-sys:
## Description

Rejects raw binary artifacts by restricting raw resolution to tar_gz entries before version selection. This keeps execute-time contracts inside checksum-verified archives and prevents binary rows from shadowing valid tarballs. Dry-run meta.toml remains preflight-only, while binary rows remain parseable but are excluded from install and update candidates.

## Related Issue

closes #1446

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] cosh (copilot-shell)
- [ ] cosh-ng (cosh-ng)
- [ ] sec-core (agent-sec-core)
- [ ] skill (os-skills)
- [ ] sight (agentsight)
- [ ] tokenless (tokenless)
- [ ] ckpt (ws-ckpt)
- [ ] memory (agent-memory)
- [x] anolisa (anolisa-cli)
- [ ] skillfs (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

- [x] My code follows the project code style
- [x] I have added tests that prove my fix is effective
- [x] For anolisa: cargo clippy, cargo fmt, and cargo test pass with the required workspace and locked flags

## Testing

- cargo fmt --all -- --check
- cargo clippy --workspace --all-targets --locked -- -D warnings
- cargo test --workspace --locked
- cargo doc --workspace --no-deps --locked
- Verified local and remote binary rows fail before artifact or sidecar download
- Verified mixed indexes select tar_gz and exclude binary-only versions from update candidates
- Verified tar.gz execution continues to use its embedded component manifest

## Additional Notes

ArtifactType::Binary remains accepted by the distribution index parser for wire compatibility, but raw install and update never select it.


src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs
```diff
@@ -241,6 +242,13 @@ pub(crate) fn available_raw_versions(
     index.matching_versions(&query)
 }
 
+fn installable_raw_index(mut index: DistributionIndex) -> DistributionIndex {
+    index.entries.retain(|entry| {
+        SUPPORTED_ARTIFACT_TYPES.contains(&artifact_type_wire(&entry.artifact_type))
+    });
```
kongche-jbw, line 248: **[P1] 请为被拒绝的 binary 条目保留可操作的迁移提示。**

这里静默过滤 `binary` 后，binary-only 索引会在上层退化成通用的 `no distribution entry matches`。这不满足 #1446 明确要求的“提示发布者改用包含 `.anolisa/component.toml` 的 tar.gz”，而新增的本地/远程测试还反向断言错误中不能出现这类提示。

建议保留被过滤条目的匹配信息：当过滤后的 tar.gz 解析失败、但原索引存在符合 query 的 binary 条目时，返回明确的迁移提示，并修改两处测试要求包含该提示；仍应保持在制品和 sidecar 下载前失败。

kongche-jbw (commented):
发现一处需要修正的问题，详见行内评论。

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs
ikunkun-sys: 当前没有使用发布任何binary-only的制品，此为早期多余设计

kongche-jbw (approved):
lgtm

PR #1485, opened 2026-07-15
fix(skillfs): stop test mount leaks

### LINKED ISSUE #1483: [skillfs] fix: prevent CLI tests leaking FUSE mounts
## Problem

Two SkillFS CLI startup tests terminate foreground mount processes with
`std::process::Child::kill()`. On Linux this sends `SIGKILL`, so SkillFS
cannot run its signal-driven unmount path. Each test leaves one dead FUSE
endpoint mounted at a `tempfile` path such as `/tmp/.tmp*`.

Repeated test runs accumulate stale mounts and can eventually destabilize
unrelated FUSE suites with blocked processes or I/O safety failures.

## Current behavior

Running:

```bash
cargo +1.86.0 test -p skillfs
```

reliably leaves two FUSE mounts owned by the current test user.

The leaks are isolated to
`src/skillfs/crates/skillfs-cli/tests/cli_startup_tests.rs`:

- `hermes_security_activation_file_passes_gate` leaks one mount.
- `hermes_layout_without_security_passes_gate` leaks one mount.

Each test starts `skillfs mount --foreground` and later calls
`child.kill()`. The same file already provides `stop_mount_child`, which
sends `SIGTERM`, waits for graceful shutdown, force-unmounts the exact
mountpoint as a fallback, and only then uses `SIGKILL`.

The following do not leak mounts when run independently:

- `scripts/test.sh`
- `cargo +1.86.0 test -p skillfs-core`
- `cargo +1.86.0 test -p skillfs-fuse`
- every `skillfs-fuse` integration test target

## Expected behavior

CLI startup tests must leave the current user FUSE mount set unchanged,
whether they pass, fail, or panic. Foreground mount children should receive
`SIGTERM` and use the existing bounded force-unmount fallback.

## Scope

- Replace the two direct `child.kill()` teardown paths with
  `stop_mount_child(&mut child, mount.path())`.
- Preserve the existing stdout and stderr assertions.
- Add a regression assertion or guard proving each exact test mountpoint is no
  longer present in `/proc/mounts` after teardown.
- Verify the full SkillFS suite from a clean FUSE environment.

## Non-goals

- Changing production mount or unmount behavior.
- Changing managed-mount recovery.
- Adding host-wide cleanup of FUSE mounts.
- Unmounting paths owned by other users.
- Changing the read-time transform pipeline or OS adapter.

## Security and compatibility

Cleanup must target only the mountpoint created by the current test. It must not
scan and unmount arbitrary `/tmp/.tmp*` paths or mounts owned by other users.
The fix should remain test-only and preserve CLI behavior.

## Acceptance criteria

- [ ] Each affected test leaves zero new FUSE mounts when run alone.
- [ ] `cargo +1.86.0 test -p skillfs` leaves the current user mount count
      unchanged.
- [ ] `cargo +1.86.0 test --workspace` passes on Linux with real FUSE.
- [ ] `scripts/test.sh` passes and leaves no mount under its temporary root.
- [ ] No foreground `skillfs mount` child from these tests remains alive.
- [ ] Cleanup never touches a mount owned by another user.

## Related work

This was discovered while validating #1450 but is a pre-existing CLI test
cleanup bug and is not caused by the read-time transform implementation.

kongche-jbw:
## Description

Two Hermes CLI startup tests terminated foreground SkillFS mounts with
`Child::kill()`, bypassing graceful shutdown and leaving dead FUSE endpoints
behind. This change routes both teardown paths through the existing bounded
shutdown helper and verifies that each test mountpoint disappears from
`/proc/mounts`. Cleanup remains scoped to the mountpoint created by the test,
and no production behavior changes.

## Related Issue

closes #1483

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [ ] `anolisa` (anolisa-cli)
- [x] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

- [x] For `skillfs`: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` pass

## Testing

The original teardown was first reproduced with:

```bash
cargo +1.86.0 test -p skillfs \
  --test cli_startup_tests \
  hermes_security_activation_file_passes_gate \
  -- --exact --nocapture
# The test passed but left one new FUSE mount owned by the test user.
# After the fix, both affected tests pass independently and leave no new mount:
cargo +1.86.0 test -p skillfs \
  --test cli_startup_tests \
  hermes_security_activation_file_passes_gate \
  -- --exact --nocapture

cargo +1.86.0 test -p skillfs \
  --test cli_startup_tests \
  hermes_layout_without_security_passes_gate \
  -- --exact --nocapture

# The complete CLI startup target also passed with no mount-count change:
cargo +1.86.0 test -p skillfs --test cli_startup_tests
# Full Linux validation with real FUSE:
cargo +1.86.0 fmt --all -- --check
cargo +1.86.0 clippy --workspace --all-targets -- -D warnings
cargo +1.86.0 test --workspace
cargo +1.86.0 doc --workspace --no-deps
scripts/test.sh
```
All commands passed. cargo doc reported only existing non-fatal rustdoc
warnings.
## Additional Notes
Only test teardown code is changed.
Cleanup targets the exact temporary mountpoint created by each test.
The fix does not scan or unmount arbitrary /tmp/.tmp* paths.
Mounts owned by other users are not inspected or modified.
No production mount, unmount, or signal-handling behavior changes.
Documentation is unchanged because this is a test-only correction.

### DIFF
--- a/src/skillfs/crates/skillfs-cli/tests/cli_startup_tests.rs
+++ b/src/skillfs/crates/skillfs-cli/tests/cli_startup_tests.rs
@@ -1796,8 +1796,13 @@ fn hermes_security_activation_file_passes_gate() {
         .expect("spawn skillfs");
 
     std::thread::sleep(std::time::Duration::from_secs(2));
-    let _ = child.kill();
+    stop_mount_child(&mut child, mount.path());
     let output = child.wait_with_output().expect("wait");
+    assert!(
+        !is_mounted(mount.path()),
+        "test mount must be removed after child shutdown: {}",
+        mount.path().display()
+    );
     let combined = format!(
         "{}{}",
         String::from_utf8_lossy(&output.stdout),
@@ -1895,8 +1900,13 @@ fn hermes_layout_without_security_passes_gate() {
         .expect("spawn skillfs");
 
     std::thread::sleep(std::time::Duration::from_secs(2));
-    let _ = child.kill();
+    stop_mount_child(&mut child, mount.path());
     let output = child.wait_with_output().expect("wait");
+    assert!(
+        !is_mounted(mount.path()),
+        "test mount must be removed after child shutdown: {}",
+        mount.path().display()
+    );
     let stderr = String::from_utf8_lossy(&output.stderr);
     assert!(
         !stderr.contains("incompatible"),


yummypeng (approved):
LGTM

Issue #1489, opened 2026-07-15
[TestAgent][cosh] bug: COSH_SESSION_ID is read from env but never exported to subprocess, breaking per-run correlation


zhangtaibo:
## Summary

`copilot-shell` (cosh) 2.7.0 reads `COSH_SESSION_ID` from the parent process env (for inheritance / `--continue` / `--resume` precedence), but **never writes the resolved session id back to `process.env`**. As a result, child processes spawned by cosh — the shell tool, MCP servers, user `!cmd` escapes — do **not** see `COSH_SESSION_ID` in their environment.

This defeats the per-run cross-plane correlation that the surrounding comment in `packages/cli/src/config/config.ts` claims to provide:

> *"Adopt a launcher-provided correlation id (exported as COSH_SESSION_ID by the cosh wrapper) so this run's own session id matches what AgentSight scrapes from /proc/<pid>/environ, enabling per-run cross-plane correlation."*

AgentSight / external observers that scrape subprocess environ cannot associate a tool invocation back to its cosh run, because the env var is only ever *consumed*, never *produced*.

## Component

copilot-shell (cosh) 2.7.0

## Reproduction

1. Install copilot-shell 2.7.0 RPM.
2. Launch the interactive TUI:
   ```bash
   cosh
   ```
3. From inside the TUI, drop into a shell escape and ask for the env var:
   ```
   !printenv COSH_SESSION_ID
   ```
4. Observe the result box.

### Expected

The result box contains a non-empty UUID (the same `sessionId` cosh uses internally for chat-file naming, telemetry, and the debug log).

### Actual

The result box is **empty** — `COSH_SESSION_ID` is not present in the subprocess environment. Only `COSH_DATA_DIR` (set by the `/usr/bin/cosh` wrapper shell script) is visible:

```
!env | grep ^COSH
COSH_DATA_DIR=/usr/share/copilot-shell
```

## Root Cause

In `src/copilot-shell/packages/cli/src/config/config.ts`, `loadCliConfig` resolves the session id with the precedence:

```
--continue / --resume  >  COSH_SESSION_ID env  >  minted uuid (Config constructor)
```

…then passes it to `new Config({ sessionId, ... })`. But the resolved value is **never written back** to `process.env`. The only reference to `COSH_SESSION_ID` in the codebase is the read at `config.ts:965`:

```ts
if (!sessionId) {
  const envSessionId = process.env['COSH_SESSION_ID'];
[...]

PR #1490, opened 2026-07-15
fix(cosh): export COSH_SESSION_ID to subprocess env

zhangtaibo:
## Summary

`loadCliConfig` in `packages/cli/src/config/config.ts` reads `COSH_SESSION_ID` from the parent env (for `--continue` / `--resume` / inheritance precedence) but never writes the resolved id back to `process.env`. As a result, child processes spawned by cosh — the shell tool, MCP servers, user `!cmd` escapes — do **not** see `COSH_SESSION_ID` in their environment, defeating the per-run cross-plane correlation that the surrounding code comment claims to provide.

This PR writes the resolved id back to `process.env` right after `Config` is constructed. Idempotent — no-op when the id was already inherited from the parent env.

Refs: #1489

## Root cause

`packages/cli/src/config/config.ts` resolves the session id with precedence:

```
--continue / --resume  >  COSH_SESSION_ID env  >  minted uuid (Config constructor)
```

…then passes it to `new Config({ sessionId, ... })`. The only `COSH_SESSION_ID` reference in the codebase is the read at `config.ts:965`; there is no corresponding `process.env['COSH_SESSION_ID'] = sessionId` setter anywhere in `packages/cli` or `packages/core`. So:

- The cosh node process itself does not have `COSH_SESSION_ID` in its environ (environ is frozen at `exec` time, before the id is minted).
- Every child process cosh spawns inherits from `process.env`, which still lacks `COSH_SESSION_ID`.

## Reproduction

1. Install copilot-shell 2.7.0 RPM.
2. Launch the TUI: `cosh`
3. Drop into a shell escape: `!printenv COSH_SESSION_ID`

### Before fix

Result box is **empty**. Only `COSH_DATA_DIR` (set by the `/usr/bin/cosh` wrapper) is visible:

```
!env | grep ^COSH
COSH_DATA_DIR=/usr/share/copilot-shell
```

### After fix

Result box contains a stable UUID that matches the id used for the chat jsonl file under `~/.copilot-shell/`. Repeated invocations return the same value within a single cosh session.

## Verification

Applied the patch, rebuilt `copilot-shell-2.7.0-1.alnx4.x86_64.rpm`, reinstalled, and reran the reproduction. A regression test (TC-3.1: *子进程 `!printenv COSH_SESSION_ID` 存在且稳定*) that previously had to be marked `it.fails()` now passes; the remaining `it.fails()` in the same suite (TC-3.3) is blocked by a separate saved-session fixture problem, not by this export bug.

## Notes

This fix makes child subprocess environ carry `COSH_SESSION_ID`, which is what the code comment and downstream consumers (AgentSight) actually want. Scraping `/proc/<cosh_node_pid>/environ` directly still won't show the id (environ is frozen at exec); a wrapper-level fix would be needed if that is also desired, but that is out of scope here.

## Test plan

- [x] Rebuild copilot-shell RPM with patch
- [x] Install + rerun `!printenv COSH_SESSION_ID` reproduction — returns stable UUID
- [x] Re-run cosh `vitest` TC-3 suite — TC-3.1 now passes (was `it.fails()`)
- [ ] CI on this PR branch passes (typecheck / lint / unit tests)

### DIFF
--- a/src/copilot-shell/packages/cli/src/config/config.ts
+++ b/src/copilot-shell/packages/cli/src/config/config.ts
@@ -1109,6 +1109,15 @@ export async function loadCliConfig(
     }
   }
 
+  // Export the resolved session id so child processes (shell tool, MCP
+  // servers, user `!cmd` escapes) inherit COSH_SESSION_ID, enabling per-run
+  // cross-plane correlation with AgentSight. Idempotent: no-op when the id
+  // was already inherited from the parent env.
+  const resolvedSessionId = config.getSessionId();
+  if (resolvedSessionId) {
+    process.env['COSH_SESSION_ID'] = resolvedSessionId;
+  }
+
   return config;
 }
 


samchu-zsl (dismissed):
No new code blocker found for the `COSH_SESSION_ID` export change. The resolved session id is written back before shell execution and MCP stdio transports copy `process.env`, so child shell/MCP/user command paths inherit the same id.

Verification run locally in an isolated worktree:
- `bash -n scripts/rpm-build.sh`
- `npm ci --ignore-scripts`
- `npm run generate`
- `npm run build --workspace=packages/core`
- `npm test --workspace=packages/cli -- src/config/config.test.ts` (160 passed, 2 skipped)
- `npm run build --workspace=packages/cli`
- `npm run typecheck --workspace=packages/cli`
- one lightweight runtime assertion that a minted session id is exported to `process.env.COSH_SESSION_ID`

GitHub CI also has copilot-shell build and CLI/core tests passing. The remaining pending CLA status is outside this code-review verdict.

kongche-jbw:
感谢 @zhangtaibo 定位 `COSH_SESSION_ID` 未向子进程传递的问题，并提供完整的复现步骤和验证结果，这对确认根因非常有帮助。

基于本 PR 的 #1491 已经合入并解决了 #1489。它与本 PR 的目标一致，通过 Config 作用域的子进程环境传递关联 ID，而不是写入进程级的全局 `process.env`。同时也补充了相应的回归测试。

因此本 PR 将作为已被 #1491 替代关闭。再次感谢对原始问题的发现和验证。

PR #1491, opened 2026-07-15
fix(cosh): pass session ID to subprocesses

### LINKED ISSUE #1489: [TestAgent][cosh] bug: COSH_SESSION_ID is read from env but never exported to subprocess, breaking per-run correlation

kongche-jbw:
## Description

Propagate `COSH_SESSION_ID` explicitly to child processes created by shell execution and stdio MCP transports. The correlation ID is kept Config-scoped and pinned to the launcher ID instead of mutating global `process.env`, which prevents independently addressable ACP chats from reusing one internal session ID. When no launcher ID exists, the initial Config session ID is used as the fallback correlation ID.

## Related Issue

closes #1489

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [ ] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

## Testing

- `npm run build`
- `npm run lint`
- `npm run typecheck`
- `npm test --workspace=packages/core -- --reporter=dot`
  - 3,783 passed, 2 skipped
- `npm test --workspace=packages/cli -- --reporter=dot`
  - Full CLI suite passed with no failures or errors
- Targeted session propagation tests
  - 424 passed, 2 skipped

Verified the following edge cases:

- A regular TUI session adopts the launcher-provided correlation ID.
- ACP chat sessions retain unique internal session IDs.
- Loading Config does not mutate global `process.env`.
- Shell tool, `!cmd`, prompt shell injection, PTY, and fallback execution receive `COSH_SESSION_ID`.
- Stdio MCP servers receive the Config-scoped environment.
- MCP server configuration cannot override the correlation ID.

## Additional Notes

This intentionally avoids the global `process.env` mutation proposed in #1490. A global mutation would cause subsequent ACP session configurations to re-adopt the same ID, breaking ACP session isolation.

No dependency or documentation files were changed.

### DIFF
--- a/src/copilot-shell/packages/cli/src/config/config.test.ts
+++ b/src/copilot-shell/packages/cli/src/config/config.test.ts
@@ -606,6 +606,10 @@ describe('loadCliConfig', () => {
     // Discriminating: without the env-adopt block, sessionId stays undefined and
     // Config mints a random uuid, so this exact-match assertion fails.
     expect(config.getSessionId()).toBe('sight-corr-abc-123');
+    expect(config.getChildProcessEnv()['COSH_SESSION_ID']).toBe(
+      'sight-corr-abc-123',
+    );
+    expect(process.env['COSH_SESSION_ID']).toBe('sight-corr-abc-123');
   });
 
   it('ignores an empty COSH_SESSION_ID and mints a fresh id', async () => {
@@ -617,6 +621,30 @@ describe('loadCliConfig', () => {
     // Empty value must be treated as absent (guards the `.length > 0` check).
     expect(config.getSessionId()).not.toBe('');
     expect(config.getSessionId().length).toBeGreaterThan(0);
+    expect(config.getChildProcessEnv()['COSH_SESSION_ID']).toBe(
+      config.getSessionId(),
+    );
+    expect(process.env['COSH_SESSION_ID']).toBe('');
+  });
+
+  it('keeps ACP chat sessions distinct from launcher correlation', async () => {
+    vi.stubEnv('COSH_SESSION_ID', 'launcher-correlation');
+    process.argv = ['node', 'script.js', '--acp'];
+    const argv = await parseArguments();
+
+    const firstConfig = await loadCliConfig({}, argv);
+    const secondConfig = await loadCliConfig({}, argv);
+
+    expect(firstConfig.getSessionId()).not.toBe('launcher-correlation');
+    expect(secondConfig.getSessionId()).not.toBe('launcher-correlation');
+    expect(firstConfig.getSessionId()).not.toBe(secondConfig.getSessionId());
+    expect(firstConfig.getChildProcessEnv()['COSH_SESSION_ID']).toBe(
+      'launcher-correlation',
+    );
+    expect(secondConfig.getChildProcessEnv()['COSH_SESSION_ID']).toBe(
+      'launcher-correlation',
+    );
+    expect(process.env['COSH_SESSION_ID']).toBe('launcher-correlation');
   });
 
   describe('Proxy configuration', () => {
--- a/src/copilot-shell/packages/cli/src/config/config.ts
+++ b/src/copilot-shell/packages/cli/src/config/config.ts
@@ -956,22 +956,23 @@ export async function loadCliConfig(
     }
   }
 
-  // Adopt a launcher-provided correlation id (exported as COSH_SESSION_ID
-  // by the cosh wrapper) so this run's own session id matches what AgentSight
-  // scrapes from /proc/<pid>/environ, enabling per-run cross-plane correlation.
-  // Precedence: --continue/--resume > COSH_SESSION_ID env > minted uuid
-  // (Config mints one when sessionId is left undefined). Unauthenticated hint.
-  if (!sessionId) {
-    const envSessionId = process.env['COSH_SESSION_ID'];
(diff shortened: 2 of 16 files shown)


samchu-zsl (approved):
本轮 assigned review 未发现新的 merge blocker。

本地验证覆盖了 Config-scoped `COSH_SESSION_ID`、shell execution、prompt shell、PTY/fallback execution 和 stdio MCP transport 的变更路径；targeted tests、core/cli build、core/cli typecheck 和 diff whitespace 检查均通过。

PR #1505, opened 2026-07-15
fix(sight): preserve user config on schema migration

### LINKED ISSUE #1496: [sight] agentsight 0.8.0: serve schema migration overwrites user cmdline/https/codex_offsets with built-in defaults — user customizations lost on upgrade
### Component

sight

## Summary

`agentsight serve` triggers a schema migration that writes `schema_version` 1 -> 2 back to `/etc/agentsight/config.json`. The migration does **not** only bump the version + add new default fields -- it also **overwrites** the user's existing `cmdline.allow`, deletes `cmdline.deny`, partially rewrites `https`, and structurally replaces `codex_offsets`. User customizations made before the upgrade are silently lost.

## Environment

- agentsight 0.8.0 (RPM install, binary at `/usr/local/bin/agentsight`)
- ALinux x86_64 ECS test box
- Config path: `/etc/agentsight/config.json`

### 1. Backup the current config

```bash
cp /etc/agentsight/config.json /tmp/cfg.original.bak
```

### 2. Write an old schema v1 config with user customizations

```bash
cat > /etc/agentsight/config.json <<'EOF'
{"schema_version":1,"verbose":1,"cmdline":{"allow":[{"rule":["*claude*"],"agent_name":"claude-code"},{"rule":["*cosh*"],"agent_name":"cosh"}],"deny":[{"rule":["*agentsight*"]}]},"https":[{"rule":["api.anthropic.com","api.openai.com"]}],"codex_offsets":{"ssl_capture_offset":"0x12345","sse_buffer_size":8192}}
EOF
```

This config has:
- `schema_version: 1`
- `cmdline.allow`: 2 user rules (`*claude* -> claude-code`, `*cosh* -> cosh`)
- `cmdline.deny`: 1 user rule (`*agentsight*`)
- `https`: 1 user rule (`api.anthropic.com + api.openai.com`)
- `codex_offsets`: 2 user fields (`ssl_capture_offset: 0x12345`, `sse_buffer_size: 8192`)

### 3. Start `agentsight serve` in the background for 5 seconds, then kill

```bash
rm -f /tmp/agentsight-serve-repro.log
nohup agentsight serve --port 7399 > /tmp/agentsight-serve-repro.log 2>&1 &
SERVE_PID=$!
sleep 5
kill -TERM $SERVE_PID 2>/dev/null
sleep 1
pkill -f "agentsight serve --port 7399" 2>/dev/null || true
```

serve startup log confirms it actually came up:

```
Dashboard auth enabled. Token: e9de6b6a****
AgentSight API server listening on http://127.0.0.1:7399
Dashboard UI: http://127.0.0.1:7399/
```

### 4. Read back the config -- observe what changed

```bash
cat /etc/agentsight/config.json | python3 -m json.tool
```

## Before / after

| Field | Before (user-written v1) | After `agentsight serve` | Impact |
|---|---|---|---|
| `schema_version` | `1` | `2` | Expected upgrade |
| `cmdline.allow` | 2 rules: `*claude* -> claude-code`, `*cosh* -> cosh` | **19 built-in rules** (Hermes/Codex/Runloop/Cosh/OpenClaw/Claude) | User's 2 rules lost -- replaced entirely |
| `cmdline.deny` | 1 rule: `*agentsight*` | **field absent** | Whole `deny` field deleted |
| `https` | 1 rule: `api.anthropic.com + api.openai.com` | 2 rules: `dashscope.aliyuncs.com`, `api.openai.com` | User's `api.anthropic.com` dropped; `dashscope` injected |
| `codex_offsets` | `{ssl_capture_offset: 0x12345, sse_buffer_size: 8192}` | `{schema_version:1, entries:[3 default codex versions]}` | Structure entirely replaced, user values gone |

### 5. Restore

```bash
cp /tmp/cfg.original.bak /etc/agentsight/config.json
```

## Expected behavior

Schema migration should:
1. Bump `schema_version` 1 -> 2
2. Add new v2-only fields with defaults (`runtime_limits`, `features.session_mapping`, etc.)
3. **Preserve** user-defined rules in pre-existing fields (`cmdline.allow`, `cmdline.deny`, `https`, `codex_offsets`) -- possibly merging built-in defaults with user values, but never silently dropping user entries

## Actual behavior

Migration overwrites the user's `cmdline.allow` with 19 built-in defaults, deletes `cmdline.deny` entirely, partially rewrites `https` (drops `api.anthropic.com`, adds `dashscope.aliyuncs.com`), and structurally replaces `codex_offsets`.

## Why this matters

- **Data loss on upgrade**: a user who customized cmdline rules (especially `deny` rules, which are security-relevant -- e.g. excluding `*agentsight*` from monitoring) silently loses them the first time `agentsight serve` runs after upgrade.
- **Silent failure**: there's no warning, no diff, no backup. The user has no way to know their config was rewritten.
- **Non-obvious trigger**: `agentsight discover` does **not** trigger migration (schema_version stays at 1, cmdline untouched). Only `serve` does. So a user testing with `discover` won't see the issue until they start the server in production.

## Notes

- Only `agentsight serve` and `agentsight discover` were probed in this repro. Other subcommands (`audit`, `dashboard`, `trace`, `metrics`) were not tested -- it's possible some of them also trigger the destructive migration.
- `agentsight discover` confirmed NOT to trigger: schema_version stays at 1, cmdline untouched after running `discover`.

## Suggested fix

Migration logic should:
- For pre-existing fields that exist in both v1 and v2 schemas (cmdline/https/codex_offsets), preserve user values; only add built-in defaults for fields the user didn't define.
- For `cmdline.deny` specifically, never delete the field -- if v2 has a default deny list, merge it with the user's.
- Consider emitting a warning log when migration modifies user-defined fields.

jfeng18:
## Summary

Replace the destructive config overwrite in `ensure_default_agents_config` with a **shallow merge**: start from the embedded default, overlay all top-level keys the user has set (except `schema_version`), bump to `CURRENT_SCHEMA_VERSION`.

Fixes #1496.

## Problem

When upgrading from schema v1 → v2, `ensure_default_agents_config` (added in 98158d24) backed up the old config and **overwrote it entirely** with the embedded default. User customizations — cmdline rules, https rules, codex_offsets, feature overrides — were silently lost.

## Fix

- Parse both the on-disk config and the embedded default as `serde_json::Value`
- Shallow-merge: for each top-level key in the user's config (except `schema_version`), overwrite the default's corresponding key
- Set `schema_version = CURRENT`, write back pretty-printed
- Backup (`.bak`) is still created (using `fs::copy` instead of `fs::rename` so the merged file has a fresh inode)

This preserves user data while adding any new sections (features, runtime_limits, etc.) from the default that the old config was missing.

## Testing

Two discriminating tests added:
- `ensure_default_agents_config_preserves_user_cmdline_rules_on_merge`: custom agent rule + https + deny rule survive migration
- `ensure_default_agents_config_adds_missing_sections_on_merge`: missing `features`/`runtime_limits` are pulled from default

Both tests **fail** if reverted to the old overwrite logic (discriminating).

```
cargo test --lib -- config::tests::ensure_default_agents_config
# 8 passed, 0 failed
```

## Verification

| Check | Status |
|-------|--------|
| cargo fmt --check | pass |
| cargo clippy --all-targets -D warnings | pass |
| cargo test --lib (1144 tests) | pass |
| Discriminating: revert to overwrite → new tests fail | yes |

### DIFF
--- a/src/agentsight/src/config.rs
+++ b/src/agentsight/src/config.rs
@@ -534,6 +534,32 @@ pub fn ensure_default_agents_config(path: &Path) -> anyhow::Result<()> {
         return Ok(());
     }
 
+    // Shallow-merge: start from the embedded default, overlay all top-level
+    // keys the user has set (except schema_version itself), then bump version.
+    // This preserves user customizations (cmdline rules, https rules,
+    // codex_offsets, feature overrides) while adding any NEW sections from the
+    // default that the old config was missing. Fixes #1496 — the previous
+    // implementation did a destructive overwrite that silently lost user data.
+    let mut base: serde_json::Value = serde_json::from_str(DEFAULT_AGENTS_JSON)
+        .expect("embedded DEFAULT_AGENTS_JSON must be valid");
+    let user: serde_json::Value =
+        serde_json::from_str(&content).expect("JSON validity already confirmed above");
+
+    if let (Some(base_obj), Some(user_obj)) = (base.as_object_mut(), user.as_object()) {
+        for (key, value) in user_obj {
+            if key == "schema_version" {
+                continue;
+            }
+            base_obj.insert(key.clone(), value.clone());
+        }
+        base_obj.insert(
+            "schema_version".to_string(),
+            serde_json::Value::Number(CURRENT_SCHEMA_VERSION.into()),
+        );
+    }
+
+    let merged = serde_json::to_string_pretty(&base).expect("merged config must serialize");
+
     let backup = {
         let ts = std::time::SystemTime::now()
             .duration_since(std::time::UNIX_EPOCH)
@@ -541,12 +567,12 @@ pub fn ensure_default_agents_config(path: &Path) -> anyhow::Result<()> {
             .unwrap_or(0);
         path.with_extension(format!("json.bak.{ts}"))
     };
-    std::fs::rename(path, &backup)
+    std::fs::copy(path, &backup)
         .with_context(|| format!("Failed to back up {path:?} to {backup:?}"))?;
-    std::fs::write(path, DEFAULT_AGENTS_JSON)
-        .with_context(|| format!("Failed to write updated agents config to {path:?}"))?;
+    std::fs::write(path, &merged)
+        .with_context(|| format!("Failed to write merged config to {path:?}"))?;
     log::info!(
-        "Config schema_version {:?} < {}, overwrote {path:?} (backup at {backup:?})",
+        "Config schema_version {:?} < {}, merged user config with defaults at {path:?} (backup at {backup:?})",
         on_disk_version,
         CURRENT_SCHEMA_VERSION
     );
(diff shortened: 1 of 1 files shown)


jfeng18:
Responding to the bot review:

**Point 1 (`.expect` panic on re-parse): Dismissed — false positive.**

`content` is an immutable `let` binding (line 517, no `mut`). Line 527 already returns early if the JSON is invalid. Between 527 and 546, only shared references are taken — no file re-read, no mutation, no interior mutability. The same bytes that passed the parse gate are parsed again at 546. The `.expect` path is unreachable for any runtime input.

**Point 2 (shallow merge loses nested default keys): Acknowledged as follow-up, not a blocker.**

The shallow merge is intentional and documented (comment at line 537). Deep merge risks overwriting user intentional nested changes with defaults — the opposite of what #1496 set out to fix. New nested keys introduced by future schema bumps can be handled by a dedicated deep-merge enhancement if real-world breakage surfaces. Not blocking the current fix which is strictly better than the destructive overwrite it replaces.

**Point 3 (timestamp collision in backup naming): Acknowledged as cosmetic nit, not blocking.**

Requires two processes to start in the same second AND both find outdated schema_version (a once-per-upgrade event). Even if triggered, one valid backup still exists. Not worth the diff churn.

chengshuyi (approved):
LGTM. 浅合并逻辑清晰正确，测试 discriminating，对 bot review 的回应到位。可以合入。

PR #1530, opened 2026-07-16
feat(cosh-ng): add MCP tool support

### LINKED ISSUE #1539: [cosh-ng] feat: add MCP server integrations
### Component

cosh-ng

### Problem Statement

`cosh-shell` can render results returned for unknown `mcp__server__tool` calls, but cosh-ng does not yet provide an end-to-end MCP server, client, registry, and execution layer. Users therefore cannot configure and connect external tool servers, discover their tools, authorize remote services, or invoke those tools through the normal approval path.

### Proposed Solution

Add a complete MCP integration layer to cosh-ng:

- support MCP server configuration and connection lifecycle management;
- discover, register, refresh, and expose tools from connected servers;
- route discovered tools through the existing tool execution and result-rendering paths;
- support OAuth authorization for remote servers;
- enforce server and tool allowlists;
- reuse the existing shell approval flow for sensitive MCP tool calls;
- surface actionable configuration, connection, discovery, authorization, and execution errors.

### Alternatives Considered

Continue accepting externally injected MCP results without owning the server connection and invocation lifecycle. This would preserve result rendering but would not make MCP tools usable directly from cosh-ng.

### Acceptance Criteria

- [ ] Users can configure, connect, inspect, and disconnect supported MCP servers.
- [ ] Tools can be discovered, refreshed, and displayed with their server identity.
- [ ] Discovered tools use the standard execution and result-rendering paths.
- [ ] OAuth credentials never appear in logs, sessions, or provider summaries.
- [ ] Server and tool allowlists are enforced before execution.
- [ ] Sensitive MCP calls pass through the existing approval policy.
- [ ] Tests cover configuration, connection, discovery, authorization, approval, invocation, and failure handling.

Zhilinlinlin:
## Description

Adds trusted MCP servers over stdio and Streamable HTTP, with OAuth login, token refresh, and short-lived lifecycle commands for listing, inspecting, refreshing, connecting, and disconnecting servers. Disconnected servers stay disabled until a successful reconnect; status output exposes credential presence only. Remote endpoints require HTTPS, and MCP/OAuth clients reject redirects so credentials cannot be forwarded to an unvalidated URL.

## Related Issue

closes #1539

## Type of Change

- [x] New feature
- [x] Documentation update

## Scope

- [x] `cosh-ng` (cosh-ng)

## Checklist

- [x] For `cosh-ng`: `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` pass

## Testing

- `cargo clippy --locked -p cosh-core --all-targets -- -D warnings`
- `cargo test --locked -p cosh-core` (263 unit, 24 integration tests)
- Manually authorized and initialized the Alibaba MCP server; 64 tools were discovered.

## Additional Notes

None.


samchu-zsl (changes requested):
Merge blocker: saved OAuth credentials need to be bound to the configured resource/endpoint before they are reused.\n\n`HttpConnection::new` loads OAuth credentials solely by `server_name`, and `send()` then attaches `credentials.access_token` to whichever `self.endpoint` is currently configured. The saved credential does store `resource`, but this path never compares it with the current `url` or `oauth.resource`. A user can log in to one MCP resource, later change `[mcp.servers.<name>].url` while keeping the same server name, and the next `cosh-core mcp connect/inspect` will send the old bearer token to the new endpoint before any fresh OAuth discovery or login. That breaks the PR's stated guarantee that credentials are not forwarded to an unvalidated URL.\n\nPlease reject or ignore stored credentials unless their saved resource matches the current configured resource, and require `cosh-core mcp login <server>` again when it changes. A regression test should cover changing the configured endpoint for an existing server name and assert that the old token is not placed on the first request to the new endpoint.\n\n

src/cosh-ng/crates/cosh-core/src/tool/mcp/oauth.rs
```diff
@@ -0,0 +1,921 @@
...
+        })?
+    );
+    let client = oauth_client()?;
+    let discovery = discover(&client, &endpoint, &server.oauth).await?;
+    if !discovery.code_challenge_methods_supported.is_empty()
+        && !discovery
+            .code_challenge_methods_supported
+            .iter()
+            .any(|method| method == "S256")
+    {
```
samchu-zsl: [P1] Reject missing PKCE metadata instead of treating it as compatible. MCP 2025-11-25 requires clients to refuse authorization when `code_challenge_methods_supported` is absent or does not contain `S256`; this guard currently allows an empty/missing field through. Please require `S256` explicitly and add regressions for both an absent field and an empty array.

src/cosh-ng/crates/cosh-core/src/tool/mcp/oauth.rs
```diff
@@ -0,0 +1,921 @@
...
+                .clone()
+                .unwrap_or_else(|| endpoint.to_string()),
+            config.scopes.clone(),
+        )
+    } else {
+        let response = client
+            .post(endpoint.clone())
+            .header(ACCEPT, "application/json, text/event-stream")
+            .header(CONTENT_TYPE, "application/json")
+            .json(&json!({"jsonrpc":"2.0","id":0,"method":"initialize","params":{}}))
```
samchu-zsl: [P1] Send a valid `InitializeRequest` during OAuth discovery. `params: {}` omits the required `protocolVersion`, `capabilities`, and `clientInfo` fields. A server that validates the request before issuing its auth challenge can return 400, or a JSON-RPC invalid-params error with HTTP 200 that this code misclassifies as "OAuth login is not required." Please reuse the normal initialize payload and cover that server ordering in a test.

src/cosh-ng/crates/cosh-core/src/tool/mcp.rs
```diff
@@ -0,0 +1,1117 @@
...
+                .ok_or_else(|| "MCP server closed stdout before responding".to_string())?;
+            let message: Value = serde_json::from_slice(&line)
+                .map_err(|error| format!("MCP server emitted invalid JSON-RPC: {error}"))?;
+
+            let messages = match message {
+                Value::Array(messages) => messages,
+                message => vec![message],
+            };
+            for message in messages {
+                if message.get("id").and_then(Value::as_u64) != Some(id) {
```
samchu-zsl, line 712: [P1] Do not silently discard server-initiated JSON-RPC requests. A server may send `ping` while this client is waiting for its response, and MCP requires the receiver to reply promptly. Ignoring that request can make a conforming server wait or disconnect, leaving this call to time out. Please dispatch server requests (at minimum, respond to `ping` with an empty result) and add a stdio regression.

src/cosh-ng/crates/cosh-core/src/tool/mcp/http.rs
```diff
@@ -0,0 +1,542 @@
...
+                event_line.pop();
+            }
+            if event_line.is_empty() {
+                if !event_data.is_empty() {
+                    let data = event_data.join("\n");
+                    event_data.clear();
+                    let message: Value = serde_json::from_str(&data).map_err(|error| {
+                        format!("MCP SSE response to '{method}' is invalid JSON: {error}")
+                    })?;
+                    if message.get("id").and_then(Value::as_u64) == Some(id) {
```
samchu-zsl: [P1] The Streamable HTTP path has the same server-request problem: SSE may carry JSON-RPC requests and notifications before the response for `id`, but every non-matching message is dropped here. In particular, a server `ping` needs a JSON-RPC response sent as a new POST. Please route these messages instead of discarding them and test a ping-before-result stream.

src/cosh-ng/crates/cosh-core/src/tool/mcp/http.rs
```diff
@@ -0,0 +1,542 @@
...
+                    return Err(format!(
+                        "MCP SSE response exceeds maximum size of {MAX_MCP_MESSAGE_BYTES} bytes"
+                    ));
+                }
+                event_data.push(data.to_string());
+            }
+        }
+    }
+
+    Err(format!(
```
samchu-zsl: [P1] Resume a resumable SSE response instead of failing at EOF. Streamable HTTP permits the server to close an SSE connection before the final JSON-RPC response after providing event `id`/`retry`; the client should reconnect with GET plus `Last-Event-ID` and must respect `retry`. This parser ignores both fields and always errors when the first connection closes. Please implement resumption or explicitly narrow the claimed transport support.

src/cosh-ng/crates/cosh-core/src/tool/mcp.rs
```diff
@@ -0,0 +1,1117 @@
...
+use crate::config::{CoreConfig, McpServerConfig};
+use crate::state::{self, MCP_SERVERS_STATE};
+
+use super::{Tool, ToolContext, ToolKind, ToolRegistry, ToolResult};
+
+const CLIENT_NAME: &str = "cosh-ng";
+const CLIENT_VERSION: &str = env!("CARGO_PKG_VERSION");
+const MCP_PROTOCOL_VERSION: &str = "2025-03-26";
+pub(super) const HTTP_MCP_PROTOCOL_VERSION: &str = "2025-11-25";
+const HTTP_MCP_COMPATIBLE_PROTOCOL_VERSIONS: &[&str] = &["2025-06-18", "2025-03-26", "2024-11-05"];
```
samchu-zsl, line 27: [P2] `2024-11-05` used the deprecated HTTP+SSE transport, but this client implements only Streamable HTTP and has no GET/endpoint-event fallback for a legacy server. Either remove this version from the compatible set or implement the specification's legacy transport fallback; accepting the version alone does not provide interoperability.

src/cosh-ng/crates/cosh-core/src/tool/mcp.rs
```diff
@@ -0,0 +1,1117 @@
...
+        let mut seen_cursors = HashSet::new();
+        let mut tools = Vec::new();
+
+        for _ in 0..MAX_TOOL_LIST_PAGES {
+            let mut params = serde_json::Map::new();
+            if let Some(ref cursor) = cursor {
+                params.insert("cursor".to_string(), Value::String(cursor.clone()));
+            }
+            let result = self
+                .request_with_timeout(self.startup_timeout, "tools/list", Value::Object(params))
```
samchu-zsl, line 443: [P2] Bound the whole discovery operation, not only each page. With 100 pages, the default 30-second timeout is reset for every request and each response may approach 1 MiB, so one server can hold startup for nearly 50 minutes and accumulate roughly 100 MiB of tool descriptions/schemas. Add a total deadline plus cumulative tool/byte limits.

src/cosh-ng/crates/cosh-core/src/tool/mcp/http.rs
```diff
@@ -0,0 +1,542 @@
...
+                "MCP HTTP request '{method}' failed with status {}{login_hint}",
+                response.status()
+            ));
+        }
+        if let Some(session_id) = response
+            .headers()
+            .get("MCP-Session-Id")
+            .and_then(|value| value.to_str().ok())
+        {
+            self.session_id = Some(session_id.to_string());
```
samchu-zsl, line 165: [P2] Add an explicit session termination path after accepting an `MCP-Session-Id`. Short-lived `inspect`, `refresh`, and `connect` commands currently drop the client without sending DELETE, so a stateful server retains each session until its own timeout. Please add an async close operation and invoke it when these management commands finish (tolerating 405 as allowed by the spec).

src/cosh-ng/crates/cosh-core/src/core.rs
```diff
@@ -128,12 +128,22 @@ impl CoshCore {
...
+
         let kind = tool.kind();
 
         if kind == ToolKind::ReadOnly {
             return Outcome::Allow;
         }
 
+        // MCP servers are external programs. Do not infer their side effects
+        // from a server-provided description or schema.
+        if kind == ToolKind::Mcp {
```
samchu-zsl, line 171: [P2] Please add an MCP-specific approval regression. The linked issue requires tests for approval, but the added test only proves that an allowlisted `shell` bypasses strict mode; nothing verifies that an `Mcp` tool emits the approval flow and cannot execute before authorization in auto/balanced/suggest/strict, or that an exact `allowed_tools` entry intentionally bypasses it.

samchu-zsl (changes requested):
Requesting changes for the remaining MCP protocol, OAuth, and lifecycle issues. The previous OAuth credential-to-endpoint binding issue is fixed on this head and is not repeated here.

The inline comments cover the merge-blocking interoperability/security findings and the important lifecycle/test gaps. The branch also currently conflicts with `main` in the English and Chinese configuration guides and `crates/cosh-core/src/config.rs`; those conflicts must be resolved and CI rerun before merge.

src/cosh-ng/crates/cosh-core/src/tool/mcp/oauth.rs
```diff
@@ -0,0 +1,933 @@
...
+            reqwest::StatusCode::UNAUTHORIZED | reqwest::StatusCode::FORBIDDEN
+        ) {
+            return Err(format!(
+                "MCP server returned {} during OAuth discovery",
+                response.status()
+            ));
+        }
+        let challenge = authorization_challenge(response.headers(), endpoint)?;
+        let metadata = load_protected_resource_metadata(client, endpoint, &challenge).await?;
+        let resource = config
```
samchu-zsl, line 330: [P1] Validate the protected-resource metadata `resource` before using it. RFC 9728 Sections 3.3 and 7.3 require this value to be present and exactly match the protected-resource identifier used for discovery (normally the MCP endpoint); missing or mismatched metadata MUST be rejected. Here a server-controlled value is fed into the authorization and token requests, after which the resulting bearer token is sent to the configured MCP endpoint. A malicious server can therefore induce issuance for a different resource and capture that token. Please make the field required, compare it exactly against the expected resource identifier, and add missing/mismatch regressions.

src/cosh-ng/crates/cosh-core/src/tool/mcp/oauth.rs
```diff
@@ -0,0 +1,933 @@
...
+        authorization_server_metadata_urls(server_or_metadata_url)?
+    };
+    for (index, url) in urls.iter().enumerate() {
+        let response = client.get(url.clone()).send().await.map_err(|error| {
+            format!("failed to load OAuth authorization-server metadata: {error}")
+        })?;
+        if response.status() == reqwest::StatusCode::NOT_FOUND && index + 1 < urls.len() {
+            continue;
+        }
+        return response
```
samchu-zsl: [P1] Require and validate the authorization-server metadata `issuer`. RFC 8414 Section 3.3 requires the returned issuer to be identical to the issuer identifier used to construct the metadata URL; otherwise the metadata MUST NOT be used. `AuthorizationServerMetadata` currently omits `issuer`, so mismatched metadata is accepted and its authorization/token/registration endpoints are trusted, defeating the issuer mix-up protection. Please deserialize the issuer, validate it exactly for both RFC 8414 and OIDC discovery paths, and cover missing and mismatched issuers.

samchu-zsl (commented):
Two additional OAuth metadata-validation issues found on the latest head:

samchu-zsl (approved):
No major issues. LGTM, thx!

PR #1518, opened 2026-07-16
feat(anolisa): add qwencode framework driver

Forrest-ly:
## Summary

Add QwenCodeDriver to support `anolisa adapter enable tokenless qwencode` for raw installs.

**Closes #1515**

## Changes

- **`src/anolisa/crates/anolisa-core/src/adapter/qwencode.rs`**: New driver module implementing the extension-based delivery model (mirrors CoshDriver)
- **`claim.rs`**: Add `QwenCodeClaim` struct and `DriverPayload::QwenCode` variant
- **`registry.rs`**: Register `QwenCodeDriver` in `builtin()`
- **`manager.rs`**: Extend `allowed_adapter_types` to accept qwencode with extension type
- **`adapter.rs`**: Export qwencode module

## Implementation Details

QwenCodeDriver follows the same extension model as CoshDriver:

- **Detection**: Checks for `qwen` CLI on PATH or existing `~/.qwen` directory (weak signal)
- **Enable**: Copies component extension tree to `~/.qwen/extensions/<plugin_id>/`
- **Disable**: Removes only ANOLISA-managed directories (verified via `.anolisa-adapter` ownership marker)
- **Safety**: Refuses to overwrite user-installed extensions without the ownership marker

## Testing

All 232 adapter tests pass, including 5 new qwencode-specific tests:

- `enable_status_disable_round_trip`: Full lifecycle test
- `enable_refuses_to_overwrite_non_anolisa_extension`: Ownership marker protection
- `read_bundle_rejects_missing_manifest`: Bundle validation
- `detect_uses_home_when_cli_absent`: Weak detection signal
- `detect_false_without_cli_or_home`: Negative detection

`cargo fmt --check` and `cargo clippy` pass with no warnings.

## Related

This PR enables the anolisa CLI side. The tokenless component contract update (declaring qwencode in the RPM manifest) is in a separate PR.

PR #1533, opened 2026-07-16
feat(anolisa): manage qwencode via qwen CLI

### LINKED ISSUE #1515: [anolisa] feat: support tokenless qwencode adapter for raw installs
### Component

anolisa

### Problem Statement

The anolisa CLI ships a closed, compiled-in framework driver registry (openclaw, hermes, cosh, codex, claude-code, qoder). qwencode has no built-in driver. tokenless already provides a qwencode adapter (Qwen Code extension, since tokenless 0.6.0), but after installing tokenless through the raw backend, anolisa adapter enable tokenless qwencode cannot register/enable the integration because the CLI has no qwencode driver — the framework-type gate defers qwencode to UnknownFramework. Qwen Code users therefore cannot wire up tokenless through anolisa.

### Proposed Solution

Add a qwencode FrameworkDriver to the built-in DriverRegistry (adapter/registry.rs) plus a matching qwencode module under adapter/, mirroring the existing cosh/qoder drivers. Register it in DriverRegistry::builtin(), extend validate_adapter_type_for_framework to accept qwencode with adapter_type = "extension" (Qwen Code loads tokenless as an extension, same shape as cosh), and ensure resource resolution works for raw-installed tokenless. Cover it with the framework-type matrix tests.Alternatives Considered
Keep qwencode as an unknown framework and require manual install of the tokenless qwencode extension — poor UX, inconsistent with other frameworks.
Support qwencode only for RPM-backed tokenless — raw deployment is a primary path and should be first-class.

### Alternatives Considered

_No response_

### Additional Context

registry.rs currently asserts reg.get("qwencode").is_none().
manager.rs test framework_type_matrix_is_silent_for_unknown_framework treats qwencode as unknown.
tokenless CHANGELOG 0.6.0: "add qwencode adapter for Qwen Code extension".

kongche-jbw:
## Description

Add a built-in `qwencode` framework driver so `anolisa adapter` can
manage Qwen Code extensions for raw-installed components. The driver
delegates registration, activation, and uninstall to Qwen's public CLI
instead of writing extension state directly. It persists native ownership
receipts and verifies registration plus activation postconditions, keeping
the receipt whenever cleanup cannot be proven.

## Related Issue

closes #1515

Related: #1519 (the required contract is present on `main` via `492127ca`)

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

- [x] I have added tests that prove the feature works
- [x] For `anolisa`: `cargo clippy --all-targets --locked -- -D warnings`,
      `cargo fmt --all --check`, and `cargo test --locked` pass

## Testing

Environment: macOS arm64, Rust 1.88.0.

Commands:

- `cargo check --locked`
- `cargo test --locked qwencode`
- `cargo test --locked run_capture_`

Covered edge cases:

- Re-enable repairs a disabled native activation policy without relinking.
- A successful CLI exit without native registration is rejected.
- Foreign same-name extensions are never overwritten or uninstalled.
- Missing CLI and incomplete cleanup retain the ownership receipt.
- Unsupported Qwen versions fail before mutation.
- Relative and tilde-based `QWEN_HOME` paths follow Qwen semantics.
- Empty process values, colon assignments, and duplicate dotenv keys match Qwen.
- Disable never relinks a missing registration to clean a stale policy.
- Case-insensitive extension-name collisions block name-only uninstall.
- Child stdin failures kill, wait for, and reap the supervised process.
- Malformed receipt resource references fail before CLI invocation.

## Additional Notes

- This is an independent alternative to #1518; only one driver PR should
  be merged.
- The tokenless `qwencode` component contract is already present on current
  `main` via `492127ca`; #1519 is no longer an integration blocker.
- No changelog entry is included; changelog updates belong in the release
  version-bump PR.

### Ship Note

- Changed: ANOLISA can enable, inspect, and disable Qwen Code adapters
  through Qwen's native extension lifecycle.
- Known limitations: Requires Qwen Code 0.17 or newer.
- Not covered: Live extension mutation against a real user Qwen home.
- Verification: full workspace tests and strict clippy passed locally.

### DIFF
--- a/src/anolisa/crates/anolisa-core/src/adapter.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter.rs
@@ -30,6 +30,7 @@ pub mod hermes;
 pub mod manager;
 pub mod openclaw;
 pub mod qoder;
+pub mod qwencode;
 pub mod registry;
 mod util;
 
--- a/src/anolisa/crates/anolisa-core/src/adapter/claim.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/claim.rs
@@ -389,6 +389,9 @@ pub enum DriverPayload {
     /// Qoder (qodercli) driver payload.
     #[serde(rename = "qoder")]
     Qoder(QoderClaim),
+    /// Qwen Code driver payload.
+    #[serde(rename = "qwencode")]
+    QwenCode(QwenCodeClaim),
 }
 
 /// OpenClaw driver payload. Holds only [`ClaimResource::id`] references —
@@ -506,6 +509,19 @@ pub struct QoderClaim {
     /// Full hook entries ANOLISA owns in `settings.json`.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub managed_hook_specs: Vec<QoderManagedHook>,
+}
+
+/// Qwen Code driver payload. Qwen owns extension artifacts and activation
+/// state through its CLI; the receipt references the exact native extension
+/// entry and plugin identity ANOLISA verifies before enabling or uninstalling.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct QwenCodeClaim {
+    /// Resource id of the Qwen-managed extension entry
+    /// ([`ClaimResourceKind::ExternalPath`]).
+    pub extension_dir_resource: String,
+    /// Resource id of the installed extension
+    /// ([`ClaimResourceKind::FrameworkPlugin`]).
+    pub plugin_resource: String,
 }
 
 // ---------------------------------------------------------------------------
@@ -1359,6 +1375,68 @@ mod tests {
         sample_qoder_claim()
             .validate(&layout, &allowed)
             .expect("qoder claim under allowed roots must pass");
+    }
+
+    #[test]
+    fn qwencode_claim_round_trips_and_validates() {
+        let claim = AdapterClaim {
+            claim_schema: CLAIM_SCHEMA_VERSION,
+            component: "tokenless".to_string(),
+            framework: "qwencode".to_string(),
+            plugin_id: Some("tokenless".to_string()),
+            adapter_type: Some("extension".to_string()),
+            enabled_at: "2026-07-16T10:30:45Z".to_string(),
+            resource_root: PathBuf::from("/usr/local/share/anolisa/adapters/tokenless/qwencode"),
+            bundle_digest: Some("sha256:0wen".to_string()),
+            driver_schema: DRIVER_SCHEMA_VERSION,
+            status: ClaimStatus::Enabled,
+            resources: vec![
+                ClaimResource {
+                    id: "qwencode_extension_dir".to_string(),
+                    purpose: "qwencode_extension_dir".to_string(),
+                    kind: ClaimResourceKind::ExternalPath {
+                        path: PathBuf::from("/home/alice/.qwen/extensions/tokenless"),
+                    },
+                },
+                ClaimResource {
+                    id: "qwencode_plugin".to_string(),
+                    purpose: "qwencode_plugin".to_string(),
+                    kind: ClaimResourceKind::FrameworkPlugin {
+                        framework: "qwencode".to_string(),
+                        plugin_id: "tokenless".to_string(),
+                    },
+                },
+            ],
+            driver_payload: DriverPayload::QwenCode(QwenCodeClaim {
+                extension_dir_resource: "qwencode_extension_dir".to_string(),
+                plugin_resource: "qwencode_plugin".to_string(),
+            }),
+        };
+
+        #[derive(Serialize, Deserialize, PartialEq, Debug)]
+        struct Wrapper {
+            adapter_claims: Vec<AdapterClaim>,
+        }
+        let wrapper = Wrapper {
+            adapter_claims: vec![claim.clone()],
+        };
+        let text = toml::to_string_pretty(&wrapper).expect("serialize Qwen Code to TOML");
+        let parsed: Wrapper = toml::from_str(&text).expect("parse Qwen Code from TOML");
+        assert_eq!(
+            wrapper, parsed,
+            "Qwen Code round-trip mismatch; TOML:\n{text}"
+        );
+
+        let json = serde_json::to_string(&claim).expect("serialize Qwen Code JSON");
+        let back: AdapterClaim = serde_json::from_str(&json).expect("parse Qwen Code JSON");
+        assert_eq!(claim, back);
+
+        claim
+            .validate(
+                &FsLayout::system(None),
+                &[PathBuf::from("/home/alice/.qwen")],
+            )
+            .expect("Qwen Code claim under allowed root must pass");
     }
 
     /// A forged qoder receipt pointing its settings resource at `~/.ssh` or
--- a/src/anolisa/crates/anolisa-core/src/adapter/claude_code.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/claude_code.rs
@@ -561,6 +561,7 @@ fn base_cmd(args: Vec<String>) -> FrameworkCommand {
     FrameworkCommand {
         program: claude_bin(),
         args,
+        stdin: None,
         env_set: Vec::new(),
         env_remove: Vec::new(),
         path_prepend: Vec::new(),
--- a/src/anolisa/crates/anolisa-core/src/adapter/codex.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/codex.rs
@@ -623,6 +623,7 @@ fn base_cmd(args: Vec<String>) -> FrameworkCommand {
     FrameworkCommand {
         program: codex_bin(),
         args,
+        stdin: None,
         env_set: Vec::new(),
         env_remove: Vec::new(),
         path_prepend: Vec::new(),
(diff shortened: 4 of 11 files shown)


ikunkun-sys:
Review result: **request changes**.

I found four blocking issues in the Qwen Code lifecycle integration:

1. **Fresh enable cannot complete noninteractively.** `QwenCodeDriver` runs `qwen extensions link <resource_root>`, but `ManagerOps::run_capture` gives framework commands `stdin = /dev/null`. Qwen Code v0.17.0 always requests consent for a new link and the `link` command has no `--consent` option. With EOF on stdin, the consent callback is never invoked, no registration metadata is written, and the driver fails its postcondition while leaving a `cleanup_failed` receipt. The current `SimOps` tests fabricate the native state and therefore miss this behavior.

2. **Re-enable can destroy the shared activation configuration.** The driver calls `qwen extensions enable` before probing `extension-enablement.json`. Qwen v0.17.0 treats malformed JSON as an empty map and rewrites the whole file, so policies belonging to other extensions can be lost before the later `Unknown` postcondition runs. Please probe before every mutation and fail closed on unreadable state.

3. **The driver ignores Qwen's supported user-level `.env` bootstrap for `QWEN_HOME`.** It resolves only the parent process environment and then always injects a default `QWEN_HOME`. This prevents Qwen from honoring `QWEN_HOME` configured in `~/.qwen/.env` or `~/.env`, so ANOLISA can install and report health against a different profile from the one normal Qwen sessions use.

4. **The receipt identity can diverge from the extension Qwen actually installs.** `read_bundle` accepts an arbitrary `bundle.entry`, while Qwen always reads the root-level `qwen-extension.json` during `extensions link`. If those manifests name different plugins, Qwen installs A while ANOLISA records and later cleans up B, leaving A installed and untracked. The driver should require and validate the exact native manifest.

One additional correctness issue: `status()` does not check the `>= 0.17.0` version floor, so a downgraded Qwen installation can still be reported as `Healthy` even though enable and cleanup refuse to run.

Verification completed at `5a55b32e`: fmt, check, clippy, docs, 683 core unit tests, 29 framework integration tests, 15 manager integration tests, and the focused qwencode tests all pass. I also reproduced the closed-stdin consent behavior with a minimal Node readline probe: it prints the prompt, exits 0, and never runs the response callback.

Finally, this PR still has an explicit merge dependency on #1519 for the tokenless `qwencode` component contract; without it, `AdapterManager` returns `AdapterNotDeclared` before reaching this driver.

kongche-jbw:
@ikunkun-sys Thanks for the detailed review. I addressed all five correctness issues in [`95356406`](https://github.com/alibaba/anolisa/pull/1533/commits/95356406b92f61e4a5d9b9876414ee19240e02fe):

1. **Non-interactive link consent:** `FrameworkCommand` now supports bounded stdin input. The Qwen link command writes `y\n` to the native consent prompt and then closes stdin, while preserving direct argv execution with no shell. A real child-process test verifies the stdin path, and the Qwen simulator asserts that every link receives explicit consent.
2. **Fail closed before activation mutations:** enable now parses `extension-enablement.json` before linking, probes it again after link, and only runs `qwen extensions enable` when the native policy is actually disabled. Disable also refuses relink/uninstall when the shared activation state is unreadable, so malformed JSON cannot be rewritten as an empty map.
3. **QWEN_HOME bootstrap parity:** home resolution now keeps the parent-process environment at highest priority, then reads `~/.qwen/.env` and `~/.env` in Qwen's bootstrap order. Tilde and relative-path resolution continue to match Qwen semantics, and the resolved home is used consistently for receipts, status, and CLI invocations.
4. **Native manifest identity:** the driver now rejects any declared bundle entry other than the root-level `qwen-extension.json` and always derives the receipt plugin identity from that exact manifest used by `extensions link`.
5. **Version-aware status:** status now runs the same Qwen version probe used by lifecycle mutations. A version below `0.17.0` makes `VerificationSupported` false and the adapter reports `Degraded` instead of `Healthy`.

Verification completed on the updated head:

- `cargo check --locked`
- `cargo test --locked`
- 17 focused `qwencode` tests plus the child-stdin regression test
- GitHub `Test anolisa`, PR lint, commit lint, and CLA checks are passing

#1519 remains an explicit merge dependency for the tokenless `qwencode` component contract.

ikunkun-sys:
Re-review result at `95356406`: **request changes**.

The previous consent, malformed-enablement preflight, fixed native manifest, and version-aware status issues are resolved. The QWEN_HOME fix is only partial, and I found three additional lifecycle/runner issues:

1. **[P1] QWEN_HOME resolution can still select a different profile from Qwen 0.17.** The handwritten parser discards an explicitly present empty `QWEN_HOME`, accepts only `=`, and returns the first duplicate assignment. Qwen preserves key presence during bootstrap and uses dotenv 17.1, which accepts `KEY: value` and keeps the last duplicate within a file. Reproductions:
   - inherited `QWEN_HOME=` plus `~/.qwen/.env` containing `QWEN_HOME=/custom`: native Qwen uses the default `~/.qwen`, while ANOLISA links under `/custom`;
   - `QWEN_HOME=/first` followed by `QWEN_HOME=/last`: ANOLISA uses `/first`, native Qwen uses `/last`.
   
   Both CLI scenarios completed successfully and persisted receipts in the wrong profile. Please match Qwen's absent-vs-empty distinction and dotenv 17.1 parsing/precedence, with regression tests.

2. **[P1] Disable recovery can install and enable an unrelated extension without a receipt.** When registration is absent but activation policy remains, lines 443-463 relink `claim.resource_root` without re-reading the root manifest or checking the recorded bundle digest. If the receipt names `tokenless` but that directory now declares `tokenless-v2`, Qwen installs/enables `tokenless-v2`; the subsequent probe checks only `tokenless`, reports incomplete cleanup, and leaves `tokenless-v2` installed and untracked. Revalidate manifest identity/digest before recovery, or avoid relinking during disable.

3. **[P1] Name-only uninstall can delete a foreign case-variant extension.** ANOLISA proves ownership of an exact extension directory, then runs `qwen extensions uninstall <name>`. Qwen 0.17 permits case-distinct installs but uninstall selects the first case-insensitive name/source match. With owned `foo` and foreign `Foo`, uninstalling `foo` may delete `Foo`. Detect case-fold collisions and fail closed unless Qwen exposes an exact identity operation.

4. **[P2] Child stdin errors bypass timeout and cleanup.** `run_capture` writes stdin synchronously before starting output drains and the timeout clock. An EPIPE returns without killing or waiting for the child, allowing it to continue mutating state after ANOLISA reports failure. Start supervision before stdin delivery and kill/wait/join on all writer failures.

Verification at `95356406`: focused qwencode tests (17), full core/framework/manager suites, fmt, clippy, docs, GitHub Test anolisa, Miri Level 1/2, and fake-Qwen CLI lifecycle scenarios passed. No live Qwen binary was available, so native contract checks were cross-verified against the Qwen v0.17.0 source.

The tokenless qwencode contract is now present on current `main` via `492127ca`, so the still-open #1519 is no longer an effective integration blocker.

kongche-jbw:
@ikunkun-sys Thanks for the second pass. I agree that all four findings are valid against the Qwen Code v0.17.0 implementation. They are addressed in [`b1ed34f7`](https://github.com/alibaba/anolisa/pull/1533/commits/b1ed34f77bcb16d6a213819fb59de79159063e49):

1. **QWEN_HOME / dotenv parity:** process-environment key presence is now preserved, so an inherited `QWEN_HOME=` blocks user `.env` overrides and resolves to the default `~/.qwen` profile. The user-env parser now accepts dotenv's `KEY: value` syntax and keeps the last duplicate assignment within a file while retaining Qwen's `~/.qwen/.env` then `~/.env` file priority. Regression tests cover the empty inherited value, colon syntax, duplicate-last-wins, and a final empty duplicate.
2. **Disable recovery:** disable no longer runs `extensions link` when registration is absent but an activation policy remains. It now fails closed, keeps the receipt, and reports incomplete cleanup, so a changed source manifest cannot install an unrelated extension during cleanup.
3. **Case-insensitive uninstall ambiguity:** before name-only uninstall, the driver enumerates installed extension directories and rejects any exact-name ownership proof that has a distinct case-fold-equivalent sibling. Both extensions remain untouched and the receipt is kept. The collision predicate is covered on all filesystems; the full case-distinct directory lifecycle is exercised where the filesystem permits those entries (including Linux CI).
4. **Supervised stdin delivery:** stdout/stderr draining and the timeout clock now start before stdin delivery, which runs on a supervised writer thread. Every writer failure kills and waits for the child, joins the writer and drain threads, and only then returns the framework error. The EPIPE regression test records the child PID and verifies that it is no longer alive when `run_capture` returns.

Verification on the updated head:

- `cargo check --locked`
- focused `qwencode` tests (21) and `run_capture_` tests (5)
- GitHub `Test anolisa`, PR lint, commit lint, and CLA checks pass

I also updated the PR description: the tokenless `qwencode` contract is present on current `main` via `492127ca`, so #1519 is no longer listed as an integration blocker.

ikunkun-sys (approved):

PR #1547, opened 2026-07-17
feat(anolisa): make OpenClaw enable host-aware

### LINKED ISSUE #1534: [anolisa] feat(adapter): make OpenClaw enable compatibility-aware
## Component

anolisa

## Problem Statement

ANOLISA owns OpenClaw adapter lifecycle through the built-in Rust driver, but
the current enable path assumes one OpenClaw CLI contract for every host
version.

Current behavior includes:

- `openclaw plugins install` always receives both `--force` and
  `--dangerously-force-unsafe-install`.
- `[adapters.compat].framework_version` is parsed from `component.toml` but is
  not enforced before adapter mutation.
- OpenClaw config entries are applied unconditionally, even when older hosts do
  not recognize the config key.
- Adapter health is primarily inferred from `plugins list`; enable does not
  verify that the installed plugin can reach the `loaded` runtime state.
- Capability support and authorization to bypass an install security policy
  are not modeled separately.

The sec-core deployment script already demonstrates the compatibility behavior
needed by real components:

- enforce a minimum OpenClaw version;
- inspect `plugins install --help` before selecting install flags;
- gate version-specific config such as
  `plugins.entries.agent-sec.hooks.allowConversationAccess`;
- select `plugins inspect --runtime --json` only when supported;
- execute the mutating install command once rather than retrying with different
  arguments.

Evidence:

- Current hard-coded install command:
  https://github.com/alibaba/anolisa/blob/a0f12e37d1e8bd94802545222acfa42ccf118ffb/src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs#L623-L641
- Existing compatibility metadata:
  https://github.com/alibaba/anolisa/blob/a0f12e37d1e8bd94802545222acfa42ccf118ffb/src/anolisa/crates/anolisa-core/src/manifest.rs#L610-L630
- sec-core compatibility deployment:
  https://github.com/alibaba/anolisa/blob/a0f12e37d1e8bd94802545222acfa42ccf118ffb/src/agent-sec-core/openclaw-plugin/scripts/deploy.sh

## Expected User-Visible Behavior

`anolisa adapter enable <component> openclaw` should resolve a read-only
OpenClaw host profile before performing any mutation.

The CLI should:

1. reject unsupported framework versions with the detected and required
   versions in the error;
2. select only install flags advertised by the current OpenClaw CLI;
3. apply only config entries compatible with the detected host;
4. require explicit operator authorization before using an unsafe install
   override;
5. verify that the installed plugin reaches `loaded`;
6. record only resources that were actually applied in the adapter claim.

No component-owned deployment script should be executed.

### 1. Resolve an OpenClaw host profile

Add an internal, typed host profile resolved through read-only commands:

- parsed OpenClaw version;
- `plugins install --force` support;
- `--dangerously-force-unsafe-install` support;
- `plugins inspect --json` support;
- `plugins inspect --runtime` support.

Use the existing OpenClaw environment contract for every probe. Version parsing
must cover the date-based release, correction, build, and prerelease forms
accepted by OpenClaw.

Profile resolution must complete before the first framework or filesystem
mutation.

### 2. Enforce adapter compatibility metadata

Enforce `[adapters.compat].framework_version` in the OpenClaw enable preflight.

A mismatch must fail before plugin installation, skill delivery, config writes,
or claim persistence.

The version requirement belongs to the component contract. The driver must not
hard-code sec-core or component-specific minimum versions.

### 3. Support minimal conditional OpenClaw config

Extend typed OpenClaw config declarations with a framework-version requirement.

Illustrative shape:

```toml
[[adapters.openclaw.config]]
key = "plugins.entries.agent-sec.hooks.allowConversationAccess"
value = true
framework_version = ">=2026.4.24"
```

A non-matching config condition should skip that entry. Components that require
the setting for all supported hosts should raise the adapter-level compatibility
floor instead.

The condition must remain typed and version-only. Do not add a generic
expression, command, or installation DSL.

Claims must include only config resources selected for the detected host.

### 4. Define an unsafe override policy

Capability support must not imply authorization.

Required policy:

- normal enable does not add `--dangerously-force-unsafe-install`;
- use of the override requires explicit operator opt-in, for example
  `--allow-unsafe-plugin-install`;
- component TOML, including `trust = "first-party"`, cannot self-authorize the
  override;
- dry-run output and the central operation log disclose when the override was
  selected;
- when opt-in is requested but the host does not expose the flag, fail before
  mutation with an actionable error;
- do not retry a failed normal installation automatically with the unsafe flag.

If installation without the override is blocked by OpenClaw, report how the
operator can explicitly retry after reviewing the finding.

### 5. Verify runtime loading

After successful installation:

- use `plugins inspect <id> --runtime --json` when runtime inspection is
  supported;
- otherwise use `plugins inspect <id> --json`;
- tolerate legacy diagnostic output before the JSON payload;
- require `.plugin.status == "loaded"`;
- surface OpenClaw diagnostics when verification fails.

Do not restart the OpenClaw gateway automatically.

## Scope

- OpenClaw driver host-profile and preflight behavior.
- Enforcement of adapter-level framework version requirements.
- Minimal framework-version conditions for OpenClaw config entries.
- Explicit unsafe override authorization.
- Post-install runtime verification.
- Claim accuracy for conditionally applied config.
- Focused unit and fake-CLI integration tests.
- CLI help/documentation for any new operator-facing option.

## Non-Goals

- Executing sec-core or other component-owned deployment scripts.
- Adding version-specific argv arrays to `component.toml`.
- Adding a generic adapter operation DSL.
- Changing the sec-core deployment script.
- Updating the sec-core component contract in this issue.
- Changing Hermes, Codex, Qoder, or other framework drivers.
- Restarting or otherwise managing the OpenClaw gateway.

## Acceptance Criteria

- [ ] OpenClaw host profiling uses read-only commands and finishes before any mutation.
- [ ] Adapter enable enforces `[adapters.compat].framework_version`.
- [ ] Unsupported or unparseable host versions fail with actionable context.
- [ ] `--force` is used only when advertised and required by the driver contract.
- [ ] Unsafe-override support and authorization are represented separately.
- [ ] The unsafe flag is absent by default and requires explicit operator opt-in.
- [ ] Unsafe opt-in is visible in dry-run output and central operation logs.
- [ ] Version-incompatible config entries are skipped before claim preparation.
- [ ] Claims record only config entries actually applied to the host.
- [ ] Runtime verification selects `--runtime --json` or `--json` by capability.
- [ ] Enable fails when inspect does not report `loaded`, preserving diagnostics.
- [ ] No failed install is retried automatically with a different security policy.
- [ ] No component-owned executable behavior or generic manifest DSL is introduced.
- [ ] Existing OpenClaw enable, disable, status, skill-bundle, and dry-run behavior remains covered.

## Verification Plan

Add a fake OpenClaw CLI matrix covering:

| Host profile | Expected result |
|---|---|
| below the declared minimum | fail before mutation |
| minimum version with `--force` | install with supported normal flags |
| unsafe flag advertised without opt-in | do not pass unsafe flag |
| unsafe flag advertised with opt-in | pass it on the single install attempt |
| unsafe opt-in requested but unsupported | fail before mutation |
| config condition does not match | skip config and omit it from the claim |
| config condition matches | apply config and record it in the claim |
| runtime inspect advertised | verify with `--runtime --json` |
| runtime inspect absent | verify with `--json` |
| diagnostics precede JSON | extract and parse the JSON payload |
| runtime status is not `loaded` | fail with OpenClaw diagnostics |
| dry-run | allow read-only probes but perform no state mutation |

Run:

```bash
cd src/anolisa
cargo fmt --all -- --check
cargo check --locked
cargo clippy --all-targets --locked -- -D warnings
cargo test --locked
```

## Alternatives Considered

1. **Execute each component's deployment script**

   Rejected because it bypasses AdapterManager locking, claims, auditing,
   dry-run guarantees, and safe execution boundaries.

2. **Declare per-version commands in `component.toml`**

   Rejected because framework behavior would be duplicated across components
   and the manifest would evolve into an executable DSL.

3. **Maintain a version-to-command table in the OpenClaw driver**

   Rejected as the primary mechanism because CLI capabilities may be added,
   removed, or backported independently of a single version boundary. Use
   capability detection where possible and version requirements only for
   compatibility that cannot be safely probed.

## Related

- #998
- #1040
- sec-core compatibility change:
  https://github.com/alibaba/anolisa/commit/dccdaf0ede4a9d2781ae998159dba7f756fb37b9
- OpenClaw plugin CLI contract:
  https://github.com/openclaw/openclaw/blob/main/docs/cli/plugins.md

kongche-jbw:
## Description

This PR makes OpenClaw adapter enable compatibility-aware by resolving a
typed, read-only host profile before any framework or adapter mutation. It
enforces component-declared version requirements, selects version-compatible
config, separates unsafe capability from explicit operator authorization, and
verifies that the plugin reaches the `loaded` runtime state after one install
attempt. A transient `PreparedEnable` value carries probed capabilities from
prepare to apply, keeping each probe to one execution without introducing
version-specific command arrays or an executable manifest DSL. Existing
non-OpenClaw drivers retain their previous behavior.

## Related Issue

closes #1534

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

## Testing

Run from `src/anolisa`:

```bash
cargo fmt --all -- --check
cargo check --locked
cargo clippy --workspace --all-targets --locked -- -D warnings
cargo test --locked --workspace
git diff --check
```

Focused coverage includes:

- OpenClaw version parsing, ordering, malformed constraints, and corrections.
- Host version and install/inspect capability probing before mutation.
- Missing, timed-out, and non-zero probe failures.
- Adapter-level version gating for plugins and skill bundles.
- Version-conditioned config selection and claim accuracy.
- Default-safe, explicitly authorized, and unsupported unsafe-install paths.
- Dry-run and central-log disclosure of unsafe authorization.
- Single-install behavior with no automatic unsafe retry.
- Runtime verification with `--runtime --json` and JSON-only fallback.
- Legacy diagnostics before inspect JSON and non-loaded runtime failures.
- Prepared-state mismatch rejection before any mutation.
- Existing enable, disable, status, skill-bundle, and dry-run behavior.

## Additional Notes

- No sec-core deployment script or component contract was changed.
- No component-owned executable behavior or generic installation DSL was added.
- OpenClaw gateway restart remains intentionally outside the driver contract.
- Changelog updates remain in the release version-bump PR.

### Ship Note

- Changed: OpenClaw adapter enable now probes host compatibility, applies only
  compatible config, requires explicit unsafe-install authorization, and
  verifies that installed plugins reach `loaded`.
- Known limitations: None beyond the explicit non-goals in Issue #1534.
- Not covered: Live OpenClaw E2E validation of shipping help text, security
  rejection wording, and inspect JSON variants.
- Verification: `cargo fmt --all -- --check`, `cargo check --locked`,
  `cargo clippy --workspace --all-targets --locked -- -D warnings`,
  `cargo test --locked --workspace`, and
  `cargo doc --workspace --no-deps`.

### DIFF
--- a/src/anolisa/crates/anolisa-core/src/adapter/manager.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/manager.rs
@@ -37,7 +37,8 @@ use super::AdapterError;
 use super::claim::{AdapterClaim, ClaimStatus};
 use super::driver::{
     AdapterCondition, AdapterConditionKind, AdapterOps, AdapterStatusReport, AdapterSummary,
-    CliOutput, ConditionStatus, DisableReport, DriverCtx, DriverPlan, FrameworkCommand, HostEnv,
+    CliOutput, ConditionStatus, DisableReport, DriverCtx, DriverPlan, EnableProgress,
+    FrameworkCommand, HostEnv,
 };
 use super::registry::DriverRegistry;
 use crate::central_log::{CentralLog, LogKind, LogRecord, LogStatus, Severity};
@@ -59,6 +60,22 @@ pub enum EnableOutcome {
     Planned(DriverPlan),
     /// Enable ran; the persisted receipt.
     Enabled(Box<AdapterClaim>),
+}
+
+/// Typed knobs for [`AdapterManager::enable_with_options`].
+///
+/// A distinct struct (rather than extra positional parameters) keeps the
+/// default-safe [`AdapterManager::enable`] wrapper stable for its many
+/// callers while letting a new authorization be threaded to the driver as
+/// typed data — never an environment variable or free-form map.
+#[derive(Debug, Clone, Default)]
+pub struct EnableOptions {
+    /// The caller explicitly authorized an unsafe plugin install
+    /// (`--allow-unsafe-plugin-install`). Only honored for the OpenClaw
+    /// plugin adapter; rejected for any other framework or a `skill_bundle`
+    /// adapter. Even when set, the driver adds the framework's unsafe flag
+    /// only if the host's install help exposes it.
+    pub allow_unsafe_plugin_install: bool,
 }
 
 /// Outcome of [`AdapterManager::disable`].
@@ -461,6 +478,25 @@ impl AdapterManager {
         framework: Option<&str>,
         dry_run: bool,
     ) -> Result<EnableOutcome, AdapterError> {
+        self.enable_with_options(component, framework, dry_run, EnableOptions::default())
+    }
+
+    /// Enable with explicit typed [`EnableOptions`]. [`Self::enable`] is the
+    /// default-safe wrapper over this; callers that need to pass an explicit
+    /// safety-bypass authorization use this form.
+    ///
+    /// # Errors
+    ///
+    /// Same as [`Self::enable`], plus [`AdapterError::UnsafeInstallNotApplicable`]
+    /// when an unsafe-install authorization is passed for a framework or
+    /// adapter type that does not support it.
+    pub fn enable_with_options(
+        &self,
+        component: &str,
+        framework: Option<&str>,
+        dry_run: bool,
+        options: EnableOptions,
+    ) -> Result<EnableOutcome, AdapterError> {
         let _lock = InstallLock::acquire(&self.layout.lock_file)?;
         let mut state = InstalledState::load(&self.state_path)?;
 
@@ -485,9 +521,24 @@ impl AdapterManager {
         // so the wrong driver code path can never run.
         validate_adapter_type_for_framework(component, &framework, adapter_type.as_deref())?;
 
+        // An unsafe-install authorization is only meaningful for the OpenClaw
+        // plugin adapter. Reject it for any other framework, and for a
+        // skill_bundle adapter (which never installs a plugin), before doing
+        // any work — the authorization must never silently no-op.
+        if options.allow_unsafe_plugin_install
+            && (framework != "openclaw" || adapter_type.as_deref() == Some("skill_bundle"))
+        {
+            return Err(AdapterError::UnsafeInstallNotApplicable {
+                component: component.to_string(),
+                framework: framework.clone(),
+                adapter_type: adapter_type.clone(),
+            });
+        }
+
         let declared_plugin_id = declared_plugin_id(&manifest, &framework);
         let skill_specs = declared_skills(&manifest, &framework);
         let config = declared_config(&manifest, &framework);
+        let framework_version_req = declared_framework_version_req(&manifest, &framework);
         let bundle_entry = declared_bundle_entry(&manifest, &framework);
         if adapter_type.as_deref() == Some("skill_bundle") && !config.is_empty() {
             return Err(AdapterError::InvalidAdapterInput {
@@ -573,6 +624,8 @@ impl AdapterManager {
             declared_skills: Vec::new(),
             declared_config: Vec::new(),
             declared_bundle_entry: None,
+            framework_version_req: None,
+            allow_unsafe_plugin_install: false,
             dry_run,
             ops: &probe_ops,
         };
@@ -610,6 +663,8 @@ impl AdapterManager {
             declared_skills: skills,
             declared_config: config,
             declared_bundle_entry: bundle_entry,
+            framework_version_req,
+            allow_unsafe_plugin_install: options.allow_unsafe_plugin_install,
             dry_run,
             ops: &ops,
         };
@@ -632,21 +687,41 @@ impl AdapterManager {
             });
         }
 
-        let claim = driver.prepare_enable(&bundle, &ctx)?;
+        let (mut claim, prepared) = driver.prepare_enable(&bundle, &ctx)?;
+        let claim_allowed_roots = driver.allowed_external_roots(&ctx);
+        if let Some(prior) = state.find_adapter_claim(component, &framework).cloned() {
+            // A forged prior receipt must not gain authority merely because a
+            // driver preserves facts from it during re-enable.
+            prior.validate_with_owned_roots(
+                &self.layout,
+                &claim_allowed_roots,
+                &self.all_datadir_roots,
+            )?;
+            driver.preserve_reenable_facts(&prior, &mut claim)?;
+        }
         // Defense in depth: the driver must not emit a claim that points
         // outside its own declared roots. Reject before persisting.
         claim.validate_with_owned_roots(
             &self.layout,
-            &driver.allowed_external_roots(&ctx),
+            &claim_allowed_roots,
             &self.all_datadir_roots,
         )?;
 
         state.upsert_adapter_claim(claim.clone());
         state.save(&self.state_path)?;
-        if let Err(err) = driver.apply_enable(&claim, &ctx) {
-            let mut failed_claim = claim.clone();
-            failed_claim.status = ClaimStatus::CleanupFailed;
-            state.upsert_adapter_claim(failed_claim);
+        let apply_result = {
+            let mut progress = ManagerEnableProgress {
+                state: &mut state,
+                state_path: &self.state_path,
+                layout: &self.layout,
+                allowed_external_roots: &claim_allowed_roots,
+                extra_owned_roots: &self.all_datadir_roots,
+            };
+            driver.apply_enable(&mut claim, &prepared, &ctx, &mut progress)
+        };
+        if let Err(err) = apply_result {
+            claim.status = ClaimStatus::CleanupFailed;
+            state.upsert_adapter_claim(claim.clone());
             if let Err(save_err) = state.save(&self.state_path) {
                 self.log_operation(
                     &label,
@@ -788,6 +863,8 @@ impl AdapterManager {
             declared_skills: Vec::new(),
             declared_config: Vec::new(),
             declared_bundle_entry: None,
+            framework_version_req: None,
+            allow_unsafe_plugin_install: false,
             dry_run,
             ops: &probe_ops,
         };
@@ -815,6 +892,8 @@ impl AdapterManager {
             declared_skills: Vec::new(),
             declared_config: Vec::new(),
             declared_bundle_entry: None,
+            framework_version_req: None,
+            allow_unsafe_plugin_install: false,
             dry_run,
             ops: &ops,
         };
@@ -938,6 +1017,8 @@ impl AdapterManager {
                 declared_skills: Vec::new(),
                 declared_config: Vec::new(),
                 declared_bundle_entry: None,
+                framework_version_req: None,
+                allow_unsafe_plugin_install: false,
                 dry_run: false,
                 ops: &probe_ops,
             };
@@ -965,6 +1046,8 @@ impl AdapterManager {
                 declared_skills: Vec::new(),
                 declared_config: Vec::new(),
                 declared_bundle_entry: None,
+                framework_version_req: None,
+                allow_unsafe_plugin_install: false,
                 dry_run: false,
                 ops: &ops,
             };
@@ -1577,6 +1660,30 @@ struct ManagerOps {
     /// under. Populated from the driver's `allowed_external_roots` plus
     /// the resource root.
     allowed_roots: Vec<PathBuf>,
+}
+
+/// Persists incremental receipt facts while the Manager holds the enable
+/// lock. Drivers never receive the state path or write installed state
+/// directly.
+struct ManagerEnableProgress<'a> {
+    state: &'a mut InstalledState,
+    state_path: &'a Path,
+    layout: &'a FsLayout,
+    allowed_external_roots: &'a [PathBuf],
+    extra_owned_roots: &'a [PathBuf],
+}
+
+impl EnableProgress for ManagerEnableProgress<'_> {
+    fn persist_claim(&mut self, claim: &AdapterClaim) -> Result<(), AdapterError> {
+        claim.validate_with_owned_roots(
+            self.layout,
+            self.allowed_external_roots,
+            self.extra_owned_roots,
+        )?;
+        self.state.upsert_adapter_claim(claim.clone());
+        self.state.save(self.state_path)?;
+        Ok(())
+    }
 }
 
 impl ManagerOps {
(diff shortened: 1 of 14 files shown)


src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs
```diff
@@ -566,6 +716,805 @@ impl OpenClawDriver {
...
+/// `-correction`/`-prerelease`/`+build` suffix. This deliberately rejects
+/// unrelated numeric versions in the output — a dependency or runtime version
+/// (`node 22.14.0`) or a warning number will not be mistaken for OpenClaw's
+/// own version, so an output with no genuine OpenClaw version returns `None`
+/// (and callers then fail closed).
+fn parse_openclaw_version_output(output: &str) -> Option<OpenClawVersion> {
+    for token in output.split_whitespace() {
+        let trimmed = token.trim_matches(|c: char| matches!(c, '(' | ')' | ',' | 'v' | 'V'));
+        if is_calendar_shaped_token(trimmed)
+            && let Some(version) = OpenClawVersion::parse(trimmed)
```
ikunkun-sys: [P1] 这里返回组合输出中的第一个日历形态 token，却没有确认它来自 OpenClaw 的版本行。实测 `warning ... 2099.1.1\nopenclaw 2026.4.14` 会按 `2099.1.1` 选择未来配置；真实版本不可解析时，诊断中的日期也可能让 enable 继续执行。请把候选锚定到明确的 OpenClaw 版本行，并对缺失、多个或歧义候选 fail closed。

src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs
```diff
@@ -566,6 +716,805 @@ impl OpenClawDriver {
...
+        let ord = version.cmp(&required);
+        let ok = match op {
+            ReqOp::Gte => ord != Ordering::Less,
+            ReqOp::Gt => ord == Ordering::Greater,
+            ReqOp::Lte => ord != Ordering::Greater,
+            ReqOp::Lt => ord == Ordering::Less,
+            ReqOp::Eq => ord == Ordering::Equal,
+        };
+        if !ok {
+            return Ok(false);
```
ikunkun-sys: [P1] 这里在第一个不满足的 clause 上直接返回，导致后续 clause 从未被解析。实测要求 `>=2027.0.0, >=not.a.version`、host 为 `2026.4.14` 时，CLI 仍 exit 0 并静默跳过配置，而不是报告 malformed requirement。应先完整解析并验证所有 clause，再计算整体是否匹配。

src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs
```diff
@@ -566,6 +716,805 @@ impl OpenClawDriver {
...
+
+        let (version, version_display) = self.probe_version(ctx)?;
+
+        let install_help = self.run_read_probe(
+            ctx,
+            build_install_help_cmd(&home, user_home),
+            "openclaw plugins install --help",
+        )?;
+        let supports_install_force = help_lists_flag(&install_help, "--force");
+        let supports_unsafe_install =
```
ikunkun-sys: [P1] help 中出现该 flag 已不能证明它仍是有效的安全绕过能力。OpenClaw `main` 目前继续接受 `--dangerously-force-unsafe-install`，但已将其定义为 deprecated no-op，并要求使用 `security.installPolicy`（[CLI docs](https://github.com/openclaw/openclaw/blob/main/docs/cli/plugins.md#L196-L204)，[source](https://github.com/openclaw/openclaw/blob/main/src/cli/plugins-cli.ts#L178-L184)）。当前逻辑会记录“unsafe authorization”并给出无效的重试建议。请检测实际语义/版本；对 no-op host 应拒绝该授权，或改为指引有效的策略配置。

ikunkun-sys (commented):
结论：建议修改后合并。整体设计与测试覆盖比较扎实，但下面 3 个问题会破坏本 PR 强调的 fail-closed 和 host-aware 兼容性语义，建议在合并前修复。

我基于 head `eb4b99b967567674b2306d5f7651b09464d7f4a6` 做了验证：

- `cargo fmt --all -- --check`、workspace clippy、`cargo doc --workspace --no-deps` 通过；
- workspace 全量 1,464 个测试通过，OpenClaw 相关 focused tests 通过；
- 真实 OpenClaw CLI 的只读 version/help 探测通过（未执行真实安装）；
- 两个版本处理问题已通过构建后的 CLI / fake host 稳定复现。

具体问题见 3 条 inline comments。

src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs
```diff
@@ -223,22 +274,24 @@ impl FrameworkDriver for OpenClawDriver {
         }
 
         let mut config_resources = Vec::new();
```
ikunkun-sys: **[P1] Record config only after it is successfully applied**

Issue #1534 explicitly requires claims to record only config entries actually applied to the host. This code adds every version-selected config resource during `prepare_enable`, and the Manager persists the claim before `apply_enable` runs. If the first or a later `config set` fails, `apply_enable` returns immediately and the Manager only changes the status to `cleanup_failed`; the failed and not-yet-attempted config resources remain recorded as applied.

I reproduced this with a temporary fake-CLI regression that exits 8 on the first `config set`: the expected assertion that the failed config was absent from the claim fails. Please distinguish selected/pending config from applied config, persist config resources only after each successful write (while retaining any separate cleanup intent needed by the receipt-first model), and cover both first-entry and mid-sequence failures.

ikunkun-sys (commented):
Re-reviewed head `2e220dd174846308d8a6a97eaf3d43d00356a087`. The three previously reported P1 issues are fixed, and the QwenCode `PreparedEnable::None` integration is sound. One remaining Issue #1534 acceptance blocker is noted inline.

Validation: GitHub checks are green; `cargo test --locked` passed with 1506 tests and 0 failures; formatting, Clippy with warnings denied, and workspace docs passed.

src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs
kongche-jbw: Fixed in `efb0419f`. Version-selected config now remains pending in `PreparedEnable`; each config resource is added to the claim and persisted only after its `config set` succeeds. Added first-entry and mid-sequence failure tests to verify failed/unattempted entries are excluded and only the successful prefix is retained.

src/anolisa/crates/anolisa-core/src/adapter/manager.rs
```diff
@@ -632,21 +687,31 @@ impl AdapterManager {
...
         // Defense in depth: the driver must not emit a claim that points
         // outside its own declared roots. Reject before persisting.
         claim.validate_with_owned_roots(
             &self.layout,
-            &driver.allowed_external_roots(&ctx),
+            &claim_allowed_roots,
             &self.all_datadir_roots,
         )?;
 
         state.upsert_adapter_claim(claim.clone());
```
ikunkun-sys, line 710: [P1] Preserve config facts across re-enable

This unconditional upsert replaces an existing successful OpenClaw receipt with the newly prepared claim, whose config list is empty. If re-enable then fails during installation or before config replay, the error path persists that empty claim as `cleanup_failed` even though the previously applied configuration remains on the host. I reproduced this with a successful enable followed by an install-stage failure on re-enable: the final receipt lost its prior `FrameworkConfig`. Preserve validated prior config facts until the replacement transaction safely supersedes them, and add this regression case.

src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs
```diff
@@ -293,19 +427,32 @@ impl FrameworkDriver for OpenClawDriver {
...
+            // Apply only the entries selected during prepare. A selected
+            // entry remains transient intent until its command succeeds;
+            // then append its index-anchored resource and persist immediately.
+            // This keeps a failed receipt limited to the successfully applied
+            // prefix, including when same-key version windows are present.
+            for i in selected_config_indices {
+                let cfg = &ctx.declared_config[i];
+                let cmd = build_config_set_cmd(&cfg.key, &cfg.value, &home, user_home);
                 let program = cmd.program.clone();
                 let output = ctx.ops.run_framework_cli(cmd)?;
```
ikunkun-sys, line 455: [P1] Journal config intent before mutation

The host mutation completes before any durable pending or applied fact is written. A timeout, a command that mutates and then exits non-zero, or process termination before `persist_claim` leaves applied host configuration absent from the receipt because the selected indices exist only in transient prepared state. Persist a typed pending/uncertain entry before running `config set`, then promote it to confirmed applied state after success.

src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs
```diff
@@ -566,6 +713,922 @@ impl OpenClawDriver {
...
+    indices: &[usize],
+) -> Result<(), AdapterError> {
+    let DriverPayload::OpenClaw(payload) = &claim.driver_payload else {
+        return Err(prepared_state_mismatch(
+            "openclaw apply requires an OpenClaw receipt payload",
+        ));
+    };
+    for &index in indices {
+        let resource_id = format!("openclaw_config_{index}");
+        if claim.resource(&resource_id).is_some()
```
ikunkun-sys: [P2] Keep `apply_enable` idempotent

After one successful application, the persisted claim contains each selected config resource, so retrying `apply_enable` with that receipt now fails here before mutation. This contradicts `FrameworkDriver::apply_enable`'s documented idempotency contract. Treat an existing matching resource as already recorded without appending a duplicate, and cover repeated application with a regression test.

ikunkun-sys (commented):
Incremental review of `efb0419f`: the fresh-enable first/mid-config failure cases are now fixed and the committed test suite is green. I found one reproducible re-enable receipt regression, plus two receipt/idempotency gaps introduced by the incremental persistence model.

kongche-jbw:
Addressed in `888c5de3`.

- Re-enable now preserves validated config facts from the prior receipt, so an early installation failure no longer loses configuration already present on the host.
- Added typed `pending`/`applied` config state. Intent is persisted before `config set` and promoted only after confirmed success; uncertain outcomes remain `cleanup_failed`.
- Repeated enable reuses matching config resources and payload references, avoiding rejection or duplication.

Added regressions for failed re-enable, mutate-then-fail recovery, mid-sequence failure, and successful repeated enable. Formatting, Clippy, workspace tests, docs, and `git diff --check` all pass.

ikunkun-sys (commented):
Re-reviewed head `888c5de38239f4b74af5b1c3729584fa3e59abef`. The three findings from the previous round are substantively fixed, but one reproducible P1 recovery path remains; see the inline comment.

Additional [P2]: `claim_config_count` counts only payload-confirmed Applied entries. After a mutate-then-fail command leaves a config `Pending`, `disable` emits no "left in place" warning, still reports cleanup complete, and the Manager removes the receipt even though the host may retain that config. Please report pending/uncertain config separately (or count all config facts) before discarding the receipt.

Validation: the stale-pending scenario was reproduced with a temporary fake-CLI integration test; the committed adapter-manager suite passes 55/55. Workspace fmt, check, Clippy with warnings denied, docs, and the full workspace tests pass.

src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs
```diff
@@ -566,6 +727,1189 @@ impl OpenClawDriver {
...
+    if next.resources.iter().any(|resource| {
+        matches!(
+            resource.kind,
+            ClaimResourceKind::FrameworkConfig {
+                state: ConfigApplyState::Pending,
+                ..
+            }
+        )
+    }) {
+        next.status = ClaimStatus::CleanupFailed;
```
ikunkun-sys, line 1426: [P1] Fail re-enable when a preserved pending config is no longer selected

Preserving every prior `Pending` resource here creates an unrecoverable state when the next manifest removes that key or its version condition no longer matches: `apply_enable` only revisits the current `selected_config_indices`, so the stale entry is never confirmed or cleared. I reproduced a mutate-then-fail on `removed.key`, removed that key from the manifest, and re-enabled; the call returned `Ok(EnableOutcome::Enabled)`, but the returned and persisted claim remained `CleanupFailed`, immediate status was `CleanupFailed`, and the retry argv never replayed `removed.key`. Every later enable repeats the same false-success path. Please explicitly reconcile unmatched pending entries or return an actionable error before reporting success, and add a removal/version-gate regression.

ikunkun-sys (approved):

PR #1556, opened 2026-07-17
feat(sight): add cmdline and normalize agent_name case in FFI LLM data

chengshuyi:
## Description

C consumers of the AgentSight FFI receive agent identity through `AgentsightLLMData`, but the existing fields are insufficient or inconsistent: `process_name` is the 16-byte kernel comm (e.g. every Node.js agent shows up as "node"), and `agent_name` arrives with config-file casing ("Claude", "OpenClaw", ...), forcing every C consumer to reimplement case handling.

This PR extends the FFI LLM data contract in two ways: a new `cmdline[128]` field carrying the space-joined argv read from `/proc/<pid>/cmdline`, and lowercase normalization of `agent_name` at the FFI boundary. Both changes are confined to the FFI export layer — SQLite storage, SLS export, and the Dashboard keep their existing values, so no historical data is affected.

## Related Issue

no-issue: FFI output contract enhancement driven by C consumer integration needs

## Type of Change

## Scope

- [x] `sight` (agentsight)

## Key Changes

- `src/ffi.rs`: add `cmdline: [c_char; 128]` to `AgentsightLLMData`, populated from `/proc/<pid>/cmdline` (space-joined argv, empty when the process has exited)
- `src/ffi.rs`: normalize `agent_name` with `to_lowercase()` in `build_llm_data` so C consumers observe a consistent value regardless of config-file casing
- `src/ffi.rs`: generalize the fixed-buffer copy into `copy_to_fixed_buf<const N: usize>`, which truncates at a UTF-8 char boundary (never emits invalid bytes) and always NUL-terminates
- `src/ffi.rs`: add unit tests covering ASCII truncation, multi-byte char boundary, and empty input
- `docs/design-docs/c-ffi-api.md`: document the new `cmdline` field and the lowercase `agent_name` contract

Note: the struct layout change is an ABI break — C consumers must recompile (no source change needed unless they use the new field).

## Checklist

- [ ] `cargo clippy --all-targets -- -D warnings` pass — fails on 7 pre-existing lints in untouched files (`unified.rs`, `genai/call_builder.rs`, `genai/openai_parse.rs`, `skill_metrics/metrics.rs`, `storage/sqlite/audit.rs`, `storage/sqlite/token.rs`) triggered by the Rust 1.96 toolchain; `src/ffi.rs` itself is lint-clean
- [x] `cargo test` pass (1154 lib + 35 integration/doc tests, 0 failures)

## Testing

- New unit test `test_copy_to_fixed_buf_truncates_at_char_boundary` verifies: ASCII strings truncate to N-1 bytes with NUL termination; a multi-byte char straddling the limit is dropped whole (never split into invalid UTF-8); empty input yields an all-zero buffer.
- Rebuilt to confirm cbindgen regenerates `include/agentsight.h` with `char cmdline[128];` and the build.rs FFI drift guard passes.
- End-to-end: run an agent that issues an LLM call while a C consumer (e.g. `examples/agentsight_example.c`) is attached; the LLM callback now receives `data->cmdline` as the space-joined argv (truncated to 127 bytes) and `data->agent_name` in lowercase (e.g. `claude`).

### DIFF
--- a/src/agentsight/docs/design-docs/c-ffi-api.md
+++ b/src/agentsight/docs/design-docs/c-ffi-api.md
@@ -35,7 +35,8 @@ typedef struct {
     /* 进程 */
     int32_t     pid;
     char        process_name[16];
-    const char* agent_name;           /* may be NULL */
+    char        cmdline[128];         /* 空格连接的 argv，截断到 127 字节；进程已退出时为空串 */
+    const char* agent_name;           /* may be NULL；统一为小写 */
 
     /* 时间与延迟 */
     uint64_t    timestamp_ns;
--- a/src/agentsight/src/ffi.rs
+++ b/src/agentsight/src/ffi.rs
@@ -1144,6 +1144,84 @@ mod tests {
         assert!(buf.iter().all(|&c| c == 0));
     }
 
+    // ─── build_llm_data tests ───────────────────────────────────────────────
+
+    fn make_llm_call(agent_name: Option<&str>, pid: i32) -> LLMCall {
+        use crate::genai::semantic::{LLMRequest, LLMResponse};
+        LLMCall {
+            call_id: "call-1".to_string(),
+            start_timestamp_ns: 1_000_000_000,
+            end_timestamp_ns: 2_000_000_000,
+            duration_ns: 1_000_000_000,
+            provider: "openai".to_string(),
+            model: "gpt-4".to_string(),
+            request: LLMRequest {
+                messages: vec![],
+                temperature: None,
+                max_tokens: None,
+                frequency_penalty: None,
+                presence_penalty: None,
+                top_p: None,
+                top_k: None,
+                seed: None,
+                stop_sequences: None,
+                stream: false,
+                tools: None,
+                raw_body: None,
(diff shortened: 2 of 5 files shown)


jfeng18 (approved):
LGTM. Three cohesive changes to enrich the FFI LLM data contract.

**cmdline[128]**: copy_to_fixed_buf with UTF-8 boundary safety is well-designed. /proc read per LLM call is negligible cost.

**agent_name lowercase**: confined to FFI boundary, does not affect cmdline rule matching or SQLite/Dashboard casing. Safe.

**/proc/{pid}/comm fallback**: fixes the real problem of SSL event thread comm ("HTTP client") leaking as process_name/agent_name.

**Nits (non-blocking)**:
1. builder.rs reads /proc/{pid}/comm twice for the same pid in build_pending_from_request (line ~356 + ~410) — a local `let proc_comm = read_comm(pid)` would dedup.
2. No runtime ABI version guard (pre-existing debt, not introduced here) — consider adding `AGENTSIGHT_ABI_VERSION` in the generated header as a follow-up.

PR #1632, opened 2026-07-22
fix(cosh-ng): [shell] parse session commands

### LINKED ISSUE #1621: [TestAgent][cosh-ng] bug: /session status extra misroutes to select_session('status') via match-guard iterator side effect
### Component

cosh-ng

### Bug Description

`/session status extra` (and `/session list extra`) silently misroutes to `select_session("status")` instead of rendering the usage panel. The user sees a "Session recovery requires the cosh-core backend." error panel — they cannot tell that their command was unrecognized.

### Steps to Reproduce

```shell
tmux new-session -d -s bug1 "/usr/libexec/anolisa/cosh-ng/cosh-shell raw fake"
sleep 2
tmux send-keys -t bug1 "/session status extra" Enter
sleep 1
tmux capture-pane -t bug1 -p | tail -10
tmux kill-session -t bug1
```

### Actual Behavior

```
╭ Session recovery ────────────────────────────────────────────────────────────╮
│ Session recovery requires the cosh-core backend.                             │
╰──────────────────────────────────────────────────────────────────────────────╯
```

`/session status extra` is an unrecognized subcommand form (the `status` subcommand takes no arguments), so it should render usage. Instead it routes to `select_session("status")` — confirmed by comparing with `/session bogus_id` which produces the **identical** "Session recovery requires the cosh-core backend." panel via the bare-resume arm.

### Expected Behavior

Per `crates/cosh-shell/src/slash/session/panel.rs:300-307` `render_usage`:

```
╭ Agent sessions ──────────────────────────────────────────────────────────────╮
│ Usage: /session [status|list|resume <id>|clear <id>...|clear --all]          │
╰──────────────────────────────────────────────────────────────────────────────╯
```

The bare-resume arm `Some(session_id) if parts.next().is_none()` at `crates/cosh-shell/src/slash/session.rs:61` should only fire when the first token is a real session id (i.e. not "status" or "list") and the iterator has exactly one element.

### Root Cause

`crates/cosh-shell/src/slash/session.rs:31-69` builds `parts` as a single `split_whitespace()` iterator and matches on `parts.next()` with match guards that also call `parts.next()`:

```rust
let mut parts = arguments.split_whitespace();
match parts.next() {
    None => open_session_manager(...),
    Some("status") if parts.next().is_none() => { render_session_status(...); Ok(true) }
    Some("list") if parts.next().is_none() => { render_session_list(...); Ok(true) }
    ...
    Some(session_id) if parts.next().is_none() => { select_session(session_id, ...); Ok(true) }
    Some(_) => { render_usage(...); Ok(true) }
}
```

For input `"status extra"`:

1. `parts.next()` returns `Some("status")` — iterator advances past "status".
2. The guard on the `Some("status")` arm calls `parts.next()` which returns `Some("extra")`, so the guard fails. **But the iterator has been advanced past "extra" as a side effect of evaluating the guard.**
3. Falls through to `Some(session_id) if parts.next().is_none()` — `session_id` is bound to `"status"` (from the original outer match), and the guard `parts.next()` returns `None` (iterator exhausted by the failed earlier guard), so the guard is **true**.
4. `select_session("status", ...)` is called.

The match-guard side effect on the shared iterator is the classic Rust footgun here.

### Suggested Fix

Collect the iterator into a `Vec<&str>` **before** matching, so guard evaluation does not mutate iterator state:

```rust
let parts: Vec<&str> = arguments.split_whitespace().collect();
match parts.as_slice() {
    [] => open_session_manager(blocks, adapter, state, output),
    ["status"] => { render_session_status(...); Ok(true) }
    ["list"] => { render_session_list(...); Ok(true) }
    ["resume", id] => { select_session(id, ...); Ok(true) }
    ["clear", ids @ ..] if !ids.is_empty() => begin_explicit_clear(...),
    [id] if !is_reserved_keyword(id) => { select_session(id, ...); Ok(true) }
    _ => { render_usage(...); Ok(true) }
}
```

Alternatively, keep the iterator but avoid match guards that consume it: eagerly collect the first two tokens, then match on (first, second).

Also worth adding a `is_reserved_keyword` guard on the bare-resume arm so that even if the iterator side-effect bug were somehow re-introduced, "status"/"list"/"resume"/"clear" would never be treated as session ids.

### Environment

- OS: Anolis OS 8.x (alinux)
- Component version: cosh-shell 0.12.0
- Package/RPM: cosh-ng-0.12.0-1.alnx4.x86_64

### Additional Context

- Reproduction stability: 3/3 consecutive runs reproduce the same misroute.
- The same bug fires for `/session list extra` (also a "no-arg" subcommand) — the misroute calls `select_session("list")`.
- `/session --all` (without `clear`) also falls into the bare-resume arm with session_id="--all" — related symptom, same root cause.

kongche-jbw:
## Description

Fix `/session status extra` and `/session list extra` incorrectly routing
to session recovery. Session arguments are now parsed into typed commands
by matching complete token slices, avoiding iterator side effects in match
guards. Invalid forms render usage while existing picker, resume, and clear
behavior remains unchanged.

## Related Issue

closes #1621

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `cosh-ng` (cosh-ng)

## Checklist

- [x] I have added tests that prove the fix is effective
- [x] For `cosh-ng`: Clippy and formatting checks pass
- [x] Lock files are up to date (unchanged)

## Testing

- `cargo test --package cosh-shell --lib` — 820 passed
- `cargo test --package cosh-shell --test logic` — 7 passed
- `cargo clippy --workspace --all-targets -- -D warnings`
- `git diff --check`
- `check-layout.sh` — no new violation groups compared with `origin/main`

## Additional Notes

No PTY test was added because the change is isolated to pure argument parsing
and render routing, both directly covered by unit tests.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/slash/session.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/slash/session.rs
@@ -1,8 +1,10 @@
+mod command;
 mod panel;
 mod state;
 #[cfg(test)]
 mod tests;
 
+use self::command::{parse_session_command, SessionCommand};
 use self::panel::{
     close_session_panel, core_adapter, partition_protected, redraw_session_panel,
     render_current_session_panel, render_not_ready, render_session_error, render_unavailable,
@@ -27,42 +29,24 @@ pub(crate) fn render_session_command<W: Write>(
     state: &mut InlineState,
     output: &mut W,
 ) -> std::io::Result<bool> {
-    let mut parts = arguments.split_whitespace();
-    match parts.next() {
-        None => open_session_manager(blocks, adapter, state, output),
-        Some("status") if parts.next().is_none() => {
+    match parse_session_command(arguments) {
+        SessionCommand::OpenPicker => open_session_manager(blocks, adapter, state, output),
+        SessionCommand::Status => {
             render_session_status(blocks, adapter, state, output)?;
             Ok(true)
         }
-        Some("list") if parts.next().is_none() => {
+        SessionCommand::List => {
             render_session_list(blocks, adapter, state, output)?;
             Ok(true)
         }
-        Some("resume") => {
-            let Some(session_id) = parts.next() else {
-                return open_session_manager(blocks, adapter, state, output);
-            };
-            if parts.next().is_some() {
-                render_usage(state, output)?;
-                return Ok(true);
-            }
+        SessionCommand::Resume(session_id) => {
             select_session(session_id, blocks, adapter, state, output)?;
             Ok(true)
         }
-        Some("clear") => {
-            let requested = parts.map(ToOwned::to_owned).collect::<Vec<_>>();
-            if requested.is_empty() {
-                render_usage(state, output)?;
-                return Ok(true);
-            }
+        SessionCommand::Clear(requested) => {
             begin_explicit_clear(requested, blocks, adapter, state, output)
         }
-        // `/resume ID` is parsed through the same command family.
-        Some(session_id) if parts.next().is_none() => {
-            select_session(session_id, blocks, adapter, state, output)?;
-            Ok(true)
-        }
-        Some(_) => {
+        SessionCommand::Usage => {
             render_usage(state, output)?;
             Ok(true)
         }
--- a/src/cosh-ng/crates/cosh-shell/src/slash/session/command.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/slash/session/command.rs
@@ -1,0 +1,84 @@
+//! Deterministic grammar for session slash-command arguments.
+
+#[derive(Debug, PartialEq, Eq)]
+pub(super) enum SessionCommand<'a> {
+    OpenPicker,
+    Status,
+    List,
+    Resume(&'a str),
+    Clear(Vec<String>),
+    Usage,
+}
+
+pub(super) fn parse_session_command(arguments: &str) -> SessionCommand<'_> {
+    let tokens = arguments.split_whitespace().collect::<Vec<_>>();
+    match tokens.as_slice() {
+        [] | ["resume"] => SessionCommand::OpenPicker,
+        ["status"] => SessionCommand::Status,
+        ["list"] => SessionCommand::List,
+        ["resume", session_id] => SessionCommand::Resume(session_id),
+        ["clear", "--all"] => SessionCommand::Clear(vec!["--all".to_string()]),
+        ["clear", session_ids @ ..]
+            if !session_ids.is_empty() && !session_ids.contains(&"--all") =>
+        {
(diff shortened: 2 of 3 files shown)


SunnyQjm:
PR number: #1632
head_sha: 592e6f4a38c1c8dd4da71e30102f250643e57835
reviewed_at: 2026-07-22T09:31:57Z

Findings
- 未发现 blocking package/module/public API 组织问题；本轮未发现新的 blocking session command parsing finding。

Validation
- Verified live PR #1632 head `592e6f4a38c1c8dd4da71e30102f250643e57835`; there was no previous same-marker top-level comment, so this review is posted as a new top-level comment.
- Reviewed `c7d94891856b1bb9de3ee557d8480098f925c1d4...592e6f4a38c1c8dd4da71e30102f250643e57835` in `/tmp/anolisa-monitor-review-20260722.7y9DF7/repo`; diff scope is 3 files under `crates/cosh-shell/src/slash/session*`.
- `git diff --check refs/codex/base-c7d9489...refs/codex/pr-1632 -- src/cosh-ng` passed, and the diff contains no lockfile, `package-lock.json`, or `.env` changes.
- `git merge-tree --write-tree refs/codex/base-c7d9489 refs/codex/pr-1632` returned 0 with tree `338284a47eb22e4a8b3e40c4ff7c520cc9bbd6f5`.
- Ran `crates/cosh-shell/scripts/check-layout.sh` with local `specs/cosh-ng-code-organization` and `specs/shell-test-organization` linked at the expected path; layout audit passed. Root `cosh-shell/src` remains only `lib.rs`/`main.rs`, public API remains classified, no `cosh_shell::` src self-crate paths or forbidden dependency-direction hits were found.
- Focused local test passed: `cargo test -q -p cosh-shell slash::session:: -- --test-threads=1` ran 11 matching tests successfully.
- Source-inspected `parse_session_command()`, `render_session_command()`, and `begin_explicit_clear()`: `status/list extra` now render usage, bare session IDs still route to resume, `clear --all` remains valid, and mixed `--all` with explicit IDs remains usage-only.
- GitHub checks are green for `Test cosh-ng`, `PR Checks`, `Commit Message Lint`, and `license/cla`; `Test cosh-ng` completed success at `2026-07-22T09:16:59Z`.

Remaining risk / Not run
- I did not run local full workspace fmt/clippy/tests or local real PTY tests in this monitor pass; full workspace cargo evidence is from GitHub CI plus the focused local session tests.
- No PTY-specific regression was added or run locally; the changed surface is pure slash argument parsing and session command routing.

ikunkun-sys (approved):

PR #1619, opened 2026-07-22
feat(anolisa): opt-out telemetry with data tiers

KaiLongZhou:
Switch ANOLISA telemetry from opt-in to opt-out, with tiered data collection controls:
- L1: persistent anonymous telemetry_id, always injected for scale counting.
- L2: personal identifiers (instance_id, uid) gated by named-reporting link_id.

The legacy ilogtail-based path is replaced by a self-hosted SLS PutWebtracking uploader. A new `anolisa telemetry` command surface manages enable/disable, link/unlink, status, and the resident upload loop. `register` is soft-deprecated and forwards to telemetry commands for collection state.

Key changes:
- Add `Uploader` with single-instance flock, per-component offsets, logrotate rotation handling, and 404/4xx skip logic.
- Add `TelemetryChannel` for the opt-out marker and ops filesystem layout.
- Split telemetry internals into common/instance/metadata/ops/uploader submodules.
- Add `anolisa-telemetry.service.in` with `Restart=on-failure`.
- Update RPM spec to enable telemetry by default in %post.

Assisted-by: Qoder

## Related Issue

closes #

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional change)
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `sec-core` (agent-sec-core)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] Multiple / Project-wide

## Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [x] My code follows the project's code style
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the documentation accordingly
- [ ] For `cosh`: Lint passes, type check passes, and tests pass
- [ ] For `cosh-ng`: `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` pass
[...]

PR #1680, opened 2026-07-23
fix(telemetry): address PR #1619 review items

Forrest-ly:
## Summary

Fixes for review items identified in PR #1619 (opt-out telemetry with data tiers).

### Changes

**Critical**
- **#3** `disable_unit_deferred`: log unexpected stop errors via `eprintln` instead of silently swallowing them

**Medium**
- **#5** `stringify_log_values`: preserve `null` values as literal `"null"` string instead of silently dropping fields like `{"error": null}`
- **#6** Region probe: cache last successful region to disk (`region.cache`) and use it as fallback when metadata API is temporarily unreachable, instead of always defaulting to `cn-hangzhou`
- **#7** `collect_component`: add `eprintln` to distinguish file-not-found from no-new-data in debug output

**Low / Nits**
- **#8** `register --yes`: emit deprecation warning when flag is used
- **#12** `do_link`: validate `link_id` is non-empty before persisting (added `InvalidInput` error variant)

### Items already fixed by PR author
- Critical #1: RPM %post telemetry notice (commit 67bb84c4)
- Critical #2: %postun path documentation (commit 67bb84c4)
- Medium #4: telemetry_id atomic write (already using tmp+rename pattern)

### Items not addressed (acceptable per review)
- Low #9: `find_product_type_in_release` return type change — acceptable, low call frequency
- Low #10: PR body `closes #` empty — PR metadata, not code
- Low #11: `handle_init` test coverage — already covered at channel level via `test_ensure_ops_channel_preserves_disable_marker`

### Testing
All 168 telemetry/register/platform/CLI unit tests pass (`cargo test -p anolisa-core -p anolisa-platform -p anolisa-cli`).

Related to ANO-58, supplements PR #1619.

### DIFF
--- a/src/anolisa/crates/anolisa-core/src/telemetry/uploader.rs
+++ b/src/anolisa/crates/anolisa-core/src/telemetry/uploader.rs
@@ -116,6 +116,10 @@ pub struct UploaderConfig {
     pub source: String,
     /// Persistent telemetry id file (UUID generated once, reused across reboots).
     pub telemetry_id_path: PathBuf,
+    /// Cached region from last successful probe; used as fallback when the
+    /// metadata API is temporarily unreachable instead of defaulting to
+    /// cn-hangzhou (which may route to the wrong SLS endpoint).
+    pub region_cache_path: PathBuf,
 }
 
 impl Default for UploaderConfig {
@@ -136,6 +140,7 @@ impl Default for UploaderConfig {
             topic: "anolisa-telemetry".to_string(),
             source: "anolisa".to_string(),
             telemetry_id_path: PathBuf::from("/var/lib/anolisa/telemetry/telemetry-id"),
+            region_cache_path: PathBuf::from("/var/lib/anolisa/telemetry/region.cache"),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/anolisa/crates/anolisa-platform/src/systemd.rs
```diff
@@ -149,7 +149,11 @@ pub fn disable_unit_deferred(unit: &str) -> Result<(), SystemdError> {
     }
     // Best-effort, non-blocking stop; the subsequent `disable` reports the
     // authoritative outcome, so a not-loaded/inactive unit here is ignored.
-    let _ = run_systemctl(&["stop", "--no-block", unit]);
+    // However, log unexpected stop failures (e.g. permission denied) so they
+    // are not silently swallowed.
+    if let Err(e) = run_systemctl(&["stop", "--no-block", unit]) {
```
ikunkun-sys, line 154: **[P2] Check the command exit status, not only spawn errors.** `run_systemctl()` returns `Err` only when the process cannot be spawned. A normal `systemctl` invocation that reports permission denied or another failure returns `Ok(Output)` with a non-zero status, so this branch still silently discards the failures it is intended to expose. Please inspect `out.status` and classify/log unexpected non-zero outputs while continuing to ignore expected `NotFound` cases.

src/anolisa/crates/anolisa-core/src/telemetry/uploader.rs
```diff
@@ -438,6 +464,7 @@ impl Uploader {
         lines.append(&mut fresh);
 
         if lines.is_empty() {
+            eprintln!("[anolisa] telemetry: {component}.jsonl has no new data");
```
ikunkun-sys, line 467: **[P2] Avoid logging the normal idle state on every poll.** Telemetry initialization pre-creates seven `.jsonl` files and the uploader polls every 120 seconds. Once offsets are caught up, this emits seven messages per round—about 5,040 journal entries per day on an idle system. Please remove this message, gate it behind an explicit debug mode, or rate-limit it.

ikunkun-sys (commented):
I found two behavioral issues that should be addressed before merge: the new systemd diagnostic does not observe normal non-zero command exits, and the idle uploader path produces sustained journal noise.

PR #1704, opened 2026-07-23
feat(sight): add 6 new interruption types

Daydreamer-Li:
## Description
Extend `InterruptionType` with 6 new variants: `tool_failure`, `empty_response`, `resource_exhaustion`, `slow_response`, `state_machine_error`, and `unauthorized_action`, bringing coverage to 18/20 of the fault detection checklist.

Tool-related failures are detected from both standalone `ToolUse` events and `tool_call_response`/`tool_result` entries embedded in `LLMCall` request/response history, since some agents only surface tool execution results via the latter path. Errors containing permission-related keywords (`EACCES`, `EPERM`, `permission denied`, `sandbox`, etc.) are classified as `unauthorized_action`; others fall back to `tool_failure`.

## Related Issue
no-issue: internal fault-detection coverage improvement

## Type / Scope
- [x] Feature
- Scope: `sight` (agentsight)

## Testing
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --locked` all pass
- Verified on a live test machine:
  - `resource_exhaustion`, `empty_response`, `slow_response` via HTTP mock LLM responses
  - `tool_failure`, `unauthorized_action` via real OpenClaw tool calls (missing file read → ENOENT; write to a `chattr +i` immutable file → EPERM)
- Known limitation: `state_machine_error` keyword matching could not be reliably triggered via local HTTPS mocks; end-to-end verification for this type is pending further capture-pipeline investigation.

### DIFF
--- a/src/agentsight/dashboard/src/utils/apiClient.ts
+++ b/src/agentsight/dashboard/src/utils/apiClient.ts
@@ -597,6 +597,12 @@ export const INTERRUPTION_TYPE_CN: Record<string, string> = {
   safety_filter: '安全过滤',
   retry_storm: '重试风暴',
   dead_loop: '死循环',
+  tool_failure: '工具调用失败',
+  empty_response: '空响应',
+  resource_exhaustion: '资源耗尽',
+  slow_response: '响应过慢',
+  state_machine_error: '状态机异常',
+  unauthorized_action: '未授权操作',
 };
 
 /**
--- a/src/agentsight/src/interruption/detector.rs
+++ b/src/agentsight/src/interruption/detector.rs
@@ -3,9 +3,13 @@
 //! # Online path (called immediately after each LLMCall is built)
 //! `InterruptionDetector::detect(call)` checks a single call against all
 //! single-call rules and returns any detected interruption events.
+//!
+//! # Tool-use path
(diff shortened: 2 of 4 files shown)


Daydreamer-Li:
Addressed all 3 review points in `b1699092`:

- **[P1] `eprem` typo**: Fixed to `eperm` in both `is_unauthorized_error_text` and `text_has_tool_error_signal`. Confirmed this was previously masked in real-world testing only because EPERM messages also happened to contain "operation not permitted".
- **[P2] Overly broad keyword matching**: Removed generic words (`exception`, `failed`, `error:`) from `text_has_tool_error_signal` that commonly appear in non-error tool output; kept only specific/structured signals.
- **[P2] Missing conversation-level dedup for ToolUse**: Added the same `exists_for_conversation` check used by the LLMCall path to the ToolUse branch, so a tool failing repeatedly in a loop no longer floods the store with one row per attempt. Deferred full RetryStorm escalation parity (it's keyed on `genai_events` call counts, which don't map 1:1 to tool invocations) as a follow-up.

Re-verified with `cargo fmt --check`, `cargo clippy -D warnings`, and `cargo test` (143 interruption tests + 28 unified tests, all passing).

chengshuyi (approved):

PR #1740, opened 2026-07-24
feat(anolisa): pin raw install versions

ikunkun-sys:
## Summary

Make `anolisa install <component> --version <v>` a first-class version pin on the **raw** backend, symmetric with the recently landed rpm pin (#43858f7a's `feat(anolisa): pin rpm install versions`):

- **Exact resolution**: the raw resolver already exact-matches the pinned version against the distribution index; the artifact's embedded manifest version is enforced against the resolved entry before any file lands.
- **Pin disclosure**: a pinned install/dry-run now reports `requested_version`, `resolved_version`, `artifact` (exact artifact URL — the raw analog of a NEVRA), and `source_repo` in the JSON envelope and human dry-run details, using the same additive wire contract as the delegated pin. Unpinned installs are unchanged (fields omitted).
- **Actionable refusal**: pinning an unpublished version now fails with a dedicated INVALID_ARGUMENT that lists the actually published versions for the host tuple, instead of the generic resolver error.

## Testing

- `cargo test -p anolisa-cli`: 762 passed (new e2e: pinned install places the exact non-latest version; pinned dry-run stays side-effect free; unpublished pin lists published versions; unit tests for the JSON/pin detail contract).
- `cargo fmt` / `clippy --all-targets` / `cargo doc --no-deps` clean.
- **Real-host verification** on Alibaba Cloud Linux 4 (x86_64) against the production OSS repo (`anolisa.oss-cn-hangzhou.aliyuncs.com/anolisa-releases`), isolated under `--prefix`:
  - pinned `tokenless --version 0.6.0` installed 0.6.0 while 0.7.1 was latest; state recorded the pinned artifact URL
  - same-version reinstall → idempotent no-op; different version → I5 refusal pointing at `update`
  - `agentsight --version 9.9.9` → refusal listing `0.8.1, 0.8.0, 0.7.0`
  - unpinned install still resolves latest (0.7.1) with no pin fields
  - uninstall left no component state; real system paths untouched

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
@@ -172,6 +172,45 @@ pub(crate) struct DelegatedPin {
     pub(crate) source_repo: Option<String>,
 }
 
+/// Resolved metadata for a version-pinned owned (raw) install, the owned
+/// analog of [`DelegatedPin`]. Built from the distribution entry the exact
+/// version query selected — the raw resolver only returns an entry whose
+/// `version` equals the request, and the contract validator later refuses an
+/// artifact whose embedded manifest disagrees with that entry, so these
+/// fields prove what will actually be placed.
+#[derive(Debug, Clone)]
+pub(crate) struct RawPin {
+    /// The `--version` value the caller requested.
+    requested_version: String,
+    /// Version of the resolved distribution entry (equals
+    /// `requested_version`, restated for an unambiguous JSON contract).
+    resolved_version: String,
+    /// Exact artifact URL the pinned entry downloads from — the raw analog
+    /// of the delegated pin's NEVRA.
+    artifact: String,
+    /// Repository base URL the distribution index was fetched from.
+    source_repo: String,
+}
+
+impl RawPin {
+    /// Capture the pin evidence from a settled raw resolution.
+    ///
+    /// Ordering is enforced by ownership, not convention:
+    /// [`validate_owned_install`] takes the [`RawResolution`] by value, so
+    /// the borrow here cannot compile after validation has consumed it. Any
+    /// future reshuffle that moves the resolution earlier surfaces as a
+    /// borrow error at this call site rather than as silently missing pin
+    /// evidence.
+    fn from_resolution(requested: &str, resolution: &RawResolution) -> Self {
+        Self {
+            requested_version: requested.to_string(),
+            resolved_version: resolution.entry.version.clone(),
+            artifact: resolution.artifact_url.clone(),
+            source_repo: resolution.base_url.clone(),
+        }
+    }
+}
+
 struct ResolvedInstallIdentity {
     component: String,
     pinned: bool,
@@ -520,12 +559,19 @@ fn execute_planned(
 
     let plan_labels: Vec<String> = steps.iter().map(step_label).collect();
 
+    // Pin evidence for an owned `--version` install, captured while the
+    // resolution is still whole (contract validation consumes it below).
+    let raw_pin = match (args.version.as_deref(), resolution.as_ref()) {
+        (Some(requested), Some(resolution)) => Some(RawPin::from_resolution(requested, resolution)),
+        _ => None,
+    };
+
     if ctx.dry_run {
         for warning in resolution.iter().flat_map(|r| r.warnings.iter()) {
             eprintln!("warning: {warning}");
         }
-        // A pinned delegated dry-run reports the version it resolved against
-        // the repository, not the raw `--version` echo — the pin fields carry
+        // A pinned dry-run reports the version it resolved against the
+        // repository, not the raw `--version` echo — the pin fields carry
         // the exact candidate so the preview proves what would be installed.
         let base_version = resolution
             .as_ref()
@@ -547,6 +593,9 @@ fn execute_planned(
         };
         if let Some(pin) = &delegated_pin {
             payload = payload.with_pin(pin);
+        }
+        if let Some(pin) = &raw_pin {
+            payload = payload.with_raw_pin(pin);
         }
         render_result(ctx, &payload)?;
         return Ok(InstallOutcome::Installed);
@@ -570,6 +619,7 @@ fn execute_planned(
             &steps,
             &plan_labels,
             validated,
+            raw_pin.as_ref(),
             native_package.as_deref(),
             &provider,
             &command,
@@ -899,6 +949,7 @@ fn install_owned(
     steps: &[Step],
     plan_labels: &[String],
     validated: ValidatedInstall,
+    raw_pin: Option<&RawPin>,
     native_package: Option<&str>,
     provider: &DelegatedProvider,
     command: &str,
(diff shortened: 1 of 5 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/install/tests/raw_e2e.rs
```diff
@@ -780,6 +780,85 @@ fn install_unpublished_version_is_invalid_argument() {
         .expect_err("must fail to resolve");
     assert_eq!(err.code(), "INVALID_ARGUMENT");
     assert!(err.reason().contains("9.9.9"), "got: {}", err.reason());
+    // The pin refusal must name what the repository actually publishes, so
+    // the caller can correct the request without a second query.
+    assert!(
+        err.reason().contains("not published"),
```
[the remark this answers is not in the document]
ikunkun-sys: Adopted in a78ddfd8: the test now publishes two versions (0.1.0 + 0.2.0) via the multi-version fixture and asserts the refusal enumerates the complete list highest-first (`published versions: 0.2.0, 0.1.0`), which also pins the `matching_versions` ordering contract.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs
```diff
@@ -69,15 +70,39 @@ pub(crate) fn resolve_raw(
...
-            env.os,
-            env.arch,
-            ctx.install_mode.as_str(),
-        ),
+    let entry = index.resolve(&query).map_err(|err| {
+        // A pinned version that the repository simply does not publish gets a
+        // dedicated refusal naming the published alternatives, symmetric with
+        // the rpm backend's version-pin errors; every other resolution
+        // failure keeps the generic query rendering.
+        if let (Some(pinned), ResolveError::NotFound) = (version, &err) {
```
kongche-jbw, line 84: [P2] Avoid reporting an unsupported artifact as unpublished

The `index` used here has already passed through `installable_raw_index`, which removes non-`tar_gz` rows. If a repository publishes `9.0.0` only as `binary` and `0.2.0` as `tar_gz` (the existing mixed-index test covers this shape), pinning `9.0.0` reaches this branch and reports that `9.0.0` is not published even though it is present in the index. Please preserve or query the unfiltered index so this can distinguish published-but-unsupported versions, or change the refusal to describe only installable versions without claiming the pin was unpublished.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -1462,6 +1510,21 @@ impl InstallResultPayload {
         }
         self
     }
+
+    /// Copy the resolved-candidate fields from an owned (raw) version pin:
+    /// the resolved entry version, the exact artifact URL, and the source
+    /// repository — same additive contract as the delegated pin, so agents
+    /// read one shape for both backends.
+    fn with_raw_pin(mut self, pin: &RawPin) -> Self {
```
kongche-jbw, line 1530: [P3] Update the shared pin-field documentation for raw semantics

This method makes the shared fields backend-dependent: for raw installs, `resolved_version` is a distribution version and `artifact` is a URL. The `InstallResultPayload` rustdoc above still says these fields are delegated-only and describes them as an EVR and NEVRA; the `dry_run_detail_lines` rustdoc similarly says owned installs contribute no pin fields. Please update those protocol comments so they document both backend shapes.

kongche-jbw (commented):
Found one behavioral edge case and one documentation mismatch. The implementation and test suite otherwise look sound.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -1661,6 +1724,79 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn raw_pinned_payload_json_exposes_requested_resolved_repo_and_artifact() {
+        // The owned pin must fill the same additive envelope fields as the
+        // delegated pin: requested/resolved version, source repo, and the
+        // exact artifact (the URL, the raw analog of a NEVRA).
+        let payload = InstallResultPayload {
+            component: "agentsight".to_string(),
```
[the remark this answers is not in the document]
ikunkun-sys: Adopted in a78ddfd8, folded into the existing test instead of adding a new one: `raw_pinned_payload_json_exposes_requested_resolved_repo_and_artifact` now starts with `version: None` so the assertion on `json["version"]` exercises the pin-fallback path directly (same pattern as the delegated `pinned_payload_json` test). One test now covers both the additive fields and the compatible-field fill.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -172,6 +172,39 @@ pub(crate) struct DelegatedPin {
...
+    /// of the delegated pin's NEVRA.
+    artifact: String,
+    /// Repository base URL the distribution index was fetched from.
+    source_repo: String,
+}
+
+impl RawPin {
+    /// Capture the pin evidence from a settled raw resolution. Must be built
+    /// before the resolution is consumed by contract validation.
+    fn from_resolution(requested: &str, resolution: &RawResolution) -> Self {
```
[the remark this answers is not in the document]
ikunkun-sys: Partially adopted in a78ddfd8. The ordering is not a bare convention: `validate_owned_install` takes `RawResolution` by value, so building the pin after validation cannot compile — a future reshuffle fails as a borrow error at this call site, never as silently missing pin evidence. Amended the `RawPin::from_resolution` doc to state this ownership guarantee explicitly. Moving construction into `resolve_raw` was considered and rejected: `resolve_raw` is shared with the update path, which has no `--version` pin, so the capture point stays in `execute_planned` where the request context lives.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs
ikunkun-sys: Good catch — adopted in 8422a108. `resolve_raw` now keeps the unfiltered index for error attribution: a pinned version present there but absent from the installable view is reported as "published … but only with artifact types the raw backend cannot install (supported: tar_gz)", while a genuinely absent version keeps the not-published wording. Both branches now list `installable versions:` (renamed from `published versions:` for honesty). New e2e `install_pinned_unsupported_artifact_type_is_reported_as_not_installable` pins the mixed-index shape you described (9.0.0 binary-only + 0.2.0 tar_gz) and asserts the refusal never claims "not published".

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
ikunkun-sys: Adopted in 8422a108: the `InstallResultPayload` rustdoc and the `resolved_version`/`source_repo`/`artifact` field docs now document both backend shapes (delegated: EVR / dnf repo id / NEVRA; raw: distribution version / repository base URL / artifact URL), and the `dry_run_detail_lines` rustdoc no longer claims owned installs contribute no pin fields.

kongche-jbw (approved):
lgtm

PR #1750, opened 2026-07-24
fix(sight): fix syscall tracepoint probe attach on PREEMPT_LAZY kernels

chengshuyi:
## Description

On kernels carrying the PREEMPT_LAZY backport (e.g. RHEL/Anolis 9, kernel 5.14),
`struct trace_entry` gains an extra `preempt_lazy_count` field, widening the
tracepoint common header. The execve enter/exit probes read their context via
vmlinux's `trace_event_raw_sys_{enter,exit}`, which declare the syscall number as
`long` (8 bytes). CO-RE then relocated `ret`/args past the actual `sys_exit_execve`
tracepoint record (only 24 bytes), so the kernel rejected the BPF attach with
`EACCES` (`max_ctx_offset > record size`) and `agentsight trace` aborted with
"Failed to attach probes". `agentsight serve` was unaffected since it uses no
tracepoints.

The fix switches the execve enter/exit probes to the kernel's own
`struct syscall_trace_{enter,exit}`, which declare the syscall number as `int nr`
(4 bytes) — matching the real record layout — so CO-RE relocates `ret`/args to the
correct offset on every kernel. This is the same fix adopted upstream by
inspektor-gadget (#2444) and BCC (#4920), and it is what a syscall tracepoint
program actually receives. proctrace's stdout probe already used
`syscall_trace_enter`, so this also unifies the context types.

## Related Issue

no-issue: kernel-specific probe-attach failure found during on-host validation

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `sight` (agentsight)

## Key Changes

- `src/agentsight/src/bpf/procmon.bpf.c`: `trace_execve_exit` now takes `struct syscall_trace_exit *` instead of `trace_event_raw_sys_exit *`.
- `src/agentsight/src/bpf/proctrace.bpf.c`: `trace_execve_enter` / `trace_execve_exit` now take `struct syscall_trace_enter *` / `struct syscall_trace_exit *`; execve args are read from `ctx->args[0]` (filename) and `ctx->args[1]` (argv), consistent with the existing stdout probe.

## Checklist

- [x] `cargo clippy --all-targets -- -D warnings` pass
- [x] `cargo test` pass (1239 passed, 0 failed)

## Testing

Verified on an Anolis 9 host (kernel `5.14.0-710.el9.x86_64`, PREEMPT_DYNAMIC with
`common_preempt_lazy_count`) and locally on `5.10.134` (standard layout):

- **Before**: `agentsight trace -v` on 5.14 reproducibly failed at
  `trace_execve_exit -> syscalls/sys_exit_execve: -EACCES` -> "Failed to attach
  probes". Confirmed independent of `perf_event_paranoid` (retested at 1 and -1).
- **After (5.14)**: all probes attach cleanly (no EACCES); `RUST_LOG=trace` shows
  procmon exec events flowing, and a process matching an agent rule (`*codex*`) is
  discovered -> `Attached to agent: Codex`. A traced parent spawning a child exec
  ran through proctrace's enter/exit with no attach/poll/panic errors.
- **After (5.10)**: no regression — probes attach and `Attached to agent: Claude`.

Note: attach success itself validates the context accesses, since the BPF verifier
checks every `ctx` offset against the tracepoint record at load time. An automated
regression test is not included because the failure only manifests on kernels with
the PREEMPT_LAZY tracepoint layout; validation was done manually on such a host per
the project's real-kernel eBPF testing rule.

### DIFF
--- a/src/agentsight/src/bpf/procmon.bpf.c
+++ b/src/agentsight/src/bpf/procmon.bpf.c
@@ -12,6 +12,16 @@
 #define NO_CGROUP_FILTER
 #include "common.h"
 
+// Use the kernel's `struct syscall_trace_exit` (what a syscall tracepoint
+// program actually receives), NOT `trace_event_raw_sys_exit`. The latter
+// declares the syscall number as `long id` (8 bytes); on kernels carrying the
+// PREEMPT_LAZY backport (e.g. RHEL/Anolis 9 5.14) `struct trace_entry` grows a
+// `preempt_lazy_count` field, and CO-RE then relocates `ret` past the real
+// tracepoint record, so the kernel rejects the attach with EACCES.
+// `syscall_trace_exit` uses `int nr` (4 bytes), matching the record layout, so
+// `ret` relocates to the correct offset on every kernel. See
+// inspektor-gadget#2444 / BCC#4920.
+
 // Tracepoint for execve exit - captures process execution after it completes
 // NOTE: We use sys_exit_execve instead of sys_enter_execve because:
 // - At sys_enter_execve, the process hasn't completed execve yet
@@ -19,7 +29,7 @@
 // - Reading /proc/[pid]/comm returns old values
 // - At sys_exit_execve, execve has completed and process info is updated
 SEC("tp/syscalls/sys_exit_execve")
-int trace_execve_exit(struct trace_event_raw_sys_exit *ctx)
+int trace_execve_exit(struct syscall_trace_exit *ctx)
 {
     // Check execve return value - skip if failed
     // ret == 0: success, ret < 0: error (e.g., -ENOENT, -EACCES)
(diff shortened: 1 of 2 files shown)


jfeng18 (approved):
LGTM. v2 is cleaner — using vmlinux syscall_trace_exit/enter (int nr, not long id) keeps CO-RE working while avoiding the PREEMPT_LAZY offset drift. Verified: syscall_trace_exit has int nr (4 bytes) so ret never relocates past the record boundary. Also cleans up proctrace.bpf.c by removing the hand-rolled struct in favor of the same vmlinux types + ctx->args[]. Good references to inspektor-gadget#2444 / BCC#4920.

PR #1812, opened 2026-07-25
fix(cosh-ng): [shell] raw action relay watchdog

SunnyQjm:
## Summary

`shell_host` raw relay action harness could hang **indefinitely** when its trailing `exit` was swallowed by a foreground child of the interactive shell, dragging the whole suite (observed 32+ minutes in the cosh-lab container during PR #1787 validation). This PR adds a finish watchdog scoped to the action-driven harness entry points, turning the unbounded hang into a bounded, attributable test failure.

Fixes #1792

## Root Cause

- `spawn_raw_action_relay` (`src/raw_input/spawn.rs`) finishes by writing a trailing `exit\n` to the PTY. (The `finish_input_relay` mentioned in the issue has since been inlined here.)
- If a foreground child of the interactive bash is still reading stdin (e.g. the `ctrl-backslash` test's `read` loop surviving a rare missed `0x1c → SIGQUIT → trap` chain), that `exit` line is consumed by the child instead of the shell.
- `read_raw_until_exit` (`src/shell_host/raw_relay.rs`) only returns when `child.try_wait()` reports an exit status — it has **no timeout**, so the harness never returns.

## Changes

- `ShellHostConfig` gains `raw_action_watchdog: Duration` (default 120s, as suggested in the issue; tests can lower it).
- `run_raw_relay_with_driver` monitors the action driver thread; once it completes (trailing `exit` relayed), the watchdog window starts. Only the three `*_with_actions*` entry points enable it.
- On expiry, `read_raw_until_exit` kills the session child, reaps it, and returns `io::ErrorKind::TimedOut` with an explicit message — the flake stays visible as a failing test instead of being masked or hanging.
- Interactive entry points (`run_raw_interactive_*` / `spawn_raw_input_relay`) pass no watchdog: product interactive behavior is byte-for-byte unchanged. When the watchdog never fires, action-path behavior is also unchanged (fail-safe).
- Follow-up commit `refactor(cosh-ng): [shell] split raw_relay emit`: the watchdog addition pushed `raw_relay.rs` past the 700-line layout gate (first CI run failed on `check-layout.sh`), so the PTY handoff emission domain (`resolve_pty_emit` / `emit_to_pty` / `restore_prompt_display_before_handoff` / `write_handoff_request`) moved verbatim into the `raw_relay/pty_emit.rs` owner submodule, following the existing `input_events`/`terminal_recovery`/`terminal_size` pattern. No behavior change; the layout audit passes again.

## Tests

[...]

PR #1818, opened 2026-07-25
fix(cosh-ng): [shell] restore healthy startup health row

### LINKED ISSUE #1817: Startup banner silently drops the healthy health status row
# Startup banner silently drops the healthy health status row

## Summary

On a fully healthy host, the startup banner renders no health information at all. The startup
health scan still runs, but an all-green report is discarded without emitting the compact
`Health: ok · ...` status row that the startup health scan design requires. This is a rendering
regression introduced while refactoring the startup banner for prompt recommendations, not a
deliberate design change.

## Current behavior

- `render_startup_banner` (`crates/cosh-shell/src/runtime/startup/recommendations.rs`) waits up to
  150ms for the scan report. When the report is all-green (`health_uses_startup_row` returns
  `true`: severity `ok`, no findings, no unavailable entries, no try items), the branch only sets
  `startup_health.rendered = true` and emits nothing.
- The deferred path `render_startup_health_banner` (report arriving after the banner) also returns
  without output for all-green reports.
- As a result, four very different states are visually indistinguishable at startup: healthy host,
  `health.enabled = false`, unsupported platform (live scan gated to Linux), and a late all-green
  report. Users cannot tell "the machine is healthy" from "the scan never ran".
- The fixed resource overview (Load / Mem used / Disk used) that was designed as a stable per-login
  reference is gone for the most common case (healthy host).

## Expected behavior

Restore the startup health scan contract for the `ok` severity:

```text
╭ cosh-shell ──────────────────────────────────────────────────────╮
│ Backend: cosh-core · Shell: bash · Approval: auto                │
│ cwd: /srv/app                                                    │
│ Health: ok · 12ms · Load 1m 0.1 / 4 cores (0.0x) · Mem used 38%  │
│ Disk / used 41%                                                  │
╰──────────────────────────────────────────────────────────────────╯
```

- Healthy report → one logical health status row merged into the startup card (at most 2 visual
  rows at 80 columns), showing status, scan elapsed time, and the fixed overview metrics with
  direction words. No gauge, no prompt suggestions, no standalone full-width panel.
- Warning/critical/degraded → unchanged: standalone health attention panel (already works today).
- Silence stays correct only when the scan genuinely did not run (disabled via config/env or
  unsupported platform). A skipped scan must not be presented as healthy.
- Late all-green report (after the 150ms banner budget) → render the same compact row right after
  the banner instead of dropping it.

## Root cause

- `ccaf6df8` (feat(cosh-shell): add startup health scan) implemented the contract: for an
  all-green report the banner body received a section separator plus
  `health_startup_row_lines(...)`.
- `8d4d74c1` (feat(cosh-ng): add shell prompt recommendations) moved the banner assembly into
  `runtime/startup/recommendations.rs` and dropped the output in the all-green branch; it now only
  marks the report as rendered. `health_startup_row_lines` itself is still alive (used by
  `/health`), so only the startup wiring is broken.
- The raw CLI startup smoke only asserts that a healthy host does not render an empty full-width
  green panel; nothing asserts the presence of the health status row, which let the regression
  land silently. A regression test pinned to the `linux-healthy` fixture should assert the row.

## Environment

- Latest `main` (observed at `ca7abe28`).
- Linux (Anolis dev container), default configuration: `health.enabled = true`, live scan mode,
  `COSH_SHELL_HEALTH_SCAN` unset.

SunnyQjm:
## Summary

Restores the healthy (`ok` severity) health status row inside the startup banner. Since the prompt-recommendation refactor moved banner assembly into `runtime/startup/recommendations.rs`, an all-green health report was silently dropped: the scan still ran, but nothing was rendered, making a healthy host indistinguishable from a host where the scan never ran (disabled config, unsupported platform, or late report).

Fixes #1817

## Root cause

- `render_startup_banner`: the all-green branch (`health_uses_startup_row == true`) only set `startup_health.rendered = true` without emitting the row that the original startup health scan implementation appended to the banner body.
- `render_startup_health_banner` (deferred path, report arriving after the 150ms banner budget): returned early for all-green reports.

## Changes

- Merge the `ok` health row back into the startup card body (separator + `health_startup_row_lines`): status, scan elapsed, fixed overview (Load / Mem used / Disk used). No gauge, no prompt suggestions, no standalone panel — matching the startup health scan spec's severity rendering matrix.
- Deferred path now always renders via `write_health_banner`, which emits the compact row shape for `ok` reports and the attention panel otherwise.
- Scan-not-run stays silent (a skipped scan must not be presented as healthy).

## Tests

- Re-anchored `raw_cli_startup_health_healthy_fixture_keeps_only_default_startup_card` → `raw_cli_startup_health_healthy_fixture_renders_startup_row_in_banner`: asserts `Health: ok`, `Mem used`, `Disk` present; no `Health check` attention panel; no `Suggested prompts`.
- Added `raw_cli_startup_health_disabled_renders_no_row`: `COSH_SHELL_HEALTH_SCAN=disabled` renders neither row nor panel.

## Verification (focused scope)

- `cargo fmt --check` / `cargo clippy --workspace --all-targets -- -D warnings`: green (cosh-lab pr preflight ok).
- `cargo test -p cosh-shell --test raw_cli startup`: 35 passed, 1 failed — `raw_cli_startup_health_context_is_not_attached_to_plain_agent_request` also fails identically on unmodified `main` (`ca7abe28`) in this local environment (agent routing input swallowed by bash, exit 127); unrelated to this change, deferring to CI.
- `cargo test -p cosh-shell --lib health` (98 passed) and `--lib startup` (1 passed): green.
- FAIL→PASS reproduction with `COSH_SHELL_HEALTH_SCAN=fixture:linux-healthy` against a real `cosh-shell raw cosh-core` startup:
  - Before: `grep -c "Health"` = 0 (banner rendered, row absent).
  - After (en-US): `│ Health: ok · 1ms · Load 1m 0.8 / 4 cores (0.2x) · Mem used 51% · Disk / used 42% │` inside the startup card.
  - After (zh-CN): `健康: 正常 · 1ms · 1分钟负载 0.8 / 4核（0.2倍） · 内存已用 51% · 磁盘 / 已用 42%`.

Full workspace suite not run locally; relying on CI for the complete gate.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/runtime/startup/recommendations.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/runtime/startup/recommendations.rs
@@ -54,6 +54,12 @@ pub(crate) fn render_startup_banner<W: Write>(
     let recommendation_notice = append_recommendation_notice(state, &mut body);
     state.startup_health.wait_ready(STARTUP_HEALTH_ROW_WAIT);
     let suggestions = prepare_startup_suggestions(state, cwd);
+    if let Some(report) = state.startup_health.report.as_ref() {
+        if health_uses_startup_row(report) {
+            body.push(renderer.startup_section_separator_line());
+            body.extend(renderer.health_startup_row_lines(HealthBannerModel::new(report)));
+        }
+    }
     if let Some(markdown) = startup_hook.markdown {
         body.push(String::new());
         body.push(startup_hook.summary);
@@ -199,17 +205,14 @@ pub(crate) fn render_startup_health_banner<W: Write>(
         .and_then(|path| path.to_str().map(str::to_string))
         .unwrap_or_else(|| ".".to_string());
     let suggestions = prepare_startup_suggestions(state, &cwd);
-    let show_health = !health_uses_startup_row(&report);
-    if !show_health && suggestions.candidates.is_empty() {
-        return Ok(());
-    }
     write!(output, "\r\x1b[2K")?;
-    if show_health {
-        let mut facts = report.clone();
-        facts.try_items.clear();
-        renderer.write_health_banner(output, HealthBannerModel::new(&facts))?;
-        writeln!(output)?;
-    }
+    // Late all-green reports still surface the compact startup row here;
+    // `write_health_banner` renders the row shape for `ok` reports and the
+    // attention panel otherwise, so both stay visible after the banner.
+    let mut facts = report.clone();
+    facts.try_items.clear();
+    renderer.write_health_banner(output, HealthBannerModel::new(&facts))?;
+    writeln!(output)?;
     write_startup_suggestion_card(
         state,
         &renderer,
--- a/src/cosh-ng/crates/cosh-shell/tests/raw_cli/startup.rs
+++ b/src/cosh-ng/crates/cosh-shell/tests/raw_cli/startup.rs
@@ -222,6 +222,24 @@ fn raw_cli_startup_health_fixture_renders_when_enabled() {
 }
 
 #[test]
+fn raw_cli_startup_health_disabled_renders_no_row() {
+    let output = run_raw_cli_with_env(
+        "fake",
+        "exit\n",
+        &[
+            ("COSH_SHELL_STARTUP_BANNER", "1"),
+            ("COSH_SHELL_HEALTH_SCAN", "disabled"),
+            ("COSH_SHELL_LANG", "en-US"),
+            ("TERM", "xterm-256color"),
+        ],
+    );
+
+    assert!(output.contains("/help"), "{output}");
+    assert!(!output.contains("Health: ok"), "{output}");
+    assert!(!output.contains("Health check"), "{output}");
+}
+
+#[test]
 fn raw_cli_startup_health_cards_share_configured_width() {
     let cwd = temp_shell_home("startup-health-card-width");
     let suppression_store = cwd.join("health-suppression");
@@ -344,7 +362,7 @@ fn raw_cli_startup_health_degraded_fixture_is_read_only() {
 }
 
 #[test]
-fn raw_cli_startup_health_healthy_fixture_keeps_only_default_startup_card() {
+fn raw_cli_startup_health_healthy_fixture_renders_startup_row_in_banner() {
     let cwd = temp_shell_home("startup-health-healthy-fixture");
     let suppression_store = cwd.join("health-suppression");
     let suppression_store = suppression_store.to_string_lossy().into_owned();
@@ -366,7 +384,9 @@ fn raw_cli_startup_health_healthy_fixture_keeps_only_default_startup_card() {
     );
 
     assert!(output.contains("cosh-shell"), "{output}");
-    assert!(!output.contains("Health:"), "{output}");
+    assert!(output.contains("Health: ok"), "{output}");
+    assert!(output.contains("Mem used"), "{output}");
+    assert!(output.contains("Disk"), "{output}");
     assert!(!output.contains("Health check"), "{output}");
     assert!(!output.contains("Suggested prompts"), "{output}");
     assert_inline_before_followup(&output, "╭ cosh-shell", "exit");


Forrest-ly (approved):
Code review 通过。代码质量良好，变更聚焦且测试充分。

**审查摘要**

- **正确性**：修复了 `render_startup_banner` 中 all-green 健康报告不渲染的问题，通过在 banner body 中正确添加 separator + health row。deferred path 也移除了错误的 early return，确保延迟到达的健康报告仍能渲染。逻辑清晰正确。
- **代码质量**：使用 `as_ref()` 避免不必要的 clone，`try_items.clear()` 只读模式保持一致。注释准确描述了 `write_health_banner` 对 ok/非 ok 报告的分支行为。
- **测试覆盖**：healthy fixture 测试从断言"不存在"改为断言"存在具体字段"（`Health: ok`, `Mem used`, `Disk`），新增 disabled-scan 负面测试。覆盖充分。
- **安全性/性能**：无问题。变更范围小，无新的 clone 开销。

整体是一个高质量的 bug fix PR，描述详尽，验证充分。

SunnyQjm:
PR number: #1818
head_sha: 1dcee9e8998ea2aa7919df1ba5a015bc5ab10c43
reviewed_at: 2026-07-25T09:50:18Z


# 评审结论

未发现 blocking package/module/public API 组织问题。变更范围仅限 `crates/cosh-shell/`，owner 归属正确（`runtime/startup/recommendations.rs` 属 `runtime` owner 的 startup orchestration 职责），未触碰 `lib.rs` public surface、root `src` 入口、forbidden dependency direction 或 self-crate path。未新增 crate、未新增依赖、未新增 `pub mod`/`pub use`。

## Findings

- **[P2] `runtime/startup/recommendations.rs:205-218` — 延迟路径移除早退后的潜在双重渲染风险**

  原代码在 `render_startup_health_banner` 中通过 `!health_uses_startup_row(&report) && suggestions.is_empty()` 做早退，阻止 all-green report 在已由 banner body 渲染后再次输出。新代码移除该早退，改为始终调用 `write_health_banner`。

  若 report 在 `STARTUP_HEALTH_ROW_WAIT` 内到达，`render_startup_banner` 新增的 body 渲染分支（第 57-62 行）会输出健康行；此时若 `render_startup_health_banner` 仍被调用且不依赖 `startup_health.rendered` 标志做跳过，则同一 all-green report 会被渲染两次（一次在 banner body 内，一次在延迟覆写区）。

  建议：确认延迟路径的调用方是否已通过 `rendered` 标志保证"banner body 已渲染则不再进入延迟路径"；若无此保护，应保留对已渲染报告的跳过逻辑。

- **[P2] `tests/raw_cli/startup.rs:386-388` — 测试断言硬编码英文文案**

  `Health: ok`、`Mem used`、`Disk` 等断言与 i18n catalog 文案耦合，后续文案调整或 locale 切换会导致测试破裂。与上一轮 Qoder review 的 P2 建议一致，建议改为结构性断言或按 locale 分支断言。

## Open Questions

- CI `Test cosh-ng` 结论为 FAILURE，PR 描述提到其中一项失败在 main 基线同样复现且与本变更无关，但需确认 CI 失败集是否仅限该已知项，无其他回归。
- 新增的 `raw_cli_startup_health_disabled_renders_no_row` 测试覆盖了 `COSH_SHELL_HEALTH_SCAN=disabled` 场景，但未覆盖 report 在 banner budget 内到达时延迟路径不被调用的契约——建议补一个断言确认 `rendered` 标志在同步路径中被置位。

## Validation

- **Not run**：本轮为静态评审，未执行 `cargo check`、`cargo test`、`check-layout.sh`、`inventory-public-api.sh`。
- **Not run**：文件行数检查（无法确认 `recommendations.rs` 是否接近 700 行阈值）。
- 依赖 CI 完成完整 `cosh-shell` focused gates。

SunnyQjm:
Thanks for the review. All findings and open questions verified and closed out below; the two actionable items landed in follow-up PR #1827.

## [P2] Deferred path double-render risk (`recommendations.rs:205-218`)

**Verification result**: no double render is possible. The synchronous banner path sets `state.startup_health.rendered = true` whenever a report was available (all-green or not), and the deferred `render_startup_health_banner` entry guard bails out on that flag (`if !state.rendered_startup_banner || state.startup_health.rendered { return Ok(()); }`). The two paths are mutually exclusive by construction: the deferred path only runs when the report missed the 150ms banner budget, in which case the banner body never contained the row.

**Hardening**: #1827 pins the contract in the healthy fixture test with `assert_eq!(output.matches("Health: ok").count(), 1)`, so any future regression that re-renders an all-green report after the banner path would fail deterministically.

## [P2] Hardcoded English assertions (`tests/raw_cli/startup.rs:386-388`)

**Verification result**: this follows the established file-wide convention. All 35+ startup raw CLI tests pin `COSH_SHELL_LANG=en-US` and assert English catalog strings (`Health check`, `Findings`, `Suggested prompts`, `Approval: trust`, ...). Changing only the new assertions to structural form would make this one test inconsistent with its siblings without removing the coupling class. If the catalog-coupled assertion style should change, it deserves a dedicated sweep across the suite rather than a one-off here — happy to file that as a separate issue if desired.

## Open Question: CI failure set

**Verified — the CI failure was NOT the known local flake, and it was caused by this PR**: the `Test cosh-ng` job failed on the test inventory gate (`violation: cosh-shell source inventory changed from 2467 to 2468; update the necessity audit`), because this PR added one source test without refreshing the floor. Compounded by #1812 landing first (floor → 2468 covering its own +1), `main` at `90b10965` is currently red (actual 2469 vs floor 2468). Follow-up #1827 refreshes the floor to 2469; no registry change is needed since the new test matches the existing `shell-raw-cli crates/cosh-shell/tests/raw_cli/*` pattern. Apologies for the gate slip — the local focused run did not include the inventory script, which is now part of the checklist.

## Open Question: `rendered` flag coverage

Covered by the exactly-once assertion in #1827 (see P2-1 above): if the sync path failed to set `rendered` and the deferred path re-emitted the row, the count would be 2 and the test fails.

## Evidence

Before (all-green silently dropped, `main` prior to this PR):

healthy quiet before

After (en-US, real `cosh-shell raw cosh-core` startup, `fixture:linux-healthy`):

healthy row after en

After (zh-CN):

healthy row after zh

Issue #1814, opened 2026-07-25
[nightly] anolisa uninstall --dry-run regression: rejects uninstalled component with INVALID_ARGUMENT (regression from #1471)


zhangtaibo:
### Component

anolisa

### Bug Description

`anolisa uninstall <component> --dry-run` for an **uninstalled** component returns `ok: false` with `INVALID_ARGUMENT` ("component 'cosh' is not installed — nothing to uninstall"), instead of returning a dry-run preview with `ok: true` and `data.dry_run: true`.

This is a regression from issue #1471 (CLOSED), which previously fixed the dry-run uninstall behavior to return a plan-view. The product has since reverted to rejecting the dry-run uninstall of uninstalled components entirely, breaking the contract that `--dry-run` should always succeed and show what would happen.

The test `test_dry_run_uninstall_carries_dry_run_flag` expects `r.ok` to be `True` with `data.dry_run: true`, but the product returns `ok: false`.

This issue was discovered by AgenticOS Nightly automated testing on 2026-07-24 (nightly-20260724-020046, anolisa main 681b3253).

### Steps to Reproduce

```shell
# 1. Ensure cosh is NOT installed via anolisa
anolisa status --json
# data.components should be empty

# 2. Dry-run uninstall cosh (not installed)
anolisa uninstall cosh --dry-run --json
# Expected: ok=true, data.dry_run=true (preview of what would happen)
# Actual: ok=false, error.code=INVALID_ARGUMENT, "component 'cosh' is not installed — nothing to uninstall"
```

### Expected Behavior

`uninstall --dry-run` should always return `ok: true` with a dry-run preview, even if the component is not installed. The preview should show `data.dry_run: true` and either an empty plan or a "nothing to uninstall" warning, but should NOT return an error.

### Suggested Fix

In `src/anolisa/crates/anolisa-cli/src/commands/tier1/uninstall.rs`, the `plan_error_to_cli` function maps `PlanError::NotInstalled` to `CliError::InvalidArgument` unconditionally. When `ctx.dry_run` is true, this should instead render a dry-run preview (empty plan with `dry_run: true`), not an error.

The dry-run path (line 283) is reached only after the plan succeeds (line 277), but `PlanError::NotInstalled` is returned as `Err` before the dry-run check. The fix: intercept `PlanError::NotInstalled` when `ctx.dry_run` is true and render a preview instead of erroring.

### Environment

- ALinux 4 ECS (cn-hongkong)
[...]

PR #123, opened 2026-04-07
feat(agent-sec-core): add bundled loongshield build flow

chenzongyao200127:
## Description

 This PR adds a bundled `loongshield` build flow to `agent-sec-core` and updates the build/docs flow around it.

`agent-sec-core` now exposes a single `make build-all` entrypoint that always builds `linux-sandbox` and attempts to build the bundled `third_party/loongshield` submodule when the host environment is supported. If the host does not pass `loongshield`'s `env-check`, the `loongshield` build is skipped and `agent-sec-core` still builds successfully.

### Changes

  - add `loongshield` as a git submodule under `src/agent-sec-core/third_party/loongshield`
  - add `make loongshield` and `make build-all` targets in `agent-sec-core`
  - update `make install` to install bundled `loongshield` when the host supports building it
  - keep build/install behavior non-blocking on unsupported hosts by checking `loongshield` build prerequisites first
  - update the top-level `scripts/build-all.sh` flow to initialize the submodule and use `make build-all`
  - update root and component docs to use the new build/install commands and describe submodule initialization

 ### Why

  Before this change, `agent-sec-core` only documented/builds `linux-sandbox` directly. This PR makes the bundled hardening dependency part of the standard build flow while preserving compatibility
  on hosts that cannot build `loongshield`.

 ### Testing

  - ran `make -n build-all` in `src/agent-sec-core` to verify the new build flow and fallback behavior
  - reviewed updated documentation and top-level build script references for consistency

## Related Issue

https://github.com/alibaba/anolisa/issues/106

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional change)
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [ ] `cosh` (copilot-shell)
- [x] `sec-core` (agent-sec-core)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] Multiple / Project-wide

## Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [x] My code follows the project's code style
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly
- [ ] For `cosh`: Lint passes, type check passes, and tests pass
[...]

PR #1915, opened 2026-07-27
fix(anolisa): give an absent component its own error code

ikunkun-sys:
## Description

Every command that acts on an installed component reported an absent target as
`INVALID_ARGUMENT` — the same code a bad invocation gets. A scripted caller could
not tell "there was nothing to act on" from "the invocation was wrong" without
matching on the human-readable reason string.

This adds `CliError::NotInstalled` → `NOT_INSTALLED` and routes the condition to
it from `uninstall`, `update`, `repair`, `forget`, `restart`, and `adapter`. All
six report the same underlying state, so they must report the same code; leaving
any of them behind would make the code depend on which command the caller
happened to run. Genuine invocation errors keep `INVALID_ARGUMENT`: enabled
adapter claims, a quarantined record needing `repair`, and
`--remove-system-package` with several targets all still route there.

**Scope of the code — state absence only.** It is not a statement about the name.
Resolution falls back to the literal input when the component index has no entry
for it, so an unknown name and a known-but-absent one arrive indistinguishably.
The docs say this explicitly and an integration test pins it.

**Exit code stays 2** for backward compatibility alone — shell callers already
branch on 2. Sharing it does not mean the two conditions call for the same
response, and what to do about an absent target depends on the command: an absent
`uninstall` target is already the desired end state, an absent `restart` target is
a stale assumption. The CLI reports the state and leaves that judgement to the
caller. The distinction is carried by `error.code` for `--json` consumers; reason
strings are byte-for-byte unchanged.

### Note on how this differs from what #1814 requested

#1814 asks for `uninstall <absent> --dry-run` to return `ok: true` with an empty
preview, and describes the current refusal as a regression from #1471. This PR
deliberately does not do that. The reasoning:

1. **It is not a regression from #1471.** #1471 was about *response shape* — the
   plan-view missing `data.dry_run`, `phases[0].mode = "execute"` being
   misleading, and `warnings` saying "plan is empty" while `phases[]` was
   non-empty. None of its four suggested fixes asked for dry-run to always
   succeed. Its fix landed in 0f97c779; the current behaviour came later from
   2d40b9a2, a declared breaking lifecycle-pipeline rebuild, not a silent revert.

2. **`--dry-run` is documented as "Print plan without executing"**, and the
   architecture is observe → plan → execute. `PlanError::NotInstalled` is a
   *planning* refusal, so dry-run reporting it is dry-run doing its job.

3. **A preview that succeeds where the real run fails is worse for automation.**
   The typical agent sequence is `uninstall X --dry-run` → decide → `uninstall X`.
   If the preview returns `ok: true` and the real run then returns `ok: false`
   with exit 2, the preview actively misled the caller. That also reintroduces
   the plan-view/execution-view asymmetry #1471 set out to remove — just moved
   from field shape to exit status.

4. **Fixing only `uninstall` would create a fresh inconsistency**, since
   `update.rs` and `repair.rs` map `PlanError::NotInstalled` identically.

What #1814 genuinely exposes is that `INVALID_ARGUMENT` is the wrong code for
this condition — the argument is valid, only the state precondition is unmet, and
the caller has no machine-readable way to tell the two apart. That is what this
PR fixes, without giving up the property that a preview never disagrees with the
run it previews.

## Related Issue

Relates to #1814 — deliberately **not** a closing reference; see the section above
for why this does not implement that issue's stated acceptance contract. Leaving
the issue open for maintainers to decide, per review feedback.

no-issue: this is a contract-level fix split out from the #1814 discussion rather
than an implementation of its requested behaviour.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

Flagged as breaking for honesty: a `--json` consumer that asserts
`error.code == "INVALID_ARGUMENT"` for an absent component will now see
`NOT_INSTALLED`. Exit code, `schema_version`, envelope shape, and every reason
string are unchanged, so shell callers branching on exit status are unaffected.

## Scope

- [ ] `cosh` (copilot-shell)
- [ ] `cosh-ng` (cosh-ng)
- [ ] `skill` (os-skills)
- [ ] `sight` (agentsight)
- [ ] `tokenless` (tokenless)
- [ ] `ckpt` (ws-ckpt)
- [ ] `memory` (agent-memory)
- [x] `anolisa` (anolisa-cli)
- [ ] `skillfs` (SkillFS)
- [ ] `blaze` (blaze)
- [ ] Multiple / Project-wide

## Checklist

Documentation: `src/anolisa/CHANGELOG.md` `[Unreleased]` → `Changed`. The exit-code
contract lives in the `response.rs` module doc rather than `docs/`, and is updated
there. No `docs/` page enumerates CLI error codes today, so none needed a change.

## Testing

All three gate commands run on `aarch64-unknown-linux-gnu`, toolchain 1.88.0:

```
cargo fmt --all --check                          # exit 0, no diff
cargo clippy --all-targets --locked -- -D warnings   # exit 0, no warnings
cargo test --locked                              # 1885 passed, 0 failed
```

New tests:

- `not_installed_does_not_distinguish_a_known_name_from_an_unknown_one` — drives
  the real binary with a seeded component index and asserts that an index-known
  name (`cosh`) and an unknown one (`coshh`) produce the identical code. This
  pins the corrected contract rather than the wording, so if identity validation
  ever lands, this is the test that fails and flags every doc describing the code.
- `response::tests::not_installed_has_its_own_code_but_keeps_exit_two` — locks
  both halves of the contract: the code differs from `INVALID_ARGUMENT`, the exit
  code does not.
- `response::tests::not_installed_survives_command_restamping` — `with_command`
  re-stamps every variant; a missing arm would silently leave the wrong verb in
  the envelope.
- `commands::adapter::tests::component_not_installed_maps_to_not_installed` —
  cross-command consistency, so a caller never has to know which command
  surfaced the condition.

Updated to assert the new code (behaviour otherwise unchanged):
`uninstall_unknown_component_routes_to_not_installed_exit_2`,
`uninstall_dry_run_on_unknown_component_reports_not_installed`,
`update::unknown_component_routes_to_not_installed`,
`repair::absent_component_is_not_installed`,
`forget_unknown_component_routes_to_not_installed`,
`restart_unknown_component_returns_not_installed`, and the subprocess
wire-contract test `uninstall_dry_run_json_absent_component_reports_not_installed`.

Manual reproduction of the issue's command, before and after:

```
$ anolisa uninstall cosh --dry-run --json
# before: error.code = INVALID_ARGUMENT, exit 2
# after:  error.code = NOT_INSTALLED,    exit 2   (reason text identical)
```

## Additional Notes

If maintainers prefer the idempotent-`uninstall` route instead (absent target is
a no-op success, `rm -f` / `kubectl delete --ignore-not-found` semantics), that is
a defensible design — but it should change the real execution path and dry-run
together, and be applied across `update`/`repair` as well. That is a larger
breaking change than this one and is deliberately left out of scope here.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/response.rs
+++ b/src/anolisa/crates/anolisa-cli/src/response.rs
@@ -13,14 +13,22 @@
 //!   exact value, so we pick a non-zero reserved code and document it
 //!   here for future tightening).
 //! - `INVALID_ARGUMENT` -> 2 (POSIX convention shared with clap).
+//! - `NOT_INSTALLED` -> 2 (shares `INVALID_ARGUMENT`'s exit code only so
+//!   shell callers already branching on 2 keep working; `--json` callers
+//!   read the distinction off `error.code`). Reports one fact and only
+//!   one: the target is absent from state, so the operation had nothing
+//!   to act on. It says nothing about whether the name was valid — an
+//!   unknown name resolves to itself and lands here identically. What to
+//!   do about it is the caller's call and depends on the command: an
+//!   absent `uninstall` target is already the desired end state, an
+//!   absent `restart` target is a stale assumption.
 //! - `EXECUTION_FAILED` -> 1 (generic non-zero "the command ran but the
 //!   underlying operation failed at runtime"). Distinct from
 //!   `INVALID_ARGUMENT` so callers can tell "I gave you bad input" apart
 //!   from "you tried and something on the machine refused": download
 //!   IO, install IO, state-write IO, log-write IO, lock IO. Plan-time
-//!   refusals (e.g. blocked plan, unknown component) stay
-//!   `INVALID_ARGUMENT` — they tell the caller to fix the input or the
-//!   environment before retrying.
+//!   refusals (e.g. a blocked plan) stay `INVALID_ARGUMENT` — they tell
+//!   the caller to fix the input or the environment before retrying.
 
 use std::process::ExitCode;
 
@@ -69,6 +77,26 @@ pub enum CliError {
     /// Caller-supplied arguments violated a contract.
     #[error("invalid argument: {reason}")]
     InvalidArgument { command: String, reason: String },
+
+    /// The target is absent from ANOLISA state, so the operation had
+    /// nothing to act on.
+    ///
+    /// Kept separate from [`CliError::InvalidArgument`] so a scripted
+    /// caller can tell a state precondition from a bad invocation —
+    /// enabled adapters, a quarantined record, or
+    /// `--remove-system-package` with several targets are all things the
+    /// caller got wrong, whereas this one may be nothing to fix at all.
+    ///
+    /// **Not** a claim about the name. Resolution falls back to the literal
+    /// input when the component index has no entry for it, so an unknown
+    /// name and a known-but-absent one arrive here indistinguishably.
+    /// Callers must not read this code as "the name was valid".
+    ///
+    /// Shares exit code 2 with [`CliError::InvalidArgument`] for backward
+    /// compatibility alone — shell callers already branch on 2. It does not
+    /// imply the two call for the same response; see the module docs.
+    #[error("not installed: {reason}")]
+    NotInstalled { command: String, reason: String },
 
     /// The command was well-formed but the underlying operation failed
     /// at runtime (download IO, install IO, state-write IO, log-write
@@ -119,6 +147,7 @@ impl CliError {
         match self {
             Self::NotImplemented { .. } => "NOT_IMPLEMENTED",
             Self::InvalidArgument { .. } => "INVALID_ARGUMENT",
+            Self::NotInstalled { .. } => "NOT_INSTALLED",
             Self::Runtime { .. } => "EXECUTION_FAILED",
             Self::Degraded { .. } => "DEGRADED",
             Self::PermissionDenied { .. } => "PERMISSION_DENIED",
@@ -131,6 +160,7 @@ impl CliError {
         match self {
             Self::NotImplemented { .. } => 64,
             Self::InvalidArgument { .. } => 2,
+            Self::NotInstalled { .. } => 2,
             Self::Runtime { .. } => 1,
             Self::Degraded { .. } => 2,
             Self::PermissionDenied { .. } => 5,
@@ -143,6 +173,7 @@ impl CliError {
         match self {
             Self::NotImplemented { command, .. } => command,
             Self::InvalidArgument { command, .. } => command,
+            Self::NotInstalled { command, .. } => command,
             Self::Runtime { command, .. } => command,
             Self::Degraded { command, .. } => command,
             Self::PermissionDenied { command, .. } => command,
@@ -155,6 +186,7 @@ impl CliError {
         match self {
             Self::NotImplemented { hint, .. } => hint.as_deref(),
             Self::InvalidArgument { .. } => None,
+            Self::NotInstalled { .. } => None,
             Self::Runtime { .. } => None,
             Self::Degraded { .. } => None,
             Self::PermissionDenied { hint, .. } => hint.as_deref(),
@@ -169,6 +201,7 @@ impl CliError {
                 format!("command '{command}' is not implemented")
             }
             Self::InvalidArgument { reason, .. } => reason.clone(),
+            Self::NotInstalled { reason, .. } => reason.clone(),
             Self::Runtime { reason, .. } => reason.clone(),
             Self::Degraded { reason, .. } => reason.clone(),
             Self::PermissionDenied { reason, .. } => reason.clone(),
@@ -202,6 +235,7 @@ impl CliError {
         match &mut self {
             Self::NotImplemented { command: c, .. }
             | Self::InvalidArgument { command: c, .. }
+            | Self::NotInstalled { command: c, .. }
             | Self::Runtime { command: c, .. }
             | Self::Degraded { command: c, .. }
             | Self::PermissionDenied { command: c, .. }
@@ -367,6 +401,45 @@ mod tests {
             }
             other => panic!("expected CliError::Runtime, got {other:?}"),
         }
+    }
+
+    /// `NOT_INSTALLED` is the whole point of the variant: a caller that
+    /// wants to skip an absent target must be able to key off `code`
+    /// alone. The exit code deliberately stays 2, shared with
+    /// `INVALID_ARGUMENT`, so existing shell callers keep working.
+    #[test]
+    fn not_installed_has_its_own_code_but_keeps_exit_two() {
+        let err = CliError::NotInstalled {
+            command: "uninstall cosh".to_string(),
+            reason: "component 'cosh' is not installed — nothing to uninstall".to_string(),
+        };
+
+        assert_eq!(err.code(), "NOT_INSTALLED");
+        assert_ne!(
+            err.code(),
+            CliError::InvalidArgument {
+                command: "uninstall".to_string(),
+                reason: "bad".to_string(),
+            }
+            .code(),
+            "an absent target must not be indistinguishable from a bad argument",
+        );
+        assert_eq!(err.exit_code(), 2);
+        assert!(err.reason().contains("not installed"));
+    }
+
+    /// `with_command` re-stamps every variant; a missing arm would silently
+    /// leave the wrong command verb in the envelope.
+    #[test]
+    fn not_installed_survives_command_restamping() {
+        let err = CliError::NotInstalled {
+            command: "install".to_string(),
+            reason: "component 'cosh' is not installed".to_string(),
+        }
+        .with_command("update cosh");
+
+        assert_eq!(err.command(), "update cosh");
+        assert_eq!(err.code(), "NOT_INSTALLED");
     }
 
     #[test]
--- a/src/anolisa/crates/anolisa-cli/tests/uninstall_dry_run_json.rs
+++ b/src/anolisa/crates/anolisa-cli/tests/uninstall_dry_run_json.rs
@@ -2,10 +2,12 @@
 //!
 //! Plain uninstall runs the planner pipeline: for an absent component the
 //! dry-run reports the same refusal as a real run would (an error envelope,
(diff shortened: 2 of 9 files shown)


ikunkun-sys:
Thanks for the review. Addressed point 2; declining 1 and 3, with reasoning below. Pushed as an amend: `8c6ec6d9` → `fed3f45d`.

### 2. `NOT_INSTALLED` exit-code semantics — valid, fixed (differently)

You caught a genuine contradiction. The module doc said an absent target is
"usually safe to skip", while the variant doc justified the shared exit code with
"both mean *fix the input or the environment before retrying*". Those pull in
opposite directions, and the second phrasing was mine — I reached for it to
justify sharing exit 2 and it undercut the actual guidance.

I did not apply the suggested wording, though, because "推荐调用方将该操作视为可安
全跳过的缺失状态" is not true as a blanket rule:

- `uninstall X` on an absent target → skipping is right; the desired end state
  already holds.
- `update X` on an absent target → skipping is probably wrong; the caller most
  likely wanted it installed.
- `restart X` on an absent target → usually means the caller's playbook is stale.

A single recommendation across all six commands would trade one inaccuracy for
another. So instead of adding a sentence, I removed the contradiction at its
source: the exit code is now documented as shared *for backward compatibility
alone*, explicitly not as a claim that the two conditions warrant the same
response, and the docs say the right response is command-dependent — the CLI
reports the state and leaves the judgement to the caller.

```
//! - `NOT_INSTALLED` -> 2 (shares `INVALID_ARGUMENT`'s exit code only so
//!   shell callers already branching on 2 keep working; `--json` callers
//!   read the distinction off `error.code`). Split out because the two
//!   warrant different responses: a wrong name has to be corrected, an
//!   absent target may already be the state the caller wanted. Which one
//!   it is depends on the command — an absent `uninstall` target is the
//!   desired end state, an absent `restart` target is a stale assumption
//!   — so the CLI reports the state and leaves that judgement to the
//!   caller.
```

The same stale phrasing was in the commit message body; the amend fixes it there
too.

### 1. CHANGELOG should call out adapter — declining, the premise doesn't hold

Two factual corrections:

- `adapter` is already named in the entry — `src/anolisa/CHANGELOG.md:15` lists
  "`uninstall`, `update`, `repair`, `forget`, `restart`, and `adapter`".
- The comment says adapter's behaviour "之前行为不同" (differed before). It did
  not. `AdapterError::ComponentNotInstalled` mapped to `INVALID_ARGUMENT`,
  identical to the other five. It was uniform before and is uniform after.

Singling adapter out would assert a distinction that never existed and imply
adapter callers need to do something the other five don't.

### 3. JSON example in the test-file header — declining, conflicts with AGENTS.md

`AGENTS.md:143` is explicit: **"No 'fixes issue #123' in comments — put that in
the PR description."** The suggestion is to move the PR's #1814 expectation
comparison into a code comment, which is the direction that rule prohibits.
`AGENTS.md:127` ("If one line suffices, do not write two") points the same way.

There's also a practical problem: that contract is already stated executably
~40 lines below, in `uninstall_dry_run_json_absent_component_reports_not_installed`,
which asserts `ok == false` and `error.code == "NOT_INSTALLED"` against the real
binary. A hand-written JSON blob in the header would duplicate an assertion that
cannot drift, using prose that can.

The underlying concern — that a caller should be able to understand the contract
without reading the PR — is fair, but a test file is not where API consumers
look. The place for it is `response.rs`, which is exactly what point 2's fix
improves.

### Verification

Re-ran the full gate set on the amended commit (`aarch64-unknown-linux-gnu`, 1.88.0):

```
cargo fmt --all --check                              exit 0, no diff
cargo clippy --all-targets --locked -- -D warnings   exit 0, no warnings
cargo test --locked                                  1884 passed, 0 failed
cargo doc -p anolisa-cli --no-deps                   no warnings (intra-doc links resolve)
```

Only doc comments changed in this round; no functional diff from the reviewed revision.

src/anolisa/crates/anolisa-cli/tests/uninstall_dry_run_json.rs
```diff
@@ -105,7 +107,8 @@ fn uninstall_dry_run_json_absent_component_reports_not_installed() {
     let error = value.get("error").expect("envelope must carry error");
     assert_eq!(
         error.get("code").and_then(|v| v.as_str()),
-        Some("INVALID_ARGUMENT"),
+        Some("NOT_INSTALLED"),
```
kongche-jbw, line 117: [P1] Keep #1814's dry-run contract

#1814 explicitly requires an absent `uninstall --dry-run` to return `ok: true` with `data.dry_run: true`. This assertion instead locks in `ok: false` and exit 2, so the reported regression remains unchanged while the PR declares `closes #1814`. Either implement the issue's acceptance contract or stop closing #1814 and treat this error-code change as a separate proposal.

src/anolisa/crates/anolisa-cli/src/response.rs
```diff
@@ -70,6 +78,21 @@ pub enum CliError {
     #[error("invalid argument: {reason}")]
     InvalidArgument { command: String, reason: String },
 
+    /// The named target is well-formed but absent from ANOLISA state, so
+    /// the lifecycle operation has nothing to act on.
+    ///
+    /// A refinement of [`CliError::InvalidArgument`]: the argument is
+    /// valid, only the state precondition is unmet. Kept separate so a
+    /// scripted caller can tell `uninstall coshh` (typo — escalate) apart
```
kongche-jbw: [P1] Do not promise typo detection

The documented `coshh` versus `cosh` distinction is not implemented. `resolve_mutation_target` falls back to the literal input when the component index has no match, so an unknown typo reaches the same absent-state path and returns `NOT_INSTALLED`; the integration test confirms this with an empty component index and `definitely-missing`. Automation following this guidance can silently skip typos. Validate known identities before emitting `NOT_INSTALLED`, or define the code as state absence without claiming it detects typos.

kongche-jbw (changes requested):
Requesting changes for two contract-level issues:

1. The behavior required by #1814 remains unchanged: an absent `uninstall --dry-run` still returns `ok: false` instead of a successful preview.
2. `NOT_INSTALLED` cannot currently distinguish an unknown typo from a known-but-absent component, despite the documented safe-skip use case.

The focused tests and CI pass, but the tests currently lock in these semantics rather than validating the stated contracts.

ikunkun-sys:
@kongche-jbw Both points are correct. Pushed as an amend: `fed3f45d` → `8e84204f`.

### P1: "Do not promise typo detection" — you're right, and I verified it

I reproduced it against the built binary rather than taking it on reading alone.
Seeded a component index containing `cosh` and ran both names:

```
$ anolisa --json --dry-run --install-mode system --prefix $P uninstall cosh
{ "ok": false, "command": "uninstall cosh",
  "error": { "code": "NOT_INSTALLED",
             "reason": "component 'cosh' is not installed — nothing to uninstall …" } }

$ anolisa --json --dry-run --install-mode system --prefix $P uninstall coshh
{ "ok": false, "command": "uninstall coshh",
  "error": { "code": "NOT_INSTALLED",
             "reason": "component 'coshh' is not installed — nothing to uninstall …" } }
```

Identical but for the echoed name. The path is `component_alias_from_repo_index`
(`commands/common.rs:379-380`): `lookup_component_alias(...).unwrap_or_else(|| input.to_string())`
falls back to the literal input when the index has no entry, and the planner then
returns `PlanError::NotInstalled` either way.

So the claim was false, and it was not a stray sentence — it was the headline
justification in the docs, the commit message, and the PR description. That is the
worse failure here: a PR whose entire purpose is to define a contract described
that contract inaccurately.

I took your second option — define the code as state absence — rather than
implementing identity validation, because validation has correctness hazards that
deserve their own change: the component index is loaded best-effort and may be
absent (offline, no repo config), in which case the classification silently
degrades; and a component adopted out-of-band but missing from the index would be
misreported as a typo. Doing that properly across all six commands is a larger
design change than this one, and I would rather not bundle it. Noted as a possible
follow-up rather than done badly here.

What changed:

- `response.rs` — the variant now documents state absence explicitly, with a
  **"Not a claim about the name"** paragraph naming the literal-input fallback.
  The module doc likewise.
- Removed the claim from the `adapter.rs` comment, the `uninstall.rs` test doc,
  the CHANGELOG entry, the commit message, and the PR description.
- The CHANGELOG now says what the code *does* distinguish — "there was nothing to
  act on" vs "the invocation was wrong" — which is true and still useful:
  `AdapterClaimsActive`, `NeedsAttention`, and
  `RemoveSystemPackageRequiresSingleTarget` all remain `INVALID_ARGUMENT`.

On your point that the tests lock in semantics rather than validating the stated
contract: fair, and I added one that validates the corrected contract by
falsifying the old claim —
`not_installed_does_not_distinguish_a_known_name_from_an_unknown_one`. It drives
the real binary with a seeded index and asserts an index-known name and an unknown
one yield the identical code. If identity validation ever lands, that test is the
one that fails, which is the signal to revisit every doc describing the code.

### P1: "Keep #1814's dry-run contract" — dropped the closing reference

Agreed, and done: `closes #1814` is gone from the PR description, replaced with a
plain reference plus a pointer to the section explaining why the issue's stated
acceptance contract is not implemented here. #1814 stays open for you to decide.

I have not changed the design position — an absent `uninstall --dry-run` still
refuses, so a preview never disagrees with the run it previews — but you're right
that declining an issue's acceptance criteria while auto-closing it is not a
defensible combination. Treating this as a separate contract fix is the honest
framing.

### Verification

```
cargo fmt --all --check                              exit 0, no diff
cargo clippy --all-targets --locked -- -D warnings   exit 0, no warnings
cargo test --locked                                  1885 passed, 0 failed
```

No behaviour change in this round — the corrections are to documentation and the
new test. The one thing I would flag for your judgement: with the typo claim gone,
`NOT_INSTALLED` earns its keep purely on separating state absence from invocation
errors. I think that is still worth a distinct code, but if you read it as too
thin a distinction to justify a wire-contract change, that is a reasonable call and
I would rather hear it now than defend the code on a premise that already turned
out to be wrong once.

kongche-jbw (approved):
lgtm

PR #2061, opened 2026-07-31
fix(anolisa): relax rpm probe on deb-family hosts

ikunkun-sys:
## Description

`anolisa install` refused raw-backend installs on hosts without RPM tooling (e.g. Ubuntu), failing with `rpm not found on PATH` even though the raw backend does not need rpm/dnf. The failure came from the planner's I3 system-RPM presence probe, which ran unconditionally for fresh raw system-scope installs.

The fix separates "the probe cannot run" from "the probe found evidence": the probe still runs whenever rpm tooling is present, and any `Present`/`MultipleVersions` evidence keeps driving the I3 refusal and the locked recheck on every host. Only a missing rpm binary is reinterpreted through the host package family — now formally exposed as `anolisa_env::package_family` over `/etc/os-release` `ID` + `ID_LIKE`, so derivatives declaring deb lineage classify without name lists. Deb-family hosts degrade to `NotProbed`; rpm-family and unrecognized hosts (Alpine, Arch, ...) keep failing closed. Host facts are injected into install planning (`plan_component`), so the family-sensitive branches are covered deterministically in tests instead of depending on the runner's real os-release.

## Related Issue

no-issue: user-reported runtime bug fixed directly

## Type of Change

## Scope

- [x] `anolisa` (anolisa-cli)

## Testing

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo doc --workspace --no-deps` all pass
- `cargo test --workspace` passes; the review's six-case matrix is covered deterministically via injected `EnvFacts` on a single runner:
  - deb + `CommandMissing` → raw install succeeds (`deb_host_missing_rpm_tooling_does_not_block_raw_install`)
  - deb + installed RPM evidence → refuses, points at adopt (`deb_host_present_system_rpm_still_points_at_adopt`)
  - deb + RPM appears under lock → refuses (`deb_host_raw_install_rechecks_native_absence_under_lock`)
  - rpm-family + `CommandMissing` → fail closed (`system_install_without_rpm_tooling_warns_and_exits`)
  - unknown + `CommandMissing` → fail closed (`unknown_host_missing_rpm_tooling_fails_closed`)
  - explicit `--backend rpm` + `CommandMissing` → still fails (`explicit_rpm_without_tooling_warns_and_exits`, pinned to a deb env)
- `anolisa_env::package_family` unit tests cover ID priority, ID_LIKE token fallback, and unknown classification

## Checklist

- [x] I have updated the documentation accordingly (CHANGELOG `[Unreleased]`)

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
@@ -8,7 +8,7 @@
 //! executor with a [`StoreRecordSink`]. No lifecycle policy lives here.
 
 use std::collections::HashSet;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use anolisa_core::executor::{DelegatedExecutionTarget, execute_delegated_steps};
 use anolisa_core::facts::{
@@ -70,10 +70,14 @@ pub(crate) fn handle_one(
 ) -> Result<InstallOutcome, CliError> {
     let mut activity = install_activity(&component, ctx);
     let (query, txn) = host_backends(&component, &args, ctx)?;
+    let env = anolisa_env::EnvService::detect();
+    let rpmdb = RpmdbProbe::for_host(&env);
     install_component_with_deps_and_planned(
         &component,
         &args,
         ctx,
+        &env,
+        &rpmdb,
         &query,
         &txn,
         privilege::is_root(),
@@ -92,10 +96,14 @@ pub(crate) fn handle_one_with_planned_components(
 ) -> Result<InstallOutcome, CliError> {
     let mut activity = install_activity(&component, ctx);
     let (query, txn) = host_backends(&component, &args, ctx)?;
+    let env = anolisa_env::EnvService::detect();
+    let rpmdb = RpmdbProbe::for_host(&env);
     install_component_with_deps_and_planned(
         &component,
         &args,
         ctx,
+        &env,
+        &rpmdb,
         &query,
         &txn,
         privilege::is_root(),
@@ -291,11 +299,41 @@ pub(crate) fn install_component_with_deps(
     txn: &dyn PackageTransaction,
     is_root: bool,
 ) -> Result<InstallOutcome, CliError> {
+    let env = anolisa_env::EnvService::detect();
+    install_component_with_deps_and_env(
+        input,
+        args,
+        ctx,
+        &env,
+        &RpmdbProbe::absent(),
+        query,
+        txn,
+        is_root,
+    )
+}
+
+/// Test variant with the host facts injected, so family-sensitive branches
+/// (deb degrade, rpm fail-closed, unknown fail-closed) are covered
+/// deterministically instead of depending on the runner's /etc/os-release.
+#[cfg(test)]
+#[expect(clippy::too_many_arguments)]
+pub(crate) fn install_component_with_deps_and_env(
+    input: &str,
+    args: &InstallArgs,
+    ctx: &CliContext,
+    env: &anolisa_env::EnvFacts,
+    rpmdb: &RpmdbProbe,
+    query: &dyn PackageQuery,
+    txn: &dyn PackageTransaction,
+    is_root: bool,
+) -> Result<InstallOutcome, CliError> {
     let mut activity = install_activity(input, ctx);
     install_component_with_deps_and_planned(
         input,
         args,
         ctx,
+        env,
+        rpmdb,
         query,
         txn,
         is_root,
@@ -311,17 +349,21 @@ fn install_component_with_deps_and_planned(
     input: &str,
     args: &InstallArgs,
     ctx: &CliContext,
+    env: &anolisa_env::EnvFacts,
+    rpmdb: &RpmdbProbe,
     query: &dyn PackageQuery,
     txn: &dyn PackageTransaction,
     is_root: bool,
     planned_components: &HashSet<String>,
     reporter: &mut dyn ProgressReporter,
 ) -> Result<InstallOutcome, CliError> {
-    let planned = plan_component(input, args, ctx, query, txn)?;
+    let planned = plan_component(input, args, ctx, env, rpmdb, query, txn)?;
     execute_planned(
         planned,
         args,
         ctx,
+        env,
+        rpmdb,
         query,
         txn,
         is_root,
@@ -338,6 +380,8 @@ pub(crate) fn plan_component(
     input: &str,
     args: &InstallArgs,
     ctx: &CliContext,
+    env: &anolisa_env::EnvFacts,
+    rpmdb: &RpmdbProbe,
     query: &dyn PackageQuery,
     txn: &dyn PackageTransaction,
 ) -> Result<PlannedComponent, CliError> {
@@ -350,7 +394,6 @@ pub(crate) fn plan_component(
         InstallMode::User => InstallationScope::User { uid },
     };
     let now = now_iso8601();
-    let env = anolisa_env::EnvService::detect();
     let repo_config = common::load_repo_config(ctx, &layout, COMMAND, RepoPersistPolicy::Require)?;
 
     // Resolve identity across all visible roots, but bind install planning to
@@ -442,7 +485,8 @@ pub(crate) fn plan_component(
                 InstallationScope::System => Some(system_probe_package(
                     args,
                     &layout,
-                    &env,
+                    env,
+                    rpmdb,
                     &repo_config,
                     &component,
                     query,
@@ -474,7 +518,7 @@ pub(crate) fn plan_component(
                 let fresh = resolve_fresh_delegated(
                     args,
                     &layout,
-                    &env,
+                    env,
                     &repo_config,
                     &component,
                     query,
@@ -498,23 +542,48 @@ pub(crate) fn plan_component(
     };
     // A missing rpm/dnf binary is a hard error whenever a probe was needed:
     // without it the host cannot prove the component is not an unobserved
-    // system RPM, and a raw install over one could corrupt it (I3).
-    let facts = assemble_facts(
+    // system RPM, and a raw install over one could corrupt it (I3). The one
+    // exception is the raw family on a host whose package authority is not
+    // RPM — there is no rpmdb to protect, so the facts degrade to NotProbed.
+    // The package identity is kept: the executor's locked recheck re-probes
+    // it and applies the same CommandMissing policy, so tooling (and an RPM)
+    // appearing between planning and placement is still caught.
+    let facts = match assemble_facts(
         &observe_request,
         &store,
         Some(&provider),
         &layout,
         &journal_dir,
-    )
-    .map_err(|err| match err {
-        FactsError::Probe(ProviderError::Query(PackageQueryError::CommandMissing {
+    ) {
+        Ok(facts) => facts,
+        Err(FactsError::Probe(ProviderError::Query(PackageQueryError::CommandMissing {
             command: bin,
-        })) => rpm_tooling_missing_error(&command, &bin, &component),
-        err => CliError::Runtime {
-            command: command.clone(),
-            reason: err.to_string(),
-        },
-    })?;
+        }))) => {
+            if family != "raw" || missing_rpm_tooling_is_fatal(env, rpmdb) {
+                return Err(rpm_tooling_missing_error(&command, &bin, &component));
+            }
+            let observe_request = ObserveRequest {
+                kind: ObjectKind::Component,
+                name: &component,
+                scope,
+                native_package: None,
+                observed_at: &now,
+                verify_owned_files: false,
+            };
+            assemble_facts(&observe_request, &store, None, &layout, &journal_dir).map_err(
+                |err| CliError::Runtime {
+                    command: command.clone(),
+                    reason: err.to_string(),
+                },
+            )?
+        }
+        Err(err) => {
+            return Err(CliError::Runtime {
+                command: command.clone(),
+                reason: err.to_string(),
+            });
+        }
+    };
 
     let request = InstallRequest {
         target,
@@ -574,6 +643,8 @@ fn execute_planned(
     planned: PlannedComponent,
     args: &InstallArgs,
     ctx: &CliContext,
+    env: &anolisa_env::EnvFacts,
+    rpmdb: &RpmdbProbe,
     query: &dyn PackageQuery,
     txn: &dyn PackageTransaction,
     is_root: bool,
@@ -593,7 +664,6 @@ fn execute_planned(
         route,
     } = planned;
     let layout = common::resolve_layout(ctx);
-    let env = anolisa_env::EnvService::detect();
     let repo_config = common::load_repo_config(ctx, &layout, COMMAND, RepoPersistPolicy::Require)?;
     let state_path = layout.state_dir.join("installed.toml");
     let journal_dir = rpm_install::journal_dir(&layout);
@@ -629,7 +699,7 @@ fn execute_planned(
                 args,
                 ctx,
                 &layout,
-                &env,
+                env,
                 &repo_config,
                 ResolvedInstallIdentity {
                     component: component.clone(),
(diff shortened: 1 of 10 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -498,23 +521,47 @@ pub(crate) fn plan_component(
...
-        err => CliError::Runtime {
-            command: command.clone(),
-            reason: err.to_string(),
-        },
-    })?;
+        }))) => {
+            if family != "raw" || missing_rpm_tooling_is_fatal(env) {
+                return Err(rpm_tooling_missing_error(&command, &bin, &component));
+            }
+            native_package = None;
```
kongche-jbw: [P1] 这里仍然把 package identity 与 probe 失败策略绑定在一起。Deb-family 主机若 planning 时返回 `CommandMissing`，`native_package` 会被清空；在后续 artifact 下载/获锁前，另一个进程若安装了 rpm 工具并写入同名 RPM，`install_owned` 仍会以 `None` 调用 `revalidate_native_absence`，直接跳过查询并继续 raw placement。当前 `deb_host_raw_install_rechecks_native_absence_under_lock` 只覆盖“planning 时 tooling 已可用”的情况，没有覆盖 missing → available/present 的转换。请保留 package identity，并在 planning 与锁内重查各自仅将 `CommandMissing` 按 deb-family 降级；一旦重查可以运行并得到 `Present`/`MultipleVersions`，仍必须执行 I3 拒绝。

src/anolisa/CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `anolisa install` no longer refuses raw-backend installs on deb-family
```
kongche-jbw: [P2] 请从本 PR 移除这条 CHANGELOG 更新。`specs/documentation-standard.md` §5 明确规定，日常 feature/fix PR 不写 CHANGELOG；CHANGELOG 只在 release version bump PR 中汇总用户可感知变更。

kongche-jbw (changes requested):
上一轮的主要回归已经修复，新增的 family 分类与确定性测试也验证通过；当前 CI 全绿。仍有一处 I3 竞态需要补齐，另外 CHANGELOG 修改不符合仓库最新文档规范，请处理后再合入。

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
ikunkun-sys: Fixed in `ade92976`. The planning degrade no longer drops the probe identity: `system_probe_package` keeps the component name as fallback identity when candidate resolution hits `CommandMissing`, and only the facts degrade to `NotProbed`. `revalidate_native_absence` now takes the same CommandMissing policy as the planning probe (`missing_tooling_fatal`): the raw executor passes the deb-family verdict, while the delegated and batch paths keep the unconditional hard error. So under the lock the re-probe always runs with the package identity — if tooling appeared and reports `Present`/`MultipleVersions`, I3 refuses; only a *still* missing binary is forgiven on deb-family hosts. New test `deb_host_tooling_appearing_under_lock_still_refuses` covers exactly the missing → available/present transition (FakeInstaller returns `CommandMissing` until the install lock is held, then presence evidence), asserting the refusal and that no owned record is claimed.

src/anolisa/CHANGELOG.md
ikunkun-sys: Removed in `ade92976` — `CHANGELOG.md` is restored to the main version, per `specs/documentation-standard.md` §5 the entry will be aggregated in the next release version bump PR. No README/user-guide update is triggered: this PR changes no CLI command, flag, or config option.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -886,20 +957,23 @@ fn system_probe_package(
...
-    .map_err(|err| match err {
-        PackageQueryError::CommandMissing { command: bin } => {
-            rpm_tooling_missing_error(command, &bin, component)
+    ) {
+        Ok(candidates) => candidates,
+        Err(PackageQueryError::CommandMissing { command: bin }) => {
+            if missing_rpm_tooling_is_fatal(env) {
+                return Err(rpm_tooling_missing_error(command, &bin, component));
+            }
+            return Ok(component.to_string());
```
kongche-jbw: [P1] `CommandMissing` 时不能无条件回退到 component 名：若用户显式传了 `--package custom-rpm`，这个 override 本身就是已知的 probe identity，但这里会把它丢成 `component`。随后 tooling 与 `custom-rpm` 在锁前出现时，锁内重查会查询错误的 component 名，得到 `Absent` 后继续 raw placement，I3 竞态仍然存在。`ComponentResolver::resolve_rpm` 的 override 分支会调用 provider query，因此该路径确实可能在 tooling 缺失时落到这里。请至少优先保留 `args.package`（再回退 component），并增加 missing → available/present + `--package` 不同于 component 的回归测试。

kongche-jbw (changes requested):
上一轮两项请求已验证修复，旧线程已解决，当前 CI 与定向测试均通过。仍有一个 `--package` override 下的同类 identity 丢失问题，请补齐后再合入。

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
ikunkun-sys: Fixed in `7cc92a71`. `system_probe_package` now shares one fallback for both the CommandMissing degrade and the non-unique-candidates arm: an explicit `--package` override wins, then the component name — so the locked recheck re-probes exactly the RPM the user named. Regression test `deb_host_package_override_survives_planning_degrade` covers the full path you described: `--package custom-rpm` differing from the component, `CommandMissing` at planning (reached through the `resolve_rpm` override branch), then tooling plus `custom-rpm` appearing before the lock — asserting the refusal names `custom-rpm` (not the component) and no owned record is claimed. Verified red/green: the test fails against the previous component-name fallback.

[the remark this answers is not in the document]
ikunkun-sys: This was reviewed against the previous revision `ade92976`; the override loss was fixed in `7cc92a71` together with the human review round — `system_probe_package` now falls back to `args.package` before the component name in both the CommandMissing degrade and the non-unique-candidates arm, and `deb_host_package_override_survives_planning_degrade` covers exactly this race (verified red/green).

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -875,7 +925,29 @@ fn resolve_owned_artifact(
...
+/// RPM is the host's package authority. When rpm tooling is present the
+/// probe always runs; this only decides what a missing binary means. A
+/// host whose `/etc/os-release` `ID`/`ID_LIKE` classifies as deb-family
+/// has no rpmdb to protect, so the probe degrades to NotProbed instead of
+/// demanding rpm/dnf the distro does not ship. Rpm-family and unrecognized
+/// hosts keep the hard error, so a genuinely rpm-based host with broken
+/// tooling still fails closed.
+pub(crate) fn missing_rpm_tooling_is_fatal(env: &anolisa_env::EnvFacts) -> bool {
+    anolisa_env::package_family(env.os_id.as_deref(), env.os_id_like.as_deref()).as_deref()
+        != Some("deb")
```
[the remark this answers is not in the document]
ikunkun-sys: Fixed in `4edd6ec3`. `missing_rpm_tooling_is_fatal` now consults filesystem evidence before trusting the family inference: `host_rpmdb_exists` checks `var/lib/rpm` and `usr/lib/sysimage/rpm` under the layout prefix for `rpmdb.sqlite`/`Packages`/`Packages.db` (sqlite, bdb, ndb backends), deliberately file-based so it works exactly when the rpm binary does not. A deb-family host with a leftover rpmdb keeps the hard error in both the planning probe and the locked recheck. Covered by `rpmdb_on_disk_overrides_the_deb_family_relaxation` (all backend/location combinations) and the e2e `deb_host_with_leftover_rpmdb_still_fails_closed`; also verified in an Ubuntu 24.04 container — creating `/var/lib/rpm/rpmdb.sqlite` flips the install to the fail-closed refusal, removing it restores the raw install.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -885,24 +974,32 @@ fn system_probe_package(
...
-    .map_err(|err| match err {
-        PackageQueryError::CommandMissing { command: bin } => {
-            rpm_tooling_missing_error(command, &bin, component)
+    ) {
+        Ok(candidates) => candidates,
+        Err(PackageQueryError::CommandMissing { command: bin }) => {
+            if missing_rpm_tooling_is_fatal(env, layout) {
+                return Err(rpm_tooling_missing_error(command, &bin, component));
+            }
+            return Ok(fallback());
```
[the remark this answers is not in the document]
ikunkun-sys: Valid scenario; fixed in `e1c263ef` via evidence rather than re-resolution. Two observations shaped the fix: (1) the "provider already installed before planning" variant is closed by the previous round — an installed RPM implies an on-disk rpmdb, which `host_rpmdb_exists` now makes fatal at planning; (2) for the remaining window (provider + tooling appear between planning and lock), any RPM installation necessarily writes the rpmdb, so the locked recheck now re-verifies `host_rpmdb_exists` under the degraded policy and refuses when a database appeared — regardless of whether the probe identity matches what was installed, which also covers the differently-named `Provides`-only case. The retry then re-plans with working tooling and full candidate resolution, yielding the precise I3 adopt guidance. This is strictly stronger than re-running resolution under the lock: it also catches the case where tooling never becomes visible to this process. Unit test `locked_recheck_refuses_on_rpmdb_evidence_even_when_probe_reports_absent` pins exactly the "query says Absent for a fallback identity, evidence refuses anyway" behavior.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -1329,7 +1443,11 @@ pub(crate) fn revalidate_native_absence(
             reason: format!("locked system-RPM probe for '{package}' did not run"),
         }),
         Err(ProviderError::Query(PackageQueryError::CommandMissing { command: bin })) => {
-            Err(rpm_tooling_missing_error(command, &bin, target))
+            if missing_tooling_fatal {
+                Err(rpm_tooling_missing_error(command, &bin, target))
+            } else {
+                Ok(())
```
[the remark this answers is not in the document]
ikunkun-sys: Correct — the boolean was a pre-lock snapshot. Fixed in `e1c263ef`: `revalidate_native_absence` now takes the layout and re-runs `host_rpmdb_exists` under the lock whenever the degraded policy is in effect, both before the probe and again in the `CommandMissing` arm, refusing with a dedicated "rpm database appeared" error (nothing changed; a retry re-plans against the database). Covered by the e2e `deb_host_rpmdb_growing_under_lock_still_refuses`, which simulates exactly the PATH-divergence you describe: every query keeps failing with `CommandMissing` while the rpmdb file appears once the install lock is held (verified red/green — removing the locked recheck fails the test). Delegated and batch paths pass the fatal policy and are unaffected.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
```diff
@@ -875,7 +925,46 @@ fn resolve_owned_artifact(
...
+        != Some("deb")
+        || host_rpmdb_exists(layout)
+}
+
+/// Filesystem evidence of an rpmdb under the layout's prefix, covering the
+/// classic and `sysimage` locations across the bdb, ndb, and sqlite
+/// backends. Deliberately file-based: the check must work exactly when the
+/// rpm binary does not.
+fn host_rpmdb_exists(layout: &FsLayout) -> bool {
+    ["var/lib/rpm", "usr/lib/sysimage/rpm"].iter().any(|dir| {
```
kongche-jbw: [P1] 这里查到的并不一定是 `PackageQuery` 所对应的宿主 rpmdb。首先，Debian Bookworm 自带 rpm 的 `%_dbpath` 默认就是 `~/.rpmdb`，不是这两个系统目录（Debian package source: https://sources.debian.org/src/rpm/4.18.0%2Bdfsg-1%2Bdeb12u1/macros.in/#L107-L110）。这正是本 PR 放宽的目标 family，因此“安装过 RPM、后来 tooling 消失”的 Debian 主机仍会被漏检。其次，`FsLayout::prefix` 是 `--prefix` 重定位后的 ANOLISA 安装根；传入 `/opt/x` 时这里查的是 `/opt/x/var/lib/rpm`，但 `PackageQuery` 仍查询当前宿主，真实数据库位于宿主 `/var/lib/rpm`、`/usr/lib/sysimage/rpm` 或该执行用户的 `~/.rpmdb`。于是自定义 prefix 也会错误降级并绕过 I3。请把 host rpmdb evidence 与 install layout 分离：至少覆盖实际宿主的两个系统路径及 `env.home/.rpmdb`，测试需要隔离时注入独立的 probe roots，而不要复用安装 prefix；并补充 Debian 默认 dbpath 与 non-root/custom-prefix 回归测试。

kongche-jbw (changes requested):
已验证上一轮 `--package` 修复、锁内 rpmdb 增长处理和相关测试，旧线程已解决，当前 CI 全绿。新加入的 rpmdb evidence 探测仍使用了错误且不完整的根路径，会在 Debian 默认配置和自定义 `--prefix` 下漏检，请修正后再合入。

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
ikunkun-sys: Both points were correct — fixed in `1a4efbc3` by decoupling the evidence probe from the install layout entirely. New `RpmdbProbe` type: production (`RpmdbProbe::for_host`) probes the real host root — `/var/lib/rpm` and `/usr/lib/sysimage/rpm` — plus `env.home/.rpmdb` for Debian's default `%_dbpath`; the planning check and the locked recheck share the same probe. `FsLayout` no longer participates: a `--prefix` relocates ANOLISA's own tree, not the host database. Tests inject isolated probe roots (`RpmdbProbe::with_roots` / `::absent()`) instead of reusing the install prefix, so they never read the runner's real rpmdb. New regressions: `debian_default_home_dbpath_also_counts_as_rpmdb_evidence` (~/.rpmdb with all three db file names), `rpmdb_evidence_ignores_the_install_prefix` (db inside the prefix is not host evidence; host db refuses regardless of prefix), and the existing leftover/growing/race tests re-staged onto isolated host roots. Also verified in an Ubuntu 24.04 container: `~/.rpmdb/Packages` alone now flips the install to fail-closed, `/var/lib/rpm` likewise, and removal restores the raw install.

kongche-jbw (approved):
已复核最新提交 `1a4efbc3`，没有剩余的阻塞问题。

`RpmdbProbe` 已将宿主 rpmdb evidence 与安装 prefix 解耦，覆盖 `/var/lib/rpm`、`/usr/lib/sysimage/rpm` 和 `env.home/.rpmdb`，并在锁内重新检查；这也补齐了 `--package` override、Provides-only identity 以及 rpmdb/tooling 并发出现时的 I3 防护。

验证结果：本地定向回归测试全部通过；当前 head 的 PR Lint 与完整 CI 均成功。

非阻塞：PR 描述仍勾选了“已更新 CHANGELOG”，但本轮已按仓库规范移除 CHANGELOG 修改，合并前请同步修正该勾选。

PR #2063, opened 2026-07-31
chore(ci): centralize repository governance metadata

kongche-jbw:
## Why

Component names, commit scopes, issue options, routing labels, CI keys, and
release prefixes need one reviewed source of truth. Keeping these rules in
separate workflows allows repository governance to drift.

## What Changed

- add `.github/components.json` as the reviewed component metadata inventory
- define component owners through `issue_triagers`
- add and validate the planned governance label taxonomy
- validate component uniqueness, Issue Form options, CODEOWNERS annotations,
  maintainer mappings, component status values, and effective triager fallback
- enforce bidirectional consistency so removed components cannot leave stale
  commit scopes, Issue Form options, CODEOWNERS labels, or maintainer scopes
- validate every public component path against its CODEOWNERS label mapping
- make PR scope guidance consume the shared component metadata

## User / Agent Impact

No product runtime behavior changes. Repository automation can validate Codex
component decisions without hard-coded component or owner lists.

## Related Issue

no-issue: repository governance groundwork approved as part of the community
workflow cleanup

## Risk and Compatibility

- [ ] Public CLI/API/config changed
- [ ] Privileged/security-sensitive path changed
- [x] Cross-workflow metadata contract added

This PR does not change existing Issue automation by itself. PR #2066 consumes
the metadata only after this PR is merged.

## Validation

- `python3 .github/scripts/validate-repository-metadata.py`
- governance metadata regression suite — 21 passed
- local validation: 13 components and 15 governance labels

## Rollout and Rollback

Merge this PR before #2066. Reverting it removes the metadata contract and must
be paired with pausing the external Issue triage dispatcher.

### DIFF
--- a/.github/components.json
+++ b/.github/components.json
@@ -1,0 +1,275 @@
+{
+  "schema_version": 1,
+  "defaults": {
+    "issue_triagers": [
+      "kongche-jbw"
+    ]
+  },
+  "components": [
+    {
+      "id": "anolisa",
+      "display_name": "ANOLISA CLI",
+      "aliases": [
+        "anolisa-cli"
+      ],
+      "issue_option": "anolisa",
+      "path_prefixes": [
+        "src/anolisa/"
+      ],
+      "label": "component:anolisa",
+      "commit_scope": "anolisa",
+      "commit_scope_aliases": [],
+      "release_tag_prefix": "anolisa/v",
+      "ci_key": "anolisa",
+      "issue_triagers": [
+        "kongche-jbw",
+        "ikunkun-sys"
+      ],
+      "status": "active"
+    },
+    {
+      "id": "cosh",
+      "display_name": "Cosh",
+      "aliases": [
+        "copilot-shell"
+      ],
+      "issue_option": "cosh",
+      "path_prefixes": [
+        "src/copilot-shell/"
+      ],
+      "label": "component:cosh",
+      "commit_scope": "cosh",
+      "commit_scope_aliases": [
+        "copilot-shell"
+      ],
+      "release_tag_prefix": "cosh/v",
+      "ci_key": "copilot_shell",
+      "issue_triagers": [
+        "kongche-jbw",
+        "samchu-zsl"
+      ],
+      "status": "active"
+    },
+    {
+      "id": "cosh-ng",
+      "display_name": "Cosh NG",
+      "aliases": [],
+      "issue_option": "cosh-ng",
+      "path_prefixes": [
+        "src/cosh-ng/"
+      ],
+      "label": "component:cosh-ng",
+      "commit_scope": "cosh-ng",
+      "commit_scope_aliases": [],
+      "release_tag_prefix": "cosh-ng/v",
+      "ci_key": "cosh_ng",
+      "issue_triagers": [
+        "KaiLongZhou",
+        "SunnyQjm"
+      ],
+      "status": "active"
+    },
+    {
+      "id": "sec-core",
+      "display_name": "Agent Security Core",
+      "aliases": [
+        "agent-sec-core"
+      ],
+      "issue_option": "sec-core",
+      "path_prefixes": [
+        "src/agent-sec-core/"
+      ],
+      "label": "component:sec-core",
+      "commit_scope": "sec-core",
+      "commit_scope_aliases": [
+        "agent-sec-core"
+      ],
+      "release_tag_prefix": "sec-core/v",
+      "ci_key": "agent_sec_core",
+      "issue_triagers": [
+        "edonyzpc",
+        "kid9",
+        "RemindD"
+      ],
+      "status": "active"
+    },
+    {
+      "id": "skill",
+      "display_name": "OS Skills",
+      "aliases": [
+        "os-skills"
+      ],
+      "issue_option": "skill",
+      "path_prefixes": [
+        "src/os-skills/"
+      ],
+      "label": "component:skill",
+      "commit_scope": "skill",
+      "commit_scope_aliases": [
+        "os-skills"
+      ],
+      "release_tag_prefix": "skill/v",
+      "ci_key": null,
+      "issue_triagers": [
+        "Ziqi002"
+      ],
+      "status": "active"
+    },
+    {
+      "id": "sight",
+      "display_name": "AgentSight",
+      "aliases": [
+        "agentsight"
+      ],
+      "issue_option": "sight",
+      "path_prefixes": [
+        "src/agentsight/"
+      ],
+      "label": "component:sight",
+      "commit_scope": "sight",
+      "commit_scope_aliases": [
+        "agentsight"
+      ],
+      "release_tag_prefix": "sight/v",
+      "ci_key": "agentsight",
+      "issue_triagers": [
+        "chengshuyi",
+        "jfeng18"
+      ],
+      "status": "active"
+    },
+    {
+      "id": "tokenless",
+      "display_name": "Tokenless",
+      "aliases": [],
+      "issue_option": "tokenless",
+      "path_prefixes": [
+        "src/tokenless/"
+      ],
+      "label": "component:tokenless",
+      "commit_scope": "tokenless",
+      "commit_scope_aliases": [],
+      "release_tag_prefix": "tokenless/v",
+      "ci_key": "tokenless",
+      "issue_triagers": [
+        "Forrest-ly",
+        "shiloong",
+        "ikunkun-sys"
+      ],
+      "status": "active"
+    },
+    {
+      "id": "ckpt",
+      "display_name": "Workspace Checkpoint",
+      "aliases": [
+        "ws-ckpt"
+      ],
+      "issue_option": "ckpt",
+      "path_prefixes": [
+        "src/ws-ckpt/"
+      ],
+      "label": "component:ckpt",
+      "commit_scope": "ckpt",
+      "commit_scope_aliases": [
+        "ws-ckpt"
+      ],
+      "release_tag_prefix": "ckpt/v",
+      "ci_key": "ws_ckpt",
+      "issue_triagers": [
+        "Ziqi002"
+      ],
+      "status": "active"
+    },
+    {
+      "id": "memory",
+      "display_name": "Agent Memory",
+      "aliases": [
+        "agent-memory"
+      ],
+      "issue_option": "memory",
+      "path_prefixes": [
+        "src/agent-memory/"
+      ],
+      "label": "component:memory",
+      "commit_scope": "memory",
+      "commit_scope_aliases": [
+        "agent-memory"
+      ],
+      "release_tag_prefix": "memory/v",
+      "ci_key": "agent_memory",
+      "issue_triagers": [
+        "shiloong",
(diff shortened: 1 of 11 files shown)


.github/scripts/validate-repository-metadata.py
```diff
@@ -0,0 +1,303 @@
...
+codeowners_path = GITHUB / "CODEOWNERS"
+try:
+    codeowners_text = codeowners_path.read_text(encoding="utf-8")
+except FileNotFoundError:
+    errors.append("missing .github/CODEOWNERS")
+    codeowners_text = ""
+
+codeowner_mappings: dict[str, set[str]] = {}
+for line in codeowners_text.splitlines():
+    annotation = re.search(r"#\s*auto-label:\s*(component:[a-z0-9-]+)", line)
```
ikunkun-sys: [P2] Validate the complete auto-label token

This regex stops at the last `[a-z0-9-]`, so an annotation such as `component:cosh!` is recorded as `component:cosh` and validation succeeds. `.github/workflows/pr-opened.yml` instead captures the complete non-whitespace token and will try to apply `component:cosh!`; because that label does not exist, the PR auto-label job fails. The validator and consumer therefore accept different grammars for the same contract.

.github/scripts/validate-repository-metadata.py
```diff
@@ -0,0 +1,303 @@
...
+    for c in components
+    if isinstance(c, dict) and c.get("release_tag_prefix")
+]
+for field, values in (
+    ("component id", ids),
+    ("component label", labels),
+    ("path prefix", paths),
+    ("release tag prefix", release_prefixes),
+):
+    for duplicate in sorted(duplicates(values)):
```
ikunkun-sys, line 227: [P2] Reject ambiguous component routing keys

This uniqueness check omits `issue_option`, `commit_scope`, `commit_scope_aliases`, and `ci_key`. I could assign the same `ci_key` to two components, or make two components share a commit scope or Issue Form option while synchronizing commitlint/the forms, and the validator still exited 0. Those keys route humans and automation, so duplicate values make the new source of truth ambiguous.

.github/components.json
```diff
@@ -0,0 +1,278 @@
...
+      ],
+      "issue_option": "anolisa",
+      "path_prefixes": [
+        "src/anolisa/"
+      ],
+      "label": "component:anolisa",
+      "commit_scope": "anolisa",
+      "commit_scope_aliases": [],
+      "release_tag_prefix": "anolisa/v",
+      "ci_key": "anolisa",
```
ikunkun-sys, line 23: [P2] Cross-check CI and release keys against their consumers

`ci_key` is not validated at all, and `release_tag_prefix` is checked only for being non-empty and unique. Changing this component to `ci_key: "not_a_real_ci_output"` and `release_tag_prefix: "wrong/v"` still passes, while `.github/workflows/ci.yaml` and `.github/workflows/release.yaml` remain hard-coded to the real values. The metadata can therefore drift from actual CI and release behavior despite being presented as the single source of truth.

ikunkun-sys (changes requested):
Three correctness gaps remain in the governance metadata contract. The existing regression suite and GitHub checks pass, but targeted mutation tests show that each inline scenario is still accepted. Please address these findings before merging.

Also, `c23321b9` fixes behavior introduced by `03c82887` in this PR, so it should be autosquashed into the originating commit per `AGENTS.md` §13.

.github/scripts/validate-repository-metadata.py
```diff
@@ -0,0 +1,192 @@
...
+    for prefix in component["path_prefixes"]:
+        require(prefix.startswith("src/") and prefix.endswith("/"), f"{component_id}: invalid path prefix {prefix!r}")
+
+    if component["status"] == "active":
+        require(bool(component["issue_option"]), f"{component_id}: active components need an issue_option")
+        require(bool(component["commit_scope"]), f"{component_id}: active components need a commit_scope")
+        require(bool(component["release_tag_prefix"]), f"{component_id}: active components need a release_tag_prefix")
+        require(bool(component["issue_triagers"]), f"{component_id}: active components need at least one issue_triager")
+    elif not component["issue_triagers"]:
+        warnings.append(f"{component_id}: no issue triager configured ({component['status']})")
```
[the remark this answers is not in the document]
kongche-jbw: Resolved in latest revision: the validator now rejects any status outside `active`, `incubating`, and `internal`, with regression coverage for unknown values.

.github/scripts/validate-repository-metadata.py
```diff
@@ -0,0 +1,192 @@
...
+try:
+    codeowners_text = codeowners_path.read_text(encoding="utf-8")
+except FileNotFoundError:
+    errors.append("missing .github/CODEOWNERS")
+    codeowners_text = ""
+
+annotated_labels = set(re.findall(r"#\s*auto-label:\s*(component:[a-z0-9-]+)", codeowners_text))
+for component in components:
+    if component.get("status") == "active" and component["label"] not in annotated_labels:
+        errors.append(f"CODEOWNERS: no # auto-label annotation for {component['label']}")
```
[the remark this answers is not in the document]
kongche-jbw: Resolved in latest revision: the validator now checks every public path-to-label mapping in both directions and rejects stale CODEOWNERS paths.

.github/scripts/validate-repository-metadata.py
kongche-jbw: Resolved in latest revision: the validator now captures the complete `component:*` token, so malformed suffixes such as `component:cosh!` are rejected.

.github/scripts/validate-repository-metadata.py
kongche-jbw: Fixed in `eabf2547`: non-empty issue options and CI keys must be unique, while primary and alias commit scopes share one globally unique namespace, with regression coverage for each contract.

.github/components.json
kongche-jbw: Fixed in `eabf2547`: component CI keys are checked against declared, emitted, consumed, and path-routed outputs, release prefixes are checked against trigger, parse, and publish routes, and the metadata workflow now runs when either consumer changes.

.github/scripts/validate-repository-metadata.py
```diff
@@ -0,0 +1,451 @@
...
+    ("component id", ids),
+    ("component label", labels),
+    ("path prefix", paths),
+    ("issue option", issue_options),
+    ("commit scope", commit_scopes),
+    ("release tag prefix", release_prefixes),
+    ("CI key", ci_keys),
+):
+    for duplicate in sorted(duplicates(values)):
+        errors.append(f"duplicate {field}: {duplicate}")
```
ikunkun-sys, line 228: [P2] Keep routing identifiers disjoint from reserved namespaces

The new duplicate checks cover collisions within selected component fields, but not across primary or reserved names. I could add `anolisa` to another component's `aliases`, add `docs` to a component's `commit_scope_aliases`, or use the fallback Issue option `other` as a component option while synchronizing the forms; the validator exited 0 in all three cases. Those tokens then have two meanings, so the metadata still cannot resolve them unambiguously.

.github/scripts/validate-repository-metadata.py
```diff
@@ -0,0 +1,451 @@
...
+    expected_ci_keys = set(ci_keys)
+    for contract, workflow_keys in ci_contracts.items():
+        for key in sorted(expected_ci_keys - workflow_keys):
+            errors.append(f"ci.yaml: missing {contract} for component CI key {key!r}")
+        for key in sorted(workflow_keys - expected_ci_keys):
+            errors.append(f"ci.yaml: unknown {contract} {key!r}")
+
+    ci_path_routes = re.findall(
+        r'if echo "\$CHANGED" \| grep -qE? "([^"]+)"; then\s+([A-Z0-9_]+)=true',
+        ci_job_text,
```
ikunkun-sys, line 361: [P2] Do not count comments as live CI contracts

These expressions scan raw workflow text and therefore accept comments as active consumers and routes. Replacing the real ANOLISA job condition with `if: false` while leaving the old `needs.detect-changes.outputs.anolisa` expression in a YAML comment still passed validation. Removing the real path-detection block and leaving a commented `if echo ...; then ANOLISA=true` line also passed. The metadata check can therefore stay green while ANOLISA changes no longer schedule their tests.

ikunkun-sys (changes requested):
Re-reviewed `eabf2547`. The three previously reported cases are addressed and the expanded 16-test regression suite passes, but two boundary gaps remain; targeted mutations for both still exit 0. Please address the inline findings before merging.

Non-blocking: the PR description still reports 8 governance tests, while the current suite contains 16.

.github/scripts/validate-repository-metadata.py
```diff
@@ -0,0 +1,476 @@
...
+    )
+    for component in components:
+        ci_key = component.get("ci_key")
+        if not ci_key:
+            continue
+        route_variables: set[str] = set()
+        for prefix in component.get("path_prefixes", []):
+            matching_variables = {
+                variable for pattern, variable in ci_path_routes if re.search(pattern, prefix)
```
ikunkun-sys: [P2] Reject stale path alternatives in CI routes

This loop validates each declared metadata prefix against `ci_path_routes`, but never verifies the reverse direction. I removed `src/anvil/` from `.github/components.json` and CODEOWNERS while keeping `^src/(anvil|blaze)/` in `ci.yaml`; the validator still exited 0, so live CI behavior can drift from the source of truth. Please reject route alternatives or prefixes that are not declared by the routed component (or derive the routes from metadata), and add a regression test for removing a legacy path.

ikunkun-sys (changes requested):
Re-reviewed `10b18e69`. The two previously reported gaps are addressed and the 21-test regression suite passes, but one reverse-consistency gap remains. A targeted mutation shows that a stale CI path alternative can survive removal from both component metadata and CODEOWNERS while validation still exits 0. Please address the inline finding before merging.

.github/scripts/validate-repository-metadata.py
kongche-jbw: Fixed in b343a1ce. Component IDs and aliases now share one uniqueness namespace, component commit scopes are checked against the reserved non-component scopes, and component Issue options are checked against fallback options. Regression tests cover all three collision cases.

.github/scripts/validate-repository-metadata.py
kongche-jbw: Fixed in b343a1ce. CI consumers and path routes are now recognized only from active, line-anchored workflow statements, so commented copies cannot satisfy the contract. Regression tests replace each live statement with a comment and verify validation fails.

.github/scripts/validate-repository-metadata.py
kongche-jbw: Fixed in b343a1ce. Active Anvil compatibility was removed from components.json, CODEOWNERS, CI, Qoder routing, and the onboarding example; Blaze is now the only live name and path. CI route validation now accepts only exact declared path patterns in both directions, with a regression test that rejects an undeclared legacy alternative.

ikunkun-sys (approved):
lgtm

PR #2066, opened 2026-07-31
feat(ci): dispatch issue triage

kongche-jbw:
## Why

Issue reporters often cannot identify the correct component. Repository-side
heuristics therefore assign the wrong owners, while using assignees as a
notification list obscures who is actually implementing the work.

## What Changed

- replace automatic `issues.opened` classification with a parameterized
  `workflow_dispatch` contract for the dedicated Codex triager
- validate component, confidence, decision id, and live Issue state before writes
- preserve every conflicting manual component label and keep low-confidence or
  dry-run decisions non-mutating
- apply the component label, owner notification, and public summary in one
  GitHub Actions run so comments remain authored by `github-actions[bot]`
- read reporter assignment exceptions from the externally managed
  `ISSUE_TRIAGE_POLICY` GitHub Secret
- default to no automatic assignment when the policy is absent
- trust workflow state markers only from first-line `github-actions[bot]`
  records and HTML-escape Codex-authored public text
- require DingTalk HTTP success and `errcode == 0` before recording delivery
- render a fixed DingTalk Issue link and separately escape the untrusted title
- track DingTalk delivery independently so failed notifications can be retried
- make claim and release marker-first operations that reconcile after reruns
- reject fresh claims carrying any non-ready `status:*` label
- restore `status:accepted` on release only when no other workflow status remains
- generate release availability text from the status that will be preserved
- reject unknown automation modes without mutating Issues
- run regression tests with the Node 20 production runtime on every relevant PR
- stop rewriting Issue titles and make Issue Form component fields optional hints
- retain community `/claim` and `/unclaim` commands
- document the server-to-Action trust boundary and shadow-mode rollback

## Related Issue

no-issue: community triage workflow approved as part of the repository workflow
cleanup

## User / Agent Impact

Most new Issues remain unassigned while component owners are notified. Any
operational assignment exception can change without a repository commit. Human
labels always take precedence over a conflicting Codex decision.

## Risk and Compatibility

- [ ] Public CLI/API/config changed
- [x] Repository automation changes labels, comments, notifications, and narrowly
  scoped assignments
- [ ] Privileged product runtime behavior changed

This PR depends on the `.github/components.json` contract from #2063 and must
not merge first. The external dispatcher remains trusted to invoke this workflow
with `ref=main`; technical enforcement of that ref is outside this PR.

## Validation

- Node 20 syntax checks for the triage and claim modules
- triage and claim state-machine tests — 32 passed
- DingTalk rendering and API response tests — 4 passed
- parsed all changed workflows as YAML
- `bash scripts/docs-lint.sh`
- `git diff --check 6642f7d8..HEAD`
- scanned the rewritten commit range for embedded assignment policy values
- live dry-run and end-to-end dispatch remain gated on #2063 merging

## Rollout and Rollback

Merge #2063 first, then rebase and merge this PR. Keep
`COMMUNITY_AUTOMATION_MODE=shadow` during the first server dry runs. Pausing
the server timer stops new dispatches; setting the repository variable to
`shadow` makes any accepted dispatch non-mutating. Unknown mode values fail
closed without changing Issues.


.github/scripts/issue-triage.js
```diff
@@ -0,0 +1,375 @@
...
+  const comments = await github.paginate(github.rest.issues.listComments, {
+    owner,
+    repo,
+    issue_number: decision.issueNumber,
+    per_page: 100,
+  });
+  const triageComment = comments.find(
+    (comment) => hasTrustedFirstLineMarker(comment, marker)
+  );
+  if (triageComment) {
```
ikunkun-sys, line 290: [P1] Keep notification retries behind the apply gate

This existing-marker branch runs before the `applyRequested && repositoryMode === 'apply'` and confidence gate. After an earlier notification failure, rerunning the same decision in `shadow` mode or with `apply=false` still sets `notification_required=true`; the workflow then sends DingTalk and appends the notification marker. That makes the documented rollback/dry-run mode externally and repository-mutating.

.github/scripts/issue-triage.js
```diff
@@ -0,0 +1,375 @@
...
+    throw new Error('ISSUE_TRIAGE_MIN_CONFIDENCE must be between 0 and 1');
+  }
+
+  const response = await github.rest.issues.get({
+    owner,
+    repo,
+    issue_number: decision.issueNumber,
+  });
+  const issue = response.data;
+  if (issue.pull_request) throw new Error('issue_number refers to a pull request');
```
ikunkun-sys, line 232: [P1] Reject closed issues before any write

The workflow fetches the live Issue but only rejects pull requests. If the Issue is closed between dispatcher discovery and execution, this path still proceeds to label, assign, comment, and notify it, contrary to the stated live-state validation contract.

.github/scripts/issue-claim.js
```diff
@@ -0,0 +1,303 @@
...
+      body,
+    });
+  }
+
+  const { data: issue } = await github.rest.issues.get({
+    owner,
+    repo,
+    issue_number: issueNumber,
+  });
+  const labels = labelNames(issue);
```
ikunkun-sys, line 131: [P1] Prevent claims on closed issues

GitHub permits comments on closed, unlocked Issues. Because the fetched Issue state is not checked, a closed Issue that retains a claimable label can receive an authoritative claim marker, lose its ready label, gain `status:in-progress`, and add an assignee.

.github/scripts/issue-triage.js
```diff
@@ -0,0 +1,375 @@
...
+  let assigned = [];
+  if (shouldAutoAssign(policy, issue.user?.login || '') && owners.length) {
+    try {
+      await github.rest.issues.addAssignees({
+        owner,
+        repo,
+        issue_number: decision.issueNumber,
+        assignees: owners,
+      });
+      assigned = owners;
```
ikunkun-sys: [P2] Report the actual assignee state

GitHub can silently ignore users who cannot be assigned while returning success, but this records every requested owner as assigned without inspecting the response. Conversely, the existing-marker retry path leaves `assigned` empty even after a successful first assignment, so the public/DingTalk output can be wrong in either direction. See: https://docs.github.com/en/enterprise-cloud@latest/rest/issues/assignees?apiVersion=2022-11-28#add-assignees-to-an-issue

ikunkun-sys (changes requested):
Found two blocking correctness issues and one assignment-reporting issue. Pending notification retries bypass the rollback/dry-run gate, and neither state machine rejects closed issues. Assignment output can also diverge from GitHub's actual assignee state. The existing CI and unit tests pass, but these state transitions are not covered.

.github/scripts/issue-triage.js
kongche-jbw: Fixed in d891ea02. The apply and confidence gates now run before idempotent marker lookup, so apply=false, shadow mode, and low-confidence reruns keep notification_required=false. Added Node 20 regression coverage for both dry-run modes.

.github/scripts/issue-triage.js
kongche-jbw: Fixed in d891ea02. Triage now rejects issue.state != open immediately after the live fetch, before labels, assignment, comments, or notification. Added a closed-Issue no-mutation regression test.

.github/scripts/issue-claim.js
kongche-jbw: Fixed in 8af9a151. Claim handling now returns for closed Issues before reading claim markers or mutating comments, labels, and assignees. Added coverage for both claim and unclaim commands on closed Issues.

.github/scripts/issue-triage.js
kongche-jbw: Fixed in d891ea02. Assignment output now intersects configured owners with the live or returned GitHub assignee list. Initial runs use the addAssignees response, while notification retries use the freshly fetched Issue state. Added partial-acceptance and retry regression tests.

ikunkun-sys (approved):
Re-reviewed commit `8a4651b6`. All four prior findings are addressed: notification retries now remain behind the apply/confidence gate, both state machines reject closed Issues before mutation, and assignment reporting uses the live/API-returned assignee state. I re-ran 36 Node state-machine tests, 4 Python tests, JavaScript syntax checks, YAML parsing, docs lint, and `git diff --check`; all passed, and the latest GitHub CI runs are green. No additional blocking findings. Approval covers this diff; retain the documented merge-order dependency on #2063.

PR #2200, opened 2026-08-04
fix(cosh-ng): support ID_LIKE fallback

### LINKED ISSUE #2111: [TestAgent][cosh-ng] bug: Distro::detect 只读 /etc/os-release,无 ID_LIKE fallback
### Component

cosh-ng

### Bug Description

`Distro::detect` 只读 `/etc/os-release` 的 `ID` 字段,不读 `ID_LIKE`(衍生发行版的父系声明,如 `ID=rocky` + `ID_LIKE=rhel fedora`),也不 fallback 到 `/usr/lib/os-release`。当 `/etc/os-release` 不存在(被删/容器最小化)或 `ID` 是未识别的衍生发行版时,直接返回 `Unknown`,导致 `pkg_manager()` 返回 `Unknown`,所有 `cosh pkg` 操作 fail。这违反 distro 抽象层应在常见衍生发行版上正确识别契约。

### Steps to Reproduce

```shell
ssh root@<ECS-IP>   # 授权测试机,读者自行替换
# 场景 1:删除 /etc/os-release(模拟容器最小化或误删)
mv /etc/os-release /tmp/os-release.bak
cosh pkg search bash    # 报 UnsupportedDistro,Distro::detect 返回 Unknown
mv /tmp/os-release.bak /etc/os-release

# 场景 2:衍生发行版(如 Rocky/RHEL),ID=rocky 不在白名单
# /etc/os-release 内容:ID=rocky\nID_LIKE="rhel fedora"\n
# cosh pkg search bash → detect 返回 Unknown(rocky 不识别),ID_LIKE=fedora 被忽略
```

### Actual Behavior

- 删除 `/etc/os-release`:`Distro::detect_from_path("/etc/os-release")` 返回 `Err` → 返回 `Distro::Unknown("unknown")` → `pkg_manager()` 返回 `PkgManager::Unknown` → `cosh pkg` 报 `UnsupportedDistro`
- 衍生发行版(如 `ID=rocky` + `ID_LIKE="rhel fedora"`):detect 只看 `ID=rocky`,不在白名单 → 返回 `Unknown`,忽略 `ID_LIKE=fedora` 本可 fallback 到 dnf 后端
- 连续 3 次稳定复现

### Expected Behavior

- detect 应在 `/etc/os-release` 不存在时 fallback 到 `/usr/lib/os-release`(systemd 标准 fallback 路径)
- detect 应解析 `ID_LIKE` 字段,当 `ID` 未识别时按 `ID_LIKE` 列表逐项匹配(如 `ID=rocky` + `ID_LIKE=rhel fedora` → 识别为类 fedora → 用 dnf 后端)
- 不应在常见衍生发行版(Rocky/Alma/Oracle/RHEL/centos-stream)上返回 Unknown

### Root Cause

`src/cosh-ng/crates/cosh-platform/src/detect.rs:17-50`:

```rust
pub fn detect() -> Self {
    if cfg!(target_os = "macos") {
        return Self::detect_macos();
    }
    Self::detect_from_path("/etc/os-release")   // 无 /usr/lib/os-release fallback
}

fn detect_from_path(path: &str) -> Self {
    match fs::read_to_string(path) {
        Ok(content) => Self::detect_from_content(&content),
        Err(_) => Self::Unknown("unknown".into()),   // /etc/os-release 不存在直接 Unknown
    }
}

pub(crate) fn detect_from_content(content: &str) -> Self {
    let mut id = None;
    let mut version_id = None;
    // 只解析 ID 和 VERSION_ID,不解析 ID_LIKE
    for line in content.lines() {
        ...
        if key.eq_ignore_ascii_case("ID") { id = ...; }
        else if key.eq_ignore_ascii_case("VERSION_ID") { version_id = ...; }
    }
    match id.as_deref() {
        Some("alinux") => ...,
        Some("ubuntu") => ...,
        ...
        Some(other) => Distro::Unknown(other.into()),   // 衍生发行版直接 Unknown,不 fallback ID_LIKE
        None => Distro::Unknown("unknown".into()),
    }
}
```

触发条件:`/etc/os-release` 不存在,或 `ID` 是衍生发行版(rocky/alma/oracle/centos-stream 等)且 `ID_LIKE` 指向已识别的父系。

设计取舍:detect 只看 `ID` 是早期简化,把识别发行版等同于匹配 ID 白名单。但 systemd 规范允许 `ID_LIKE` 声明父系(rocky → rhel fedora),衍生发行版用 `ID_LIKE` 表达我行为像 X。修复需要解析 `ID_LIKE`,按空格拆分,逐项匹配白名单。同时 `/usr/lib/os-release` 是 systemd 标准的 fallback 路径(基础发行版元数据),`/etc/os-release` 是 override,两者应合并读取。

### Suggested Fix

- `detect()` 改为先读 `/etc/os-release`,失败则 fallback 到 `/usr/lib/os-release`
- `detect_from_content` 解析 `ID_LIKE` 字段,当 `ID` 不在白名单时按 `ID_LIKE` 列表逐项匹配(如 `ID=rocky` + `ID_LIKE="rhel fedora"` → 匹配 `fedora` → 用 dnf 后端)
- 把 Rocky/Alma/Oracle/RHEL/centos-stream 等常见衍生发行版 ID 显式加入白名单(直接映射到对应 Distro variant)
- 同步加 unit test `detect_uses_id_like_fallback` 和 `detect_falls_back_to_usr_lib_os_release`

### Environment

- OS: Alibaba Alinux 4, kernel 6.6.102
- Component: cosh-ng
- Package/RPM: cosh-ng-0.13.0-2.alnx4

### Additional Context

- 稳定性: 连续 3 次稳定复现
- 被测版本 main HEAD: a96622bd

Zhilinlinlin:
## Why

Derived Linux distributions can expose an unrecognised ID while declaring a supported package family through ID_LIKE. This currently makes all pkg commands fail as unsupported.

## What changed

Read /usr/lib/os-release when /etc/os-release is unavailable. Preserve an unlisted distribution ID while routing its package commands through the first supported ID_LIKE family.

## Related issue

Closes #2111

## User / Agent impact

cosh-cli pkg commands now work on compatible distributions such as Rocky Linux and report their actual distro ID.

## Risk and compatibility

Low risk: compatible distributions now select their declared package-manager family instead of failing as unsupported.

## Validation

- cargo test -p cosh-platform --lib: 285 passed
- clippy-driver check for cosh-platform --all-targets with warnings denied
- cargo doc -p cosh-platform --no-deps
- Rocky 9 container: cosh-cli pkg list --installed returned 146 packages with meta.distro=rocky

## Documentation and rollback

No user documentation change. Revert this commit to restore the previous detection behavior.

### DIFF
--- a/src/cosh-ng/crates/cosh-platform/src/detect.rs
+++ b/src/cosh-ng/crates/cosh-platform/src/detect.rs
@@ -1,28 +1,52 @@
 use std::fs;
+use std::path::Path;
 use std::process::Command;
 
 /// Operating system / distribution variants supported by cosh.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Distro {
-    Alinux { version: String },
-    Ubuntu { version: String },
-    Debian { version: String },
-    CentOS { version: String },
-    Fedora { version: String },
-    OpenSUSE { version: String },
-    MacOS { version: String },
+    Alinux {
+        version: String,
+    },
+    Ubuntu {
+        version: String,
+    },
+    Debian {
+        version: String,
+    },
+    CentOS {
+        version: String,
+    },
+    Fedora {
+        version: String,
+    },
+    OpenSUSE {
+        version: String,
+    },
+    MacOS {
+        version: String,
+    },
+    /// An unlisted distribution with a supported package-manager family.
+    Compatible {
+        id: String,
+        version: String,
+        pkg_manager: PkgManager,
+    },
     Unknown(String),
 }
 
 impl Distro {
     /// Detect the current OS. On macOS uses `sw_vers`; on Linux reads
-    /// /etc/os-release.
+    /// /etc/os-release, falling back to /usr/lib/os-release.
     pub fn detect() -> Self {
         // Check macOS first (compile-time or runtime)
         if cfg!(target_os = "macos") {
             return Self::detect_macos();
         }
-        Self::detect_from_path("/etc/os-release")
+        Self::detect_from_paths(
+            Path::new("/etc/os-release"),
+            Path::new("/usr/lib/os-release"),
+        )
     }
 
     /// Detect macOS version via `sw_vers -productVersion`.
@@ -42,15 +66,23 @@ impl Distro {
         Distro::MacOS { version }
     }
 
-    fn detect_from_path(path: &str) -> Self {
-        match fs::read_to_string(path) {
+    fn detect_from_paths(primary: &Path, fallback: &Path) -> Self {
+        match fs::read_to_string(primary) {
             Ok(content) => Self::detect_from_content(&content),
+            Err(error) if error.kind() == std::io::ErrorKind::NotFound => {
+                // os-release gives /etc exclusive precedence whenever it exists.
+                match fs::read_to_string(fallback) {
+                    Ok(content) => Self::detect_from_content(&content),
+                    Err(_) => Self::Unknown("unknown".into()),
+                }
+            }
             Err(_) => Self::Unknown("unknown".into()),
         }
     }
 
     pub(crate) fn detect_from_content(content: &str) -> Self {
         let mut id = None;
+        let mut id_like = None;
         let mut version_id = None;
 
         for line in content.lines() {
@@ -61,10 +93,12 @@ impl Distro {
 
             if let Some((key, value)) = line.split_once('=') {
                 let key = key.trim();
-                let value = value.trim().trim_matches('"');
+                let value = strip_matching_outer_quotes(value.trim());
 
                 if key.eq_ignore_ascii_case("ID") {
                     id = Some(value.to_lowercase());
+                } else if key.eq_ignore_ascii_case("ID_LIKE") {
+                    id_like = Some(value.to_lowercase());
                 } else if key.eq_ignore_ascii_case("VERSION_ID") {
                     version_id = Some(value.to_string());
                 }
@@ -73,18 +107,49 @@ impl Distro {
 
         let version = version_id.unwrap_or_else(|| "unknown".into());
 
-        match id.as_deref() {
-            Some("alinux") => Distro::Alinux { version },
-            Some("ubuntu") => Distro::Ubuntu { version },
-            Some("debian") => Distro::Debian { version },
-            Some("centos") => Distro::CentOS { version },
-            Some("fedora") => Distro::Fedora { version },
-            Some("opensuse-leap") | Some("opensuse-tumbleweed") | Some("sles") => {
-                Distro::OpenSUSE { version }
-            }
-            Some(other) => Distro::Unknown(other.into()),
-            None => Distro::Unknown("unknown".into()),
+        let id = id.unwrap_or_else(|| "linux".into());
+
+        if let Some(distro) = Self::from_known_id(&id, &version) {
+            return distro;
         }
+
+        if let Some(pkg_manager) = id_like
+            .as_deref()
+            .and_then(|ids| ids.split_whitespace().find_map(pkg_manager_from_id_like))
+        {
+            return Distro::Compatible {
+                id,
+                version,
+                pkg_manager,
+            };
+        }
+
+        Distro::Unknown(id)
+    }
+
+    fn from_known_id(id: &str, version: &str) -> Option<Self> {
+        let distro = match id {
+            "alinux" => Distro::Alinux {
+                version: version.into(),
+            },
+            "ubuntu" => Distro::Ubuntu {
+                version: version.into(),
+            },
+            "debian" => Distro::Debian {
+                version: version.into(),
+            },
+            "centos" => Distro::CentOS {
+                version: version.into(),
+            },
+            "fedora" => Distro::Fedora {
+                version: version.into(),
+            },
+            "opensuse-leap" | "opensuse-tumbleweed" | "sles" => Distro::OpenSUSE {
+                version: version.into(),
+            },
+            _ => return None,
+        };
+        Some(distro)
     }
 
     /// Returns the distro identifier string for JSON output.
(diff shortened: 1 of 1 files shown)


src/cosh-ng/crates/cosh-platform/src/detect.rs
```diff
@@ -73,18 +100,51 @@ impl Distro {
...
+            .and_then(|ids| ids.split_whitespace().find_map(pkg_manager_from_id_like))
+        {
+            return Distro::Compatible {
+                id,
+                version,
+                pkg_manager,
+            };
         }
+
+        Distro::Unknown(id)
```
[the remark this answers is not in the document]
Zhilinlinlin: Kept as `Unknown`: `Compatible` already identifies distributions with a supported `ID_LIKE` family, while the remaining case follows the existing unsupported-distro path. Adding another public error variant is outside this issue.

src/cosh-ng/crates/cosh-platform/src/detect.rs
```diff
@@ -124,11 +186,21 @@ impl Distro {
             Distro::Ubuntu { .. } | Distro::Debian { .. } => PkgManager::Apt,
             Distro::OpenSUSE { .. } => PkgManager::Zypper,
             Distro::MacOS { .. } => PkgManager::Brew,
+            Distro::Compatible { pkg_manager, .. } => *pkg_manager,
             Distro::Unknown(_) => PkgManager::Unknown,
         }
     }
 }
 
+fn pkg_manager_from_id_like(id: &str) -> Option<PkgManager> {
```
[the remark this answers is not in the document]
Zhilinlinlin: `ID_LIKE` is used only for Linux `os-release` detection. macOS is detected earlier via `sw_vers`, so a `darwin` to Brew mapping here would be misleading.

src/cosh-ng/crates/cosh-platform/src/detect.rs
```diff
@@ -65,6 +90,8 @@ impl Distro {
 
                 if key.eq_ignore_ascii_case("ID") {
                     id = Some(value.to_lowercase());
+                } else if key.eq_ignore_ascii_case("ID_LIKE") {
```
kongche-jbw, line 100: [P1] Valid single-quoted `ID_LIKE` values are not recognized. The shared parser strips only `"`, so `ID_LIKE='rhel fedora'` is split into `"'rhel"` and `"fedora'"`; neither token matches and the distro remains `Unknown`. The systemd `os-release` format permits either single or double quotes, so this leaves a valid form of the newly supported field broken.

src/cosh-ng/crates/cosh-platform/src/detect.rs
```diff
@@ -42,15 +66,16 @@ impl Distro {
         Distro::MacOS { version }
     }
 
-    fn detect_from_path(path: &str) -> Self {
-        match fs::read_to_string(path) {
+    fn detect_from_paths(primary: &Path, fallback: &Path) -> Self {
+        match fs::read_to_string(primary).or_else(|_| fs::read_to_string(fallback)) {
```
kongche-jbw: [P1] This falls back on every `read_to_string` error, not only when the primary file is missing. If `/etc/os-release` exists but is unreadable, is a directory, or contains invalid UTF-8, the code silently uses `/usr/lib/os-release` and may route package operations using different vendor data. The `os-release` contract gives `/etc/os-release` exclusive precedence whenever it exists and permits fallback only when it is missing.

src/cosh-ng/crates/cosh-platform/src/detect.rs
```diff
@@ -73,18 +100,51 @@ impl Distro {
...
-            Some("centos") => Distro::CentOS { version },
-            Some("fedora") => Distro::Fedora { version },
-            Some("opensuse-leap") | Some("opensuse-tumbleweed") | Some("sles") => {
-                Distro::OpenSUSE { version }
-            }
-            Some(other) => Distro::Unknown(other.into()),
-            None => Distro::Unknown("unknown".into()),
+        let Some(id) = id else {
+            return Distro::Unknown("unknown".into());
+        };
```
kongche-jbw: [P2] A missing `ID` returns before consulting `ID_LIKE`. `ID` is optional in the `os-release` format (readers may default it to `linux`), so a valid file such as `ID_LIKE=debian` still produces `Unknown` and all package commands fail even though a supported family is declared.

kongche-jbw (changes requested):
Found three `os-release` contract gaps that can still leave valid compatible distributions unroutable or select the fallback data incorrectly. The targeted platform tests, clippy, rustfmt check, cargo doc, and GitHub CI all pass, but these boundary cases are not covered.

src/cosh-ng/crates/cosh-platform/src/detect.rs
Zhilinlinlin: Fixed in 8329f09: fallback now occurs only for NotFound; every other primary read error returns Unknown. A regression test uses an unreadable directory as the primary path.

src/cosh-ng/crates/cosh-platform/src/detect.rs
Zhilinlinlin: Fixed in 8329f09: a missing ID now defaults to linux before ID_LIKE evaluation, with regression coverage for ID_LIKE=debian.

src/cosh-ng/crates/cosh-platform/src/detect.rs
Zhilinlinlin: Fixed in 8329f09: matching single or double quotes are stripped before ID_LIKE tokenization, with a regression test for a single-quoted RHEL family value.

kongche-jbw (approved):

PR #2209, opened 2026-08-04
fix(cosh-ng): [core] decode sse events per spec and fail loud

### LINKED ISSUE #2182: [TestAgent][cosh-ng] bug: shell tool 转发 command 到 PTY 时字符串被截断插入换行,bash 续行卡死 LLM 无法重新触发
### Component

cosh-ng

### Bug Description

cosh-shell 在把 LLM 生成的 tool_use `command` 字符串转发到 bash PTY 时,**字符串被截断并在截断点插入了换行符**,导致 bash 收到不完整的命令、单引号未闭合、进入续行 prompt(`> `)等 stdin 输入。bash 永远等不到补全输入,cosh-shell busy-loop spin(`read EAGAIN + wait4 WNOHANG=0` 反复),cosh-core idle 等 tool_result,LLM 无法被重新触发,Agent turn 永久卡死。

LLM 实际生成的 command 是**完整合法**的(现场 agentsight `genai_events.db` 的 `event_json` 反序列化可证),但 bash 收到的是截断版本 —— 截断发生在 command 字符串中间的某个字符之后,且 cosh-shell 在截断点插入了 `\n`。

### Steps to Reproduce

**前置**:已装 cosh-ng-0.14.0-2,Provider 配置为 qwen3.8-max / qwen3.7-max 等(openai 兼容接口),approval mode = auto / trust 均可复现。

**Step 1**:在授权测试机上(`ssh root@<ECS-IP>`),让 Agent 跑一个调 hermes config + 验证 DashScope 端点的多步任务。Agent 生成形如下面的 shell 命令(完整、合法、单引号闭合):

```shell
~/.local/bin/hermes config set model.base_url https://dashscope.aliyuncs.com/compatible-mode/v1 2>&1 | tail -2; echo "---"; rtk grep -n -A3 '^model:' ~/.hermes/config.yaml
```

```shell
KEY=$(grep '^DASHSCOPE_API_KEY=' ~/.hermes/.env | tail -1 | cut -d= -f2); curl -s -m 20 -X POST https://dashscope.aliyuncs.com/compatible-mode/v1/chat/completions -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' -d '{"model":"qwen3.8-max","messages":[{"role":"user","content":"hi"}],"max_tokens":5}' | head -c 500
```

**Step 2**:观察 Agent 行为:
- 第一条命令(req-20)正常执行,输出 `✓ Set model.base_url = ...` + `model:` 配置回显
- **第二条命令(req-21)在 bash 终端显示为截断版本**:`KEY=$(grep '^DASHSCOPE_API_KEY=`,后面是 `\n` + bash 续行 prompt `> `
- bash 进入续行模式,永远等 stdin 输入补全单引号 + `)` + 命令体
- Agent 显示"Approved for this turn req-21"后,UI 卡死,LLM 不会被重新触发

### Actual Behavior

现场(2026-08-04)抓取的进程状态和 strace 证据:

```
进程树:
cosh-shell raw <pid-1> ── bash <pid-2> --rcfile /tmp/cosh-shell-raw-session-<pid-1>/cosh-marker.bash -i
                       └─ 状态 S(sleeping),深睡在 read(0) 等 stdin 补全单引号
cosh-core <pid-3> (headless, --approval-mode auto)
                └─ idle futex_wait + eventfd write 循环(tokio runtime 空转)

strace cosh-shell <pid-1>:
  read(6, ..., 8192) = -1 EAGAIN            ← PTY master 读不到数据(bash 没输出)
  wait4(<pid-2>, ..., WNOHANG, NULL) = 0    ← bash 没退出
  read(6, ..., 8192) = -1 EAGAIN            ← 反复(busy-loop spin)
  wait4(<pid-2>, ..., WNOHANG, NULL) = 0
  ... (5s 内多次反复)

strace cosh-core <pid-3>:
  futex(FUTEX_WAIT_PRIVATE) = 0             ← tokio idle 循环
  write(4, "\1\0\0...", 8) = 8              ← eventfd 唤醒(没真活)
  ... (5s 内反复 idle,LLM 不会被重新触发)
```

agentsight `genai_events.db` 最近 LLM call 状态(确认 LLM 收到完整 tool_result 之前不会发起新 call):

```
#101 [10:30:53] status=200 finish_reason=tool_calls  ← LLM 生成了完整合法的 KEY=$(grep ...) 命令
#102 [10:31:01] status=None event=llm_call            ← LLM 进入下一轮但等 tool_result
#103 [10:31:01] status=None event=llm_call            ← 同上
(之后无新 LLM call —— cosh-core 在 idle 等 shell tool_result)
```

### Expected Behavior

按 cosh-ng 的产品契约("Agent 调 shell 命令应能完成或退出,而不是永久卡死"),至少应满足其中一条:

1. **command 字符串完整转发**:cosh-shell 把 LLM 生成的完整 command 字符串原封不动写入 PTY master,不在中途截断、不插入换行符;
2. **bash 续行检测**:cosh-shell 检测到 bash 进入续行 prompt(`> ` 出现在 PTY 输出且无 newline 终结),主动 close PTY slave 写端让 bash 收到 EOF,或发 SIGTERM 给 bash;
3. **stdin idle 超时**:bash 子进程 stdin `read` block 超 N 秒(如 30s)无数据,cosh-shell 主动 close PTY slave 写端,bash 收到 EOF 退出续行模式,tool_result 回传给 cosh-core,LLM 重新触发。

### Root Cause

**LLM 生成的完整 command**(agentsight `event_json` 反序列化):

```
KEY=$(grep '^DASHSCOPE_API_KEY=' ~/.hermes/.env | tail -1 | cut -d= -f2); curl -s -m 20 -X POST https://dashscope.aliyuncs.com/compatible-mode/v1/chat/completions -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' -d '{"model":"qwen3.8-max","messages":[{"role":"user","content":"hi"}],"max_tokens":5}' | head -c 500
```

单引号 `'^DASHSCOPE_API_KEY='` 是闭合的,`$(...)` 也是闭合的,命令整体合法。

**bash 收到的截断版本**(用户终端实际显示):

```
[root@<INTERNAL-HOST> ~]# KEY=$(grep '^DASHSCOPE_API_KEY=
> 
> 
```

截断点在第二个 `'`(闭合单引号)之后的 `=` 字符。**cosh-shell 在转发 command 到 PTY 时,在 `=` 之后插入了 `\n` 换行符,丢弃了 `~/.hermes/.env | tail -1 | cut -d= -f2); curl ...` 等后续内容**。bash 检测到单引号没闭合(实际是闭合的,但 bash 只看到 `'^DASHSCOPE_API_KEY=` 部分),进入续行 prompt `> ` 等 stdin 补全。

**触发条件**:

| 维度 | 条件 |
|---|---|
| Command 长度 | 较长的复合命令(本例 ~280 字符) |
| Command 内容 | 含单引号 + `$(...)` + `;` + `|` + JSON 内嵌引号 |
| 截断位置 | 不固定,本例在 `=` 字符之后(单引号闭合之后) |

**file:line 疑似位置**(待研发确认):

- `src/cosh-ng/crates/cosh-shell/src/raw_input/spawn.rs:71` `spawn_raw_input_relay` —— input relay 启动入口
- `src/cosh-ng/crates/cosh-shell/src/raw_input/spawn.rs:273-292` `relay_input_bytes` —— 把字节写入 PTY master 的核心函数,可能存在 buffer split / 换行插入逻辑

**设计取舍**:

cosh-shell 用 PTY 跑 bash 是为了支持 TUI / 颜色 / 行编辑(readline)。但 readline 在 canonical 模式下按行 buffer,如果 cosh-shell 写入 PTY master 时分多次 write 或在某个字符处插入 `\n`,bash readline 就会把 command 当成多行处理,触发续行 prompt。

当前实现没有:
- 把 command 字符串作为整体原子写入 PTY master(避免 split)
- 检测 bash 续行 prompt(`> `)主动 close stdin
- stdin idle 超时机制兜底

### Suggested Fix

按短期缓解 → 长期修复的优先级:

**方向 A(短期,推荐先做)**:cosh-shell 加 stdin idle 超时机制。bash 子进程 stdin `read` block 超 N 秒(建议默认 30s,可配置)无数据,cosh-shell 主动 close PTY slave 写端,bash 收到 EOF 退出续行模式,tool_result(含续行错误)回传给 cosh-core,LLM 重新触发。

**方向 B(中期,根因修复)**:审查 `relay_input_bytes`(spawn.rs:273-292)的 PTY master 写入逻辑,确保:
- command 字符串作为整体原子 write(不分多次)
- 不在 command 字符串中间插入 `\n`(除非 LLM 生成的 command 本身含 `\n`)
- 如果必须分多次 write(如 buffer 大小限制),用 `writev` 或保证顺序 + 不插入额外字节

**方向 C(辅助)**:cosh-shell 检测 bash 续行 prompt(`> ` 在 PTY 输出且无 newline 终结),主动 close PTY slave 或发 SIGINT 给 bash,让续行命令被中断,tool_result 回传错误给 LLM。

**方向 D(长期,防御)**:cosh-core 在转发 tool_use command 给 bash 前,做 `bash -n` 语法检查 + 完整性校验,确保 command 字符串到 bash 时仍合法。同时 cosh-shell 在写入 PTY 后,读 PTY 输出对比 echo 回显,确认 bash 收到的 command 与写入一致。

### Environment

- OS: linux x86_64
- Component: cosh-ng
- Package/RPM: cosh-ng-0.14.0-2.alnx4
- Provider: qwen3.8-max(openai 兼容)
- Approval mode: auto

### Additional Context

- 稳定性:基于现场 agentsight `genai_events.db` + strace + 进程树证据(非 flaky 复现),只要 LLM 生成包含单引号 + `$(...)` + 复合管道的较长命令就可能触发
- 相关 issue:
  - #2161 [CLOSED] shell tool 子进程发交互 prompt 时永久卡死 —— 同类问题(PTY stdin 等输入),但触发路径不同:#2161 是子进程自身发 prompt,本 issue 是 cosh-shell 转发 command 时截断导致 bash 续行
  - #2144 [CLOSED] foreground shell handoff 命令失败后 busy-loop spin —— 同样的 busy-loop 现象(`read EAGAIN + wait4 WNOHANG=0` 反复),但触发路径不同:#2144 是 handoff 失败后,本 issue 是 command 截断后续行
- 现场 LLM 完整 command 已在 agentsight `genai_events.db` 的 `event_json.LLMCall.response.messages[0].parts[*].tool_call.arguments.command` 字段中保留,可与 bash 实际收到的截断版本对比,定位 cosh-shell 转发逻辑的截断点

SunnyQjm:
## Summary

The OpenAI-compat SSE decoder silently dropped spec-compliant frames, and a
dropped mid-stream `tool_call` arguments delta can concatenate the remaining
head/tail deltas into syntactically valid JSON — a truncated shell command
that then passes strict argument parsing and executes. In the reported turn
the model produced a ~280-character compound command, but bash received only
the 31-character prefix `KEY=$(grep '^DASHSCOPE_API_KEY=` plus a newline:
unbalanced quotes put bash into PS2 continuation forever, and the whole turn
hung with no error anywhere.

Root cause (in `crates/cosh-core/src/provider/openai_compat.rs`):

1. The line handler only matched `"data: "` (colon **plus space**); the space
   is optional in the SSE wire format, so `data:{...}` frames were silently
   `continue`d away.
2. `if let Ok(chunk) = serde_json::from_str(data)` silently skipped any data
   line that failed to parse.
3. Bytes were appended via `String::from_utf8_lossy`, so a multi-byte
   character split across network chunks decayed into U+FFFD and corrupted
   that line's JSON (feeding failure mode 2).
4. End-of-stream without `[DONE]` was reported as a plain `MessageEnd`, so a
   connection cut mid-message looked like a clean completion.

## Changes

Single file: `crates/cosh-core/src/provider/openai_compat.rs`.

- Assemble events per the SSE wire format via a new `SseEventBuffer`: the
  space after `data:` is optional, one event may span multiple `data` lines
  (joined with newlines), blank lines are event boundaries, comment (`:`) and
  non-data field lines (`event:`/`id:`/`retry:`) are ignored, CR is stripped.
- Buffer raw bytes and decode per extracted line, so split multi-byte
  characters reassemble; a line with invalid UTF-8 fails loud.
- Surface malformed non-empty data payloads as `GenerateEvent::Error` instead
  of skipping them; a later `[DONE]` no longer masks the corruption.
- Treat EOF without `[DONE]` or a prior `finish_reason` as a stream error. A
  stream that ends after `finish_reason` but before the usage payload still
  completes (content is whole; a missing usage payload is not worth failing
  the turn).
- The decoding state (byte buffer, `SseEventBuffer`, dedup stream state,
  `exhausted` flag) is carried through the unfold tuple; pending events are
  always drained before the terminal event, and the byte stream is never
  polled again after a terminal event.

Existing semantics preserved: tool-call dedup (`argument_deltas_seen` /
`started_tool_calls` / `emitted_text`), deferred `MessageEnd` with usage,
final-message snapshot fallback, cancellation.

## Tests

10 new tests drive the real byte-stream decoding loop over a local TCP
socket (`terminal_events_from_sse*` helpers):

- `data_line_without_space_is_not_dropped` — the reported truncation shape:
  a no-space mid frame must be decoded, not dropped.
- `malformed_sse_data_reports_error`,
  `malformed_sse_data_before_done_reports_error` — fail loud on corrupted
  data; `[DONE]` cannot mask it.
- `truncated_stream_without_done_or_finish_reports_error` — EOF before
  `[DONE]`/`finish_reason` is an error, never a silent success.
- `eof_after_finish_reason_without_done_is_message_end` — completed content
  with a missing usage tail still succeeds.
- `data_line_with_two_spaces_parses_payload`,
  `multi_data_lines_join_before_parse`,
  `crlf_comments_and_field_lines_are_tolerated`,
  `multibyte_character_split_across_chunks_reassembles`,
  `invalid_utf8_line_reports_error`,
  `final_event_without_trailing_separator_is_flushed` — wire-format matrix.

## Verification

Verified (Linux container, alinux3 arm64, real dev binaries):

- `cargo test -p cosh-core --lib provider::openai_compat` — 46/46 green.
- `cargo test -p cosh-core` — lib 783/786; the 3 failures are pre-existing
  `session::store` chmod-000 permission simulations that a root container
  bypasses; the same test binary passes 3/3 as a non-root user. Unrelated to
  this diff.
- `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D
  warnings` — clean.
- `crates/cosh-shell/scripts/check-layout.sh` and
  `scripts/check-test-inventory.sh` — pass.
- End-to-end FAIL→PASS against a mock OpenAI-compat provider that replays
  the reported truncation (arguments split into head / no-space mid / tail
  frames), with the real `cosh-core` + `cosh-shell` + PTY + bash:
  - Before: 2 deltas survive, the 31-character truncated command reaches the
    approval card and bash, bash hangs on PS2 (probe input swallowed).
  - After: 3 deltas, the full command reaches the approval card and bash and
    executes; a corrupted-frame arm ends the turn with a visible
    `malformed SSE data` error and no `control_request` at all.

Not run: full cosh-shell test suite (diff does not touch cosh-shell), macOS
host build (pre-existing rustix `openat2` Linux-only limitation in
`workspace_fs.rs`).

## Evidence

Real-PTY runs (real `cosh-core` + `cosh-shell` + bash in an alinux3
container) against a mock provider replaying the reported truncation shape.

**Before** — the truncated 31-character command reaches bash; unbalanced
quotes leave bash on PS2 continuation and swallow the next input line:

before: truncated command, bash stuck on PS2

[before, full replay (GIF)](https://raw.githubusercontent.com/SunnyQjm/anolisa/f7a6a7c40cb040a22a25dbeb5862d1499e75f68c/before-truncated-ps2.gif)

**After** — the same no-space mid frame is decoded, the full ~280-character
command reaches the approval card and bash and executes normally:

after: full command decoded and executed

[after, full replay (GIF)](https://raw.githubusercontent.com/SunnyQjm/anolisa/f7a6a7c40cb040a22a25dbeb5862d1499e75f68c/after-full-command.gif)

Fixes #2182

### DIFF
--- a/src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
+++ b/src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
@@ -214,7 +214,8 @@ impl ContentGenerator for OpenAICompatProvider {
         // otherwise no usage chunk is coming and MessageEnd must not wait.
         let defer_message_end = config.include_usage && self.profile.supports_stream_usage();
         let byte_stream = response.bytes_stream();
-        let buffer = String::new();
+        let buffer = SseByteBuffer::default();
+        let sse_event = SseEventBuffer::default();
         let event_queue: Vec<GenerateEvent> = Vec::new();
         let stream_state = OpenAICompatStreamState::default();
 
@@ -222,33 +223,56 @@ impl ContentGenerator for OpenAICompatProvider {
             (
                 byte_stream,
                 buffer,
+                sse_event,
                 cancelled,
                 event_queue,
                 thinking_field,
                 stream_state,
+                false, // exhausted: terminal emitted or byte stream drained
             ),
             move |(
                 mut stream,
                 mut buf,
+                mut sse_event,
                 cancelled,
                 mut pending,
                 thinking_field,
                 mut stream_state,
+                exhausted,
             )| async move {
                 let tf = thinking_field.as_deref();
                 loop {
                     if let Some(event) = pending.pop() {
+                        // A terminal event popped from the queue seals the
+                        // stream too; otherwise a [DONE] still in flight
+                        // after e.g. a finish_reason-driven MessageEnd would
+                        // produce a second terminal event.
+                        let exhausted = exhausted
+                            || matches!(
+                                event,
+                                GenerateEvent::MessageEnd
+                                    | GenerateEvent::Error(_)
+                                    | GenerateEvent::Cancelled
+                            );
                         return Some((
                             event,
                             (
                                 stream,
                                 buf,
+                                sse_event,
                                 cancelled,
                                 pending,
                                 thinking_field,
                                 stream_state,
+                                exhausted,
                             ),
                         ));
+                    }
+
+                    // Never poll the byte stream again once a terminal event
+                    // went out or the stream already returned end-of-stream.
+                    if exhausted {
+                        return None;
                     }
 
                     if cancelled.load(Ordering::SeqCst) {
@@ -257,68 +281,153 @@ impl ContentGenerator for OpenAICompatProvider {
                             (
                                 stream,
                                 buf,
+                                sse_event,
                                 cancelled,
                                 pending,
                                 thinking_field,
                                 stream_state,
+                                true,
                             ),
                         ));
                     }
 
-                    if let Some(line_end) = buf.find('\n') {
-                        let line = buf[..line_end].to_string();
-                        buf = buf[line_end + 1..].to_string();
-
-                        let line = line.trim();
-                        if line.is_empty() || line.starts_with(':') {
-                            continue;
-                        }
-                        if let Some(data) = line.strip_prefix("data: ") {
-                            if data.trim() == "[DONE]" {
+                    // Extract one complete line (LF, CRLF, or bare CR). Bytes
+                    // are buffered raw and decoded per line so a multi-byte
+                    // character split across network chunks reassembles
+                    // instead of turning into replacement characters that
+                    // corrupt the payload.
+                    if let Some(line_bytes) = buf.take_line() {
+                        let line = match std::str::from_utf8(&line_bytes) {
+                            Ok(line) => line,
+                            Err(error) => {
+                                return Some((
+                                    GenerateEvent::Error(format!(
+                                        "SSE stream carried invalid UTF-8: {error}"
+                                    )),
+                                    (
+                                        stream,
+                                        buf,
+                                        sse_event,
+                                        cancelled,
+                                        pending,
+                                        thinking_field,
+                                        stream_state,
+                                        true,
+                                    ),
+                                ));
+                            }
+                        };
+                        let data = match sse_event.push_line(line) {
+                            Ok(Some(data)) => data,
+                            Ok(None) => continue,
+                            Err(message) => {
+                                return Some((
+                                    GenerateEvent::Error(message),
+                                    (
+                                        stream,
+                                        buf,
+                                        sse_event,
+                                        cancelled,
+                                        pending,
+                                        thinking_field,
+                                        stream_state,
+                                        true,
+                                    ),
+                                ));
+                            }
+                        };
+                        match dispatch_sse_data(&data, tf, defer_message_end, &mut stream_state) {
+                            SseDispatch::Done => {
                                 return Some((
                                     GenerateEvent::MessageEnd,
                                     (
                                         stream,
                                         buf,
+                                        sse_event,
                                         cancelled,
                                         pending,
                                         thinking_field,
                                         stream_state,
+                                        true,
                                     ),
                                 ));
                             }
-                            if let Ok(chunk) = serde_json::from_str::<Value>(data) {
-                                if let Some(mut events) = parse_sse_chunk_with_state(
-                                    &chunk,
-                                    tf,
-                                    defer_message_end,
-                                    &mut stream_state,
-                                ) {
-                                    if !events.is_empty() {
-                                        let first = events.remove(0);
-                                        events.reverse();
-                                        pending = events;
-                                        return Some((
-                                            first,
-                                            (
-                                                stream,
-                                                buf,
-                                                cancelled,
-                                                pending,
-                                                thinking_field,
-                                                stream_state,
-                                            ),
-                                        ));
-                                    }
+                            SseDispatch::Malformed(message) => {
+                                return Some((
+                                    GenerateEvent::Error(message),
+                                    (
+                                        stream,
+                                        buf,
+                                        sse_event,
+                                        cancelled,
+                                        pending,
+                                        thinking_field,
+                                        stream_state,
+                                        true,
+                                    ),
+                                ));
+                            }
+                            SseDispatch::Events(mut events) => {
+                                if events.is_empty() {
+                                    continue;
                                 }
+                                let first = events.remove(0);
+                                events.reverse();
+                                pending = events;
+                                // A chunk-parsed terminal (finish_reason in
+                                // non-deferred mode) seals the stream here as
+                                // well, so a trailing [DONE] cannot emit a
+                                // second MessageEnd.
+                                let exhausted = matches!(
+                                    first,
+                                    GenerateEvent::MessageEnd
+                                        | GenerateEvent::Error(_)
+                                        | GenerateEvent::Cancelled
+                                );
+                                return Some((
+                                    first,
+                                    (
+                                        stream,
+                                        buf,
+                                        sse_event,
+                                        cancelled,
+                                        pending,
+                                        thinking_field,
+                                        stream_state,
+                                        exhausted,
+                                    ),
+                                ));
                             }
                         }
-                        continue;
                     }
 
                     match stream.next().await {
                         Some(Ok(bytes)) => {
-                            buf.push_str(&String::from_utf8_lossy(&bytes));
+                            buf.extend(&bytes);
+                            // A line is only extracted once its terminator
+                            // arrives, so an endless line must be bounded
+                            // here before it exhausts memory. Complete lines
+                            // still in the buffer drain on the next
+                            // iterations, so only a terminator-free overflow
+                            // is an error.
+                            if buf.overflowed_without_line_ending() {
+                                return Some((
+                                    GenerateEvent::Error(format!(
+                                        "SSE line exceeds the maximum size of \
+                                         {MAX_SSE_EVENT_BYTES} bytes"
+                                    )),
+                                    (
+                                        stream,
+                                        buf,
+                                        sse_event,
+                                        cancelled,
+                                        pending,
+                                        thinking_field,
+                                        stream_state,
+                                        true,
+                                    ),
+                                ));
+                            }
                         }
                         Some(Err(e)) => {
                             return Some((
@@ -326,56 +435,91 @@ impl ContentGenerator for OpenAICompatProvider {
                                 (
                                     stream,
                                     buf,
+                                    sse_event,
                                     cancelled,
                                     pending,
                                     thinking_field,
                                     stream_state,
+                                    true,
                                 ),
                             ));
                         }
                         None => {
-                            if !buf.trim().is_empty() {
-                                let line = buf.trim().to_string();
-                                buf.clear();
-                                if let Some(data) = line.strip_prefix("data: ") {
-                                    if data.trim() != "[DONE]" {
-                                        if let Ok(chunk) = serde_json::from_str::<Value>(data) {
-                                            if let Some(mut events) = parse_sse_chunk_with_state(
-                                                &chunk,
-                                                tf,
-                                                defer_message_end,
-                                                &mut stream_state,
-                                            ) {
-                                                if !events.is_empty() {
-                                                    let first = events.remove(0);
-                                                    events.reverse();
-                                                    pending = events;
-                                                    return Some((
-                                                        first,
-                                                        (
-                                                            stream,
-                                                            buf,
-                                                            cancelled,
-                                                            pending,
-                                                            thinking_field,
-                                                            stream_state,
-                                                        ),
-                                                    ));
-                                                }
-                                            }
+                            // Flush a final line missing its trailing newline,
+                            // then the event buffer (streams may end without
+                            // the blank-line separator).
+                            let mut terminal: Option<GenerateEvent> = None;
+                            let mut flushed: Option<String> = None;
+                            if !buf.is_empty() {
+                                let tail = buf.take_tail();
+                                match std::str::from_utf8(&tail) {
+                                    Ok(line) => match sse_event.push_line(line) {
+                                        Ok(data) => flushed = data,
+                                        Err(message) => {
+                                            terminal = Some(GenerateEvent::Error(message));
                                         }
+                                    },
+                                    Err(error) => {
+                                        terminal = Some(GenerateEvent::Error(format!(
+                                            "SSE stream carried invalid UTF-8: {error}"
+                                        )));
                                     }
                                 }
                             }
+                            let mut queued: Vec<GenerateEvent> = Vec::new();
+                            if terminal.is_none() {
+                                if let Some(data) = flushed.or_else(|| sse_event.take_data()) {
+                                    match dispatch_sse_data(
+                                        &data,
+                                        tf,
+                                        defer_message_end,
+                                        &mut stream_state,
+                                    ) {
+                                        SseDispatch::Done => {
+                                            terminal = Some(GenerateEvent::MessageEnd);
+                                        }
+                                        SseDispatch::Malformed(message) => {
+                                            terminal = Some(GenerateEvent::Error(message));
+                                        }
+                                        SseDispatch::Events(events) => queued = events,
+                                    }
+                                }
+                            }
+                            // End-of-stream without [DONE] is complete only
+                            // when the model already signaled finish_reason;
+                            // otherwise the tail was lost in transit and the
+                            // partial output must not pass as success.
+                            let terminal = terminal.unwrap_or_else(|| {
+                                if stream_state.saw_finish_reason {
+                                    GenerateEvent::MessageEnd
+                                } else {
+                                    GenerateEvent::Error(
+                                        "SSE stream ended before completion: \
+                                         no [DONE] marker or finish_reason received"
+                                            .to_string(),
+                                    )
+                                }
+                            });
+                            let already_terminal = queued.iter().any(|event| {
+                                matches!(event, GenerateEvent::MessageEnd | GenerateEvent::Error(_))
+                            });
+                            if !already_terminal {
+                                queued.push(terminal);
+                            }
+                            let first = queued.remove(0);
+                            queued.reverse();
+                            pending = queued;
                             return Some((
-                                GenerateEvent::MessageEnd,
+                                first,
                                 (
                                     stream,
                                     buf,
+                                    sse_event,
                                     cancelled,
                                     pending,
                                     thinking_field,
                                     stream_state,
+                                    true,
                                 ),
                             ));
                         }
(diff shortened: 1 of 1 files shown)


src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
```diff
@@ -439,6 +517,86 @@ struct OpenAICompatStreamState {
...
+        if line.is_empty() {
+            return self.take_data();
+        }
+        if line.starts_with(':') {
+            return None; // comment line
+        }
+        if let Some(value) = line.strip_prefix("data:") {
+            // The space after the colon is optional in the SSE format.
+            self.data_lines
+                .push(value.strip_prefix(' ').unwrap_or(value).to_string());
```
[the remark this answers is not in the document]
SunnyQjm: **Finding**: `SseEventBuffer::data_lines` accumulated `data:` lines with no
byte limit; an endpoint that withholds the blank-line separator could grow
memory without bound. Verified against the code: the claim is correct — the
previous decoder was line-by-line and had no cross-line accumulation, so this
buffer introduced a new unbounded dimension (the raw line buffer had the same
exposure for a single endless line).

**Fix** (in `9d4c95c2`): introduced `MAX_SSE_EVENT_BYTES` (8 MiB) enforced at
both accumulation points — `SseEventBuffer::push_line` now returns
`Err(...)` once the assembled event exceeds the bound (buffer cleared so the
oversized payload is not kept alive), and the raw byte buffer errors on a
newline-free overflow. Both surface as `GenerateEvent::Error`, mirroring the
other fail-loud corruption paths.

**Verification**: new test
`oversized_event_reports_error_instead_of_accumulating` covers
accumulation past the bound + buffer reset; full module suite green
(47 tests), workspace clippy clean.

Old→new head: `da23e35f` → `9d4c95c2`.

src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
```diff
@@ -237,139 +242,257 @@ impl ContentGenerator for OpenAICompatProvider {
...
+                                        cancelled,
+                                        pending,
+                                        thinking_field,
+                                        stream_state,
+                                        true,
+                                    ),
+                                ));
+                            }
+                        };
+                        let data = match sse_event.push_line(line) {
```
kongche-jbw, line 320: [P1] Strip one leading UTF-8 BOM before parsing the first line. The SSE grammar permits a leading BOM, but `from_utf8` preserves it here and `push_line` only recognizes a literal `data:` prefix. Consequently `\u{FEFF}data: {...}` is treated as an unknown field and ignored; a following `[DONE]` still emits `MessageEnd`. A decode-loop repro produced only `[MessageEnd]`, silently dropping the first event and reopening the truncation class this PR is meant to close. Track stream start, remove exactly one BOM, and add a TCP regression covering BOM + first data event + `[DONE]`.

src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
```diff
@@ -237,139 +242,257 @@ impl ContentGenerator for OpenAICompatProvider {
...
-                        if line.is_empty() || line.starts_with(':') {
-                            continue;
-                        }
-                        if let Some(data) = line.strip_prefix("data: ") {
-                            if data.trim() == "[DONE]" {
+                    // Extract one complete line. Bytes are buffered raw and
+                    // decoded per line so a multi-byte character split across
+                    // network chunks reassembles instead of turning into
+                    // replacement characters that corrupt the payload.
+                    if let Some(line_end) = buf.iter().position(|&b| b == b'\n') {
```
kongche-jbw: [P2] Recognize bare CR as an SSE line ending. WHATWG allows LF, CRLF, and a single CR, but this splitter searches only for `b'\n'`; a valid `data: {...}\r\rdata: [DONE]\r\r` stream remains buffered until EOF and is then parsed as one line, producing `malformed SSE data: trailing characters` instead of the expected event. The streaming splitter should handle CR and CRLF, including a CR split from the following LF across network chunks, with a regression test for the bare-CR form.

kongche-jbw (changes requested):
当前 head 已覆盖原始 `data:{...}` 中段丢失、跨 chunk UTF-8、malformed payload 和意外 EOF 的主要截断链路；本地定向测试 `cargo test -p cosh-core --lib provider::openai_compat` 为 47/47 通过。

不过仍有一个会静默丢事件的 P1，以及一个 SSE 规范兼容性 P2，详见 inline comments。两个最小 decode-loop 复现分别得到 `[MessageEnd]`（BOM 首事件被丢弃）和 `Error("malformed SSE data: trailing characters ...")`（bare-CR 流）。建议修复后补入真实 TCP decode-loop 回归矩阵。

非阻断观察：`finish_reason=stop` 后再收到 `[DONE]` 时，完整消费 provider stream 会得到两个 `MessageEnd`，因为从 `pending` 弹出 terminal 时仍携带 `exhausted=false`。当前 core 在第一个 `MessageEnd` 处 break，因此这是既有终止语义债务，不作为本 PR 阻断项；但 PR 中“terminal 后不再 poll”的描述目前并不完全成立。

src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
SunnyQjm: **Finding**: the SSE grammar permits exactly one leading BOM; `from_utf8`
preserved it and `push_line` only matched a literal `data:` prefix, so
`\u{FEFF}data: {...}` became an unknown field and the first event was
silently dropped. Verified: correct — this reopened the silent-drop class.

**Fix** (in `50b10899`): `SseEventBuffer` tracks stream start and strips
exactly one leading BOM before field matching; a later U+FEFF stays payload
(unknown field), per the grammar.

**Verification**: TCP regression `leading_bom_does_not_drop_the_first_event`
(BOM + first data event + `[DONE]`) plus unit
`only_the_first_bom_is_stripped`; module suite 56/56 green.

src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
SunnyQjm: **Finding**: the splitter searched only for `b'\n'`, so a valid bare-CR
stream buffered to EOF and was parsed as one line, yielding
`malformed SSE data: trailing characters`. Verified: correct.

**Fix** (in `50b10899`): new `take_sse_line` splits on LF, CRLF, and bare
CR. A CR that is the last buffered byte is left in place until the next
chunk (it may be the first half of a chunk-split CRLF; consuming it early
would fabricate an empty line, i.e. an event boundary); the EOF flush
resolves a genuinely trailing CR.

**Verification**: TCP regressions `bare_cr_line_endings_are_decoded`
(the reported repro shape `data: {...}\r\rdata: [DONE]\r\r`) and
`crlf_split_across_chunks_is_one_line_ending`; module suite 56/56 green.

src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
```diff
@@ -445,6 +589,148 @@ struct OpenAICompatStreamState {
...
+        if line.is_empty() {
+            return Ok(self.take_data());
+        }
+        if line.starts_with(':') {
+            return Ok(None); // comment line
+        }
+        if let Some(value) = line.strip_prefix("data:") {
+            // The space after the colon is optional in the SSE format.
+            let value = value.strip_prefix(' ').unwrap_or(value);
+            self.buffered_bytes += value.len();
```
kongche-jbw: [P1] The assembled-event bound can still be bypassed with empty or many short `data:` lines. `buffered_bytes` counts only `value.len()`, while every `data:`/bare `data` line appends another `String` to `data_lines`; empty values therefore keep the counter at zero and retain memory without limit. A direct repro accepted 100,000 empty lines with `buffered_bytes=0` and `data_lines.len()=100000`. Separators and per-`String`/`Vec` overhead also make many short lines consume far beyond the advertised 8 MiB before the payload counter trips. Thus the previously reported endpoint-driven OOM path remains reachable even though the non-empty-payload test passes.

src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
```diff
@@ -445,6 +589,148 @@ struct OpenAICompatStreamState {
...
+    while i < buf.len() {
+        match buf[i] {
+            b'\n' => {
+                let mut line: Vec<u8> = buf.drain(..=i).collect();
+                line.truncate(i);
+                return Some(line);
+            }
+            b'\r' => {
+                if i + 1 >= buf.len() {
+                    return None; // possible CRLF split across chunks
```
kongche-jbw: [P2] A chunk-final CR is already a valid bare-CR line ending, but returning `None` here postpones it until another byte or EOF arrives. If an SSE endpoint writes and flushes a complete `data: {...}\r\r` event and then keeps the long-lived connection idle, the second CR (the blank-line boundary) stays buffered and the event never dispatches. A real TCP repro timed out waiting 250 ms for the first `TextDelta`. The current bare-CR regression sends `[DONE]` afterward, which supplies the extra byte and masks this streaming stall.

kongche-jbw (changes requested):
上一轮 BOM、CRLF 跨 chunk 和重复 `MessageEnd` 的修复已验证有效，当前定向模块测试 `cargo test -p cosh-core --lib provider::openai_compat` 为 56/56 通过。

继续复审发现 event bound 仍可被空/短 `data:` 行绕过（P1），并且 chunk 尾部 bare CR 会让已经完整的事件等待下一字节或 EOF 才 dispatch（P2），详见 inline comments。

验证证据：连续输入 100,000 个空 `data:` 行后得到 `buffered_bytes=0, retained_lines=100000`；真实 TCP 服务端发送完整 `data: {...}\r\r`、flush 后保持连接打开时，250ms 内 `stream.next()` 得到 `Err(Elapsed(()))`。现有 bare-CR 测试后面仍有 `[DONE]`，因此没有覆盖连接保持空闲的流式行为。

src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
SunnyQjm: **Finding**: `buffered_bytes` counted only payload bytes, so empty or tiny
`data:` lines grew `data_lines` (String headers, allocator slack, join
separators) while the counter stayed at or near zero — the endpoint-driven
OOM path remained reachable. Verified with your repro shape: 100k empty
lines, counter 0. Correct.

**Fix** (in `2504d82a`): every buffered data line is charged a flat
`SSE_LINE_OVERHEAD_BYTES` (64) on top of its payload, including empty and
bare-`data` lines, so payload-free floods trip the same 8 MiB bound.

**Verification**: `empty_data_lines_cannot_bypass_the_event_bound`
(empty-line flood → bound error, buffer reset); module suite 60/60.

src/cosh-ng/crates/cosh-core/src/provider/openai_compat.rs
SunnyQjm: **Finding**: holding a chunk-final CR until the next byte arrives stalls a
stream that flushes a complete `data: {...}\r\r` event and then idles —
the blank-line boundary sits buffered and the event never dispatches until
more bytes (e.g. [DONE]) show up. Verified; my earlier regression indeed
masked this by sending [DONE] immediately. Correct.

**Fix** (in `2504d82a`): reworked the splitter into `SseByteBuffer` — a
bare CR now ends its line immediately; a `skip_lf` flag swallows only the
LF half of a chunk-split CRLF on the next chunk, so no empty line (event
boundary) is fabricated.

**Verification**: TCP liveness regression
`cr_framed_event_dispatches_before_the_stream_idles` (server flushes the
CR-framed event, then idles 1s before [DONE]; the delta must arrive within
700ms) plus unit coverage
`chunk_final_bare_cr_yields_the_line_without_more_bytes` and
`chunk_split_crlf_does_not_fabricate_an_event_boundary`; module suite
60/60, workspace clippy clean.

kongche-jbw (approved):
lgtm thx

PR #1942, opened 2026-07-28
fix(cosh-ng): [shell] intercept slash-bearing natural-language prompts

SunnyQjm:
## Summary

Natural-language prompts whose IFS first token contains a slash (e.g. a
Chinese sentence glued to a path by full-width punctuation:
`你读一下，并安装这个skill：/usr/share/anolisa/runtime/skills/ws-ckpt/SKILL.md`)
never reach the natural-language interception: bash executes slash-bearing
command words as paths and reports `No such file or directory` **without
invoking `command_not_found_handle`**, where the classifier is hooked. The
user sees a raw bash error instead of the agent taking over.

Root cause is a broken evidence chain, plus a second gap the issue did not
cover: `_cosh_command_veto` unconditionally classified slash-bearing top
tokens as commands, so remounting the classifier alone (the issue's
suggested direction A/B) is a no-op — verified on a real PTY.

## Changes

- `shell_host/input_intent.sh`: `_cosh_classify_missing` /
  `_cosh_command_veto` accept an optional `missing_path` context that lifts
  **only** the slash-shape veto. The tilde veto and every other rule
  (metacharacters, option/assignment tokens, length/UTF-8 guards) stay
  active. Callers that do not pass the context keep byte-identical
  semantics (zsh untouched).
- `shell_host/marker/bash.rs`: new `_cosh_should_intercept_missing_path`
  DEBUG-trap helper with five gates — slash-bearing first word, AI
  enabled, no secret, **path does not exist** (`-e`), missing-path
  classification says `natural_language` — mounted on both the main
  `_COSH_AT_PROMPT` branch and the stale-history `BASH_COMMAND` fallback.
  On a `natural_language` verdict it emits the intercept marker (same
  shape as slash/agent-marker intercepts) and returns 1 so `extdebug`
  drops the command; every other verdict falls through to the native bash
  error, byte-identical to pre-fix behavior. Zero Rust runtime changes.

Safety envelope (fail-closed): existing paths — even non-executable ones —
are never intercepted; plain-English typo paths, bare URLs, `~/` prefixes,
veto metacharacters, secret-bearing input, and AI-disabled sessions all
keep native behavior. The worst failure mode is the fix not applying.

Design doc: `specs/cosh-1919-slash-nl-intercept/` (input-shape matrix,
invariants, decision table incl. the zsh boundary D2: zsh has no
extdebug-equivalent veto — probes confirmed `trap "false" DEBUG` does not
block execution — so zsh is tracked separately).

## Tests

[...]

PR #2210, opened 2026-08-04
fix(cosh-ng): [shell] intercept zsh slash-bearing natural-language prompts

### LINKED ISSUE #1943: [cosh-ng] zsh: slash-bearing natural-language prompts bypass interception (bash fixed in #1942)
### Component

cosh-ng

### Background

#1919 fixed slash-bearing natural-language prompt interception for the
bash host (PR #1942): bash executes slash-bearing command words as paths
without invoking `command_not_found_handle`, so the DEBUG-trap
missing-path reclassification was added there. The same defect exists on
the zsh host, but the bash mechanism cannot be ported. This issue tracks
the zsh side (declared Non-Goal D2 in `specs/cosh-1919-slash-nl-intercept/`).

### Probe evidence (2026-07-28, zsh 5.9 / bash)

| Probe | Result |
|---|---|
| bash: `shopt -s extdebug; trap "false" DEBUG; echo X` | X suppressed — the veto mechanism the bash fix relies on |
| zsh: `trap "false" DEBUG; echo X` | X still prints — zsh's DEBUG trap has **no veto-execution semantics**; the bash approach is not portable |
| zsh: `/no/such/path-probe` with `command_not_found_handler` defined | prints `no such file or directory`, handler **not** invoked (slash-free tokens do invoke it) — same broken evidence chain as bash |

### Candidate routes

**Z1 (primary): `accept-line` ZLE widget** — the canonical zsh
pre-execution interception point (used by the plugin ecosystem, e.g.
autosuggestions). The five-gate verdict from the bash fix can be reused
as-is (`input_intent.sh` is already shared by bash and zsh, and the
`missing_path` classification context landed in #1942). ZLE `$CONTEXT`
(`start`/`cont`) natively isolates heredoc/PS2 continuation lines. New
risk surface that needs its own design pass: compatibility matrix with
user ZLE customizations / other widget-wrapping plugins, and history
semantics for intercepted lines.

**Z2 (degraded fallback): post-failure recovery** — `precmd` sees exit
127/126; with a slash-bearing first word plus natural-language shape,
route through the existing failed-command insight machinery (#1742) to
auto-suggest/take over. The user briefly sees the native error (degraded
UX) but the change is small and zero-risk to native behavior.

### Expected behavior

Same contract as the bash fix: a natural-language prompt whose first
token contains a slash and does not resolve to an existing path is taken
over by the agent; existing paths, English typo paths, URLs, secrets, and
veto metacharacters keep native zsh behavior (fail-closed).

SunnyQjm:
## Summary

zsh sibling of #1919 (bash fixed in #1942): a natural-language prompt whose
first IFS token contains a slash and does not resolve to an existing path is
executed by zsh as a path — zsh never invokes `command_not_found_handler` for
slash-bearing command words, so the NL classifier hooked there is unreachable
and the user sees a raw `zsh: no such file or directory` error instead of the
agent taking over. zsh's DEBUG trap has no veto-execution semantics (probe
evidence in #1943), so the bash mechanism cannot be ported; the only
pre-execution seam zsh offers is the `accept-line` ZLE widget.

## Changes

- `marker/zsh.rs`: new `_cosh_should_intercept_missing_path` — the bash
  missing-path verdict (slash-bearing first word, AI enabled, path provably
  ENOENT, missing-path classification says natural_language) plus a URL
  veto: scheme-shaped first words (`[a-zA-Z]*://*`) always keep the native
  result. The shared `input_intent.sh` classifier is reused verbatim
  (single source of truth; classifier and bash side have zero diff).
- `marker/zsh.rs`: `_cosh_accept_line` ZLE widget mounted over the active
  `accept-line` (alias-saving chain keeps user widgets working) evaluates the
  gates at line submission. On a natural_language verdict it re-adds the line
  to history (`print -sr`; skipped for secret-bearing lines), re-echoes the
  prompt+text (ZLE erases the edit line on buffer clear; sensitive lines
  re-echo the redaction placeholder), emits the same intercept marker shape
  as the bash missing-path route (`natural_language`,
  `top_level_missing=false`, `sensitive` as-is), clears the buffer and
  accepts the empty line. Every gate failure and internal error falls open to
  the original accept-line: the worst case is the interception not firing,
  never a broken native line. `CONTEXT != start` submissions (PS2/heredoc
  continuations, vared) always pass through.

- The zsh marker script body lives in `zsh_marker.sh`
  (`include_str!`, byte-identical move) keeping `zsh.rs` under the
  700-line layout gate; slash registry tests re-anchored to the script
  file.

## Tests

Twenty-three zsh-side integration tests mirroring the bash anchors
(`tests/shell_host/marker.rs`):

- `shell_host_zsh_missing_path_natural_language_intercepts` — positive
  intercept, no native error, re-echo visible.
- `shell_host_zsh_sensitive_missing_path_natural_language_intercepts` —
  sensitive flag, journal whole-field redaction, redaction placeholder
  re-echo, raw key absent from events/journal.
- `shell_host_zsh_missing_path_counterproofs_stay_native` — existing
  executable, English typo path, dangling symlink, permission-opaque parent
  all keep native behavior; slash-free CNF route unaffected.
- `shell_host_zsh_missing_path_intercepts_with_user_accept_line_widget` —
  interception works with a user accept-line widget present and the user
  widget stays in the chain for pass-through lines.
- `shell_host_zsh_missing_path_heredoc_continuation_stays_native` —
  CONTEXT=cont lines pass through untouched.

- `shell_host_zsh_missing_path_intercepts_with_builtin_alias_accept_line` —
  `zle -A` builtin-alias customizations survive the mount and native lines
  keep executing through the preserved alias (review P2).
- `shell_host_zsh_missing_path_intercept_never_runs_user_widget_synthesis` —
  an empty-buffer-synthesizing user widget never executes a native command
  after a successful intercept (review round 2).
- `shell_host_zsh_missing_path_foreign_history_hook_stays_uninvoked` —
  a foreign `fc -p` zshaddhistory hook is never replayed and the session
  HISTFILE context stays untouched (review round 2).
- `shell_host_zsh_missing_path_url_first_word_stays_native` — URL first
  word + NL text keeps the native error (review round 3).
- `shell_host_zsh_missing_path_intercepts_with_direct_submit_key_binding` —
  `bindkey '^M'/'^J'` straight to a user widget: interception reachable,
  pass-through keeps the user widget (review round 3).
- `shell_host_zsh_missing_path_intercepted_lines_stay_out_of_history` —
  repeated intercepted prompts leave zero history entries (review round 3).
- `shell_host_zsh_missing_path_intercept_does_not_reevaluate_ps1` —
  PROMPT_SUBST side effects run exactly as often as native submission
  (review round 3).
- `shell_host_zsh_missing_path_submit_key_widgets_stay_per_keymap` —
  distinct submit widgets in the main and vicmd keymaps never cross modes
  (review round 4).
- `shell_host_zsh_missing_path_reentrant_accept_line_runs_widget_once` —
  a delegated widget finishing with the named `zle accept-line` runs
  exactly once and the line still submits (review round 5).
- `shell_host_zsh_missing_path_slash_function_stays_native` — a defined
  slash-bearing function executes natively; the gate mirrors zsh
  resolution via `whence -w` (review round 6).
- `shell_host_zsh_missing_path_han_glob_stays_native` — a Han first word
  carrying an unquoted glob expands and runs natively (review round 6).
- `shell_host_zsh_missing_path_reentrant_reaches_saved_accept_line` — a
  re-entrant named accept-line reaches the saved user wrapper, keeping
  the A → B chain intact (review round 6).
- `shell_host_zsh_missing_path_quoted_function_stays_native` — a quoted
  slash-bearing function call runs natively; the literal-word whitelist
  fails open on any lexing/expansion form (review round 7).
- `shell_host_zsh_missing_path_brace_expansion_stays_native` — a
  brace-expanded first word runs natively (review round 7).
- `shell_host_zsh_missing_path_aliased_keymap_keeps_user_widget` — with
  `bindkey -A main vicmd` the claim scan mirrors saved entries so a
  vicmd submission still reaches the user's widget (review round 7).
- `shell_host_zsh_missing_path_custom_keymap_still_intercepts` — a
  user-defined keymap selected from zle-line-init is claimed at mount;
  interception fires and the user's widget stays (review round 8).
- `shell_host_zsh_missing_path_space_keymap_still_intercepts` — a
  space-bearing keymap name is enumerated per line and claimed
  (review round 9).
- `shell_host_zsh_missing_path_alias_mirrors_shared_map_widget` — the
  two-phase mount captures bindings before any rebind, so an alias of
  main dispatches main's widget, never vicmd's (review round 9).

## Verification

Focused (alinux3 arm64 container, zsh 5.5.1):

- `cargo test -p cosh-shell --test shell_host` — new cases green; 53 zsh
  filter tests green. Pre-existing environment failures unrelated to this
  diff (`heavy::raw_relay_host_shows_isolated_sudo_prompt_and_keeps_shell_usable`,
  `termios::cosh_owned_timeout_recovery_restores_pty_without_visible_command`,
  `tools::readonly_compound_tests::*` reap-deadline pair) fail identically on
  the clean base commit (stash-control attribution).
- `cargo test -p cosh-shell --lib` / `--bin cosh-shell` — green except the
  same pre-existing `readonly_compound` pair (also fails on clean base).
- `cargo clippy --workspace --all-targets -- -D warnings` (container) green;
  `cargo fmt --check` green; `check-layout.sh` and `check-test-inventory.sh`
  green.

Real-machine acceptance (real PTY 120x40, real cosh-core adapter, real LLM):

- Container zsh 5.5.1: FAIL→PASS on both defect scenarios (f1 full-width
  single-token absolute path, f2 relative path), slash-free control scenario
  unchanged; native error gone, agent takes over.
- macOS zsh 5.9 spot check (f1): intercept + takeover + re-echo correct
  under a themed user prompt.

Not run: workspace-wide test targets beyond cosh-shell; macOS cosh-core
build (pre-existing Linux-only `rustix::openat2` usage, unrelated).

## Evidence

Fork-hosted, commit-SHA-pinned
(`SunnyQjm/anolisa@c9a65b3e312458743766d4f530ed28b5d4ff0e8c`, branch
`pr-2210-assets`). All screenshots are full-size (120x40) final frames
rendered from the real-PTY casts.

FAIL→PASS on the exact issue scenario (f1, container zsh 5.5.1):

| Before (main `b0a57a26`) | After (this PR) |
|---|---|
| head f1 | fixed f1 |

Relative-path variant (f2):

| Before | After |
|---|---|
| head f2 | fixed f2 |

Slash-free control (c1, no regression):

| Before | After |
|---|---|
| head c1 | fixed c1 |

macOS zsh 5.9 spot check (f1, themed user prompt):

macos f1

Casts:
[head-f1.cast](https://raw.githubusercontent.com/SunnyQjm/anolisa/c9a65b3e312458743766d4f530ed28b5d4ff0e8c/head-f1.cast) ·
[fixed-f1.cast](https://raw.githubusercontent.com/SunnyQjm/anolisa/c9a65b3e312458743766d4f530ed28b5d4ff0e8c/fixed-f1.cast)

Closes #1943


src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
```diff
@@ -614,6 +628,63 @@ for _cosh_sc in about agent allow answer approval-mode approve audit auth cancel
...
+  fi
+  local sensitive=false
+  local echo_line="$line"
+  if _cosh_command_has_secret "$line"; then
+    sensitive=true
+    echo_line="<redacted sensitive command>"
+  else
+    # Cleared buffers never reach zshaddhistory, so intercepted prompts
+    # are re-added manually to stay recallable like their bash siblings.
+    print -sr -- "$line" 2>/dev/null || true
```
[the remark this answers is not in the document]
SunnyQjm: Valid — accepted and fixed in 27551a84.

**Review item**: `print -sr` bypassed `HIST_IGNORE_SPACE` and the `zshaddhistory` hook chain, so explicitly excluded lines (including secrets `_cosh_command_has_secret` does not recognize) could be persisted.

**Verification**: confirmed against zsh behavior — `print -s` performs a raw history add without consulting history options or hooks.

**Fix**: new `_cosh_history_policy_allows` replays the active policy before the re-add: `histignorespace` + leading-space check, then the full `zshaddhistory_functions` chain (which includes cosh's own secret/utf8 filter). Any veto skips the add — worst case is a non-recallable intercepted prompt, never an unwanted persisted one.

**Evidence**: new test `shell_host_zsh_missing_path_history_respects_user_policy` (HIST_IGNORE_SPACE + space-prefixed intercepted prompt: interception still fires, line stays out of history).

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
```diff
@@ -614,6 +628,63 @@ for _cosh_sc in about agent allow answer approval-mode approve audit auth cancel
...
+  BUFFER=""
+  zle -I 2>/dev/null || true
+  print -r -- "${(%%)PS1}${echo_line}" 2>/dev/null || true
+  _cosh_emit_intercept_marker "$line" "natural_language" false "$sensitive"
+  _cosh_dispatch_accept_line
+}
+if [[ -n "${widgets[accept-line]:-}" && "${widgets[accept-line]}" == user:* ]]; then
+  zle -A accept-line _cosh_orig_accept_line 2>/dev/null \
+    && _COSH_HAS_ORIG_ACCEPT_LINE=1 \
+    || _COSH_HAS_ORIG_ACCEPT_LINE=0
```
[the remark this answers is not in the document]
SunnyQjm: Valid — accepted and fixed in 27551a84.

**Review item**: `zle -A <builtin> accept-line` customizations keep `widgets[accept-line]` reporting `builtin` (verified on zsh 5.5.1: after `zle -A accept-line-and-down-history accept-line`, `widgets[accept-line]` is `builtin`, not `user:*`), so the `user:*`-gated save dropped them and dispatch fell back to `.accept-line`.

**Fix**: the save is now an unconditional `zle -A accept-line _cosh_orig_accept_line`, preserving whatever accept-line currently resolves to — user widget, builtin alias, or the default; the `user:*` branch is gone. Alias-save failure still falls back to `.accept-line` (fail-open).

**Evidence**: new test `shell_host_zsh_missing_path_intercepts_with_builtin_alias_accept_line` (rcfile aliases accept-line to `accept-line-and-down-history`; interception still fires and native lines keep executing through the preserved alias), plus the existing `user:*` widget chain test still green.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh
```diff
@@ -0,0 +1,704 @@
...
+    print -sr -- "$line" 2>/dev/null || true
+  fi
+  # Clear first so ZLE erases the edit line while it still owns it, then
+  # invalidate and re-echo: invalidating before the clear leaves the old
+  # line on screen and the re-echo would duplicate it.
+  BUFFER=""
+  zle -I 2>/dev/null || true
+  print -r -- "${(%%)PS1}${echo_line}" 2>/dev/null || true
+  _cosh_emit_intercept_marker "$line" "natural_language" false "$sensitive"
+  _cosh_dispatch_accept_line
```
kongche-jbw: **[P1] Bypass the saved widget after a successful intercept**

At this point `BUFFER` has been cleared and the intercept marker has already been emitted, but `_cosh_dispatch_accept_line` invokes an arbitrary user `accept-line` widget. Such a widget may legitimately synthesize a command for an empty buffer. I reproduced this with a widget that sets an empty buffer to `echo review-unexpected-native`: the event stream became `UserInputIntercepted` followed by `CommandStarted("echo review-unexpected-native")`. This violates the interception guarantee and can execute unintended native commands. Submit the cleared line through `.accept-line` directly on the successful-intercept path; keep the saved user widget only for pass-through paths.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh
```diff
@@ -0,0 +1,704 @@
...
+# non-recallable intercepted prompt, never an unwanted persisted one.
+_cosh_history_policy_allows() {
+  local line="$1"
+  if [[ -o histignorespace ]] && [[ "$line" == ' '* ]]; then
+    return 1
+  fi
+  local hook
+  for hook in "${zshaddhistory_functions[@]}"; do
+    (( ${+functions[$hook]} )) || continue
+    "$hook" "$line"$'\n' || return 1
```
kongche-jbw: **[P1] Do not invoke `zshaddhistory` hooks outside native hook processing**

zsh automatically restores history contexts pushed with `fc -p` after native `zshaddhistory` processing. Calling each hook function directly here bypasses that restoration. With a valid per-directory-history hook that runs `fc -p local-history`, one intercepted prompt permanently changed subsequent `$HISTFILE` from `global-history` to `local-history` in the PTY session. Later commands can therefore be persisted to the wrong file. Preserve zsh's native hook-context semantics instead of directly replaying arbitrary hooks; the replacement should also retain status `2` semantics (internal history only, no history-file write).

kongche-jbw (changes requested):
Requesting changes for two blocking ZLE/history integration regressions reproduced against `27551a84` with the cosh-shell PTY harness. The PR's seven focused zsh tests pass, but they do not cover these user-customization paths.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh
SunnyQjm: Valid — accepted and fixed in 72c5da83.

**Review item**: after a successful intercept the cleared buffer was handed to the saved user accept-line widget, which may legitimately synthesize a command for an empty buffer — executing a native line the marker already claimed as intercepted.

**Verification**: confirmed on the current code path — `_cosh_dispatch_accept_line` on the successful-intercept path invokes `_cosh_orig_accept_line` unconditionally; your empty-buffer-synthesis reproduction is exactly reachable.

**Fix**: the successful-intercept path now finalizes through `zle .accept-line` directly; the saved user widget stays in play only on pass-through paths.

**Evidence**: new test `shell_host_zsh_missing_path_intercept_never_runs_user_widget_synthesis` — rcfile widget sets an empty buffer to `echo review-unexpected-native`; asserts the intercept event fires, no command block and no terminal output for the synthesized command, and pass-through lines still reach the user widget.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh
SunnyQjm: Valid — accepted and fixed in 72c5da83.

**Review item**: replaying `zshaddhistory_functions` directly bypasses zsh's automatic restoration of `fc -p` history contexts (native-only behavior) and cannot honor status-2 (internal-only) semantics — a per-directory-history hook left `$HISTFILE` permanently swapped.

**Verification**: confirmed against zsh semantics — the pushed history context is only popped automatically after *native* zshaddhistory processing; a direct call leaks it into the session.

**Fix**: foreign hooks are no longer invoked at all. `_cosh_history_policy_allows` now fails closed: any foreign entry in `zshaddhistory_functions`, or a standalone `zshaddhistory` function, skips the manual re-add entirely (worst case: a non-recallable intercepted prompt — never a polluted history context or an unwanted persisted line). Only cosh's own side-effect-free filter (`_cosh_zshaddhistory_marker`) is consulted, which also covers the status-2 concern since no foreign return value is reinterpreted anymore.

**Evidence**: new test `shell_host_zsh_missing_path_foreign_history_hook_stays_uninvoked` — a logging `fc -p` hook mirroring per-directory-history; asserts interception still fires, the hook is never invoked for the intercepted line, and `$HISTFILE` observed by the next command is not left pointing at the pushed context.

kongche-jbw (commented):
Review baseline: `6ece763ed999...72c5da8346bc`

**[P1] Keep URL-shaped tokens on the native path**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:185`

The missing-path context disables the slash veto without adding the URL counterproof required by
#1943. In a readable working directory, `https://example.invalid/path 请帮我打开` proves missing
and classifies as `natural_language`, so this branch sends a URL-shaped command to the agent
instead of preserving zsh's native result. Please reject URL schemes before this verdict and add
a PTY counterproof with a URL first token plus natural-language text.

**[P1] Wrap the widget bound to the submit key**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:710`

Replacing the widget named `accept-line` does not affect a keymap that directly binds `^J` or
`^M` to `accept-line-and-down-history` or a user widget. After this mount, `bindkey '^J'` still
points to that alternate widget, so `_cosh_accept_line` is unreachable and the slash-bearing NL
input reaches the original raw path error. Preserve and wrap the active submit-key bindings across
keymaps, then add a PTY test using a direct `bindkey` customization.

**[P2] Apply the complete in-memory history policy**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:687`

`_cosh_history_policy_allows` only honors `HIST_IGNORE_SPACE` and hooks. `print -s` also bypasses
options such as `HIST_IGNORE_DUPS`: with that option set, two identical intercepted prompts are
both inserted, whereas native submission keeps only one. This makes the new route persist history
entries the user asked zsh to suppress. Cover the applicable duplicate/reduction options or avoid
the raw history insertion, with a repeated-prompt regression test.

**[P2] Do not evaluate PS1 again while re-echoing**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:694`

`${(%%)PS1}` performs prompt expansion. With `PROMPT_SUBST` and a `PS1` containing `$(...)`, an
intercepted line runs that command substitution here and then again when zsh renders the next
prompt. Side-effecting or expensive prompt commands therefore run one extra time only on this
route. Reuse the prompt already rendered by ZLE, or re-echo without evaluating `PS1`, and add a
`PROMPT_SUBST` counter test.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh
emanuelmustea: Fixed

kongche-jbw (commented):
Review baseline: `43d939139df6...5e4d2581784b`

**[P1] Preserve submit widgets per keymap**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:733`

`_COSH_SUBMIT_KEY_WIDGETS` is indexed only by `$KEYS`, although the mount scans four keymaps.
If `emacs` and `viins` bind `^M` to different widgets, the later scan overwrites the saved
`emacs` widget. Pressing Enter in emacs then invokes the viins widget for every pass-through
line; a mode-specific widget may rewrite or submit a different buffer.

Possible direction: key saved widgets by keymap plus key (or use per-keymap wrappers), and add
a PTY test with distinct `^M` widgets in emacs and viins that exercises both active maps.

kongche-jbw (commented):
Review baseline: `4349e319a682...0c5864e1000e`

**[P1] Prevent recursive dispatch through `accept-line`**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:653`

A directly bound submit widget may legitimately finish with `zle accept-line`. On a pass-through
line, this function invokes that widget; its `accept-line` call re-enters `_cosh_accept_line`, which
dispatches the same widget again. Binding both `^M` and `^J` to such a widget and submitting
`echo ok` repeats the widget until zsh reports `maximum nested function level reached`, so native
lines can multiply widget side effects and fail submission.

Possible direction: guard delegated dispatch so a re-entrant `accept-line` call reaches
`_cosh_orig_accept_line`, and add a PTY test whose directly bound widget uses `zle accept-line`
and runs exactly once.

kongche-jbw (commented):
Review baseline: `4349e319a682...778126fdbef7`

**[P1] Preserve the saved `accept-line` chain on re-entry**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:679`

When `^M` is bound to widget A and `accept-line` is customized to widget B, A may finish with
`zle accept-line` to invoke B. The new in-progress branch instead jumps directly to
`.accept-line`, so B never runs on pass-through submissions; its buffer transforms, validation,
or side effects are silently lost. The added test leaves `accept-line` at the builtin and misses
this combined customization.

Possible direction: Track key-widget delegation separately from saved-accept-line delegation.
First re-entry should invoke `_cosh_orig_accept_line`; only a further recursive entry should reach
the builtin. Please add a PTY regression with both custom widgets active.

kongche-jbw (commented):
Review baseline: `4349e319a682...778126fdbef7`

**[P1] Preserve the saved `accept-line` target in S5**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:679`

The proposed S5 outcome still changes a valid combined setup. If Enter is bound to widget A and
`accept-line` names widget B, A may call `zle accept-line` specifically to delegate to B. The guard
at this line jumps to `.accept-line`, so B's buffer edits, validation, and side effects disappear.
This violates I-b even though A runs once: native behavior is A once, followed by B once.

Possible direction: distinguish key-widget dispatch from saved-accept-line dispatch. The first
re-entry from A should call `_cosh_orig_accept_line`; only a further recursive entry should reach
the builtin. Please add a PTY regression with distinct A and B widgets.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh
```diff
@@ -0,0 +1,769 @@
...
+    _cosh_dispatch_accept_line
+    return
+  fi
+  local line="$BUFFER"
+  local first_word="$line"
+  while [[ "$first_word" == ' '* || "$first_word" == $'\t'* ]]; do
+    first_word="${first_word#?}"
+  done
+  first_word="${first_word%%[[:space:]]*}"
+  if ! _cosh_should_intercept_missing_path "$first_word" "$line" 2>/dev/null; then
```
kongche-jbw, line 732: **[P1] Keep Zsh-resolved commands on the native path**

This gate treats the raw `BUFFER` token as only a filesystem path, but Zsh can resolve a slash-bearing token as a function or alias, and can expand it before execution. I reproduced this on Zsh 5.9:

```zsh
function 路径/run { return 0 }
路径/run 帮我运行一下
```

Native Zsh runs the function and returns 0, while the current helper reports the literal path as provably missing and classifies the line as `natural_language`, so the wrapper consumes a valid native command and sends it to the agent. A glob has the same failure: `目录*/run` can expand to an existing executable, but when the token contains Han characters the shared veto path does not reject `*`. This also violates #1943's contract that metacharacter counterproofs stay native. Please preserve Zsh command resolution/expansion on this path and add PTY counterproofs for a slash-bearing function and a matching Han-path glob.

kongche-jbw (commented):
Review baseline: `4b131c4faa2a...1cfdaccd2675`

Found one new blocking correctness issue in the current head. The previously reported P1 at `zsh_marker.sh:679` (re-entry skips the saved `accept-line` target in the combined A → B widget chain) is unchanged and remains blocking; I have not duplicated that inline comment.

Validation: reproduced the new issue with zsh 5.9 on aarch64 Linux using the exact current gate/classifier logic. Relevant GitHub CI checks are green, but they do not cover this Zsh resolution path.

kongche-jbw (commented):
Review baseline: `52eeb520147b...5b35d4058965`

**[P1] Parse command words before proving them missing**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:196`

`first_word` is still raw `BUFFER` text here. Zsh runs a defined `路径/run`
function for `"路径/run" 帮我运行`, but `whence` receives the quotes and the path
probe treats `"路径` as missing, so this gate returns `natural_language` and
consumes a valid command. Brace expansion has the same problem:
`/b{in,路径}/echo 帮我运行` executes `/bin/echo` natively but is intercepted here.
Possible direction: safely normalize a literal command word, or fail open on
quoting, escaping, and every pre-execution expansion form. Add PTY counterproofs
for quoted slash functions or paths and brace-expanded commands.

**[P1] Preserve bindings when keymaps alias the same map**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:788`

If an rcfile binds a custom Enter widget and then runs `bindkey -A main vicmd`
before this marker loads, `main` and `vicmd` share one map. The scan saves the
original only under `main`; its rebind is already visible through `vicmd`, so the
second scan skips it. In vicmd, `KEYMAP=vicmd`, the lookup at line 673 misses, and
the global accept-line target silently replaces the user's submit transform or
validation on every native line. Possible direction: capture every logical map
before rebinding aliased maps, or mirror the saved entry. Add a PTY regression
that aliases `vicmd` to `main`, enters vicmd, and verifies the widget runs once.

kongche-jbw (commented):
Review baseline: `0a190d509b8e...86e37c72df18`

**[P1] Claim submit keys in user-defined keymaps**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:811`

The mount scans only `main` and `vicmd`. If an rcfile creates a custom keymap,
binds `^M`/`^J` there to a widget that finishes with `zle .accept-line`, and
selects that map from `zle-line-init`, Enter never reaches `_cosh_accept_line`.
The issue input then still produces the native `no such file or directory`
error and no intercept marker. This setup exists before the marker loads, so it
is not the declared post-mount rebind boundary.

Possible direction: enumerate user-defined keymaps at mount time, preserve
aliases as this table already does, and add a PTY regression that submits from
a custom map.

kongche-jbw (commented):
Review baseline: `8f2416470233...97f275cb701d`

**[P1] Preserve whitespace in custom keymap names**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:818`

`bindkey -N 'review map' main` is valid, and `bindkey -l` emits that name on one line.
The unquoted command substitution here splits it into `review` and `map`, so the actual
map is never claimed. If `zle-line-init` selects that map and its submit widget calls
`zle .accept-line`, Enter bypasses `_cosh_accept_line`; the slash-bearing prompt still
ends in zsh's native `no such file or directory` instead of an intercept.

Possible direction: iterate `bindkey -l` line by line and cover a space-bearing map in PTY.

**[P1] Mirror aliases from the map they actually share**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:796`

When `main` and `vicmd` have different submit widgets and a custom map aliases `main`,
both source maps have already been rebound when the custom map is scanned. This loop
copies the first saved entry matching the key, without checking which physical map the
custom name aliases. On zsh 5.9 at this head, `bindkey -A main review-alias` stores
`_vicmd_submit` for `review-alias`. Every pass-through Enter in that map then invokes
the wrong user widget, which can apply mode-specific edits or validation.

Possible direction: capture each logical map's binding before any rebind, then add a PTY
regression with distinct `main`/`vicmd` widgets and a custom alias of `main`.

kongche-jbw (commented):
Review baseline: `8f2416470233...026f6e9fac91`

**[P1] Preserve the complete widget name from `bindkey`**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:788`

ZLE accepts widget names containing whitespace. For a direct binding to `submit widget`,
`bindkey` emits `"^M" submit widget`, but this last-field split records only `widget`.
The following `widgets[$widget]` lookup then fails, so Enter remains bound directly to the
user widget and never reaches `_cosh_accept_line`. A slash-bearing natural-language prompt
therefore still ends in the native path error in this valid configuration.

Possible direction: remove only the rendered key prefix, preserve the remaining widget name
exactly, and add a PTY regression with a space-bearing widget name.

**[P1] Do not claim Enter in non-submitting custom maps**

`src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_marker.sh:811`

Every non-builtin keymap is treated as a command-submission map, but user keymaps may be
modal. For example, a map selected by `zle-line-init` can bind Enter to a widget that edits
`BUFFER` or accepts a selection without calling `accept-line`. This scan still captures that
widget and phase 2 rebinds Enter to `_cosh_accept_line`; on a slash-bearing natural-language
buffer, the gate clears and submits the line instead of invoking the modal widget. Excluding
only zsh's builtin modal maps does not preserve equivalent user-defined maps.

Possible direction: claim custom-map bindings only with a verifiable submission contract,
fail open otherwise, and cover a non-submitting custom-map Enter widget in PTY.

PR #2221, opened 2026-08-04
fix(anolisa): support native qoder plugins

### LINKED ISSUE #2191: Qoder driver rejects native plugin bundle layouts
## Context

agent-sec-core ships a Qoder-native plugin bundle with this layout:

```text
.qoder-plugin/plugin.json
hooks/hooks.json
hooks/*.py
```

Its hook commands use `${QODER_PLUGIN_ROOT}` to refer to files inside the
plugin.

## Current behavior

The ANOLISA Qoder driver requires `hooks.json` at the adapter resource root.
Both bundle probing and bundle loading fail when only `hooks/hooks.json`
exists.

The driver also only recognizes `${QODER_TOKENLESS_HOOKS}` and resolves it to a
sibling `common/hooks` directory. It does not recognize
`${QODER_PLUGIN_ROOT}`.

Running:

```shell
anolisa adapter enable sec-core qoder
```

fails during bundle validation because the root-level `hooks.json` is missing.
If that check is bypassed, the hook commands still contain an unresolved
`${QODER_PLUGIN_ROOT}` value.

## Impact

- The sec-core Qoder payload can be installed but cannot be enabled through
  ANOLISA.
- Qoder settings and adapter receipts cannot be created successfully.
- The current driver only supports the existing Tokenless-specific layout, not
  Qoder's native plugin layout.

Related: #2133

ikunkun-sys:
## Why

Qoder native hook plugins are discovered and managed by qodercli, but the ANOLISA driver only supported the legacy Tokenless layout by copying bundles, rewriting hooks, and editing settings.json. This caused native bundles such as sec-core to be rejected or installed through an incompatible lifecycle.

## What changed

- Classify Qoder bundles as Native (`hooks/hooks.json`) or Legacy (root `hooks.json`) and reject ambiguous layouts.
- Require Qoder's fixed `.qoder-plugin/plugin.json` manifest so receipt identity matches the plugin qodercli actually registers.
- Delegate Native validation, installation, inventory verification, status, and uninstall to `qodercli plugins`.
- Reject an existing matching registration in any Qoder scope before mutation because project, local, and user registrations share the local plugin cache.
- Permit same-ID re-enable only when the prior schema-v3 receipt contains the explicit successful-install checkpoint.
- Reject Native re-enable when the Plugin ID changes, preserving the old receipt until its plugin is explicitly disabled.
- Persist an install-confirmation checkpoint immediately after qodercli succeeds and before inventory verification; failed install attempts never imply ownership.
- Make retry and disable fail closed for unconfirmed and pre-v3 Native receipts: registered plugins are left untouched, while a verified absence permits receipt cleanup.
- Capture structured CLI inventory with a separate 4 MiB stdout bound instead of the ordinary 64 KiB diagnostic limit.
- Keep Native plugin files and `settings.json` untouched, with write-ahead receipts retained on uncertain cleanup.
- Evolve the driver receipt schema to v3 while preserving Legacy v1 deserialization.
- Preserve the existing Legacy Tokenless staging and settings lifecycle unchanged.

## Related issue

Closes #2191

## User / Agent impact

ANOLISA can enable Qoder native hook plugins through the framework-supported plugin lifecycle. Existing Tokenless installations and legacy receipts continue to use their current behavior.

## Risk and compatibility

The change adds a new Qoder adapter lifecycle without removing the legacy path. ANOLISA refuses to install while the same Plugin ID exists in any scope, because qodercli backs those registrations with a shared cache. Native cleanup fails closed whenever inventory cannot prove the user-scope plugin is absent. Pre-v3 Native receipts have no durable install checkpoint and therefore never authorize replacement or uninstall; users must remove a matching registration explicitly before the stale receipt can be cleaned up.

## Validation

- Native integration coverage includes direct resource-root installation, byte-stable settings, preserved `${QODER_PLUGIN_ROOT}`, shared-cache cross-scope rejection, same-ID owned re-enable, cross-ID re-enable rejection, pre-existing ownership conflicts, failed-install/user-install retry and disable safety, schema-v2 pre-apply crash safety, post-install verification cleanup, fixed-manifest identity, inventory failures, inventory larger than 64 KiB, and uninstall verification.
- Existing Qoder Legacy tests pass unchanged.
- Qoder CLI 1.1.9 reproductions confirmed that same-ID user installs replace the active version and an alternate manifest path registers a different directory-derived ID.

## Documentation and rollback

No documentation files changed; the PR is limited to the ANOLISA driver and tests. Reverting commit `8bc15dc9` restores the prior Legacy-only Qoder behavior.

### DIFF
--- a/src/anolisa/crates/anolisa-core/src/adapter/qoder.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/qoder.rs
@@ -1,44 +1,19 @@
 //! Qoder (`qodercli`) framework driver.
 //!
-//! Qoder CLI installs a plugin from a directory whose name becomes the
-//! plugin id (`qodercli plugins install <dir>`), then activates it through
-//! two entries in `~/.qoder/settings.json`: the plugin's hooks under
-//! `.hooks`, and `<plugin>@local` under `.plugins.enabled`. ANOLISA
-//! reproduces the legacy install script's behavior entirely in this driver
-//! — it never shells out to `scripts/install.sh` / `uninstall.sh`:
+//! The driver supports two unambiguous bundle layouts. Native bundles carry
+//! `.qoder-plugin/plugin.json` plus `hooks/hooks.json`; their complete
+//! lifecycle is delegated to `qodercli plugins`, and registration,
+//! activation, and loaded hooks are verified from `plugins list --json`.
+//! ANOLISA installs the original resource root without staging or rewriting
+//! plugin files and never edits `~/.qoder/settings.json` for this mode.
 //!
-//! ```text
-//! qodercli plugins install <staging>/<plugin>   # staging = <data>/qoder-plugins
-//! # merge our hooks + <plugin>@local into ~/.qoder/settings.json in place
-//! ```
-//!
-//! The plugin bundle lives under a resource directory named `qoder`, but
-//! `qodercli` derives the plugin id from the *directory name*, so enable
-//! stages a real copy of the bundle under a directory named after the
-//! plugin id (`tokenless`) and installs from there — mirroring the legacy
-//! script's private tempdir. The staged copy's `hooks.json` is patched
-//! first: qodercli copies the plugin into its cache verbatim, and
-//! consumers that load the cached `hooks.json` directly (the Qoder IDE
-//! shares `~/.qoder` with qodercli) never expand
-//! `${QODER_TOKENLESS_HOOKS}` — staging a symlink to the raw bundle would
-//! leave them with broken hook commands whose non-zero exit is treated as
-//! a tool-call block. Staging is install-time only and is removed
-//! immediately after install.
-//!
-//! **settings.json is merged, then atomically swapped in via rename.** All
-//! reads and writes go through the Manager's controlled
-//! [`AdapterOps`](super::driver::AdapterOps); the driver only ever adds or
-//! removes ANOLISA-managed entries (the exact hook entries resolved from the
-//! bundle at enable time and persisted in the receipt, plus the
-//! `<plugin>@local` plugin entry). A settings file that exists but cannot be
-//! parsed is left untouched: enable fails closed and disable reports cleanup
-//! incomplete, so ANOLISA never clobbers a config it cannot safely merge.
-//!
-//! `qodercli plugins list` has been observed to omit freshly installed
-//! plugins, so `status` does **not** trust it: plugin registration is
-//! reported `Unknown` rather than faked healthy. The reliable, CLI-free
-//! signal is the presence of our managed entries in `settings.json`, which
-//! `status` verifies directly.
+//! Legacy bundles carry the same manifest plus a root-level `hooks.json`.
+//! That branch preserves the original Tokenless integration exactly: it
+//! stages a plugin-id-named copy, expands `${QODER_TOKENLESS_HOOKS}`, installs
+//! the copy, and atomically merges only its owned hooks and activation entry
+//! into `settings.json`. Legacy status remains settings-based because older
+//! qodercli inventories could omit freshly installed plugins, and legacy
+//! disable prunes only the exact entries persisted in its receipt.
 //!
 //! Env contract: `QODERCLI_BIN` overrides the executable (tests point it at
 //! a fake CLI); otherwise the binary is resolved in the legacy order
@@ -74,13 +49,17 @@ use settings::{
 /// Default timeout for a `qodercli` invocation.
 const CLI_TIMEOUT: Duration = Duration::from_secs(60);
 
-/// Qoder-native plugin manifest inside the bundle. The contract may override
-/// the entry, but this is the default the legacy layout ships.
+/// Qoder plugin manifest shared by native and legacy layouts.
 const QODER_PLUGIN_MANIFEST: &str = ".qoder-plugin/plugin.json";
 
-/// Hook declarations shipped alongside the plugin manifest, merged into the
-/// user's `settings.json` at enable time.
+/// Legacy hook declarations merged into the user's `settings.json`.
 const QODER_HOOKS_FILE: &str = "hooks.json";
+
+/// Qoder-native hook declarations auto-discovered by `qodercli`.
+const QODER_NATIVE_HOOKS_FILE: &str = "hooks/hooks.json";
+
+/// Native lifecycle mutations are always owned in Qoder's user scope.
+const QODER_NATIVE_SCOPE: &str = "user";
 
 /// Placeholder in `hooks.json` for the absolute hook-scripts directory,
 /// expanded to `<resource_root>/../common/hooks` before the entries are
@@ -90,6 +69,12 @@ const HOOKS_PLACEHOLDER: &str = "${QODER_TOKENLESS_HOOKS}";
 /// Resource ids used in Qoder receipts.
 const RES_PLUGIN: &str = "qoder_plugin";
 const RES_SETTINGS: &str = "qoder_settings";
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum QoderBundleKind {
+    Native,
+    Legacy,
+}
 
 /// Qoder driver. Stateless; all per-operation context arrives via
 /// [`DriverCtx`].
@@ -114,13 +99,11 @@ impl FrameworkDriver for QoderDriver {
     }
 
     fn probe_bundle(&self, resource_root: &Path, declared_entry: Option<&str>) -> bool {
-        // Mirrors read_bundle's mandatory checks: the plugin manifest (or
-        // the contract-declared entry) AND hooks.json must both be
-        // present — a manifest-only leftover is not a usable bundle.
-        resource_root
-            .join(declared_entry.unwrap_or(QODER_PLUGIN_MANIFEST))
-            .is_file()
-            && resource_root.join(QODER_HOOKS_FILE).is_file()
+        if declared_entry.is_some_and(|entry| entry != QODER_PLUGIN_MANIFEST) {
+            return false;
+        }
+        resource_root.join(QODER_PLUGIN_MANIFEST).is_file()
+            && classify_bundle(resource_root).is_ok()
     }
 
     fn detect(&self, env: &HostEnv) -> DetectResult {
@@ -161,10 +144,17 @@ impl FrameworkDriver for QoderDriver {
                 reason: "resource root does not exist or is not a directory".to_string(),
             });
         }
-        let manifest = ctx
-            .declared_bundle_entry
-            .as_deref()
-            .unwrap_or(QODER_PLUGIN_MANIFEST);
+        if let Some(entry) = ctx.declared_bundle_entry.as_deref()
+            && entry != QODER_PLUGIN_MANIFEST
+        {
+            return Err(AdapterError::BundleInvalid {
+                root: root.clone(),
+                reason: format!(
+                    "qoder bundle entry must be the native manifest '{QODER_PLUGIN_MANIFEST}', got '{entry}'"
+                ),
+            });
+        }
+        let manifest = QODER_PLUGIN_MANIFEST;
         if !root.join(manifest).is_file() {
             return Err(AdapterError::BundleInvalid {
                 root: root.clone(),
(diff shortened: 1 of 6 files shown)


kongche-jbw (commented):
Review baseline: `b826878555d0...0529df3200fa`

**[P1] Preserve ownership after a failed native enable**

`src/anolisa/crates/anolisa-core/src/adapter/qoder.rs:1192`

`prepare_enable` reads the pre-install inventory but discards whether the target already exists.
If a user-installed `name@local` plugin is present and native install or verification fails, the
write-ahead receipt is retained as `cleanup_failed`. A later disable reaches this block and
unconditionally runs `plugins uninstall ... --scope user`, deleting the pre-existing plugin.
The new failure test only proves that enable itself does not uninstall, so this follow-up path
is not covered.

Possible direction: persist pre-install ownership and refuse destructive cleanup for an adopted
or uncertain plugin. Add a regression that seeds an existing plugin, fails enable, then disables
and proves the original installation remains.

**[P1] Preserve legacy cleanup facts across a native re-enable**

`src/anolisa/crates/anolisa-core/src/adapter/qoder.rs:297`

When an installed component's Qoder bundle changes from root `hooks.json` to
`hooks/hooks.json`, the manager replaces the prior receipt before `apply_enable`, and this driver
uses the default no-op `preserve_reenable_facts`. The native receipt drops `settings_resource`
and all managed hook specs, so the legacy hooks and activation entry remain in `settings.json`;
native disable can only uninstall the plugin. Those executable hooks are then permanently
unmanaged.

Possible direction: reject cross-layout re-enable until legacy cleanup completes, or carry and
prune the legacy settings facts during the transition. Cover legacy enable -> native re-enable ->
disable with an integration test.

**[P2] Bump the driver schema for the new Qoder receipt shape**

`src/anolisa/crates/anolisa-core/src/adapter/claim.rs:594`

Making `settings_resource` optional and omitting it for native receipts adds a second serialized
`DriverPayload::Qoder` shape, but `DRIVER_SCHEMA_VERSION` remains 1 despite its documented
shape-change contract. Old and new payload generations are therefore indistinguishable to future
migration or compatibility logic.

Possible direction: bump the driver schema and add compatibility coverage for both the legacy
version-1 receipt and the new native receipt.

kongche-jbw (commented):
Review baseline: `b6a8fb57bb6c...aaf3dd761dd1`

**[P1] Track native plugin ownership per installation scope**

`src/anolisa/crates/anolisa-core/src/adapter/qoder.rs:275`

`plugins list --json` is unscoped, while Qoder permits the same plugin ID in user,
project, and local scopes. This lookup matches only `<name>@local` and drops scope.
If a project-scoped copy exists, enable marks it pre-existing, installs a user-scoped
copy, and disable then removes the receipt without uninstalling the user copy. In the
opposite direction, a project copy can also keep an owned user receipt stuck after the
user uninstall succeeds because line 1265 still sees the same ID.

Possible direction: persist and verify user-scope provenance, or reject ambiguous
inventory. Add a regression with the same plugin ID at project and user scopes.

**[P1] Accept native plugins that do not declare hooks**

`src/anolisa/crates/anolisa-core/src/adapter/qoder.rs:727`

Qoder native plugins may contain only Skills, commands, agents, or MCP resources, but
`classify_bundle` requires `hooks/hooks.json`; line 1001 also requires a non-empty hooks
inventory. A valid manifest-plus-Skills plugin is therefore rejected before `qodercli`,
and relaxing classification alone would still make post-install verification fail.

Possible direction: let Qoder CLI recognize native layouts and verify the resources the
plugin actually declares. Add a Skills-only native integration test.

kongche-jbw (commented):
Review baseline: `ed87489f05e1...bae6c33098bd`

**[P1] Reject pre-existing user plugins before replacing them**

`src/anolisa/crates/anolisa-core/src/adapter/qoder.rs:424`

`prepare_enable` records an existing user-scope plugin, but this line still installs the
bundle over it. Qoder CLI 1.1.9 accepts a second install of the same ID and changes the
active version (reproduced with 0.9.0 -> 9.9.9). `disable_native` then sees
`plugin_preexisting` and removes the receipt without uninstalling or restoring the
original plugin. Disable therefore leaves the ANOLISA bundle active and the user's
original registration lost.

Possible direction: reject a pre-existing user registration before mutation unless a
prior ANOLISA receipt proves ownership, or preserve enough provenance to restore it.
Add successful enable/disable coverage with a different pre-existing version; the
current test covers only an install failure.

**[P1] Bind native receipts to Qoder's actual plugin manifest**

`src/anolisa/crates/anolisa-core/src/adapter/qoder.rs:163`

Native ID parsing trusts `[adapters.bundle].entry`, even when it is not
`.qoder-plugin/plugin.json`. Qoder ignores such a custom manifest and, when the fixed
manifest is absent, registers the directory name instead. With Qoder CLI 1.1.9, a
`custom.json` manifest naming `agent-sec-core` installed as `custom@local`.
Post-install verification then fails, while disable looks only for the declared ID,
sees it absent, removes the receipt, and leaves the actual hook plugin active.

Possible direction: require the fixed native manifest path and matching name, or derive
and persist the actual registered ID from post-install inventory. Add a manager-level
regression proving failed enable followed by disable leaves no unexpected registration.

kongche-jbw (commented):
Review baseline: `c5b3bcf99fff...d0bd9c9bcee7`

**[P1] Bind re-enable ownership to the same Qoder plugin ID**

`src/anolisa/crates/anolisa-core/src/adapter/qoder.rs:395`

`next_payload.plugin_preexisting &= prior_preexisting` treats any prior native receipt for
the component/framework as proof that ANOLISA owns the plugin in the new bundle. If plugin A
was enabled, an upgrade changes the manifest and contract to plugin B, and B already belongs
to the user, re-enable clears B's pre-existing flag and `plugins install` replaces it. If B is
absent, the same receipt replacement loses A's cleanup claim, so disabling B leaves A
registered and unmanaged.

Possible direction: resolve both plugin resources and reject re-enable unless their IDs match,
as with a cross-layout change. Add regressions for A -> pre-existing B and A -> absent B,
proving no install occurs and the A receipt remains available for cleanup.

kongche-jbw (commented):
Review baseline: `7b01f34a1205...22c78b01ce95`

**[P1] Do not infer ownership from an uncertain failed enable**

`src/anolisa/crates/anolisa-core/src/adapter/qoder.rs:424`

When an initial enable sees no plugin, persists `plugin_preexisting=false`, and
`plugins install` fails before creating anything, the retained receipt is
`CleanupFailed`. If the user then installs the same ID and retries, current inventory
sets the next flag to true, but this `&=` resets it from the prior false value and lets
the retry replace the user's registration. Disabling the failed receipt instead also
unconditionally reaches the user-scope uninstall at line 1320, with the same data-loss
risk.

Possible direction: persist whether the install mutation actually succeeded and inherit
ownership only from that confirmed state; otherwise reject a now-present plugin. Add a
regression covering failed install -> user installs same ID -> retry and disable, proving
neither path mutates the user's plugin.

kongche-jbw (commented):
Review baseline: `7b01f34a1205...fbdd4318851f`

The previous ownership and manifest-identity findings are fixed in this revision. I found two remaining P1 ownership/compatibility issues and one P2 inventory-scaling issue; details are inline.

src/anolisa/crates/anolisa-core/src/adapter/qoder.rs
```diff
@@ -726,6 +1044,393 @@ fn managed_hook_specs(claim: &AdapterClaim) -> Option<&[QoderManagedHook]> {
...
+        })
+        .map(|item| {
+            let scope = item
+                .get("scope")
+                .and_then(Value::as_str)
+                .ok_or_else(|| format!("{expected_id} has no string 'scope' provenance"))?;
+            Ok((scope, item))
+        })
+        .collect::<Result<Vec<_>, String>>()?;
+    let mut user_plugins = matching
```
kongche-jbw, line 1118: **[P1] Do not ignore matching project/local registrations backed by the shared cache**

Filtering the inventory down to `user` scope makes a project-only plugin look absent, so enable proceeds with a user-scope install. Qoder does not isolate plugin files by scope: with an isolated HOME and the currently accepted Qoder CLI 1.1.5, I installed two different bundles with the same manifest name/version in project and user scope. Both inventory entries reported the same `~/.qoder/plugins/cache/local/<name>/<version>` installPath, and the user install replaced the marker used by the project registration. Uninstalling user scope left the project registration present but still pointing at the replacement content. Thus this path can overwrite a plugin ANOLISA does not own and cannot restore it on disable. Please reject matching non-user scopes before install (unless ownership/content equivalence can be proven), and make the cross-scope test model Qoder's shared name/version cache rather than independent per-scope content.

src/anolisa/crates/anolisa-core/src/adapter/qoder.rs
```diff
@@ -678,6 +957,46 @@ fn qoder_payload(claim: &AdapterClaim) -> Option<&QoderClaim> {
...
+    match &mut claim.driver_payload {
+        DriverPayload::Qoder(qoder) => Some(qoder),
+        _ => None,
+    }
+}
+
+/// Driver-schema-v2 Native receipts predate the explicit checkpoint, but an
+/// `Enabled` receipt could only be written after install and verification
+/// both succeeded. Failed v2 receipts remain unconfirmed and fail closed.
+fn native_install_confirmed(claim: &AdapterClaim, payload: &QoderClaim) -> bool {
```
kongche-jbw, line 984: **[P1] Treat schema-v2 `Enabled` receipts as unconfirmed ownership**

The compatibility inference here is unsafe: Manager persists the initial claim with `status = Enabled` before `apply_enable` runs (`manager.rs:780-790`). A v2 process can therefore crash after that save but before/during install, or fail to persist the later `CleanupFailed` update, leaving an `Enabled` receipt without proof that ANOLISA created the registration. If the user subsequently installs the same ID, current disable/re-enable treats that plugin as ANOLISA-owned and may uninstall or replace it. Since v2 has no durable install checkpoint, it should fail closed rather than infer confirmation from `Enabled`. Please add a regression built from a pre-apply v2 receipt, followed by an external same-ID user install, and assert that neither disable nor re-enable mutates it.

src/anolisa/crates/anolisa-core/src/adapter/qoder.rs
```diff
@@ -241,6 +252,78 @@ impl FrameworkDriver for QoderDriver {
...
+                require_cli_success(
+                    &program,
+                    "plugins validate",
+                    ctx.ops.run_framework_cli(build_native_validate_cmd(
+                        &program,
+                        &bundle.resource_root,
+                    ))?,
+                )?;
+            }
+            let inventory = ctx.ops.run_framework_cli(build_native_list_cmd(&program))?;
```
kongche-jbw: **[P2] Avoid parsing silently truncated global inventory JSON**

This new path requires the complete output of `plugins list --json`, but the shared framework runner silently keeps only the first 64 KiB of stdout (`manager.rs:54-57`, `spawn_drain`). A hook-heavy Qoder inventory entry is already about 1.8 KiB in a real CLI run, so a few dozen plugins can turn otherwise valid inventory into truncated, invalid JSON. Enable, status, and disable then report the inventory as unavailable even though Qoder succeeded. Please use an appropriate structured-output capture limit (or an explicit truncation error) for inventory commands, and cover an inventory larger than 64 KiB.

src/anolisa/crates/anolisa-core/src/adapter/qoder.rs
ikunkun-sys: Fixed in `8bc15dc9`.

The preflight now retains all matching scope provenance and rejects the enable before receipt persistence or install whenever the same `<name>@local` exists outside `user`. This applies even if an ANOLISA-owned user receipt also exists, because the shared cache makes replacement unsafe for the other registration.

The fake qodercli now models registration separately from the shared cache. The regression seeds a project-only registration and project-owned cache marker, then asserts no user install/uninstall runs, no receipt is written, and the marker is byte-for-byte unchanged.

src/anolisa/crates/anolisa-core/src/adapter/qoder.rs
ikunkun-sys: Fixed in `8bc15dc9`.

Ownership confirmation now requires both driver schema v3 or newer and the explicit `plugin_install_confirmed` checkpoint. Receipt status is no longer used as evidence, so every Native v2 receipt fails closed.

The regression constructs the pre-apply failure shape, rewrites it as a schema-v2 `Enabled` receipt, then adds a user-owned same-ID plugin. Re-enable is rejected without a second install, disable does not uninstall and keeps the receipt, and the user ownership marker remains unchanged.

src/anolisa/crates/anolisa-core/src/adapter/qoder.rs
ikunkun-sys: Fixed in `8bc15dc9`.

`AdapterOps` now has a structured-JSON runner. The production Manager captures JSON stdout with a separate bounded 4 MiB limit while stderr retains the normal 64 KiB diagnostic cap; every Qoder `plugins list --json` path (preflight, post-install verification, status, and disable verification) uses it.

The Manager-level regression emits 900 filler entries with the target after more than 64 KiB and verifies the complete enable/status/disable lifecycle, including absence verification after uninstall.

kongche-jbw (commented):
Review baseline: `1f422996030a...8bc15dc9f1f5`

未发现阻塞问题。

Remaining risks:

- 未使用真实 Qoder CLI 复测跨 scope 共享缓存；行为仍依赖 CLI 1.1.x 合约。

Verification scope:

- Reviewed the full six-file diff, adjacent lifecycle code, and every review thread.
- Local: Qoder integration 35/35, unit 25/25, fmt, diff check, and cargo doc passed.
- GitHub: ANOLISA tests, PR checks, commit lint, and component-version checks passed.

kongche-jbw (approved):

PR #2226, opened 2026-08-04
fix(cosh-ng): [shell] serialize handoffs

### LINKED ISSUE #2225: [cosh-ng] fix(cosh-ng): serialize batched shell handoffs
### Component

cosh-ng

### What happened?

Follow-up to #2151.

PR #2151 fixed token-based closure for a single secret-redacted foreground shell handoff, but multiple approved shell handoffs can still be emitted concurrently. When two executable Bash requests from one turn are covered by batch consent, the second request may be written to the PTY before the first request reaches preexec.

The second emission overwrites the singleton handoff request/token sidecars and the OSC parser's singleton pending-origin claim slot. The resulting command blocks no longer match the front of the FIFO pending queue:

- the first block is reported as `UserInteractive`, or carries the second request's token;
- the second block carries the second request's token;
- closure examines the first pending request, rejects the explicit token mismatch, and stops;
- untracked fallback also refuses closure because a `CommandStarted` event was observed.

The first request therefore remains pending with no normal terminal transition. `has_active_handoff()` suppresses the owning agent poll, so the turn and later input can remain stuck until the session is manually recovered.

Expected behavior: foreground shell handoffs are emitted one at a time. The next approved request is not staged or written until the current pending request has closed.

Observed on current upstream `main` at `b6a8fb57bb6c1994f173129c6ac970d18d9a3909`.

Relevant paths:

- Batch consent queues every covered request in one sweep:
  https://github.com/alibaba/anolisa/blob/b6a8fb57bb6c1994f173129c6ac970d18d9a3909/src/cosh-ng/crates/cosh-shell/src/approval/runtime.rs#L305-L332
- The controller calls `emit_next_approved()` without checking whether `pending` is non-empty:
  https://github.com/alibaba/anolisa/blob/b6a8fb57bb6c1994f173129c6ac970d18d9a3909/src/cosh-ng/crates/cosh-shell/src/runtime/controller.rs#L48-L60
- Each emission replaces the singleton parser claim and request/token sidecars:
  https://github.com/alibaba/anolisa/blob/b6a8fb57bb6c1994f173129c6ac970d18d9a3909/src/cosh-ng/crates/cosh-shell/src/shell_host/raw_relay/pty_emit.rs#L253-L259
- FIFO closure stops when no block matches the front request:
  https://github.com/alibaba/anolisa/blob/b6a8fb57bb6c1994f173129c6ac970d18d9a3909/src/cosh-ng/crates/cosh-shell/src/activity/shell_handoff.rs#L149-L167
- The untracked fallback defers permanently after `CommandStarted`:
  https://github.com/alibaba/anolisa/blob/b6a8fb57bb6c1994f173129c6ac970d18d9a3909/src/cosh-ng/crates/cosh-shell/src/activity/shell_handoff.rs#L14-L32

The multi-request state was previously identified during #2151 review:
https://github.com/alibaba/anolisa/pull/2151#discussion_r3703445450

### How can we reproduce it?

Minimal controller regression:

1. Create two `ShellHandoffRequest` values.
2. Enqueue both with `enqueue_approved_request()`.
3. Call `render_raw_inline_events()` once. It returns `EmitToPty(first)`.
4. Before supplying a preexec/completion event for the first request, call `render_raw_inline_events()` again.
5. Assert that the second call remains `RawPassthrough`.

Current result:

```text
assertion failed
  actual:   EmitToPty(second)
  expected: RawPassthrough
```

Command used for the temporary targeted regression:

```shell
cargo test --package cosh-shell --bins \
  pending_shell_handoff_serializes_a_second_approved_request -- --nocapture
```

Two existing tests confirm the terminal stuck-state behavior once the overlap occurs:

```shell
cargo test --package cosh-shell --bins \
  block_with_another_requests_token_does_not_close_the_front_handoff -- --nocapture

cargo test --package cosh-shell --bins \
  untracked_shell_handoff_defers_to_tracked_command_path -- --nocapture
```

Both pass: the mismatched block leaves the front request pending, and the untracked fallback leaves it pending after `CommandStarted`.

A user-facing reproduction is a provider turn containing two non-high-risk Bash calls followed by **Allow all this turn**. The bug is timing-dependent at the PTY boundary: a relay callback caused by the first command's echo can emit the second request before the first preexec marker is consumed.

### Diagnostic output

```shell
running 1 test

thread 'runtime::controller::tests::pending_shell_handoff_serializes_a_second_approved_request' panicked:
assertion `left == right` failed
  left: EmitToPty(ShellHandoffRequest { command: "echo second", ... })
 right: RawPassthrough

test result: FAILED. 0 passed; 1 failed
```

### Anything else?

Suggested invariant and acceptance criteria:

- At most one emitted-but-unclosed shell handoff exists: `pending.len() <= 1`.
- `emit_next_approved()` leaves requests in `approved` while `pending` is non-empty.
- Closing the current handoff allows the next request to emit in the same or next dispatch.
- Add real bash and zsh PTY regressions for:
  - two different commands;
  - two identical commands;
  - two secret-bearing commands whose marker text is redacted.
- Each result is delivered exactly once and in request order.
- After both results are delivered, follow-up user input starts a new agent run.

kongche-jbw:
## Why

PR #2151 introduced one-time claim tokens for shell handoff closure, but the runtime could still emit multiple approved handoffs in one batch. Because the transport and parser each retain only one active claim, later emissions could overwrite the first handoff state and leave the batch statically blocked.

## What changed

Serialize approved shell handoffs so the next request remains queued until the current foreground handoff closes. Add controller, activity, and raw CLI regressions covering ordered completion and post-batch follow-up turns on bash and zsh.

## Related issue

Closes #2225

## User / Agent impact

Batched shell calls complete in order without losing the active handoff claim or blocking later agent turns.

## Risk and compatibility

Low risk. The change narrows handoff scheduling to match the existing single-active-claim transport contract; public CLI and configuration remain unchanged.

## Validation

- `cargo test --package cosh-shell --bins shell_handoff -- --nocapture`
- `cargo test --package cosh-shell --test raw_cli turn_batch_consent -- --nocapture`

## Documentation and rollback

No documentation changes are required. Revert commit `b22c1593` to restore the previous scheduling behavior.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/activity/runtime_tests.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/activity/runtime_tests.rs
@@ -2579,12 +2579,9 @@ fn untracked_shell_handoff_closure_leaves_interactive_handoffs_untouched() {
 }
 
 #[test]
-fn untracked_shell_handoffs_share_prompt_boundary_when_emitted_in_same_cycle() {
-    // Boundary contract: handoffs emitted within the same prompt cycle (no
-    // ShellReady between their emission points) legitimately share the next
-    // ShellReady as their closure boundary. Forcing a distinct boundary per
-    // handoff would pin later handoffs to unrelated future prompts and
-    // misreport the evidence timeline.
+fn untracked_shell_handoffs_close_on_distinct_prompt_boundaries() {
+    // The transport owns one request sidecar and claim slot, so the second
+    // request remains approved until the first closes at its prompt boundary.
     let mut state = InlineState::default();
     for (command, approval_id) in [("ls -la /root/", "req-a"), ("df -h", "req-b")] {
         let request = ShellHandoffRequest::new(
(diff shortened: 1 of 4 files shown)


SunnyQjm (approved):
LGTM

PR #2237, opened 2026-08-05
feat(skillfs): allow skill meta passthrough

### LINKED ISSUE #2042: [skillfs] feat: support .skill-meta passthrough without Ledger integration
### Component

skillfs

## Summary

Extend SkillFS compatibility so that `.skill-meta` can be passed through as an
ordinary directory when SkillFS and Skill Ledger coexist without integrating
with or being aware of each other.

SkillFS currently protects `.skill-meta` regardless of whether Ledger
integration is active. That behavior is appropriate when the two components
cooperate, but it prevents Skill Ledger from independently managing a skill
directory exposed through the SkillFS FUSE view.

## Motivation

SkillFS and Skill Ledger need to support two coexistence models.

### Integrated mode

- Skill Ledger resolves the raw backing directory through SkillFS.
- Skill Ledger manages `.skill-meta` through that raw backing path.
- SkillFS consumes activation or snapshot metadata and protects `.skill-meta`
  in the FUSE view.

### Zero-integration coexistence mode

- SkillFS and Skill Ledger are both deployed but are unaware of each other.
- No resolver, control-socket interaction, or raw-backing integration is
  enabled.
- Skill Ledger may manage a skill directory exposed through the SkillFS FUSE
  view.
- `.skill-meta` belongs to Skill Ledger, while SkillFS treats it as ordinary
  filesystem content.

SkillFS does not currently support the second model. It hides `.skill-meta`
from read and directory operations and rejects mutations, so Ledger cannot
create, inspect, or update its metadata through the FUSE view.

## Proposed behavior

Determine the effective `.skill-meta` policy when the SkillFS mount starts:

```text
protect_skill_meta =
    active_resolver_enabled
    OR trusted_writer_configured
```

| Active resolver | Trusted writer | `.skill-meta` behavior |
|---|---|---|
| Disabled | Not configured | Pass through as an ordinary directory |
| Enabled | Not configured | Protect; Ledger uses the resolver and raw backing path |
| Enabled | Configured | Protect; trusted callers may use the existing FUSE exception |
| Disabled | Configured | Protect; retain direct-FUSE trusted-writer compatibility |

The selected policy must remain fixed for the lifetime of the mount. It must
not change dynamically because a socket connects, disconnects, or becomes
unavailable.

All read, lookup, listing, and mutation paths must use the same effective
policy so that `.skill-meta` cannot become writable but invisible, or visible
but unwritable.

## Compatibility and security impact

This is an intentional change to the default `.skill-meta` semantics.

When neither an active resolver nor a trusted writer is configured, SkillFS
will no longer provide its current implicit hiding and write protection for
`.skill-meta`. It will pass the directory through, and access control will be
determined by the backing filesystem's POSIX permissions.

Deployments that currently rely on implicit `.skill-meta` protection while
enabling neither capability must explicitly enable a protection-capable
configuration after upgrading.

No independent `skill_meta_mode` option is proposed initially. The absence of
both integration signals is intentionally interpreted as zero-integration
coexistence mode.

## Implementation scope

The effective protection state must apply consistently to `.skill-meta`
located directly under each logical skill root, and to all of its descendants,
for both flat and Hermes layouts.

It must cover every FUSE operation that can observe or affect those paths,
including:

- lookup, `getattr`, directory listing, open, and read;
- create, `mkdir`, write, unlink, and `rmdir`;
- `access` and `setattr`;
- extended-attribute operations;
- hard-link and symlink operations;
- rename checks for both source and destination paths.

Current implementation areas include:

- `src/skillfs/crates/skillfs-fuse/src/fs/mod.rs`
- `src/skillfs/crates/skillfs-fuse/src/fs/policy.rs`
- `src/skillfs/crates/skillfs-fuse/src/security/policy.rs`

Changing only the mutation policy is insufficient because read and listing
paths currently enforce separate hiding behavior.

## Acceptance criteria

- [ ] With no active resolver and no trusted writer, `.skill-meta` behaves like
      an ordinary directory through the FUSE mount.
- [ ] In that mode, Skill Ledger can complete its metadata lifecycle through
      the FUSE path, including creation, discovery, reads, updates, renames,
      and deletion.
- [ ] Passthrough behavior covers `access`, `setattr`, xattrs, hard links,
      symlinks, and both sides of rename operations.
- [ ] With an active resolver, `.skill-meta` remains hidden and protected from
      ordinary FUSE callers.
- [ ] With a trusted writer configured, existing trusted and untrusted caller
      behavior is preserved.
- [ ] Read, listing, and mutation paths use one effective protection decision.
- [ ] The decision is made at mount startup and does not change at runtime.
- [ ] Existing Current, Snapshot, and Hidden behavior remains unchanged in
      protected mode with an active resolver.
- [ ] Passthrough mode performs no activation filtering because it has no
      active resolver.
- [ ] Automated tests cover all four resolver/trusted-writer combinations for
      both flat and Hermes skill layouts.

## Related work

- #1115 established protected trusted-process access to live `.skill-meta`.
  This issue preserves that behavior in protected mode and adds the independent
  coexistence case.
- #1233 and #1235 define the resolver-based integrated mode between SkillFS
  and Skill Ledger.
- #2028 tracks the standalone SkillFS sidecar baseline on ACS. Cross-container
  Skill Ledger integration remains separate from this filesystem behavior.

## Non-goals

This issue does not address:

- control-socket communication or peer authentication between an ACS main
  container and a SkillFS sidecar;
- sharing or translating raw backing paths across containers;
- resolver fallback, routing, or multiple SkillFS instances;
- reverse notifications from SkillFS to the sec-core daemon;
- dynamic policy switching after the mount has started;
- requiring SkillFS and Skill Ledger to discover or become aware of each other
  in zero-integration mode.

kongche-jbw:
## Why

SkillFS currently hides and protects `.skill-meta` even when no Skill Ledger
integration is configured. That prevents an independently deployed Ledger from
managing metadata through the SkillFS FUSE view in zero-integration coexistence
mode.

## What changed

- Derive one mount-scoped metadata policy from the presence of an active
  resolver or enabled trusted writer.
- Treat `.skill-meta` as ordinary passthrough content when both integration
  signals are absent, consistently across reads, listings, mutations, links,
  xattrs, and both rename endpoints.
- Preserve protected Current, Snapshot, Hidden, trusted-writer, and explicitly
  injected policy behavior for flat and Hermes layouts.
- Keep Hermes in-place link classification on the fd-bypass path and revalidate
  Hidden state at both link endpoints.
- Enable kernel permission checks for `allow_other` mounts so backing POSIX
  permissions remain authoritative across UIDs.
- Add FUSE regression coverage for all configuration combinations, metadata
  lifecycle operations, links, xattrs, resolver state changes, and real
  source-equals-mountpoint Hermes topology.

## Related issue

closes #2042

## User / Agent impact

Without an active resolver or enabled trusted writer, agents and independently
deployed Ledger processes now see `.skill-meta` as ordinary filesystem content
through SkillFS. Integrated deployments retain the existing hidden and
protected metadata view.

## Risk and compatibility

This intentionally changes the default metadata behavior when neither
integration capability is configured. Deployments relying on implicit
`.skill-meta` protection must enable an active resolver or trusted writer.
`--security-mode` selects in-place topology only and does not enable metadata
protection by itself. The selected policy is fixed for the mount lifetime.

## Validation

- `cargo +1.86.0 fmt --all -- --check`
- `cargo +1.86.0 clippy --workspace --all-targets -- -D warnings`
- `cargo +1.86.0 test --workspace`
- `scripts/test.sh` with real FUSE and managed-mount smoke coverage
- focused link/FIFO tests: 23/23 passed
- metadata view and lifecycle tests: 23/23 passed
- xattr tests: 11/11 passed
- security-mode runtime tests: 8/8 passed
- `cargo +1.86.0 doc --workspace --no-deps`
- `git diff --check`

The root-only multi-UID `allow_other` regression is included; it skips when
the local test runner is not root.

## Documentation and rollback

The bilingual README and user guide, CLI help/runtime diagnostics, and
capability record now document the passthrough/protected mode matrix and
migration requirement. Operators can retain the former protected behavior by
configuring an active resolver or trusted writer. Reverting this commit
restores unconditional metadata protection.

### DIFF
--- a/src/skillfs/crates/skillfs-fuse/src/fs/callbacks/link.rs
+++ b/src/skillfs/crates/skillfs-fuse/src/fs/callbacks/link.rs
@@ -16,6 +16,25 @@ use crate::symlink_policy;
 use crate::sys::errno;
 
 impl SkillFs {
+    fn should_reject_hidden_link_endpoint(&self, path_type: &PathType) -> bool {
+        match path_type {
+            PathType::Passthrough {
+                skill_name,
+                relative_path,
+            } => self.should_reject_hidden_write(skill_name, Some(relative_path)),
+            PathType::NestedPassthrough {
+                category,
+                skill_name,
+                relative_path,
+            } => self.should_reject_hermes_nested_hidden_write(
+                category,
+                skill_name,
+                Some(relative_path),
+            ),
+            _ => false,
+        }
+    }
+
     pub(in crate::fs) fn readlink_impl(&mut self, req: &Request, ino: u64, reply: ReplyData) {
         debug!(ino, "readlink");
         let path = match self.inodes.get_path(ino) {
@@ -55,6 +74,15 @@ impl SkillFs {
                 skill_name,
                 relative_path,
             } => {
+                if let Some(false) = self.is_trusted_skill_meta_access(
+                    &PathType::Passthrough {
+                        skill_name: skill_name.clone(),
+                        relative_path: relative_path.clone(),
+                    },
+                    req,
+                ) {
+                    return reply.error(libc::ENOENT);
+                }
                 if is_skill_discover_path(&skill_name) {
                     // skill-discover virtual namespace contains no symlinks.
                     self.emit_event(
@@ -101,6 +129,15 @@ impl SkillFs {
                 skill_name,
                 relative_path,
             } => {
+                if let Some(false) = self.is_trusted_skill_meta_access(
+                    &PathType::InboxPassthrough {
+                        skill_name: skill_name.clone(),
+                        relative_path: relative_path.clone(),
+                    },
+                    req,
+                ) {
+                    return reply.error(libc::ENOENT);
+                }
                 if !Self::is_inbox_skill_name_allowed(&skill_name) {
                     self.emit_event(
                         SkillEvent::new(SkillEventKind::Readlink)
@@ -152,17 +189,39 @@ impl SkillFs {
             | PathType::CategoryPassthrough {
                 name,
                 relative_path,
-            }
-            | PathType::NestedPassthrough {
-                category: name,
-                skill_name: _,
-                relative_path,
             } => {
                 let physical = match self.resolve_physical_path(&path) {
                     Some(p) => p,
                     None => return reply.error(libc::ENOENT),
                 };
                 let _ = (&name, &relative_path);
+                match std::fs::read_link(&physical) {
+                    Ok(target) => {
+                        use std::os::unix::ffi::OsStrExt;
+                        reply.data(target.as_os_str().as_bytes());
+                    }
+                    Err(e) => reply.error(errno(&e)),
+                }
+            }
+            PathType::NestedPassthrough {
+                category,
+                skill_name,
+                relative_path,
+            } => {
+                if let Some(false) = self.is_trusted_skill_meta_access(
+                    &PathType::NestedPassthrough {
+                        category: category.clone(),
+                        skill_name: skill_name.clone(),
+                        relative_path: relative_path.clone(),
+                    },
+                    req,
+                ) {
+                    return reply.error(libc::ENOENT);
+                }
+                let physical = match self.resolve_physical_path(&path) {
+                    Some(p) => p,
+                    None => return reply.error(libc::ENOENT),
+                };
                 match std::fs::read_link(&physical) {
                     Ok(target) => {
                         use std::os::unix::ffi::OsStrExt;
@@ -203,28 +262,45 @@ impl SkillFs {
         let path_type = self.parse_fuse_path(Path::new(&path_str));
         let target_str = target.display().to_string();
 
-        // Only Passthrough leaves under an ordinary skill may host a new
-        // symlink. Virtual paths keep their existing virtual semantics,
-        // which means SymlinkDir / SymlinkMd / Root / SkillsDir / Invalid
-        // remain EROFS as in S0.
-        let (skill_name, relative_path) = match &path_type {
-            PathType::Passthrough {
-                skill_name,
-                relative_path,
-            } => (skill_name.clone(), relative_path.clone()),
-            _ => {
-                self.ro_warn("symlink", &path_str);
-                self.emit_event(
-                    SkillEvent::new(SkillEventKind::SymlinkAttempt)
-                        .with_action(SkillEventAction::Rejected)
-                        .with_errno(libc::EROFS)
-                        .with_caller(req.uid(), req.gid())
-                        .with_detail(format!("class=virtual_link target={}", target_str)),
-                );
-                reply.error(libc::EROFS);
-                return;
-            }
-        };
+        // Only passthrough leaves under a real skill may host a new symlink.
+        // Virtual paths keep their existing virtual semantics, which means
+        // SymlinkDir / SymlinkMd / Root / SkillsDir / Invalid remain EROFS.
+        // Hermes nested skills use `category/skill` as their logical ID while
+        // retaining the nested physical root for boundary classification.
+        let (skill_name, relative_path, classification_root, classification_skill) =
+            match &path_type {
+                PathType::Passthrough {
+                    skill_name,
+                    relative_path,
+                } => (
+                    skill_name.clone(),
+                    relative_path.clone(),
+                    self.source_base(),
+                    skill_name.clone(),
+                ),
+                PathType::NestedPassthrough {
+                    category,
+                    skill_name,
+                    relative_path,
+                } => (
+                    Self::hermes_skill_id(category, skill_name),
+                    relative_path.clone(),
+                    self.source_base().join(category),
+                    skill_name.clone(),
+                ),
+                _ => {
+                    self.ro_warn("symlink", &path_str);
+                    self.emit_event(
+                        SkillEvent::new(SkillEventKind::SymlinkAttempt)
+                            .with_action(SkillEventAction::Rejected)
+                            .with_errno(libc::EROFS)
+                            .with_caller(req.uid(), req.gid())
+                            .with_detail(format!("class=virtual_link target={}", target_str)),
+                    );
+                    reply.error(libc::EROFS);
+                    return;
+                }
+            };
 
         // skill-discover is virtual and read-only regardless of the
         // physical layout, so refuse before any classifier work.
@@ -239,6 +315,14 @@ impl SkillFs {
                     .with_detail(format!("class=skill_discover target={}", target_str)),
             );
             reply.error(libc::EROFS);
+            return;
+        }
+
+        // A parent directory handle may outlive an activation transition.
+        // Recheck Hidden immediately before every link mutation instead of
+        // relying on the lookup that originally produced the handle.
+        if self.should_reject_hidden_link_endpoint(&path_type) {
+            reply.error(libc::ENOENT);
             return;
         }
 
@@ -304,23 +388,44 @@ impl SkillFs {
         // space sees when it constructs an absolute target, so we use it
         // for both. Relative targets are resolved against the parent of
         // the link path.
-        let source_root = self
-            .source
+        // Enumerate Hermes siblings through `source_base()` before
+        // canonicalization. In an in-place mount this stays on the pre-opened
+        // `/proc/self/fd/N` path and cannot recurse into the FUSE callback.
+        // Only real Skill leaves participate in cross-skill classification;
+        // ordinary category directories remain outside the Skill namespace.
+        let known_skill_names = if let PathType::NestedPassthrough { category, .. } = &path_type {
+            std::fs::read_dir(&classification_root)
+                .into_iter()
+                .flatten()
+                .filter_map(|entry| {
+                    let entry = entry.ok()?;
+                    let name = entry.file_name().into_string().ok()?;
+                    self.hermes_nested_is_skill(category, &name).then_some(name)
+                })
+                .collect::<Vec<_>>()
+        } else {
+            let store_guard = self.store.read();
+            store_guard
+                .list()
+                .into_iter()
+                .map(str::to_owned)
+                .collect::<Vec<_>>()
+        };
+        let source_root = classification_root
             .canonicalize()
-            .unwrap_or_else(|_| self.source.clone());
+            .unwrap_or_else(|_| classification_root.clone());
         let link_parent_for_classifier = source_root
-            .join(&skill_name)
+            .join(&classification_skill)
             .join(relative_path.parent().unwrap_or(Path::new("")));
-        let store_guard = self.store.read();
-        let known_skill_names: Vec<&str> = store_guard.list();
+        let known_skill_name_refs: Vec<&str> =
+            known_skill_names.iter().map(String::as_str).collect();
         let class = symlink_policy::classify_symlink_target(
             &source_root,
-            &skill_name,
-            &known_skill_names,
+            &classification_skill,
+            &known_skill_name_refs,
             &link_parent_for_classifier,
             target,
         );
-        drop(store_guard);
 
         let class_label = symlink_policy::symlink_class_label(&class);
         if !matches!(class, symlink_policy::SymlinkTargetClass::SameSkill) {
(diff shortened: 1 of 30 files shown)


src/skillfs/crates/skillfs-fuse/src/fs/policy.rs
```diff
@@ -391,6 +398,9 @@ impl SkillFs {
         if !is_skill_meta_path(relative_path) {
             return None;
         }
+        if !self.protects_skill_meta() {
+            return Some(true);
```
1570005763, line 405: [P1] Preserve backing POSIX permissions under `allow_other`

This new passthrough return makes `.skill-meta/**` reachable by ordinary FUSE callbacks in no-integration mode. The mount can enable `AllowOther` without `DefaultPermissions`, while those callbacks perform backing I/O as the SkillFS daemon rather than as `req.uid()`/`req.gid()`. As a result, the existing passthrough permission weakness now reaches security-sensitive metadata: another UID may fail `access(2)` but still open, overwrite, or unlink a daemon-readable file. That contradicts #2042's requirement that the backing filesystem's POSIX permissions determine access. Please enable kernel permission checks or enforce request credentials on every relevant callback, and add a multi-UID `allow_other` regression test.

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/link.rs
```diff
@@ -304,23 +361,43 @@ impl SkillFs {
...
-            .unwrap_or_else(|_| self.source.clone());
+            .unwrap_or(classification_root);
         let link_parent_for_classifier = source_root
-            .join(&skill_name)
+            .join(&classification_skill)
             .join(relative_path.parent().unwrap_or(Path::new("")));
-        let store_guard = self.store.read();
-        let known_skill_names: Vec<&str> = store_guard.list();
+        let known_skill_names = if matches!(&path_type, PathType::NestedPassthrough { .. }) {
+            std::fs::read_dir(&source_root)
```
1570005763: [P1] Keep Hermes in-place enumeration on the fd-bypass path

For a real in-place mount, `source_base()` is `/proc/self/fd/N` specifically to bypass the FUSE overmount. Canonicalizing `/proc/self/fd/N/<category>` produces the ordinary mounted pathname, and this `read_dir` then re-enters the same FUSE mount from inside its symlink callback. `fuser::mount2` uses a single receive/dispatch loop, so the nested request cannot be serviced while the original callback is waiting; an otherwise valid Hermes `ln -s` can hang the syscall and the mount. The new fixture passes `in_place=true` with separate source and mountpoint directories, so it does not exercise this topology. Please keep the scan fd-relative and add a real source==mountpoint regression test.

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/link.rs
```diff
@@ -203,28 +243,45 @@ impl SkillFs {
...
+                PathType::Passthrough {
+                    skill_name,
+                    relative_path,
+                } => (
+                    skill_name.clone(),
+                    relative_path.clone(),
+                    self.source_base(),
+                    skill_name.clone(),
+                ),
+                PathType::NestedPassthrough {
```
1570005763, line 281: [P1] Recheck Hidden state before nested link mutations

This new `NestedPassthrough` symlink path never calls `should_reject_hermes_nested_hidden_write`; the new nested hardlink source and destination branches have the same omission. A caller can open the directory while the skill is Current, transition the resolver to Hidden, and then use the still-valid dirfd with `symlinkat` to mutate the live backing tree. Other nested mutation callbacks correctly return `ENOENT` in this state. The flat link path already lacked this check, but this PR extends the gap to Hermes while promising to preserve Hidden behavior. Please gate both nested link endpoints, including both hardlink sides, and cover the Current-to-Hidden transition.

src/skillfs/crates/skillfs-fuse/src/fs/policy.rs
```diff
@@ -79,6 +79,13 @@ impl SkillFs {
             _ => return None,
         };
 
+        // In the no-integration mount mode `.skill-meta/**` is an ordinary
+        // physical subtree. Bypass only this reserved-metadata gate; other
+        // lifecycle and path-safety checks remain in their callers.
+        if !self.protects_skill_meta() && is_skill_meta_path(relative_path) {
+            return None;
```
1570005763, line 89: [P1] Do not bypass an explicitly injected `SecurityPolicy`

This return occurs before `policy_check`, so a custom policy installed through `with_policy` or `MountConfig.policy` is never consulted for no-integration `.skill-meta` mutations. For example, an embedder's deny-delete or deny-all policy still applies to ordinary passthrough paths but is silently ignored here. Passthrough mode may disable the default metadata-protection rule, but it should not bypass every caller-supplied policy. Please preserve the policy call or distinguish the default `SkillMetaProtectionPolicy` from an explicit override.

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/xattr.rs
```diff
@@ -29,6 +29,9 @@ impl SkillFs {
         if !path_type_supports_xattr_passthrough(&path_type) {
             return reply.error(libc::EOPNOTSUPP);
         }
+        if let Some(false) = self.is_trusted_skill_meta_access(&path_type, req) {
```
1570005763, line 32: [P1] Make Hermes metadata xattrs reach this policy gate

This new check is unreachable for `PathType::NestedPassthrough`: the preceding `path_type_supports_xattr_passthrough` helper still accepts only flat `PathType::Passthrough`, so every Hermes get/list/set/remove xattr operation returns `EOPNOTSUPP` first. This limitation predates the PR, but the PR claims xattr passthrough for both layouts and #2042 explicitly requires it. Please extend the helper for nested skills and add Hermes passthrough plus protected-mode xattr coverage.

src/skillfs/crates/skillfs-fuse/src/fs/mod.rs
```diff
@@ -229,6 +238,7 @@ impl SkillFs {
             refresh_controller: None,
             notify_controller: None,
             trusted_writer: TrustedWriterConfig::disabled(),
+            skill_meta_mode: SkillMetaMode::Passthrough,
```
1570005763, line 246: [P1] Update the shipped security-mode contract for the new default

This default is consistent with #2042: `--security-mode` alone is a topology guarantee, not an integration signal, so it should not be added to `protect_skill_meta`. However, the shipped CLI help still says `--security-mode --audit-log` provides `.skill-meta protection`, and the English/Chinese README and capability guidance still describe metadata as unconditionally hidden and mutation-protected. Operators following those instructions can upgrade into writable metadata without realizing it. The PR marks documented and security-sensitive behavior as changed and migration guidance as required, but changes no public docs. Please update the CLI help, runtime warning, bilingual docs/capability record and migration guidance, and pin `security-mode` without resolver/writer as passthrough in an operator-facing test.

1570005763 (changes requested):
The no-integration passthrough direction matches #2042, but the six inline findings below should be addressed before merge. Findings 1 and 3 are pre-existing gaps newly exposed or extended by this PR; finding 5 is an unfulfilled acceptance criterion; findings 2 and 4 are newly introduced regressions; finding 6 is a security-sensitive documentation and migration contract mismatch.

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/link.rs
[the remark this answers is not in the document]
kongche-jbw: Fixed in e23c626c0. Hermes sibling enumeration now stays on the fd-backed classification root, runs before canonicalization, and filters every candidate through `hermes_nested_is_skill` (a real directory with a regular `SKILL.md`). This intentionally follows the set of physical leaves the mount can expose, while excluding ordinary category directories and avoiding the in-place FUSE recursion.

src/skillfs/crates/skillfs-fuse/tests/posix_xattr_tests.rs
```diff
@@ -322,6 +323,43 @@ fn test_user_xattr_set_get_list_remove_on_passthrough_file() {
     );
 }
 
+#[test]
```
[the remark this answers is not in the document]
kongche-jbw: Confirmed intentional and documented in e23c626c0. With neither an active resolver nor an enabled trusted writer, `.skill-meta/**` is ordinary POSIX passthrough, including `user.*` xattrs. Either integration signal selects protected mode, where untrusted metadata access/mutation remains hidden or denied. The bilingual user guide and capability record now include this mode matrix and migration guidance.

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/link.rs
kongche-jbw: Fixed in e23c626c0. Hermes sibling enumeration now performs `read_dir` on the pre-opened `/proc/self/fd/N/<category>` path before any canonicalization. A real source==mountpoint Hermes regression creates and reads a metadata symlink and completes without recursive FUSE entry.

src/skillfs/crates/skillfs-fuse/tests/trusted_skill_meta_view_tests.rs
```diff
@@ -980,12 +1097,7 @@ fn hermes_trusted_meta_symlink_hardlink_rejected() {
         .join("link-to-regular");
     let err = std::os::unix::fs::symlink("../regular.txt", &link_path)
```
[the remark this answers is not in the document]
kongche-jbw: Confirmed intentional. `EACCES` is the canonical denial for protected `.skill-meta` link mutations; `EROFS` is reserved for virtual/read-only namespaces and was only an artifact of the previously unsupported nested endpoint. No published contract distinguishes the old mixed errno result, and the protected flat/Hermes tests now pin the consistent policy denial.

src/skillfs/crates/skillfs-fuse/tests/common.rs
```diff
@@ -148,6 +148,55 @@ impl MountFixture {
         Self::mount_now(MountMode::Normal, source, Some(mountpoint))
     }
 
+    /// Mount in normal mode with an active resolver so reserved metadata stays
+    /// protected while the skill remains visible as `Current`.
+    pub fn normal_with_active_resolver<F: FnOnce(&Path)>(seed: F) -> Self {
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in e23c626c0. The shared FUSE fixture was moved from `tests/common/mod.rs` to the Rust 2018 `tests/common.rs` layout, and explicit `#[path]` users were updated accordingly.

src/skillfs/crates/skillfs-fuse/src/fs/policy.rs
kongche-jbw: Fixed in e23c626c0. `allow_other` now always adds FUSE `DefaultPermissions`, so the kernel checks the projected backing mode/uid/gid before daemon-credential callbacks run. A root-only multi-UID regression verifies UID 65534 cannot overwrite a mode-0600 `.skill-meta` file; it skips explicitly on non-root runners.

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/link.rs
kongche-jbw: Fixed in e23c626c0. Symlink mutations recheck the destination endpoint immediately before policy/mutation, and hardlinks recheck both source and destination endpoints. The helper covers flat and Hermes paths. A real in-place Current→Hidden stale-dirfd regression verifies both `symlinkat` and `linkat` return `ENOENT`.

src/skillfs/crates/skillfs-fuse/src/fs/policy.rs
kongche-jbw: Fixed in e23c626c0. Passthrough mode bypasses only the built-in `SkillMetaProtectionPolicy`; `with_policy`/`MountConfig.policy` marks an explicit override and continues through `policy_check`. The regression injects a custom metadata-create denial and verifies its errno and no backing mutation.

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/xattr.rs
kongche-jbw: Fixed in e23c626c0. `path_type_supports_xattr_passthrough` now accepts `NestedPassthrough`. New Hermes tests cover the full passthrough set/get/remove round trip and confirm protected-mode metadata xattrs remain hidden with `ENOENT`.

src/skillfs/crates/skillfs-fuse/src/fs/mod.rs
kongche-jbw: Fixed in e23c626c0. CLI help and runtime diagnostics now state that `--security-mode` is topology-only and warn when `.skill-meta` remains passthrough. The English/Chinese README and user guide, capability record, and migration guidance describe the integration-gated mode matrix. An in-place security-mode regression pins writable metadata when no resolver/writer is configured.

PR #2239, opened 2026-08-05
feat(sec-core): gate prompt scanner by env

haosanzi:
## Risk and compatibility


edonyzpc (approved):
LGTM

PR #2352, opened 2026-08-10
fix(cosh-ng): [shell] narrow interactive-cancel needles

### LINKED ISSUE #2090: [TestAgent][cosh-ng] bug: interactive_cancel_output 的 interrupted 等过宽 needles 掩盖真实失败
### Component

cosh-ng

### Bug Description

`interactive_cancel_output` 用一组过宽的 needles(如 `"interrupted"`、`"access key"`、`"default region id"`、`"operation cancelled"` 等)做子串匹配,把任何 stdout/stderr 含这些子串的 exit_code=1 命令都判定为"用户交互式取消"(`command_should_skip_failure_analysis` 返回 true),从而跳过失败分析。这掩盖了真实失败:任何含 "interrupted" 的命令输出(如 `grep -q interrupted file; exit 1`、`sed '/interrupted/d'`、网络超时"ECONNINTERRUPTED" 等)都被误判为用户取消,失败原因不上报。

### Steps to Reproduce

```shell
ssh root@<ECS-IP>
# 静态确认:看 needles 列表
sed -n '52,68p' src/cosh-ng/crates/cosh-shell/src/hooks/interrupt.rs
```

> 注:本 bug 为静态源码确认,动态复现需 `hooks::interrupt` 模块编译进 test 二进制(本仓 build 配置下 `mod.rs` 未加载 `interrupt`,见测试报告关键发现 6)。

### Actual Behavior

`src/cosh-ng/crates/cosh-shell/src/hooks/interrupt.rs:52-68`:

```rust
fn interactive_cancel_output(output: &str) -> bool {
    let normalized = output.to_ascii_lowercase();
    [
        "sudo: a password is required",
        "a password is required",
        "a terminal is required",
        "operation cancelled",
        "operation canceled",
        "keyboardinterrupt",
        "interrupted",
        "access key id",
        "access key secret",
        "default region id",
    ]
    .iter()
    .any(|needle| normalized.contains(needle))
}
```

`normalized.contains(needle)` 是子串匹配,`"interrupted"` 会命中任何含这词的命令输出(无论是否真取消)。命中后 `command_looks_like_interactive_cancel`(line 25-40)返回 true → `command_should_skip_failure_analysis`(line 7-12)返回 true → 失败分析被跳过,真实失败原因不上报。

### Expected Behavior

cancel 检测 needles 应精确匹配(整行/带边界),或加 exit_code + 时序约束(如 Ctrl-C 事件落 Block 区间内);不应单纯用子串匹配掩盖真实失败。

### Root Cause

`interrupt.rs:52-68` 用 `normalized.contains(needle)` 子串匹配,needles 含 `"interrupted"`、`"access key"`、`"default region id"` 等通用词,任何含这些词的 stdout/stderr + exit_code=1 都被误判为用户取消。触发条件:命令输出含上述任一子串 + exit_code=1。设计取舍:无文档说明 needles 为何这么宽;非 by-design,是 needle 选择过宽的疏漏。

### Suggested Fix

needles 改精确匹配(如 `normalized.lines().any(|line| line.trim() == needle)` 或加 word boundary);或加 `command_has_user_interrupt_event` 双门槛(必须同时有 Ctrl-C 事件落在 Block 区间内);把 `"interrupted"` 等过宽 needle 改为更具体的失败签名(如 `"^interrupted$"` 顶格匹配)。

### Environment

- OS: Alibaba Alinux 4, kernel 6.6.102
- Component: cosh-ng
- Package/RPM: cosh-ng-0.13.0-2.alnx4

### Additional Context

- 相关 issue/PR: 无
- 稳定性: 静态源码确认(动态复现需 `hooks::interrupt` 模块编译进 test 二进制)
- 被测版本 main HEAD: a96622bd

SunnyQjm:
## Summary

`interactive_cancel_output` in `crates/cosh-shell/src/hooks/interrupt.rs` matched a set of overly broad needles with plain substring `contains`: any `exit_code=1` command whose output merely contained `"interrupted"`, `"access key id"`, `"access key secret"` or `"default region id"` was misclassified as a user-interactive cancel, so failure analysis was silently skipped for real failures (e.g. `curl: (18) transfer was interrupted by peer`, `InvalidAccessKeyId` errors).

This PR tiers the needles by specificity instead of removing them, preserving every existing by-design detection.

Closes #2090

## Changes

- The cancel matcher is converged into a single shared `pub(crate) fn interactive_cancel_output` in `command/failure_semantics.rs` (review round 1: `classify_failure` had a second wide-substring copy which still classified these outputs as `InteractiveCancel`); `hooks/interrupt.rs` now reuses it via the hooks prelude.
- The shared matcher normalizes its input itself (review round 2: ANSI strip + lowercase via the module's `normalize_output`, idempotent for the pre-normalized `classify_failure` path) — `hooks/interrupt.rs` passes raw `output_preview` text, so colored cancels like `\x1b[31mInterrupted\x1b[0m` must not slip past the line-shape checks.
- The shared matcher applies three tiers:
  - multi-word cancel phrases (`sudo: a password is required`, `operation cancelled/canceled`, `keyboardinterrupt`, …) keep substring matching — they are specific enough that false hits are negligible;
  - the bare `interrupted` marker only counts when a line, after trim, is exactly `interrupted`;
  - credential needles (`access key id`, `access key secret`, `default region id`) only count when the line itself is the prompt label — it must start with the needle and carry at most a bracketed default before the trailing `:` (review round 3: colon-ended error headers like `Error: failed to validate access key id:` previously matched the looser contains + `ends_with(':')` shape). `Access Key Id []:` prompts from `aliyun configure` stay detected; `InvalidAccessKeyId …` errors and credential error headers stay analyzable.
- No change to the `exit_code == 1` gate, the ctrl-c event window (`command_has_user_interrupt_event`), the 4096-char `output_preview` truncation, or the `prompt_like_output` fallback.
- All production paths are covered by the one shared predicate: the skip gate (`agent/failed_command.rs::failure_analysis_disposition`, `runtime/hooks.rs::record_command_hook_findings`) and the classifier (`command/failure_semantics.rs::classify_failure`). A repo-wide grep confirms no further copies.

## Tests

Added in `hooks::interrupt::tests` (production entry `command_should_skip_failure_analysis` through the real `output_preview` file path):

- `keeps_failures_with_interrupted_substring_analyzable` — `curl: (18) transfer was interrupted by peer` must not be skipped (FAIL before this fix);
- `keeps_failures_mentioning_access_key_analyzable` — `Error: InvalidAccessKeyId, the access key id you provided does not exist` must not be skipped (FAIL before this fix);
- `detects_bare_interrupted_line_as_cancel` — a standalone `Interrupted` line is still treated as a user cancel;
- `detects_ansi_colored_interrupted_line_as_cancel` / `detects_ansi_colored_credential_prompt_as_cancel` (review round 2) — ANSI-colored cancel lines still match after in-matcher normalization (both FAIL without it);
- `keeps_credential_error_header_analyzable` (hook gate) and `credential_error_header_stays_analyzable_end_to_end` (skip gate + `classify_failure`, review round 3) — colon-ended credential error headers are not treated as prompts (both FAIL against the previous prompt-shape check).

Added in `agent::failed_command::tests` (review round 1, end-to-end through `classify_failure` + the skip gate; placed in this bin-only module so the `check-test-inventory.sh` lib/bin overlap ceiling stays at 688):

- `interrupted_substring_failure_is_not_cancel_classified_end_to_end` / `access_key_error_failure_is_not_cancel_classified_end_to_end` — skip gate stays open and `classify_failure` no longer yields `InteractiveCancel` nor an `InteractiveCancelOutput` reason (both FAIL against the wide matcher). Honest scope note: these outputs classify as `GenericRuntimeFailure`/`UnknownFailure`, which stay `SilentRecord` by the existing disposition table — the fix removes the false *cancel* attribution; it does not promise an action card for them;
- `bare_interrupted_line_stays_silent_end_to_end` / `credential_prompt_output_stays_silent_end_to_end` — true cancels still classify `InteractiveCancel` and stay `SilentRecord`.

Existing contract tests remain untouched and green: sudo password prompt, `aliyun configure` credential prompt, ctrl-c window boundaries, plain `exit 1` stays analyzable.

## Verification

- FAIL→PASS control: `cargo test -p cosh-shell --bin cosh-shell hooks::interrupt` — on the unfixed base the two regression tests fail (`5 passed; 2 failed`); with the fix all pass (`8 passed; 0 failed`). Note the module is compiled into the **bin** target only (the lib target maps `hooks` via `hooks/public.rs` to model+engine), hence `--bin cosh-shell`.
- Round-1 FAIL→PASS control: with the wide matcher temporarily restored, the two new end-to-end regressions fail; with the shared tiered matcher they pass (`end_to_end` filter: `4 passed; 0 failed`; `--lib failure_semantics`: `16 passed; 0 failed`).
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `crates/cosh-shell/scripts/check-layout.sh`, `scripts/check-test-inventory.sh` all green locally; no upstream drift at submission time.
- Not run locally (excluded scope, delegated to CI): full workspace test suite, release build.

## Evidence

Pure classification-logic fix with no UI rendering path; evidence is anchored in-repo by the regression test names above and the FAIL→PASS control command. No screenshot/cast assets apply.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/command/failure_semantics.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/command/failure_semantics.rs
@@ -505,21 +505,50 @@ use failure_signatures::{
     BoundedOutput,
 };
 
-fn interactive_cancel_output(output: &str) -> bool {
-    [
+/// Shared cancel-output matcher for failure classification and the
+/// failure-analysis skip gate (`hooks::interrupt`). Needles are tiered by
+/// specificity so real failures whose output merely mentions these words
+/// stay analyzable (issue #2090). Input is normalized here (ANSI stripped,
+/// lowercased) because raw callers pass terminal output with control
+/// sequences; the pass is idempotent for pre-normalized callers.
+pub(crate) fn interactive_cancel_output(output: &str) -> bool {
+    let normalized = normalize_output(output);
+    let phrase_hit = [
         "sudo: a password is required",
         "a password is required",
         "a terminal is required",
         "operation cancelled",
         "operation canceled",
         "keyboardinterrupt",
-        "interrupted",
-        "access key id",
-        "access key secret",
-        "default region id",
     ]
     .iter()
-    .any(|needle| output.contains(needle))
+    .any(|needle| normalized.contains(needle));
+    if phrase_hit {
+        return true;
+    }
+
+    normalized.lines().any(|line| {
+        let line = line.trim();
+        // Single-word cancel markers must stand alone on their own line so
+        // failures like "transfer was interrupted by peer" stay analyzable.
+        if line == "interrupted" {
+            return true;
+        }
+        // Credential needles only signal a cancel when the line itself is the
+        // prompt label, optionally with a bracketed default, ending with ':'
+        // (e.g. "access key id []:"); error headers that merely mention the
+        // needle and end with ':' stay analyzable.
+        ["access key id", "access key secret", "default region id"]
+            .iter()
+            .any(|needle| {
+                line.strip_prefix(needle)
+                    .and_then(|rest| rest.strip_suffix(':'))
+                    .is_some_and(|middle| {
+                        let middle = middle.trim();
+                        middle.is_empty() || (middle.starts_with('[') && middle.ends_with(']'))
+                    })
+            })
+    })
 }
 
 fn push_reason_once(reasons: &mut Vec<FailureReason>, reason: FailureReason) {
--- a/src/cosh-ng/crates/cosh-shell/src/hooks/interrupt.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/hooks/interrupt.rs
@@ -1,7 +1,8 @@
 use std::fs;
 
 use super::prelude::{
-    classify_command_interaction, CommandBlock, PtyRequirement, ShellEvent, ShellEventKind,
+    classify_command_interaction, interactive_cancel_output, CommandBlock, PtyRequirement,
+    ShellEvent, ShellEventKind,
 };
 
 pub(crate) fn command_should_skip_failure_analysis(
@@ -47,24 +48,6 @@ fn output_preview(block: &CommandBlock) -> String {
         return String::new();
     };
     content.chars().take(4096).collect()
-}
-
-fn interactive_cancel_output(output: &str) -> bool {
-    let normalized = output.to_ascii_lowercase();
-    [
-        "sudo: a password is required",
-        "a password is required",
-        "a terminal is required",
-        "operation cancelled",
-        "operation canceled",
-        "keyboardinterrupt",
-        "interrupted",
-        "access key id",
-        "access key secret",
-        "default region id",
-    ]
-    .iter()
-    .any(|needle| normalized.contains(needle))
 }
 
 fn prompt_like_output(output: &str) -> bool {
@@ -156,6 +139,68 @@ mod tests {
         block.output.terminal_output_ref = Some(write_output("Access Key Id []:"));
 
         assert!(command_should_skip_failure_analysis(&[], &block));
+    }
+
+    #[test]
+    fn detects_bare_interrupted_line_as_cancel() {
+        let mut block = block();
+        block.command = "pip install requests".to_string();
+        block.output.terminal_output_ref = Some(write_output("Collecting requests\nInterrupted\n"));
+
+        assert!(command_should_skip_failure_analysis(&[], &block));
+    }
+
+    #[test]
+    fn detects_ansi_colored_interrupted_line_as_cancel() {
+        let mut block = block();
+        block.command = "pip install requests".to_string();
+        block.output.terminal_output_ref = Some(write_output(
+            "Collecting requests\n\x1b[31mInterrupted\x1b[0m\n",
+        ));
+
+        assert!(command_should_skip_failure_analysis(&[], &block));
+    }
+
+    #[test]
+    fn detects_ansi_colored_credential_prompt_as_cancel() {
+        let mut block = block();
+        block.command = "aliyun configure".to_string();
+        block.output.terminal_output_ref = Some(write_output("\x1b[36mAccess Key Id []:\x1b[0m"));
+
+        assert!(command_should_skip_failure_analysis(&[], &block));
+    }
+
+    #[test]
+    fn keeps_credential_error_header_analyzable() {
+        let mut block = block();
+        block.command = "ossutil ls oss://bucket".to_string();
+        block.output.terminal_output_ref = Some(write_output(
+            "Error: failed to validate access key id:\n  the key contains invalid characters\n",
+        ));
+
+        assert!(!command_should_skip_failure_analysis(&[], &block));
+    }
+
+    #[test]
+    fn keeps_failures_with_interrupted_substring_analyzable() {
+        let mut block = block();
+        block.command = "curl https://example.com/api".to_string();
+        block.output.terminal_output_ref = Some(write_output(
+            "curl: (18) transfer was interrupted by peer\n",
+        ));
+
+        assert!(!command_should_skip_failure_analysis(&[], &block));
+    }
+
+    #[test]
+    fn keeps_failures_mentioning_access_key_analyzable() {
+        let mut block = block();
+        block.command = "ossutil cp local oss://bucket/key".to_string();
+        block.output.terminal_output_ref = Some(write_output(
+            "Error: InvalidAccessKeyId, the access key id you provided does not exist\n",
+        ));
+
+        assert!(!command_should_skip_failure_analysis(&[], &block));
     }
 
     #[test]
--- a/src/cosh-ng/crates/cosh-shell/src/agent/failed_command_tests.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/agent/failed_command_tests.rs
@@ -1207,3 +1207,73 @@ fn manual_mode_skips_user_interrupted_failed_command_card() {
 
     assert!(output.is_empty());
 }
+
+#[test]
+fn interrupted_substring_failure_is_not_cancel_classified_end_to_end() {
+    let mut block = failed_block(1, "curl https://example.com/api");
+    let output = "curl: (18) transfer was interrupted by peer\n";
+    block.output.terminal_output_ref = Some(write_output(output.as_bytes()));
+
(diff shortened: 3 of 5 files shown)


src/cosh-ng/crates/cosh-shell/src/hooks/interrupt.rs
```diff
@@ -51,20 +51,35 @@ fn output_preview(block: &CommandBlock) -> String {
...
-        "default region id",
     ]
     .iter()
-    .any(|needle| normalized.contains(needle))
+    .any(|needle| normalized.contains(needle));
+    if phrase_hit {
+        return true;
+    }
+
+    normalized.lines().any(|line| {
```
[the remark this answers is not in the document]
SunnyQjm: Verified and adopted — fixed in 25193c78.

- **Review item**: `command/failure_semantics.rs::interactive_cancel_output` was a second wide-substring copy; `classify_failure` still classified `curl ... interrupted by peer` / `InvalidAccessKeyId ...` as `InteractiveCancel` (silently recorded), so the hardening only covered the skip gate.
- **Verification**: confirmed at source — `failure_semantics.rs` L508-523 (old) did plain `contains` on `"interrupted"` / credential needles; `failure_analysis_disposition` calls `classify_failure` right after the skip gate.
- **Fix**: converged on one shared `pub(crate) fn interactive_cancel_output` in `command/failure_semantics.rs` (tiered matching); `hooks/interrupt.rs` now reuses it via the hooks prelude and its local copy is deleted. Repo-wide grep shows no further copies.
- **Regressions on the production path** (in bin-only `agent/failed_command_tests.rs`, keeping the lib/bin overlap ceiling at 688): `interrupted_substring_failure_is_not_cancel_classified_end_to_end` and `access_key_error_failure_is_not_cancel_classified_end_to_end` assert the skip gate stays open and `classify_failure` yields neither `InteractiveCancel` nor an `InteractiveCancelOutput` reason; both FAIL against the wide matcher and PASS with the shared one. Honest scope note: these outputs classify as `GenericRuntimeFailure`/`UnknownFailure`, which remain `SilentRecord` by the existing disposition table — the fix removes the false cancel attribution rather than promising an action card. `bare_interrupted_line_stays_silent_end_to_end` / `credential_prompt_output_stays_silent_end_to_end` pin that true cancels still stay silent.

src/cosh-ng/crates/cosh-shell/src/command/failure_semantics.rs
```diff
@@ -505,21 +505,41 @@ use failure_signatures::{
...
-        "default region id",
     ]
     .iter()
-    .any(|needle| output.contains(needle))
+    .any(|needle| normalized.contains(needle));
+    if phrase_hit {
+        return true;
+    }
+
+    normalized.lines().any(|line| {
```
kongche-jbw, line 530: **[P2] Normalize terminal control sequences before the line-shape checks**

`classify_failure` reaches this matcher with ANSI already stripped by `BoundedOutput::new`, but `hooks::interrupt` passes the raw `output_preview` directly. Consequently, outputs such as `\x1b[31mInterrupted\x1b[0m` and `\x1b[36mAccess Key Id []:\x1b[0m` no longer satisfy the exact-line / `ends_with(':')` checks when the Ctrl-C event is missing. Both used to match the broad substring matcher, so `command_should_skip_failure_analysis` can now let an actual interactive cancel reach runtime hooks. Please normalize terminal control sequences inside the shared matcher (or before every raw caller) and add an ANSI regression through `command_should_skip_failure_analysis`.

kongche-jbw (changes requested):
One correctness regression remains in the raw hook path. The scoped local tests and current CI otherwise pass.

src/cosh-ng/crates/cosh-shell/src/command/failure_semantics.rs
SunnyQjm: Verified and adopted — fixed in 11e4c122.

- **Review item**: `hooks::interrupt` passes raw `output_preview` text, so ANSI-colored cancels (`\x1b[31mInterrupted\x1b[0m`, `\x1b[36mAccess Key Id []:\x1b[0m`) failed the exact-line / `ends_with(':')` checks that the old substring matcher tolerated — a real regression on the raw hook path.
- **Verification**: reproduced FAIL-first with two new regressions through `command_should_skip_failure_analysis` (`detects_ansi_colored_interrupted_line_as_cancel`, `detects_ansi_colored_credential_prompt_as_cancel`) — both FAILED on the previous head.
- **Fix**: the shared matcher now normalizes its input itself via the module's existing `normalize_output` (ANSI strip + lowercase). The pass is idempotent for the pre-normalized `classify_failure` path (`BoundedOutput` already strips ANSI per line), so both callers converge on one normalization boundary instead of each caller pre-normalizing.
- **Validation**: `hooks::interrupt` 10/10, `end_to_end` regressions 4/4, `--lib failure_semantics` 16/16; fmt/clippy/check-layout/check-test-inventory green; rebased onto latest main before push.

kongche-jbw (commented):
Review baseline: `92a67aa84d18...11e4c1221d79`

**[P2] Do not treat colon-ended credential errors as prompts**

At `src/cosh-ng/crates/cosh-shell/src/command/failure_semantics.rs:540`, any line
that contains a credential needle and ends in `:` is accepted. An exit-1 error such as
`Error: failed to validate access key id:` followed by details therefore still makes both
the hook gate and `classify_failure` treat the command as an interactive cancel, even though
it is an error header. This preserves the false-silence class this change is intended to remove.

Possible direction: anchor the prompt label itself, then add this error-header case through
both `command_should_skip_failure_analysis` and `failure_analysis_disposition`.

kongche-jbw (commented):
Review baseline: `f46ac79fd6f4...e51994c07224`

未发现阻塞问题。

Remaining risks:

- 未在本地执行完整 workspace 测试或真实 PTY 手工验证，当前 CI 已覆盖构建和测试门禁。

Verification scope:

- 核对完整 diff、相邻实现、Issue #2090、全部分页评论、Review 与 review threads。
- `cargo test -p cosh-shell --bin cosh-shell hooks::interrupt`，11/11 通过。
- `cargo test -p cosh-shell --bin cosh-shell end_to_end`，5/5 通过。
- `cargo test -p cosh-shell --lib failure_semantics`，16/16 通过。
- `cargo fmt --all -- --check` 通过；当前 cosh-ng CI 与 release build 均通过。

kongche-jbw (approved):
lgtm

PR #2411, opened 2026-08-11
feat(cosh-ng): [packaging] add raw packer

ikunkun-sys:
## Why

cosh-ng needs a component-owned raw packaging interface so release tooling can consume immutable artifacts without duplicating payload assembly. Keeping the contracts, launchers, and validation beside the component makes the shipped layout independently testable and reviewable across supported targets.

## What changed

- Add a `stage`/`package` entrypoint for deterministic cosh-ng raw archives.
- Support Linux x86_64, Linux aarch64, and macOS aarch64; explicitly reject macOS x86_64.
- Select target-specific Linux/macOS contracts so Linux keeps its OpenSSL runtime dependency without imposing it on macOS.
- Validate contract metadata, prebuilt ELF/Mach-O architecture, and cross-target build provenance before packaging.
- Add static launcher assets without lifecycle hooks, including portable macOS path resolution.
- Add fixture coverage for reproducibility, platform normalization, layout, permissions, wrapper routing, and failure cases.
- Register the raw packaging suite and ShellCheck in the cosh-ng fast CI gate.
- Extend the repository version check to keep both raw contracts synchronized with the cosh-ng workspace version.

## Related issue

no-issue: move raw packaging ownership into the cosh-ng component

## User / Agent impact

There is no change to the existing source or RPM installation path. Release tooling can invoke `src/cosh-ng/packaging/raw/package.sh` for these artifacts:

- `cosh-ng-<version>-linux-x86_64.tar.gz`
- `cosh-ng-<version>-linux-aarch64.tar.gz`
- `cosh-ng-<version>-macos-aarch64.tar.gz`

Cross-target `BIN_DIR` inputs must include `cosh-ng-build.toml` with the version, target OS/architecture, and SHA-256 of all three binaries. Release-pipeline integration is intentionally outside this PR.

## Risk and compatibility

The packer embeds a Linux contract that lets ANOLISA query the native `openssl1.1`/`libssl1.1` package, or a macOS contract without that Linux-only dependency because native-tls uses Security.framework there. The raw package has no lifecycle hooks and does not modify `/etc/shells`. Both contracts require ANOLISA 0.2.17 or newer and do not replace the existing RPM contract.

## Validation

- `shellcheck src/cosh-ng/packaging/raw/package.sh src/cosh-ng/packaging/raw/assets/bin/cosh src/cosh-ng/packaging/raw/assets/bin/cosh-switch src/cosh-ng/tests/test-package-raw.sh`
- `bash src/cosh-ng/tests/test-package-raw.sh`
- `python3 scripts/check-component-versions.py`
- `git diff origin/main...HEAD --check`

The packaging test creates synthetic Linux x86_64/aarch64 ELF and macOS aarch64 Mach-O inputs with hash-bound build metadata. It verifies all three artifact names and target contracts, rejects unsupported, mislabeled, or version-mismatched inputs, checks deterministic output and layout, exercises the launcher without GNU `readlink -f`, uses portable permission checks, and confirms that no lifecycle hook is shipped.

`cargo clippy --workspace --all-targets -- -D warnings` could not complete in the local environment because the host lacks the OpenSSL development package and `openssl.pc`; the PR CI build and cosh-ng checks remain the authoritative Rust gates.

## Documentation and rollback

No user documentation changed. Revert commit `89541c31` to remove the component-owned raw packaging interface and contracts.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1191,6 +1191,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: 'rustfmt, clippy'
@@ -1202,7 +1206,7 @@ jobs:
       - name: Install system dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y libssl-dev pkg-config ripgrep
+          sudo apt-get install -y libssl-dev pkg-config ripgrep shellcheck
 
       - name: Check formatting
         run: |
--- a/scripts/check-component-versions.py
+++ b/scripts/check-component-versions.py
@@ -23,6 +23,8 @@ TOML_CONTRACTS = (
     ("src/agentsight/Cargo.toml", "src/agentsight/component.toml"),
     ("src/copilot-shell/package.json", "src/copilot-shell/component.toml"),
     ("src/cosh-ng/Cargo.toml", "src/cosh-ng/component.toml"),
+    ("src/cosh-ng/Cargo.toml", "src/cosh-ng/.anolisa/component.toml"),
+    ("src/cosh-ng/Cargo.toml", "src/cosh-ng/.anolisa/component.macos.toml"),
     ("src/skillfs/Cargo.toml", "src/skillfs/component.toml"),
     ("src/tokenless/Cargo.toml", "src/anolisa/manifests/components/tokenless/component.toml"),
     ("src/ws-ckpt/src/Cargo.toml", "src/ws-ckpt/component.toml"),
--- a/src/cosh-ng/.anolisa/component.macos.toml
+++ b/src/cosh-ng/.anolisa/component.macos.toml
@@ -1,0 +1,80 @@
+[component]
+name = "cosh-ng"
+version = "0.15.0"
+layer = "runtime"
+domain = "tools"
+display_name = "cosh-ng"
+description = "Computable Operating System Harness (cosh-ng) - a deterministic Agent-OS interface"
+license = "Apache-2.0"
+repository = "https://github.com/alibaba/anolisa"
+conflicts = ["cosh"]
+
+[component.contract]
+schema_version = "1.0"
+min_anolisa_version = "0.2.17"
+
+[backends.rpm]
+package = "cosh-ng"
+
+[component.platform]
+os = ["macos"]
+arch = ["aarch64"]
+
+[[component.dependencies]]
+name = "bash"
+kind = "system-package"
+probe = "bash --version"
+packages = { rpm = "bash", deb = "bash" }
+
+[component.layout]
+modes = ["user", "system"]
+
+[[component.layout.files]]
+source = ".anolisa/component.toml"
+target = "{datadir}/components/cosh-ng/component.toml"
+mode = "0644"
+
+[[component.layout.files]]
+source = "bin/cosh-cli"
+target = "{bindir}/cosh-cli"
+mode = "0755"
+type = "executable"
+
+[[component.layout.files]]
+source = "bin/cosh"
+target = "{bindir}/cosh"
+mode = "0755"
+type = "executable"
+
+[[component.layout.files]]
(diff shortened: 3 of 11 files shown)


kongche-jbw (commented):
Review baseline: `7e971b13f92db35ce95142b7699ae89968b8fb4e...6501c40ce4b296e8411c261492b6e3919711d063`

**[P1] Keep the OpenSSL dependency probe shell-free**

`src/cosh-ng/.anolisa/component.toml:27` is executed by ANOLISA as a direct
`program arg...` command, not through a shell. The current `$(uname -s)` and
`||` therefore become literal arguments and `test` exits 2, so raw-install
preflight always sees `openssl1.1` as missing. User-mode installs are refused;
system-mode installs may provision a package and then fail the same recheck,
while macOS cannot provision this intentionally Linux-only dependency at all.
Possible direction: use a target-specific contract or platform-scoped
dependency, keep probes argv-only, and cover Linux/macOS raw-install preflight.

**[P2] Verify the version of cross-target binaries**

`src/cosh-ng/packaging/raw/package.sh:65` returns before running
`cosh-cli --version` whenever the target differs from the host. The remaining
check only validates file magic and architecture, so reusing an older build for
Linux aarch64 or macOS aarch64 can publish stale binaries under the current
contract version and artifact name. Possible direction: require target build
provenance carrying the binary version, and reject a mismatched cross-target
fixture in the packaging test.

kongche-jbw (commented):
Review baseline: `4dee69d705a874c4740ba48d499ffa2510a53eb8...15c1146a119abe7c955e3c2f945a12549719c247`

**[P1] Keep the metadata-free path compatible with Bash 3.2**

`src/cosh-ng/packaging/raw/package.sh:175` expands an empty `METADATA_ARGS`
under `set -u` when a native build has no `cosh-ng-build.toml`. Bash versions
before 4.4 treat `${METADATA_ARGS[@]}` as an unbound variable, so the script
exits before binary verification or staging. This breaks the default
metadata-free path on macOS's bundled Bash 3.2 and on older Linux build hosts.
The current test cannot catch it because every fixture writes build metadata.
Possible direction: avoid expanding an empty array, then exercise a native
no-metadata package on the oldest supported Bash.

kongche-jbw (commented):
Review baseline: `cdf2ad0a45c3...2b15e9286a05`

**[P1] Gate the contracts on conflict-aware ANOLISA releases**

`src/cosh-ng/.anolisa/component.toml:14` and the macOS contract declare
`min_anolisa_version = "0.1.16"` while both rely on `conflicts = ["cosh"]`.
ANOLISA 0.1.16's tolerant manifest parser has no `component.conflicts` field,
and its raw install path does not validate component conflicts. It can therefore
install cosh-ng beside a tracked or RPM-owned `cosh` instead of refusing the
mutually exclusive pair. The contract guidance also requires semantics unknown
to older CLIs to be gated and published at 0.2.17 or newer. Possible direction:
raise the minimum in both contracts, keep these artifacts out of the legacy
index, and cover the conflict through a raw-install test at the minimum version.

**[P2] Register the new packaging regression suite in CI**

`src/cosh-ng/tests/test-package-raw.sh:1` is not called by any workflow or by
`scripts/run-test-gates.sh`; the green cosh-ng jobs currently exercise only the
Cargo and E2E gates. A broken packer, launcher, or contract can therefore merge
without this 305-line suite ever running. Possible direction: add the script and
the listed shellcheck command to the cosh-ng fast or packaging gate, then verify
that a deliberately failing packaging assertion makes CI fail.

kongche-jbw (commented):
Review baseline: `a350dc19ba97...89541c310157`

未发现阻塞问题。

Remaining risks:

- 未在 macOS aarch64 主机执行产物，也未运行真实 ANOLISA raw install/uninstall。
- 本机缺少 `shellcheck`；当前 head 的 `Test cosh-ng fast checks` 已通过并包含该检查。

Verification scope:

- 完整 diff、相邻 raw installer、现有 Review 和 4 个 review threads 的静态复查。
- `bash src/cosh-ng/tests/test-package-raw.sh`
- `python3 scripts/check-component-versions.py`
- `git diff a350dc19ba97...89541c310157 --check`

kongche-jbw (approved):

PR #2419, opened 2026-08-11
fix(anolisa): verify package-owned adapters

### LINKED ISSUE #2252: [anolisa] sec-core Hook 生成 __pycache__ 导致 bundle drift 误报
## Component

anolisa

## What happened?

通过 `anolisa adapter enable sec-core qwencode` 启用 Qwen Code extension 后，初始 adapter 状态为 healthy。

首次运行 Qwen Code 并触发 sec-core Python Hook 后，Python 会在 adapter resource root 中生成：

```text
hooks/__pycache__/hook_config.cpython-311.pyc
hooks/__pycache__/pii_text.cpython-311.pyc
hooks/__pycache__/trace_context.cpython-311.pyc
```

随后执行 `anolisa adapter status sec-core`，Qwen adapter 被标记为 degraded：

```text
resource_bundle_matches=false
reason="resource bundle changed since enable"
```

但 Qwen CLI 仍显示 extension enabled，Code Scanner、Prompt Scanner 和 PII Checker Hook 也能正常产生审计事件。

当前 `digest_tree` 会对 resource root 下所有文件计算摘要，包括 Hook 运行时自动生成的 `__pycache__/*.pyc`。因此，正常执行 Hook 自身就会改变 bundle digest，被误判为资源包漂移。

期望结果：正常运行已安装的 Hook 不应使 adapter 从 healthy 变为 degraded。真正的源码、manifest 或受管理资源变化仍应被检测。

## How can we reproduce it?

1. 安装 sec-core 0.9.0，并确保 Qwen Code 0.21.5 可用。
2. 启用 Qwen adapter：

```bash
anolisa adapter enable sec-core qwencode
anolisa adapter status sec-core
```

3. 确认初始状态为 healthy。
4. 运行一次 Qwen Code，使 Prompt Scanner 或 Code Scanner Hook 被调用：

```bash
qwen -m qwen3.6-plus -o json --approval-mode=yolo \
  -p '请执行一个简单的 shell 测试命令'
```

5. 检查 resource root。

Raw 安装：

```bash
ls -la \
  /usr/local/share/anolisa/adapters/sec-core/qwencode/hooks/__pycache__
```

RPM 安装：

```bash
ls -la \
  /opt/agent-sec/qwen-code-extension/hooks/__pycache__
```

6. 再次查询状态：

```bash
anolisa adapter status sec-core
```

7. 观察 Qwen adapter 变为 degraded，并报告 `resource bundle changed since enable`。

## Diagnostic output

```shell
$ qwen extensions list
✓ agent-sec-core-qwen-code-extension (0.9.0)
Path: /opt/agent-sec/qwen-code-extension/
Enabled (User): true
Enabled (Workspace): true

$ ls hooks/__pycache__
hook_config.cpython-311.pyc
pii_text.cpython-311.pyc
trace_context.cpython-311.pyc

$ anolisa adapter status sec-core
framework: qwencode
summary: degraded
resource_bundle_matches: false
reason: resource bundle changed since enable
plugin_registered: true
activation_enabled: true
```

## Anything else?

该问题同时复现于：

- Ubuntu 22.04 + raw backend
- Alibaba Cloud Linux 4 + raw backend
- Alibaba Cloud Linux 4 + CI RPM artifact
- anolisa 0.2.17
- sec-core 0.9.0
- Qwen Code 0.21.5

运行期验证表明 adapter 功能仍然正常：

- Code Scanner 产生 `code_scan` 事件。
- Prompt Scanner 产生 `prompt_scan` 事件。
- RPM 场景中 PII Checker 产生 `pii_scan` 事件。
- Qwen extension 仍处于 enabled 状态。

相关机制：enable 时对整个 resource root 执行 `digest_tree` 并记录 digest；status 时重新对整个目录树计算 digest；`digest_tree` 当前会纳入所有文件，没有排除 `__pycache__`、`*.pyc` 等运行时派生文件。

可能的修复方向包括：

- Hook 以禁止写入 bytecode 的方式运行，例如使用 `python -B` 或设置 `PYTHONDONTWRITEBYTECODE=1`。
- 将 adapter resource root 作为只读资源交付。
- 让 bundle digest 明确排除可安全识别的运行时派生文件，同时继续检测 manifest、Hook 源码和其他受管理文件的变化。

不建议笼统忽略所有隐藏文件或可变文件，以免削弱真实篡改检测能力。

ikunkun-sys:
## Why

Adapter health currently hashes the whole resource tree, so runtime-generated files such as
`__pycache__/*.pyc` are reported as bundle drift even though no package-owned input changed.
This change aligns drift detection with the package manager's ownership boundary instead of
assuming that adapter directories are immutable.

## What changed

- Build one managed-file inventory from Raw ownership records or a read-only rpmdb query, and
  verify regular-file contents plus literal symlink targets.
- Persist schema-v2 source revisions and expected ANOLISA-materialized files while retaining
  `bundle_digest` only for legacy receipt migration.
- Replace whole-tree matching with tri-state managed, source-revision, and materialized-bundle
  conditions; unavailable or unsupported authoritative metadata remains `Unknown`.
- Apply the same authoritative revision comparison to post-update stale actions, and cover Cosh,
  Hermes, and OpenClaw copies without inventing a materialized condition for Qoder staging.

## Related issue

Fixes #2252

## User / Agent impact

Unmanaged runtime files no longer degrade adapter health. Modified, missing, or newly declared
managed inputs still degrade health, source-root changes still require enable, and unavailable
package metadata is reported as `Unknown` rather than healthy.

## Risk and compatibility

`AdapterClaim` advances to schema 2. Schema-1 receipts are read without mutation and compare only
their current authoritative managed subset; ambiguous receipts and unavailable inventories remain
`Unknown`. The rpmdb path is read-only and rejects unsupported digest algorithms. This deliberately
does not claim that every byte used by a running process is trusted; it verifies package-managed
inputs and ANOLISA's explicitly copied outputs only.

## Validation

- `cargo test --workspace --locked --quiet -- --skip systemd::tests::unimplemented_unit_operations_return_errors_instead_of_panicking`
- Raw and simulated RPM regression coverage for runtime extras, managed add/change/delete, source
  root migration, legacy receipts, symlinks, query failures, and unsupported or empty digests
- Cosh, Hermes, and OpenClaw materialized-file coverage for ignored runtime extras and changed or
  missing delivered files

A full unfiltered workspace test run passed the adapter suites but hit the existing host-dependent
systemd test above because this development host has a real `agentsight.service`; exercising that
test's enable/disable path against the real service was intentionally avoided. The filtered
workspace run passed.

## Documentation and rollback

No user documentation changed. To roll back, revert this commit; claims created with schema 2 may
need a fresh enable after rollback so the older implementation can write its legacy digest receipt.

### DIFF
--- a/src/anolisa/crates/anolisa-core/src/adapter/claim.rs
+++ b/src/anolisa/crates/anolisa-core/src/adapter/claim.rs
@@ -34,11 +34,9 @@ use serde_json::Value;
 use crate::path_safety::{PathBoundaryError, canonicalize_nearest_existing, validate_owned_path};
 use anolisa_platform::fs_layout::FsLayout;
 
-use super::util::digest_tree;
-
 /// Schema version for the generic claim shape and [`ClaimResource`].
 /// Persisted in every receipt so a future on-disk migration can branch.
-pub const CLAIM_SCHEMA_VERSION: u32 = 1;
+pub const CLAIM_SCHEMA_VERSION: u32 = 2;
 
 /// Schema version for [`DriverPayload`]. Bumped independently of
 /// [`CLAIM_SCHEMA_VERSION`] when a driver's typed payload changes shape.
@@ -82,10 +80,20 @@ pub struct AdapterClaim {
     /// Resource directory read at enable time. Kept for status display and
     /// upgrade detection; `disable` must NOT depend on it still existing.
     pub resource_root: PathBuf,
-    /// Digest of the resource tree at enable time, for drift/upgrade
-    /// detection. Optional: a driver may decline to compute one.
+    /// Legacy whole-tree digest read only to migrate pre-v2 receipts. New
+    /// receipts never write it.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub bundle_digest: Option<String>,
+    /// Package-manager-owned adapter inputs captured at enable time. The root
+    /// is the resolved source root and every file path is relative to it, so
+    /// a source-root migration changes the revision even when bytes do not.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub source_revision: Option<AdapterSourceRevision>,
+    /// Files ANOLISA explicitly materialized into framework-owned resources.
+    /// Runtime additions under those resources are intentionally absent and
+    /// therefore ignored by status.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub materialized_files: Vec<MaterializedFile>,
     /// [`DriverPayload`] schema version ([`DRIVER_SCHEMA_VERSION`] at write
     /// time).
     pub driver_schema: u32,
@@ -111,21 +119,6 @@ impl AdapterClaim {
     /// Whether this receipt represents a skill-only adapter bundle.
     pub fn is_skill_bundle(&self) -> bool {
         self.adapter_type.as_deref() == Some("skill_bundle")
-    }
-
-    /// Compare the enable-time [`Self::bundle_digest`] against a fresh digest
-    /// of [`Self::resource_root`].
-    ///
-    /// This is the drift/upgrade detection the `bundle_digest` field exists
-    /// for; the drivers' `ResourceBundleMatches` condition and post-update
-    /// adapter actions branch on the same verdict, so the comparison lives
-    /// with the receipt schema instead of being re-derived per caller.
-    pub fn bundle_match(&self) -> BundleMatch {
-        match (&self.bundle_digest, digest_tree(&self.resource_root)) {
-            (Some(recorded), Some(current)) if recorded == &current => BundleMatch::Matched,
-            (Some(_), Some(_)) => BundleMatch::Changed,
-            _ => BundleMatch::Unknown,
-        }
     }
 
     /// Find a resource by its stable `id`.
@@ -239,6 +232,7 @@ impl AdapterClaim {
                 _ => {}
             }
         }
+        validate_integrity_entries(self)?;
         Ok(())
     }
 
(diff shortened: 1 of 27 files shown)


kongche-jbw (commented):
Review baseline: `cdf2ad0a45c3...e5f00e9a1c70`
未发现阻塞问题。

Remaining risks:
- Symlinked RPM resource roots are already tracked in an existing unresolved review thread.
- No live RPM host was available; parser verification used unit coverage and upstream RPM source.

Verification scope:
- Full diff, adjacent code/tests, PR/issue context, comments, reviews, threads, and CI status.
- Targeted test suites: 40 + 7 + 57 + 74 + 5 tests passed.

src/anolisa/crates/anolisa-core/src/adapter/managed_files.rs
```diff
@@ -0,0 +1,970 @@
...
+    mappings: &[MaterializedMapping],
+) -> Result<AdapterSourceRevision, String> {
+    let lexical_root = normalize_absolute(source_root)?;
+    let canonical_root = std::fs::canonicalize(&lexical_root).map_err(|err| {
+        format!(
+            "cannot resolve adapter source root '{}': {err}; re-enable the adapter",
+            lexical_root.display()
+        )
+    })?;
+    let files = source_files_below(inventory, &lexical_root, &[])?;
```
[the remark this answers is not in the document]
ikunkun-sys: Fixed in e155a07d. Source-revision and materialized-file scoping now accept both the lexical and canonical root aliases, skip the exact root symlink entry, and deliberately avoid canonicalizing leaf inventory paths so managed symlink targets remain verifiable. symlinked_source_root_scopes_canonical_package_files covers the RPM-style root symlink plus canonical payload layout.

src/anolisa/crates/anolisa-core/src/adapter/manager.rs
```diff
@@ -3830,19 +4002,80 @@ fn source_condition(source: &SourceProbe) -> AdapterCondition {
...
+            verify_materialized_bundle(claim)
+        };
+        integrity.push(managed_condition(
+            AdapterConditionKind::MaterializedBundleMatches,
+            materialized,
+        ));
+    }
+
+    let has_false = integrity
+        .iter()
```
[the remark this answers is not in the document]
ikunkun-sys: No code change: changing an otherwise Healthy report to Unknown when authoritative evidence is unavailable is intentional for this fix and is documented as a compatibility change in the PR. Keeping Healthy or making this opt-in would present an unverifiable adapter as verified. The unavailable_authoritative_inventory_never_reports_healthy and missing_materialized_inventory_is_unknown tests explicitly lock this behavior.

src/anolisa/crates/anolisa-core/src/adapter/claim.rs
```diff
@@ -269,47 +263,74 @@ impl AdapterClaim {
...
-    /// component's current adapter resources.
-    Changed,
-    /// No digest was recorded at enable time or the resource root cannot be
-    /// digested now; drift cannot be decided either way.
-    Unknown,
+/// Follow-up reason shared by adapter status and component update actions.
+pub const SOURCE_REVISION_CHANGED_REASON: &str = "adapter source revision changed since enable";
+
+/// Package-managed file type persisted in source and materialized revisions.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
```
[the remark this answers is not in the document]
ikunkun-sys: Fixed in e155a07d. RPM Other entries are now retained until adapter-root scoping. An unsupported entry under a source or materialized root produces an explicit unverifiable reason: enable fails, while status and update become Unknown. An unrelated entry elsewhere in the same package does not affect the adapter. Directories remain excluded because lifecycle integrity owns them. The new rpm_inventory_rejects_scoped_unsupported_file_types test covers both sides.

src/anolisa/crates/anolisa-platform/src/rpm_query.rs
```diff
@@ -144,6 +157,133 @@ impl<R: CommandRunner> RpmPackageQuery<R> {
...
+    let (algorithm, rows) = stdout
+        .split_once('\n')
+        .ok_or_else(|| unexpected_file_inventory("missing digest-algorithm header"))?;
+    let algorithm = algorithm
+        .trim()
+        .parse::<u32>()
+        .map_err(|_| unexpected_file_inventory("digest algorithm is not an integer"))?;
+    let digest_algorithm = match algorithm {
+        8 => PackageFileDigestAlgorithm::Sha256,
+        other => PackageFileDigestAlgorithm::Unsupported(other),
```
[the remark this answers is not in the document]
ikunkun-sys: No code change: this case is already explicit and distinguishable. inventory_for_installation rejects any digest algorithm other than SHA-256 before constructing the managed inventory, and its reason includes Unsupported(n). Enable surfaces that as InvalidAdapterInput; status and update preserve the same reason as Unknown with a re-enable prompt. Changed would incorrectly claim a byte mismatch when the actual fact is that ANOLISA cannot verify the package digest. This is the intended fail-closed behavior.

src/anolisa/crates/anolisa-core/src/adapter/manager.rs
```diff
@@ -988,6 +1004,35 @@ impl AdapterManager {
...
+            .and_then(|installation| {
+                inventory_for_installation(&installation, self.package_files.as_ref()).map_err(
+                    |reason| AdapterError::InvalidAdapterInput {
+                        component: component.to_string(),
+                        framework: framework.clone(),
+                        reason,
+                    },
+                )
+            })?;
+        claim.source_revision = Some(
```
kongche-jbw: [P1] Verify authoritative bytes before enabling

This records the revision supplied by Raw ownership or rpmdb metadata, but the enable path never compares the current source bytes against that revision. `verify_managed_bundle` is only called later from status. If a package-managed adapter file is modified before enable, the driver can still copy or register the modified content and enable returns success; only a subsequent status reports degradation. The authoritative digest therefore does not currently act as an installation-time integrity gate.

src/anolisa/crates/anolisa-core/src/adapter/managed_files.rs
```diff
@@ -0,0 +1,1122 @@
...
+            };
+            validate_relative(&relative_path)?;
+            if mapping
+                .excluded_prefixes
+                .iter()
+                .any(|prefix| relative_path.starts_with(prefix))
+            {
+                continue;
+            }
+            let (kind, sha256, symlink_target) = integrity_metadata(file)?;
```
kongche-jbw, line 297: [P1] Do not leave copied unmanaged inputs untracked

The materialized receipt is built only from package-inventory entries, while Cosh, Hermes, and OpenClaw later call `copy_tree` over whole source directories. A pre-existing unowned file below one of those roots is therefore explicitly copied into the framework directory but omitted from `materialized_files`; changing or deleting it afterward still leaves the materialized condition healthy. Such an extra can be a discoverable hook or module, so the receipt does not actually cover every output ANOLISA copied.

src/anolisa/crates/anolisa-core/src/adapter/managed_files.rs
```diff
@@ -0,0 +1,1122 @@
...
+                    path.display()
+                )));
+            }
+            let expected = symlink_target.ok_or_else(|| {
+                ManagedMatch::Unknown(format!(
+                    "managed symlink '{}' has no recorded target; re-enable the adapter",
+                    path.display()
+                ))
+            })?;
+            let actual = std::fs::read_link(path).map_err(|err| io_verdict(path, "read", err))?;
```
kongche-jbw, line 500: [P1] Include package-owned symlink referent contents

This verifies only the literal link target. When a managed symlink below the adapter root points to a package-owned regular file outside that root, source scoping omits the referent's inventory entry. An RPM update can then change the referent bytes without changing the link text, leaving both the source revision and managed-bundle verdict matched even though direct-root frameworks consume different content.

kongche-jbw (changes requested):
Review baseline: `e155a07da8ad3c18b4d1c143dc740fbd2edf3e72`

Requesting changes for three newly identified integrity gaps. The existing review threads are resolved; these findings cover enable-time verification, the materialized-copy contract, and out-of-root symlink referents.

Verification completed on the current head: GitHub CI is green; local fmt, workspace clippy, workspace docs, and the platform/core/CLI test suites passed (2,085 passed, 1 ignored). No live RPM host was available, so the rpmdb query path remains parser/unit-tested only.

src/anolisa/crates/anolisa-core/src/adapter/manager.rs
ikunkun-sys: Fixed in 18ebaceb. Enable now completes driver read-only preflight, builds the authoritative revision, and calls verify_managed_bundle before any replacement cleanup, receipt persistence, copy, or registration. The new enable_rejects_modified_package_owned_source regression test confirms a tampered managed source leaves both framework state and the claim untouched.

src/anolisa/crates/anolisa-core/src/adapter/managed_files.rs
ikunkun-sys: Fixed in 18ebaceb. Cosh, Hermes, and OpenClaw now materialize only receipt entries derived from authoritative managed source mappings. The helper never enumerates the source tree and verifies each source and destination file. The Cosh regression creates an unowned __pycache__/runtime.pyc before enable and confirms it is not copied, while runtime extras already present at the destination remain outside health checks.

src/anolisa/crates/anolisa-core/src/adapter/managed_files.rs
ikunkun-sys: Not changing this in the current PR. Verifying the literal symlink target is the explicit #2252 trust boundary: this change does not promise that every byte a process may reach through an adapter is trusted. Following out-of-root referents would expand the schema and require explicit ownership, root, and cycle semantics. A referent that must become authoritative should be declared as its own managed source mapping in a follow-up design rather than inferred by traversal.

kongche-jbw (approved):

PR #2428, opened 2026-08-12
fix(cosh-ng): [core] expose runtime context

kongche-jbw:
## Why

Security-event queries need the provider session UUID recorded by sec-core hooks. `COSH_SESSION_ID` identifies the shell/terminal instead, while embedding the provider UUID in every system prompt would make the DashScope explicit-cache block session-specific.

## What changed

- Added the read-only `runtime_context` tool. It returns the current provider session ID, cosh-ng name/version, bound model, approval mode, workspace paths, resume state, compaction revision, bound tools, and active extensions.
- Kept all dynamic runtime values out of the system prompt so provider prefix caching remains reusable across sessions. The tool reads live bound runtime state and requires no shell or terminal evidence.
- Omitted config-derived provider name/type and hook state because reload can change configuration without rebuilding the bound provider or hook system.
- Classified `COSH_SESSION_ID` only as `shell_session_id` in policy audit records and corrected the retained `CallerInfo` documentation.
- Defined the sec-core integration contract: `security-observability` should call `runtime_context`, pass its top-level `provider_session_id` explicitly to `agent-sec-cli events` or `observability report`, and never substitute `COSH_SESSION_ID` or silently widen to `--last`.

## Related issue

Part of #2387

## User / Agent impact

Agents can retrieve current runtime and conversation identity on demand without asking the user to copy `/session status`. Security queries remain explicit and unambiguous under concurrent sessions.

## Risk and compatibility

The change adds one default read-only tool and fixes an audit identity classification. Existing session, hook, CLI, and shell identifiers are unchanged. The sec-core consumer is being developed in #2245; older runtimes remain compatible through its documented user-provided session-ID fallback.

## Validation

- `cargo test -p cosh-core --bin cosh-core runtime_context` — 8 passed
- `cargo test -p cosh-core --bin cosh-core provider_session_identity_stays_out_of_the_system_prompt` — 1 passed
- `cargo test -p cosh-platform policy_event_` — 3 passed
- `cargo clippy -p cosh-core --bin cosh-core --tests -- -D warnings`
- `git diff --check`

The runtime-context regression mutates reloadable provider and hook configuration after construction and verifies that the tool does not report those unbound values. Full workspace coverage is intentionally delegated to CI under the cosh-ng scoped-validation policy.

## Documentation and rollback

The tool declaration is self-describing; the sec-core Skill contract is owned by #2245. Roll back by reverting this commit, after which users can still provide the UUID from `/session status` explicitly.

### DIFF
--- a/src/cosh-ng/crates/cosh-core/src/core.rs
+++ b/src/cosh-ng/crates/cosh-core/src/core.rs
@@ -24,7 +24,9 @@ use crate::provider::{
     ContentGenerator, GenerateConfig, GenerateEvent, Message, MAX_TOOL_CALL_INDEX,
 };
 use crate::tool::ask_user_question;
-use crate::tool::{SessionWorkspace, ToolContext, ToolKind, ToolRegistry, ToolResult};
+use crate::tool::{
+    SessionWorkspace, ToolContext, ToolKind, ToolRegistry, ToolResult, ToolRuntimeContext,
+};
 use crate::truncator::OutputTruncator;
 
 use self::tool_execution::{
@@ -65,6 +67,7 @@ pub struct CoshCore {
     /// sees the projected effective context.
     pub compaction: CompactionRuntime,
     pub model: String,
+    session_resumed: bool,
     pub shell_context: Option<ShellContext>,
     project_root: PathBuf,
     workspace: SessionWorkspace,
@@ -90,6 +93,23 @@ pub struct CoshCore {
 impl CoshCore {
     pub fn tool_names(&self) -> Vec<String> {
         self.tools.names()
+    }
+
+    pub(crate) fn set_session_resumed(&mut self, resumed: bool) {
+        self.session_resumed = resumed;
+    }
+
+    fn tool_runtime_context(&self) -> ToolRuntimeContext {
+        let snapshot = self.extension_generation.current();
+        ToolRuntimeContext {
+            model: self.model.clone(),
+            approval_mode: self.config.agent.approval_mode.to_string(),
+            session_resumed: self.session_resumed,
+            compaction_revision: self.compaction.revision(),
+            compacted_through: self.compaction.state().map(|state| state.compacted_through),
+            tools: self.tool_names(),
+            active_extensions: snapshot.active_extensions.iter().cloned().collect(),
+        }
     }
 
     pub fn emit<W: Write>(&self, writer: &mut W, msg: &OutputMessage) {
@@ -975,11 +995,12 @@ impl CoshCore {
             self.messages
                 .push(Message::assistant_with_tool_calls(&text_buf, tc_infos));
 
-            let ctx = ToolContext::with_workspace(
+            let ctx = ToolContext::with_runtime(
                 self.cwd(),
                 self.session_id.clone(),
                 self.project_root.clone(),
                 self.workspace.clone(),
+                self.tool_runtime_context(),
             );
 
             let mut interrupted = false;
--- a/src/cosh-ng/crates/cosh-platform/src/audit.rs
+++ b/src/cosh-ng/crates/cosh-platform/src/audit.rs
@@ -65,7 +65,7 @@ impl AuditTestDir {
 /// Call-site identity captured for a policy audit event.
 #[derive(Debug, Clone)]
 pub struct CallerInfo {
-    /// Provider/session correlation identifier.
+    /// Shell/terminal correlation identifier, with a process-scoped fallback.
     pub session_id: String,
     /// Local user label used only by legacy call sites, never persisted in v1.
     pub user: String,
@@ -91,9 +91,7 @@ impl CallerInfo {
                     .filter(|value| !value.is_empty())
             })
             .unwrap_or_else(|| "unknown".to_string());
-        let session_id = std::env::var("COSH_SESSION_ID")
-            .ok()
-            .filter(|value| !value.is_empty())
+        let session_id = detected_shell_session_id()
             .unwrap_or_else(|| format!("p{}-t{}", std::process::id(), Utc::now().timestamp()));
         let sudo_user = std::env::var("SUDO_USER")
             .ok()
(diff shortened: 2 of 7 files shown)


src/cosh-ng/crates/cosh-core/src/compaction/engine.rs
```diff
@@ -491,9 +491,14 @@ fn bounded_cut_and_input(
 }
 
 /// Estimates the runtime prefix (`P`) for provider-free CLI budgets.
```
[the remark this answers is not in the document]
kongche-jbw: Resolved by the redesign in `21f3755f`. The provider UUID is no longer part of the system prompt or CLI prefix budget; the read-only `runtime_context` tool returns the live value on demand, so compaction accounting remains identical to the stable prompt and no identity can be summarized away.

src/cosh-ng/crates/cosh-core/src/context.rs
```diff
@@ -77,6 +80,11 @@ impl ContextBuilder {
             cwd.display(),
         ));
 
+        parts.push(format!(
```
yangdao479: `# Runtime Identity` 插在 `# Environment` 之后、`# Context` 之前，而 `provider_session_id` 对每个会话唯一。这会让 system prompt 从这一段开始的全部后续内容（`# Context`、扩展上下文、`# Available Skills`、工具列表）在**跨会话**之间不再拥有相同前缀，同一 workspace 下新建或切换会话时无法复用 provider 端的 prompt 前缀缓存；在此之前这些内容对同一 workspace 是稳定的。

会话内没有问题：`CoshCore.session_id` 全程不被改写，resume 时也从 `session.record.session_id` 恢复，所以单会话内 system prompt 是 byte-stable 的。

建议把这段移到组装顺序的**末尾**（`# Available Skills` 之后），让前面与会话无关的内容保持跨会话可复用；否则每个新会话都会从这一段起整体 cache miss。

如果 cosh-ng 目前并不依赖 provider 端跨请求的前缀缓存，这条可以忽略 —— 我没有在配置中确认这一点。

src/cosh-ng/crates/cosh-platform/src/audit.rs
```diff
@@ -194,7 +199,7 @@ fn record_to_log(
             version: env!("CARGO_PKG_VERSION").to_string(),
         },
         AuditIdentity {
-            provider_session_id: Some(caller.session_id),
+            shell_session_id,
```
yangdao479, line 202: 这里把 `COSH_SESSION_ID` 正确归类为 shell 身份之后，`CallerInfo` 上的字段注释就与实际语义不一致了：`audit.rs:68` 仍写 `/// Provider/session correlation identifier.`，但 `CallerInfo::detect()` 中 `session_id` 的取值来自 `detected_shell_session_id()`（即 `COSH_SESSION_ID`），且 fallback 为 `p{pid}-t{timestamp}` —— 两者都是 shell/进程身份，不是 provider session。

PR 为兼容而保留 `CallerInfo` 的形状是合理的，但建议同步把该注释改为 shell/终端会话语义（例如 `Shell/terminal session correlation identifier; falls back to a process-scoped id`），否则后续调用方仍会按注释把它当作 provider session id 使用，重新引入这次修掉的错配。

src/cosh-ng/crates/cosh-core/src/context.rs
kongche-jbw: Fixed in `21f3755f`. The concern is valid, and moving the UUID to the end of the same system message would still leave DashScope explicit caching with a session-specific cache block because cosh-ng marks the first system message as a whole. The revised design removes all dynamic identity from the system prompt and exposes it only through the read-only `runtime_context` tool. Added a provider-boundary regression test proving the session UUID is absent from the system prompt. The sec-core consumer contract is documented in the PR and will be consumed by the open `security-observability` Skill PR #2245.

src/cosh-ng/crates/cosh-platform/src/audit.rs
kongche-jbw: Fixed in `21f3755f`. The retained `CallerInfo.session_id` doc now says it is a shell/terminal correlation identifier with a process-scoped fallback, matching `detected_shell_session_id()` and the compatibility behavior.

src/cosh-ng/crates/cosh-core/src/core.rs
```diff
@@ -92,6 +95,32 @@ impl CoshCore {
         self.tools.names()
     }
 
+    pub(crate) fn set_session_resumed(&mut self, resumed: bool) {
+        self.session_resumed = resumed;
+    }
+
+    fn tool_runtime_context(&self) -> ToolRuntimeContext {
+        let resolved = self.config.resolve_provider();
```
ikunkun-sys: [P2] Report bound runtime state, not reloadable config

`tool_runtime_context()` derives the provider identity (and later `hooks_enabled`) from `self.config`, but `ShellControlRequest::ReloadConfig` only replaces `engine.config`; it does not recreate `engine.provider` or `engine.hook_system`. After reloading a changed `active_provider` or `hooks.enabled`, this tool reports the new configuration while model calls and hooks continue using the previously bound instances. That violates the advertised current-runtime contract and can mislead diagnostics or future consumers. Please source these fields from the bound runtime objects, rebuild those objects during reload, or omit the fields; a reload regression test should cover the chosen behavior.

ikunkun-sys (changes requested):
The revised on-demand tool design resolves the previous cross-session prefix-cache invalidation: `provider_session_id` no longer enters the system prompt, and the regression test locks that down. I found one correctness issue in the new generic runtime-metadata contract, described inline. Please either derive the provider/hook fields from the bound runtime state or narrow this tool to metadata that is guaranteed to remain accurate.

Separately, #2245 does not yet consume `runtime_context`, so `Closes #2387` is premature. Please use `Part of #2387`, or land and test the sec-core consumer contract before closing that issue.

The current cosh-ng CI checks are green.

src/cosh-ng/crates/cosh-core/src/core.rs
kongche-jbw: Fixed in `3024253a`. I chose the narrow bound-runtime contract: removed config-derived provider name/type and `hooks_enabled`, while retaining the model from the bound engine and tools/extensions from the bound snapshot. The regression mutates `active_provider` and `hooks.enabled` after construction and verifies that `runtime_context` does not report those unbound config values.

ikunkun-sys (approved):

PR #2446, opened 2026-08-12
fix(cosh-ng): [shell] fill CJK wrap width

### LINKED ISSUE #2437: [cosh-ng] bug: CJK wrapping wastes remaining line width
## Summary

The generic `cosh-shell` inline renderer can wrap mixed ASCII and CJK text before the current line is full.

For example, this hook message contains no explicit newline:

```text
Message: [pii-checker] 检测到 4 项高风险敏感信息；当前策略仅提醒，本次不会阻断。
```

At a width where wrapping is required, it may be rendered as:

```text
Message: [pii-checker] 检测到 4
项高风险敏感信息；当前策略仅提醒，本次不会阻断。
```

This makes the output look as if a newline was intentionally inserted after the number. The CJK text should consume the remaining display cells before continuing on the next line.

## Reproduction

A minimal renderer-level example is:

```rust
let lines = wrap_plain_line("abc 中文中文中文中", 10);
```

Current result:

```text
abc
中文中文中文中
```

Expected result:

```text
abc 中文中文中
文
```

The exact break position in the UI depends on the terminal width, but the current line should be filled before wrapping a breakable CJK sequence.

## Root cause

The issue is in:

```text
src/cosh-ng/crates/cosh-shell/src/ui/agent_render/wrap.rs
```

`split_wrap_tokens()` treats a continuous non-whitespace CJK sequence as one token. When that token does not fit the remaining width, `wrap_with_prefix()` flushes the current line before calling `wrap_long_token()`. Character-level wrapping therefore starts on an empty continuation line and leaves usable columns behind.

This remains reproducible from source at `upstream/main@a350dc19`.

## Scope

This should be fixed in the generic `cosh-ng` renderer. The PII hook only exposed the issue; its message is joined into a single line and does not insert a newline after the count.

This is related to #1703 because both touch panel wrapping, but it is not the same issue: #1703 focused on simplifying the `/skills list` presentation, while this defect affects any panel containing mixed ASCII and continuous CJK text.

## Acceptance criteria

- Continuous Chinese/CJK text uses the current line's remaining display width before wrapping.
- No rendered line exceeds the configured display width.
- Normal ASCII words are not split merely to fill the current line.
- Existing indentation, hanging prefixes, explicit newlines, ANSI handling, and display-width accounting remain unchanged.
- Add focused regression coverage for:
  - mixed ASCII and Chinese text;
  - a Governance hook notification;
  - Chinese punctuation mixed with ASCII identifiers;
  - list or indentation prefixes at narrow widths.
- No changes are required in `src/copilot-shell` or the PII Checker policy and message generation paths.

kongche-jbw:
## Why

The generic cosh-shell renderer treated a continuous CJK sequence as one
word-sized token. When it did not fit, the renderer moved the entire sequence
to a continuation line before applying character-level wrapping, leaving usable
display cells behind.

## What changed

- Treat CJK characters as breakable wrap units so they consume the current
  line's remaining display width.
- Bind common CJK opening and closing punctuation to adjacent text so quotes and
  brackets do not become stranded at line boundaries.
- Preserve non-CJK words, identifiers, accented text, emoji sequences, prefixes,
  explicit newlines, and the existing long-token fallback.
- Add focused coverage for mixed ASCII/CJK text, Governance notifications,
  punctuation with identifiers, CJK quotes and brackets, and narrow prefixes.

## Related issue

Closes #2437

## User / Agent impact

Mixed ASCII and CJK panel text now fills the current row before wrapping.
Ordinary ASCII words continue to move intact when they fit on the next row.

## Risk and compatibility

Low risk. The change is limited to generic inline text tokenization and adds no
dependencies or configuration. Existing width accounting and wrapping paths are
reused.

## Validation

Local validation at commit `c08011f9e462`:

- `cargo test --locked --package cosh-shell --lib` — 1312 passed
- `cargo test --locked --package cosh-shell --lib wrap_tests::wrap_plain_line_`
  — 6 passed
- `cargo clippy --locked --package cosh-shell --lib -- -D warnings`
- `scripts/check-test-inventory.sh` — lib/bin overlap 693/693

ECS targeted regression tests and manual terminal UX were completed on the
initial candidate before review. The user confirmed remaining-width fill,
intact ASCII identifiers, aligned panel borders, and a successful provider
response. Subsequent review refinements are covered by the automated regression
tests above and GitHub CI.

The issue's illustrative width-10 expected row has a display width of 14. The
regression test keeps the configured width at 10 and asserts both remaining-cell
use and the no-overflow acceptance criterion.

## Documentation and rollback

No documentation changes are required. Revert commit `c08011f9e462` to restore
the previous wrapping behavior.


src/cosh-ng/crates/cosh-shell/src/ui/agent_render/wrap.rs
```diff
@@ -172,6 +186,24 @@ fn split_wrap_tokens(text: &str) -> Vec<String> {
     tokens
 }
 
+fn is_cjk_breakable(ch: char) -> bool {
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in b991af54. The implementation now lives in `ui/wrap.rs`; `ui/agent_render/wrap.rs` contains compatibility re-exports only. The regression tests are lib-only, so this also avoids increasing the existing lib/bin test overlap.

src/cosh-ng/crates/cosh-shell/src/ui/agent_render/wrap.rs
```diff
@@ -172,6 +186,24 @@ fn split_wrap_tokens(text: &str) -> Vec<String> {
...
+    matches!(
+        ch,
+        '\u{2e80}'..='\u{2fff}'
+            | '\u{3000}'..='\u{303f}'
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in c5595600. A punctuation-binding pass now attaches closing punctuation to the preceding token and opening punctuation to the following token. Regression coverage verifies `abc 中文。` wraps as `["abc 中", "文。"]`, verifies opening punctuation, and asserts every line stays within the configured width.

src/cosh-ng/crates/cosh-shell/src/ui/wrap.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+pub(super) fn should_buffer_word_char(ch: char) -> bool {
+    ch.is_ascii() && !ch.is_whitespace() && !ch.is_control()
+}
+
+pub(super) fn is_line_closing_punctuation(ch: char) -> bool {
+    matches!(
+        ch,
+        '。' | '，' | '、' | '；' | '：' | '！' | '？' | ')' | ']' | '}' | '）' | '】' | '》'
```
ikunkun-sys: **[P2] Keep the remaining CJK line-closing punctuation off continuation-line starts**

This whitelist covers `。`/`，` but omits common closing quotation and bracket characters even though `is_cjk_breakable` tokenizes them independently. On this head, `wrap_plain_line("abc 中文”", 8)` returns `["abc 中文", "”"]`; `」` and `〉` reproduce the same orphaned continuation-line punctuation, while the parent returned `["abc", "中文”"]`. Opening quotes have the symmetric problem (`"abc 中「文"` becomes `["abc 中「", "文"]`). Thus the PR still introduces the same kinsoku regression for ordinary quoted CJK text that this binding pass is intended to prevent, and the current punctuation test does not cover it.

ikunkun-sys (changes requested):
Requesting changes for the inline P2 correctness issue reproduced against the current head. The current CI checks are green, and the branch also merges cleanly with the latest `main`. Separately, the PR description still cites obsolete commit `3aa33a48f089` for validation and rollback; please refresh those references.

src/cosh-ng/crates/cosh-shell/src/ui/wrap.rs
kongche-jbw: Fixed in e92b3b8d. Added the requested closing quotes/brackets (`”`, `’`, `」`, `』`, `〉`) and symmetric opening forms (`“`, `‘`, `「`, `『`, `〈`) to the punctuation binding rules. The table-driven regression now exercises every added character, including the reported `abc 中文”` and `abc 中「文` cases, and verifies that every resulting line remains within width 8. I also removed the unrelated test-file blank line and refreshed the PR description to reference the final revision, current 1312-test validation, and accurate ECS/manual scope.

src/cosh-ng/crates/cosh-shell/src/ui/wrap.rs
```diff
@@ -0,0 +1,330 @@
...
+            | ')'
+            | ']'
+            | '}'
+            | '）'
+            | '】'
+            | '》'
+            | '〉'
+            | '」'
+            | '』'
+            | '”'
```
ikunkun-sys, line 325: **[P2] Avoid overflowing streaming cards through the shared punctuation predicate**

`is_line_closing_punctuation` is also used by `StreamingAgentBlock::write_visible_char`, which skips a line break for matching characters even when `next_width > self.width`. Adding `”` here (and the other new closing quote/bracket characters) therefore lets streaming output exceed its configured width. With a 40-column rich renderer (36-column content), `"中".repeat(18) + "”"` previously produced two aligned 40-column rows, but this head emits one 41-column row; `」`/`』`/`〉` can make it 42 columns. `finish_partial_line` uses `saturating_sub`, so the right border shifts instead of recovering. This violates the no-overflow acceptance criterion, and the new `wrap_plain_line` tests do not exercise this streaming consumer.

ikunkun-sys (changes requested):
Requesting changes for the new inline P2 reproduced against the current head. The earlier `wrap_plain_line` quote/bracket finding is fixed, all current CI checks are green, and the branch is mergeable; however, expanding the shared closing-punctuation predicate now regresses streaming card width and right-border alignment.

src/cosh-ng/crates/cosh-shell/src/ui/wrap.rs
kongche-jbw: Fixed in d8e9bc95. I reproduced the reported rich-card regression first: `"中".repeat(18) + "”"` rendered a 41-column row at renderer width 40. Streaming now treats its configured width as a hard limit and always breaks before any character that would exceed it; the broader closing-punctuation predicate remains private to batch `wrap_plain_line`, where token binding can preserve kinsoku without overflowing. The existing streaming alignment test now covers `”`, `’`, `」`, `』`, and `〉`; all emitted non-empty rows remain exactly 40 columns. Full cosh-shell lib tests (1312), fmt, scoped Clippy, and test inventory (693/693) pass. The PR validation/rollback SHA is also refreshed to d8e9bc95c9b1.

src/cosh-ng/crates/cosh-shell/src/ui/agent_render/stream.rs
```diff
@@ -240,11 +239,7 @@ impl StreamingAgentBlock {
         }
 
         let next_width = self.current_line_width_with(ch);
-        if next_width > 0
-            && self.current_width > 0
-            && next_width > self.width
-            && !is_line_closing_punctuation(ch)
-        {
+        if next_width > 0 && self.current_width > 0 && next_width > self.width {
```
ikunkun-sys, line 273: **[P2] Preserve streaming line-closing punctuation without exceeding the width**

Making this break unconditional fixes the reported overflow, but it also removes the streaming renderer's pre-existing protection for closing punctuation. With a 40-column renderer (36-column content), `"中".repeat(18) + "。"` now emits the 18 ideographs on the first row and a lone `。` at the start of the continuation row; `，`, `）`, `”`, and `」` have the same boundary behavior. Before this revision these characters stayed with the preceding text, and the PR description still promises that closing punctuation will not be stranded at line boundaries. The new streaming test only checks that every row is 40 columns and that the punctuation is present, so it passes even when the punctuation is orphaned.

ikunkun-sys (changes requested):
Requesting changes for the new inline P2 reproduced against the current head. The previous streaming overflow and border-alignment issue is fixed, all CI checks are green, and the branch merges cleanly with current `main`; however, the hard-width fix removes the streaming renderer's existing protection against orphaned closing punctuation.

src/cosh-ng/crates/cosh-shell/src/ui/agent_render/stream.rs
kongche-jbw: Fixed in 8312d559. The streaming renderer now keeps a one-CJK-unit look-behind, so a following closing-punctuation run is emitted atomically with that unit. This preserves the hard width limit without orphaning `。`, `，`, `）`, `”`, `」`, or the other supported closing marks, including across `write_delta` boundaries. The regression now asserts the bound `中<punctuation>` continuation for all 18 supported marks, consecutive punctuation, normal ASCII continuation, and hard-width splitting for pathological long punctuation runs. Targeted stream tests, formatting, Clippy, and the test-inventory audit pass.

src/cosh-ng/crates/cosh-shell/src/ui/wrap.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+    bind_line_punctuation(tokens)
+}
+
+fn bind_line_punctuation(tokens: Vec<String>) -> Vec<String> {
+    let mut bound = Vec::<String>::with_capacity(tokens.len());
+
+    for token in tokens {
+        let is_text = token != "\n" && !token.chars().all(char::is_whitespace);
+        if is_text && token.chars().all(is_line_closing_punctuation) {
```
ikunkun-sys: **[P2] Bind closing punctuation before trailing ASCII text**

This condition only binds a token when every character is closing punctuation, but `split_wrap_tokens()` groups non-CJK ASCII into one token. Therefore a closer followed by an identifier is tokenized as `")request_id"` and bypasses this branch. On this head, `wrap_plain_line("abc 中文)request_id", 15)` returns `["abc 中文", ")request_id"]`; the parent returns `["abc", "中文)request_id"]`. This introduces a continuation line starting with closing punctuation, contrary to the PR's punctuation-binding contract. The current tests cover closers only at end of input, so this case remains untested.

ikunkun-sys (changes requested):
Requesting changes for one inline P2 reproduced against the current head `8312d559`. The previous streaming width and punctuation issue is fixed, all current CI checks are green, and the branch merges cleanly with current `main`; however, batch wrapping still strands a closing mark when it is followed by ASCII text.

src/cosh-ng/crates/cosh-shell/src/ui/wrap.rs
kongche-jbw: Fixed in c08011f9. `bind_line_punctuation` now treats a non-whitespace token whose first character is closing punctuation as one bound unit, so `)request_id` stays intact and attaches to the preceding CJK unit instead of starting a continuation line. The reported width-15 case now wraps as `["abc 中", "文)request_id"]`, preserving the PR’s CJK remaining-width behavior while keeping the identifier whole and every line within the configured width. The focused six-test wrap suite, formatting, scoped Clippy, and diff check pass; the PR validation and rollback SHA were refreshed.

ikunkun-sys (approved):
Re-reviewed current head `c08011f9`. The previous batch-wrapping P2 is fixed with focused regression coverage; affected local checks pass, current CI is green, and no new blocking findings remain.

PR #2451, opened 2026-08-12
fix(sec-core): add validation on input params

RemindD:
## Why

Some input params don't have validation at entrance which may cause unexpected behavior or unclear error message.

## What changed

Add validations on input params at both cli and daemon.

## Related issue

fixes https://github.com/alibaba/anolisa/issues/2390
## Risk and compatibility


edonyzpc (approved):
LGTM

PR #2449, opened 2026-08-12
feat(skillfs): authenticate container peers

kongche-jbw:
## Why

SkillFS currently authenticates a local control peer by executable identity. That
does not work across container PID and mount namespaces, while accepting an
unauthenticated Unix socket peer would weaken the existing security boundary.

This change adds an opt-in, shared-secret HMAC mode owned entirely by SkillFS.
The peer-side implementation remains a separate sec-core follow-up after its
maintainers confirm the proposed contract.

## What changed

- Add mutual HMAC-SHA256 authentication to the SkillFS control server and notify
  client, with domain-separated proofs and fixed cross-language vectors.
- Keep the existing executable-identity control path and plain notify path
  unchanged when the new key flags are absent.
- Load keys fail-closed from strict regular files and enforce one total
  authentication deadline, including slow and replay rejection coverage.
- Add the proposed peer contract, a unilateral real-environment validation plan,
  and an independent Python protocol probe.
- Do not modify any `src/agent-sec-core/**` source, test, or documentation file.

## Related issue

Refs #2439 and #2012. This draft does not close either issue because peer-side
implementation, contract ACK, and ACK-cluster evidence are still pending.

## User / Agent impact

SkillFS gains two opt-in flags: `--trusted-peer-key-file` for its control server
and `--notify-auth-key-file` for its notify client. Existing host deployments are
unchanged. This PR does not configure or implement agent-sec-core support.

## Risk and compatibility

- [x] Public CLI, API, configuration, or documented behavior changed
- [x] Privileged or security-sensitive behavior changed
- [x] Cross-component contract changed
- [x] Migration or rollback guidance is needed

The new behavior is gated by explicit key-file flags. Half-configured HMAC paths
fail during startup; the existing host executable authentication and unauthenticated
notify behavior remain the defaults. The peer contract is proposed for review and
is not presented as an implemented sec-core API.

## Validation

- `cargo +1.86.0 fmt --all -- --check`
- `cargo +1.86.0 clippy --workspace --all-targets -- -D warnings`
- `cargo +1.86.0 test --workspace`
- `cargo +1.86.0 doc --workspace --no-deps`
- `src/skillfs/scripts/test.sh` on Linux with `/dev/fuse`
- `scripts/docs-lint.sh`
- `python3 scripts/docs-link-check.py`
- Independent probe formatting/lint and `container-peer-auth-probe.py self-test`
- Manual real-FUSE probe: valid mutual auth, plain/wrong-key/replay/slow-peer
[...]

PR #2493, opened 2026-08-13
feat(sec-core): add SkillFS HMAC peer auth

1570005763:
## Why

Skill Ledger and SkillFS currently rely on same-host peer identity. In ACS, Ledger runs in
the main container while SkillFS runs as a sidecar, so the resolver and reverse-notify paths
need an explicit cross-container trust contract without silently downgrading authenticated
deployments.

## What changed

- Add a Skill Ledger-scoped HMAC-SHA256 protocol implementation compatible with SkillFS
  PR #2449.
- Authenticate the control resolver and daemon notify paths over their raw NDJSON frames.
- Preserve legacy `ENOENT` host fallback only when control authentication is not configured;
  configured authentication fails closed.
- Keep unrelated daemon APIs and deployments without HMAC configuration backward compatible.

## Related issue

no-issue: follows up on SkillFS PR #2449 for ACS sidecar interoperability

Depends on #2449.

## User / Agent impact

ACS deployments can configure authenticated Skill Ledger-SkillFS communication with:

- `AGENT_SEC_SKILLFS_CONTROL_SOCKET`
- `AGENT_SEC_SKILLFS_CONTROL_AUTH_KEY_FILE`
- `AGENT_SEC_SKILLFS_NOTIFY_AUTH_KEY_FILE`

Authentication keys must be absolute-path, non-symlink regular files owned by the current
effective UID with no group or other permissions. Configuring a control key makes a missing
or unreachable SkillFS socket an error instead of triggering host fallback.

## Risk and compatibility

The authenticated wire contract must remain compatible with SkillFS PR #2449. Legacy behavior
is unchanged when no HMAC key is configured; authenticated deployments intentionally fail
closed on socket, handshake, or frame-authentication failures.

## Validation

- `make python-code-pretty`
- `make test-python`
  - core: 3704 passed, 3 skipped, 20 subtests passed
  - CLI e2e: 87 passed, 2 skipped
  - daemon e2e: 12 passed, 6 skipped
  - Skill Ledger e2e: 50 passed
- `bash scripts/docs-lint.sh`
- `python3 scripts/docs-link-check.py`
- `git diff --check upstream/main...HEAD`
- SkillFS PR #2449 fixed-vector probe at commit `f6e979432e1bc44a4abf67af86321bb8942b59bc`:
  all handshake and business-frame HMAC vectors matched

The real SkillFS FUSE binary and Kubernetes two-container deployment were not exercised on
this macOS development host.

## Documentation and rollback

Updated the English and Chinese component READMEs, Skill Ledger user guides, and the existing
Skill Ledger-SkillFS integration design. To disable the integration, unset the three new
environment variables; to remove the implementation entirely, revert this commit. ACS HMAC
interoperation requires the corresponding SkillFS implementation from #2449.

### DIFF
--- a/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/server.py
+++ b/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/server.py
@@ -21,6 +21,7 @@ from agent_sec_cli.daemon.errors import (
     DaemonRuntimePathError,
     DaemonTimeoutError,
     InternalDaemonError,
+    PayloadTooLargeError,
     ResponseTooLargeError,
     ShutdownError,
 )
@@ -36,6 +37,7 @@ from agent_sec_cli.daemon.handlers.security_query import (
     register_security_query_methods,
 )
 from agent_sec_cli.daemon.handlers.skill_ledger import (
+    METHOD_SKILLFS_NOTIFY_CHANGE,
     register_skill_ledger_methods,
 )
 from agent_sec_cli.daemon.health import register_health_methods
@@ -45,7 +47,6 @@ from agent_sec_cli.daemon.protocol import (
     DEFAULT_MAX_REQUEST_BYTES,
     DEFAULT_MAX_RESPONSE_BYTES,
     DaemonResponse,
-    NDJSONFrameParser,
     error_response,
     generate_request_id,
     parse_request_line,
@@ -58,12 +59,82 @@ from agent_sec_cli.daemon.runtime import (
     lock_path_for_socket,
     resolve_socket_path,
 )
+from agent_sec_cli.skill_ledger.skillfs_peer_auth import (
+    AUTH_FRAME_LIMIT,
+    AUTH_HANDSHAKE_TIMEOUT_SECONDS,
+    NOTIFY_CLIENT_DOMAIN,
+    NOTIFY_SERVER_DOMAIN,
+    AuthenticatedSession,
+    SharedSecret,
+    SkillFsPeerAuthError,
+    build_auth_challenge_frame,
+    build_auth_proof_frame,
+    classify_initial_frame,
+    configured_notify_key_path,
+    verify_auth_proof_frame,
+)
 
 LOGGER = logging.getLogger("agent-sec-core.daemon")
 DEFAULT_MAX_CONNECTIONS = 64
 DEFAULT_DRAIN_TIMEOUT_SECONDS = 2.0
 DEFAULT_REQUEST_READ_TIMEOUT_MS = 5000
 SocketIdentity = tuple[int, int]
+_AUTH_WIRE_FRAME_LIMIT = AUTH_FRAME_LIMIT + 1
+
+
+class _SilentClose(Exception):
+    """Close an unauthenticated or invalid authenticated connection quietly."""
+
+
+class _ConnectionFrameReader:
+    """Read bounded NDJSON frames while retaining coalesced trailing frames."""
+
+    def __init__(self, reader: asyncio.StreamReader) -> None:
+        self._reader = reader
+        self._buffer = bytearray()
+
+    async def read_frame(
+        self,
+        max_frame_bytes_including_delimiter: int,
+        *,
+        deadline: float | None,
+        allow_eof_terminated: bool,
+    ) -> bytes:
+        """Read one frame with a wire-size limit that includes the delimiter."""
+        while True:
+            newline_index = self._buffer.find(b"\n")
+            if newline_index >= 0:
+                frame_size = newline_index + 1
+                if frame_size > max_frame_bytes_including_delimiter:
+                    raise PayloadTooLargeError(max_frame_bytes_including_delimiter)
+                frame = bytes(self._buffer[:frame_size])
+                del self._buffer[:frame_size]
+                return frame
+
+            if len(self._buffer) > max_frame_bytes_including_delimiter:
+                raise PayloadTooLargeError(max_frame_bytes_including_delimiter)
+
+            chunk = await self._read_chunk(deadline)
+            if chunk:
+                self._buffer.extend(chunk)
+                continue
+
+            if self._buffer and allow_eof_terminated:
+                frame = bytes(self._buffer)
+                self._buffer.clear()
+                return frame
+            if self._buffer:
+                raise BadRequestError("incomplete daemon request")
+            raise BadRequestError("empty daemon request")
+
+    async def _read_chunk(self, deadline: float | None) -> bytes:
+        if deadline is None:
+            return await self._reader.read(4096)
+
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            raise asyncio.TimeoutError
+        return await asyncio.wait_for(self._reader.read(4096), timeout=remaining)
 
 
 def create_default_registry() -> MethodRegistry:
@@ -142,9 +213,14 @@ class DaemonServer:
         self._drain_timeout_seconds = DEFAULT_DRAIN_TIMEOUT_SECONDS
         self._previous_umask: int | None = None
         self._socket_identity: SocketIdentity | None = None
+        self._notify_auth_secret: SharedSecret | None = None
 
     async def start(self) -> None:
         """Prepare runtime paths, bind the Unix socket, and start jobs."""
+        notify_key_path = configured_notify_key_path()
+        self._notify_auth_secret = (
+            None if notify_key_path is None else SharedSecret.load(notify_key_path)
+        )
         self._set_daemon_umask()
         try:
             self._lock = prepare_socket_path(self.socket_path)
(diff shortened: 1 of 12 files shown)


src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/skill_ledger/core/live_root.py
```diff
@@ -179,6 +273,72 @@ def resolve(self, canonical_dir: Path) -> Path | None:
...
+                    nonce=nonce,
+                )
+
+                session = AuthenticatedSession(
+                    self._control_secret,
+                    nonce,
+                    CONTROL_CLIENT_DOMAIN,
+                    CONTROL_SERVER_DOMAIN,
+                )
+                business_deadline = time.monotonic() + self.timeout_seconds
```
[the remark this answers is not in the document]
1570005763: This is the intended fail-closed boundary rather than a fallback mismatch. `default_skillfs_control_socket()` always returns a `Path`; without a control key, connect-time `ENOENT` becomes `_ResolverSocketMissing` and falls back to the canonical host path, while a configured key makes the same condition an authentication availability failure because the deployment has declared trusted SkillFS mandatory. The README matrix and `test_resolver_fails_closed_when_hmac_socket_is_missing` explicitly cover both default and explicit sockets. Falling back in the keyed case would introduce an authentication downgrade, so no code change is needed here.

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/skill_ledger/skillfs_peer_auth.py
```diff
@@ -0,0 +1,428 @@
...
+    value = os.environ.get(name)
+    if value is None:
+        return None
+    if not value.strip():
+        raise SkillFsPeerAuthError(f"{name} must not be empty")
+    return Path(value)
+
+
+def _read_bounded_secret(file_descriptor: int) -> bytes:
+    remaining = MAX_SECRET_LENGTH + 1
```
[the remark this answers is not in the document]
1570005763: The `MAX_SECRET_LENGTH + 1` read is an intentional sentinel-byte check, not truncation followed by acceptance. The loader reads at most 4097 bytes, and `SharedSecret` immediately rejects every value outside 32–4096 bytes, so any larger file is rejected before the key can be used; reading the rest would only weaken the bounded-read property. The existing oversized-key test covers the boundary, an additional 8192-byte probe was also rejected, and SkillFS PR #2449 uses the same bounded sentinel pattern. No code change is needed here.

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/server.py
```diff
@@ -286,40 +447,127 @@ async def _process_client(
...
+        )
+
+    async def _read_authenticated_request(
+        self,
+        frame_reader: _ConnectionFrameReader,
+        session: AuthenticatedSession,
+        *,
+        started: float,
+    ) -> tuple[bytes, int]:
+        deadline = started + (self.request_read_timeout_ms / 1000)
```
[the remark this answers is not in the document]
1570005763: The two-stage budget is intentional and matches both the control resolver in this PR and SkillFS PR #2449: the complete handshake is bounded by `min(request_read_timeout, 5s)`, then an authenticated business frame receives the normal request-read budget. A peer without the key cannot enter the second stage, authentication failures close the connection and release the slot, and compromised key-bearing peers are outside the current threat model. A single pre-dispatch deadline could be considered as future availability hardening, but it is not a P1 issue and changing only the daemon would make the two sides inconsistent, so we are keeping the current contract in this PR.

kongche-jbw (commented):
Review baseline: `cbd7b2489a99...bc3d29f01d72`

未发现阻塞问题。

Remaining risks:

- 未在真实 SkillFS FUSE 与 Kubernetes 双容器环境执行端到端互操作。
- 本地未构建可编辑包或运行全量套件；GitHub 的 sec-core 构建和测试检查均通过。

Verification scope:

- 已审查完整 diff、相邻实现与测试、PR 描述、全部评论和 resolved threads。
- 定向 pytest 149 passed；Black、Ruff、docs lint、link check、diff check 通过。

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/skill_ledger/skillfs_peer_auth.py
```diff
@@ -0,0 +1,428 @@
...
+    def frame_proof(self, domain: str, nonce: bytes, payload: bytes) -> bytes:
+        """Return a direction- and session-bound business-frame proof."""
+        _validate_nonce(nonce)
+        message = (
+            domain.encode("utf-8")
+            + b"\0frame\0"
+            + nonce
+            + len(payload).to_bytes(8, byteorder="big")
+            + payload
+        )
```
edonyzpc, line 121: `frame_proof` 的 MAC 输入为 `domain + "\0frame\0" + nonce + len + payload`。同一会话内所有帧使用相同 nonce，无递增序号。当前协议为单请求-单响应不可利用，但若未来扩展为多帧交换，攻击者可重放或重排帧。

建议: 在 MAC 输入中加入方向内递增序号，或在协议规范中明确限制每方向仅允许一帧。

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/skill_ledger/core/live_root.py
```diff
@@ -147,27 +197,71 @@ def __init__(
...
+            return self._parse_response(canonical_dir, response_payload)
+
+        response_line = self._resolve_legacy(payload)
+        return self._parse_response(canonical_dir, response_line)
+
+    def _load_configuration(self) -> None:
+        if self._configuration_loaded:
+            if self._configuration_error is not None:
+                raise self._configuration_error
+            return
```
edonyzpc, line 237: [NIT] 使用 `_configuration_loaded` 布尔标记实现一次性加载缓存，无 `threading.Lock` 保护。在 CPython GIL 下实际安全，但 free-threaded CPython 3.13+ 下可能出现竞态：线程 A 看到 `_configuration_loaded = True` 但 `_control_secret` 尚未写入，依赖实现细节而非语言保证。建议：添加 `threading.Lock` 或使用 `@functools.cached_property` 模式

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/server.py
```diff
@@ -247,16 +323,101 @@ async def _process_client(
...
-            bytes_in = len(line)
-            request = parse_request_line(line, max_request_bytes=self.max_request_bytes)
+            # A partial first frame cannot yet be classified safely, so a
+            # configured peer key bounds that pre-authentication hold time.
+            try:
+                initial_frame = await frame_reader.read_frame(
+                    max(self.max_request_bytes, _AUTH_WIRE_FRAME_LIMIT),
+                    deadline=initial_deadline,
+                    allow_eof_terminated=True,
+                )
```
edonyzpc, line 350: 配置 `_notify_auth_secret` 后，每个连接在首帧读取时使用 `min(request_timeout, 5s)` 超时。攻击者可开启 64 个连接（`DEFAULT_MAX_CONNECTIONS`），发送 `auth.init` 后不再发送数据，每个连接占用最多 5 秒。期间所有新连接收到 `BusyError`。

建议：
- 认证握手使用独立连接计数
- 或在 `auth.init` 之后使用更短超时（如 1 秒）等待 client proof

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/server.py
```diff
@@ -342,6 +590,12 @@ async def _write_response(
             if len(raw_response) > self.max_response_bytes:
                 raw_response = b""
 
+        if raw_response and authenticated_session is not None:
+            raw_response = authenticated_session.protect_payload(
+                raw_response[:-1],
+                "server",
+            )
+
         bytes_out = 0
```
edonyzpc, line 599: 当 response 超过 `max_response_bytes` 被截断为 `b""` 时：
1. `raw_response and ...` 为 `False`，跳过签名
2. 后续 `if raw_response:` 也为 `False`，不写任何数据
3. 客户端收到 EOF，无法区分 "response 过大" 和 "连接被恶意中断"

在认证模式下，应返回签名的错误响应帧，使客户端能验证错误确实来自合法 server。建议超限时构造一个签名的错误响应（如 `{"error": "response_too_large"}`），而非静默关闭。

edonyzpc (changes requested):

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/skill_ledger/skillfs_peer_auth.py
1570005763: Good future-proofing point. Auth v1 is intentionally single-use: each connection carries one authenticated client request and one authenticated server response. The resolver opens a fresh socket for every resolve, the daemon dispatches once and then closes the connection, and the governing SkillFS #2449 authentication specification defines the same single-use contract. A fresh nonce prevents cross-connection replay, while direction-specific domains prevent reflection; there is no second same-direction frame to reorder in v1. Adding a counter now would change the bilateral wire contract and fixed vectors. Any future multi-frame version should add per-direction sequencing together with an auth-version update, so no code change is needed for v1.

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/skill_ledger/core/live_root.py
1570005763: Thanks for raising the future runtime concern. agent-sec-core currently pins CPython 3.11.6 and does not support 3.12+, while the production resolution path creates a one-shot resolver client rather than sharing an instance across threads. The current initialization order also writes the secret or cached error before setting `_configuration_loaded = True`. Concurrent external reuse could duplicate the load, but it does not affect the supported production path or stable configuration semantics. If free-threaded Python and shared clients are supported later, this should use an explicit per-instance lock; `cached_property` does not provide an exactly-once guarantee on Python 3.12+. No change is needed in the current compatibility scope.

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/server.py
1570005763: The connection-slot exhaustion scenario is real, but it is a pre-existing daemon-wide availability property rather than an HMAC-specific regression. `upstream/main` already holds the same 64 global slots while waiting up to five seconds for the first request frame, so 64 idle or partial plaintext connections can produce the same `BusyError` behavior. The authenticated path uses one absolute handshake deadline and releases the slot in `finally`, so it does not leak or extend that hold indefinitely. A one-second proof timeout would not address empty or partial first frames, and a separate authentication counter cannot classify those connections before the first frame is available. Socket ownership and runtime-volume isolation remain the first availability boundary; broader daemon admission control can be tracked separately, so no HMAC-specific change is needed in this PR.

src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/server.py
1570005763: This path only produces EOF when even the structured `ResponseTooLargeError` response exceeds `max_response_bytes`. For a normal oversized response, the daemon first replaces it with that bounded error response and then signs it when the connection is authenticated. The empty-response fallback predates this PR, the default limit is 4 MiB, and the notify response is fixed and small, so the remaining case requires an abnormally low configured limit. EOF remains fail-closed, and HMAC cannot guarantee that a client can distinguish a local size-limit failure from an active peer dropping the connection. A minimum configuration bound could be considered as separate daemon hardening, but this is not an authentication defect introduced by the PR.

edonyzpc (approved):
LGTM

Issue #2512, opened 2026-08-13
[tokenless] flaky: test_detect_installed_framework_ready[claude-code] — detect script timing race


zhangtaibo:
The test test_detect_installed_framework_ready[claude-code] fails intermittently. Root cause: detect script has filesystem/PATH initialization timing issue on first execution. 14/15 stable reproductions, only this 1 case flaky.

Root cause: detect.sh checks $HOME/.claude directory existence, but first execution may have filesystem state/PATH initialization timing issue.

Suggestion: Add retry mechanism or pre-warm detect script before assertion.

Nightly: nightly-20260814-020042, ECS: 47.242.86.116, Component: tokenless

zhangtaibo:
经 nightly run nightly-20260814-020042 复审,本 issue 判定为 **flaky误判**,不应作为 product bug 提 GitHub issue,关闭处理。

理由:
- title/正文自述 "flaky ... detect script timing race ... intermittently",即测试 timing 不稳定,非确定性产品 bug
- flaky 测试应回 Step 1 重分类:加 retry/rerunfailures 让用例稳定,或在 case baseline 标 flaky-known-recurring 让 triage 跳过
- 不应该开 GitHub issue 让产品维护者处理测试侧的 timing 问题

后续处理走测试侧,本 issue 关闭。

Forrest-ly:
已修复，详见 PR #2519。

PR #2519, opened 2026-08-14
fix(tokenless): retry transient detect failures

Forrest-ly:
## Summary

Defensive hardening of the claude-code `detect.sh` against first-run initialization races.

Background: #2512 reported an intermittent nightly failure of `test_detect_installed_framework_ready[claude-code]`, attributed to a filesystem/PATH initialization timing race on detect.sh's first execution. That issue was closed as a **test-side flaky misclassification** (see the closing comment on #2512) and no failing trace was attached. Following the review's elimination analysis, the `~/.claude` config-dir probe is informational only and cannot change the exit status — so the only exit-status-affecting check that can race under the reported `$HOME/.claude`/PATH visibility mechanism is the claude binary lookup itself: a binary that is not yet visible makes detect.sh exit 2 (missing prerequisites), which is exactly what turns a "framework ready" assertion into a failure.

This PR therefore:

- Retries the transient first-run failures: the claude binary not yet visible under `$HOME/.local/bin` (PATH is fixed within one run, but the filesystem is not — the file itself can appear mid-execution, and `command -v` re-scans on every attempt), and a `claude plugin list` invocation failing while the CLI initializes `~/.claude`.
- Never retries definitive results: a **successful** `plugin list` that omits the plugin is a definitive not-installed and is reported after exactly one invocation — that is the ordinary pre-install state, which previously slept through the whole retry budget (~3.02s, 4 CLI invocations).
- Pins the behavior with a regression test anchored on that reported failure path (binary appearing late), with a no-retry control reproducing the nightly failure mode (exit 2, missing prerequisites), plus cases pinning the plugin probe's retry semantics.

## Changes

`src/tokenless/adapters/tokenless/claude-code/scripts/detect.sh`:

- `settle()` retry helper with bounded, tunable retries (`TOKENLESS_DETECT_RETRIES`, default 3; `TOKENLESS_DETECT_RETRY_DELAY`, default 1s). It retries only while the check reports a retryable failure (exit status 1); any other status is definitive and returned immediately (at most `1 + DETECT_RETRIES` attempts in total).
- The claude binary lookup goes through `settle()`: it catches a binary appearing while detect.sh settles. A genuinely absent CLI cannot be distinguished from that race, so it spends one retry budget (documented; `TOKENLESS_DETECT_RETRIES=0` opts out).
- `claude_plugin_listed` classifies the outcome: 0 = plugin listed; 1 = the `plugin list` invocation itself failed (retryable first-run init); 2 = a successful list that omits the plugin (definitive absent, never retried).
- The `~/.claude` config-dir check stays a single informational probe; nothing in the script creates the directory before `claude plugin list` runs.
- The tri-state exit contract (0 ready / 1 installable / 2 missing prereqs) is unchanged.

`src/tokenless/Makefile`:

- The `claude-code-install` pre-install detect call now runs with `TOKENLESS_DETECT_RETRIES=0`: that probe is informational only (`|| true`, install.sh makes the final call) and runs on a steady-state machine, so it returns immediately.
- Registers `make test-claude-code-detect` and wires it into the `test` aggregate.

`src/tokenless/tests/test-claude-code-detect-retry.sh` (new):

- Case 1: the claude binary becomes visible in `$HOME/.local/bin` only after a delay (background provisioner simulating the provisioning race); `~/.claude` does not exist until the CLI's first `plugin list`. With retries, detect.sh exits 0 and reports `installed`/`ready`; the test also asserts the config-dir probe observed the not-yet-created `~/.claude` without affecting readiness.
- Case 2 (control): the same first execution with retries disabled exits 2 with missing prerequisites — the nightly framework-ready failure mode — proving the retries are what fix it.
- Case 3: definitive plugin-absent — the CLI runs successfully but never lists the plugin → detect.sh exits 1 after **exactly one** `plugin list` invocation regardless of the retry budget.
- Case 4: a `plugin list` invocation that fails outright is transient and is retried → exit 0.

## Testing

Environment: Linux x86_64, GNU bash.

- `bash -n` on `detect.sh` and the new test script — pass.
- `make -C src/tokenless test-claude-code-detect` — pass (all 4 cases).
- Manual steady-state latency verification (default retries=3, delay=1s): plugin-absent path exits 1 in ~0.02s with exactly one CLI invocation (previously ~3.02s and 4 invocations); the `make claude-code-install` pre-install probe returns immediately (retries disabled).
- Manual first-run race verification: a binary appearing mid-execution is found on the first retry (elapsed >= one retry delay), exit 0 ready.
- Not run: the full `make -C src/tokenless test` aggregate (requires the full toolchain/npm environment); the change is confined to one adapter detect script, its dedicated test, and one opt-out env var on an informational Makefile call.

### DIFF
--- a/src/tokenless/adapters/tokenless/claude-code/scripts/detect.sh
+++ b/src/tokenless/adapters/tokenless/claude-code/scripts/detect.sh
@@ -16,6 +16,33 @@ PLUGIN_SRC="$ADAPTER_DIR/claude-code"
 CLAUDE_BIN="${CLAUDE_BIN:-}"
 export PATH="$HOME/.local/bin:/usr/local/bin:$PATH"
 
+# First-run settling retries: right after provisioning, the claude binary or
+# the plugin registry may be transiently invisible on the very first detect.sh
+# execution (filesystem/PATH init timing race). settle() only retries checks
+# that report a retryable failure (exit status 1); checks that succeed or
+# report a definitive result return immediately, so steady-state runs stay
+# fast.
+DETECT_RETRIES="${TOKENLESS_DETECT_RETRIES:-3}"
+DETECT_RETRY_DELAY="${TOKENLESS_DETECT_RETRY_DELAY:-1}"
+
+# settle <cmd...> — run cmd once; if it reports a retryable failure (exit
+# status 1), sleep DETECT_RETRY_DELAY and retry, up to DETECT_RETRIES retries
+# (that is, at most 1 + DETECT_RETRIES attempts in total). Exit status 0 is
+# success. Any other exit status is a definitive result and is returned
+# immediately without further retries. Returns the final exit status.
+# Callers must therefore reserve exit status 1 for conditions that a retry
+# may still resolve.
+settle() {
+    local retry=0 rc=0
+    "$@"; rc=$?
+    while [ "$rc" -eq 1 ] && [ "$retry" -lt "$DETECT_RETRIES" ]; do
+        retry=$((retry + 1))
+        sleep "$DETECT_RETRY_DELAY"
+        "$@"; rc=$?
+    done
+    return "$rc"
+}
+
 line()  { printf '[%s] %s\n' "$COMPONENT" "$*"; }
 field() { printf '[%s]   %-26s %s\n' "$COMPONENT" "$1" "$2"; }
 
@@ -24,8 +51,17 @@ INSTALL_MISSING=()
 note_prereq_missing()  { PREREQ_MISSING+=("$1"); }
 note_install_missing() { INSTALL_MISSING+=("$1"); }
 
-if [ -z "$CLAUDE_BIN" ]; then
+find_claude_bin() {
     CLAUDE_BIN="$(command -v claude 2>/dev/null || true)"
+    [ -n "$CLAUDE_BIN" ]
+}
+
+if [ -z "$CLAUDE_BIN" ]; then
+    # PATH/filesystem may still be settling right after install; re-check
+    # briefly before declaring the CLI missing. A genuinely absent CLI cannot
+    # be distinguished from that settling race, so it spends one retry budget
+    # (set TOKENLESS_DETECT_RETRIES=0 to opt out).
+    settle find_claude_bin || true
 fi
 
 line "${AGENT} detect"
@@ -38,7 +74,10 @@ else
 fi
 
 # Informational only: claude creates ~/.claude on first run; absence is
-# not a prerequisite failure.
+# not a prerequisite failure. Check once, without settling retries: nothing in
+# this script creates the directory before `claude plugin list` runs, and that
+# call carries its own retry budget and initializes ~/.claude itself, so
+# retrying this probe would only delay the report.
 if [ -d "$HOME/.claude" ]; then
     field "claude config dir" "present ($HOME/.claude)"
 else
@@ -58,8 +97,30 @@ else
     field "plugin.json"       "missing (run: make stamp-adapter-templates)"
 fi
 
+# claude_plugin_listed — probe the plugin registry, using the settle()
+# exit-status contract: 0 = plugin listed; 1 = `claude plugin list` itself
+# failed (the CLI may still be initializing ~/.claude on first run, so a
+# retry may still succeed); 2 = `plugin list` ran successfully but did not
+# list the plugin. That is a definitive absent result — no retry can change
+# it — so settle() must return immediately instead of sleeping out the
+# retry budget and invoking the CLI DETECT_RETRIES + 1 times.
+claude_plugin_listed() {
+    local listing
+    if ! listing="$("$CLAUDE_BIN" plugin list 2>&1)"; then
+        return 1
+    fi
+    if printf '%s\n' "$listing" | grep -qF "$PLUGIN_ID"; then
+        return 0
+    fi
+    return 2
+}
+
 if [ -n "$CLAUDE_BIN" ] && [ -x "$CLAUDE_BIN" ]; then
-    if "$CLAUDE_BIN" plugin list 2>&1 | grep -qF "$PLUGIN_ID"; then
+    # First-run race: `claude plugin list` may transiently fail while the CLI
+    # initializes ~/.claude; settle() retries while that failure (status 1)
+    # persists. A successful list that simply omits the plugin is definitive
+    # (status 2) and is reported as "not installed" without further retries.
+    if settle claude_plugin_listed; then
         field "plugin install"    "installed ($PLUGIN_ID)"
     else
         field "plugin install"    "not installed"
--- a/src/tokenless/tests/test-claude-code-detect-retry.sh
+++ b/src/tokenless/tests/test-claude-code-detect-retry.sh
@@ -1,0 +1,238 @@
+#!/usr/bin/env bash
+# Regression test for the detect.sh first-run settling retries.
+#
+# Background: GH #2512 reported test_detect_installed_framework_ready
+# [claude-code] flaking in nightly runs. The reported cause was a
+# filesystem/PATH initialization timing race on the first detect.sh
+# execution right after provisioning: the claude binary under
+# $HOME/.local/bin and the $HOME/.claude state dir are transiently
+# invisible. #2512 itself was closed as a test-side flaky misclassification
+# (issuecomment-5288391676), but the detect.sh-side weakness it exposed is
+# real, and only one exit-status-affecting check can actually race: the
+# claude binary lookup. A binary that is not yet visible makes detect.sh
+# exit 2 ("missing prerequisites"), which is exactly what turns a
+# "framework ready" assertion into a failure. (The $HOME/.claude config-dir
+# probe is informational only and never changes the exit status.)
+#
+# Scenario 1 therefore reproduces that real failure path: the claude binary
+# becomes visible in $HOME/.local/bin only after a delay (simulated with a
+# background provisioner), and $HOME/.claude does not exist until the CLI's
+# first `plugin list`. With settling retries, detect.sh must ride out the
+# race and report ready (exit 0); without retries (scenario 2) the same
+# first execution must fail with exit 2, proving the retries are what fix
+# it. Scenarios 3-4 pin the plugin probe's retry semantics: a successful
+# `plugin list` that omits the plugin is a definitive result and must not
+# consume the retry budget, while a failing `plugin list` is transient and
+# is retried.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(CDPATH='' cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+DETECT="$SCRIPT_DIR/../adapters/tokenless/claude-code/scripts/detect.sh"
+TEST_DIR="$(mktemp -d)"
+
+FAKE_HOME="$TEST_DIR/home"
+FAKE_BIN="$FAKE_HOME/.local/bin"
+SHARED_HOOKS="$FAKE_HOME/.local/share/anolisa/adapters/tokenless/common/hooks"
+PLUGIN_ID="tokenless@anolisa-tokenless"
+CALL_LOG="$TEST_DIR/claude-calls.log"
+STUB_MODE_FILE="$TEST_DIR/stub-mode"
+FLAKY_MARKER="$TEST_DIR/flaky-marker"
+PROVISIONER_PID=""
+
+cleanup() {
+    if [ -n "$PROVISIONER_PID" ]; then
+        kill "$PROVISIONER_PID" 2>/dev/null || true
+        wait "$PROVISIONER_PID" 2>/dev/null || true
+    fi
+    rm -rf "$TEST_DIR"
+}
+trap cleanup EXIT
+
+mkdir -p "$FAKE_BIN" "$SHARED_HOOKS"
+
+# detect.sh also requires `tokenless` and `rtk` on PATH (their absence is a
+# prerequisite failure). Provide isolated stubs so this test is
+# self-contained: run_detect restricts PATH to the fake home plus system
+# directories, and a fresh runner or detached worktree may not have either
+# binary installed globally.
+printf '#!/bin/sh\nexit 0\n' >"$FAKE_BIN/tokenless"
+printf '#!/bin/sh\nexit 0\n' >"$FAKE_BIN/rtk"
+chmod +x "$FAKE_BIN/tokenless" "$FAKE_BIN/rtk"
+
+fail() {
+    echo "FAIL: $1" >&2
+    printf '%s\n' "${2:-}" >&2
+    exit 1
+}
+
+# Stub claude CLI. `plugin list` behaviour is steered via $STUB_MODE_FILE:
+#   ready  — the list succeeds and contains the tokenless plugin (default)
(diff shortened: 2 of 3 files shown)


src/tokenless/adapters/tokenless/claude-code/scripts/detect.sh
```diff
@@ -16,6 +16,30 @@ PLUGIN_SRC="$ADAPTER_DIR/claude-code"
...
+# its ~/.claude state dir, or the plugin registry may be transiently
+# invisible on the very first detect.sh execution (filesystem/PATH init
+# timing race). Only failing checks retry, so steady-state runs stay fast.
+DETECT_RETRIES="${TOKENLESS_DETECT_RETRIES:-3}"
+DETECT_RETRY_DELAY="${TOKENLESS_DETECT_RETRY_DELAY:-1}"
+
+# settle <cmd...> — run cmd; if it fails, sleep DETECT_RETRY_DELAY and retry
+# up to DETECT_RETRIES extra times. Returns 0 once cmd eventually succeeds.
+settle() {
+    local attempt=1
```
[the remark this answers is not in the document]
Forrest-ly: 已采纳，修复见 5a8fa7a0：

- `settle()` 循环计数器由 `attempt` 改名为 `retry`，条件改为 `retry < DETECT_RETRIES`（每次重试前先自增），计数器语义现在明确表示"首次尝试之后已执行的重试次数"，不再与总尝试次数混淆；
- 函数注释同步更新，明确说明总尝试次数上限为 `1 + DETECT_RETRIES`（1 次首次尝试 + 至多 DETECT_RETRIES 次重试）。

行为保持不变：默认配置下仍是 1 + 3 = 4 次尝试，回归测试 `tests/test-claude-code-detect-retry.sh` 通过。

src/tokenless/adapters/tokenless/claude-code/scripts/detect.sh
```diff
@@ -39,6 +70,11 @@ fi
 
 # Informational only: claude creates ~/.claude on first run; absence is
 # not a prerequisite failure.
+if [ ! -d "$HOME/.claude" ]; then
+    # claude creates ~/.claude during first-run init; re-check briefly so a
+    # freshly provisioned environment is not reported as missing.
+    settle test -d "$HOME/.claude" || true
```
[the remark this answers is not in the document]
Forrest-ly: Adopted — fixed in 5a8fa7a0. The config-dir probe is now a single `test -d` without settling retries.

Rationale: nothing in detect.sh creates `~/.claude` before `claude plugin list` runs, and that call carries its own retry budget and initializes `~/.claude` itself, so retrying the purely informational probe only added latency without changing the outcome. Also updated the header comment accordingly (the retry budget now covers only the CLI-discovery and plugin-list checks).

Measured on a machine with no claude CLI and no `~/.claude`: detect time dropped from ~6s to ~3s. Regression test `tests/test-claude-code-detect-retry.sh` still passes (both the retry-enabled and retries-disabled control cases).

src/tokenless/tests/test-claude-code-detect-retry.sh
```diff
@@ -0,0 +1,86 @@
...
+# call fails while ~/.claude is being initialized; later calls succeed.
+cat >"$FAKE_BIN/claude" <<'STUB'
+#!/bin/sh
+MARKER="$HOME/.claude/.first-run-done"
+case "${1:-}" in
+--version)
+    echo "claude 9.9.9-test"
+    ;;
+plugin)
+    if [ ! -f "$MARKER" ]; then
```
ikunkun-sys: [P1] Ground this regression test in the actual nightly failure. The reported cause in #2512 is `$HOME/.claude`/PATH visibility, but the pre-PR config-directory probe is informational and cannot change the exit status; the issue was also closed as a test-side flaky misclassification (https://github.com/alibaba/anolisa/issues/2512#issuecomment-5288391676). This stub instead invents a first `plugin list` failure followed by success, so it proves the retry helper handles that fabricated condition without showing that the original `test_detect_installed_framework_ready` failure is fixed. Without the real failing trace or a faithful reproducer, the PR's core success criterion remains unverified.

src/tokenless/adapters/tokenless/claude-code/scripts/detect.sh
```diff
@@ -58,8 +93,14 @@ else
...
 
+claude_plugin_listed() {
+    "$CLAUDE_BIN" plugin list 2>&1 | grep -qF "$PLUGIN_ID"
+}
+
 if [ -n "$CLAUDE_BIN" ] && [ -x "$CLAUDE_BIN" ]; then
-    if "$CLAUDE_BIN" plugin list 2>&1 | grep -qF "$PLUGIN_ID"; then
+    # First-run race: `claude plugin list` may fail or omit the plugin while
+    # the CLI initializes ~/.claude; retry before declaring it not installed.
+    if settle claude_plugin_listed; then
```
ikunkun-sys, line 123: [P2] Avoid retrying a definitive plugin-absent result. `claude_plugin_listed` returns 1 both when `claude plugin list` fails and when it succeeds but `grep` does not find this plugin, so every normal pre-install detection sleeps for the full retry budget and invokes the CLI four times. I reproduced an exit-1 plugin-absent run in 3.02 seconds; the normal missing-CLI path likewise took 3.02 seconds via the retry at line 60. Because this detect script also runs before `make claude-code-install`, these are ordinary steady-state paths, contrary to the comment that steady-state runs stay fast.

ikunkun-sys (changes requested):
Requesting changes for two verified issues: the regression test does not reproduce the failure mechanism reported for #2512, and the retry wrapper adds the full default delay to normal negative detection states.

Validation on current head `5a8fa7a0a04c1f812fb6a6100eaf305f2613086f`: the targeted test and both shell syntax checks pass, remote Tokenless/PR checks are green, and both the missing-CLI and plugin-absent paths were reproduced at about 3.02 seconds.

Before merge, please also autosquash the follow-up fix into its originating PR commit and shorten both commit subjects to the repository's 50-character maximum.

src/tokenless/tests/test-claude-code-detect-retry.sh
Forrest-ly: Agreed — addressed in 3d7f239b by correcting the PR's premise rather than claiming verification it cannot provide:

- #2512 was closed as a test-side flaky misclassification (see its closing comment) and never carried a failing trace, so the original `test_detect_installed_framework_ready` failure cannot be reproduced or claimed fixed. The PR body no longer claims that (the "Fixes #2512" wording is removed); it is rescoped as defensive hardening against the observable transient failure mode — a `claude plugin list` invocation failing while the CLI initializes `~/.claude`.
- Your point that the pre-PR config-directory probe was informational and cannot change the exit status is correct, so the reported symptom could not explain the nightly failure either; the previous commit already dropped that probe's retry.
- The test header is rewritten to state honestly what it covers: it exercises the retry mechanism against the transient failure mode detect.sh is designed to ride out (cases 1/2), and a new case 3 pins that deterministic results are never retried. It no longer presents itself as a regression test for the nightly failure.

If the failing nightly log ever becomes retrievable, grounding a reproducer against the real trace would be the right follow-up; as things stand no trace exists on the issue.

src/tokenless/adapters/tokenless/claude-code/scripts/detect.sh
Forrest-ly: Adopted — fixed in 3d7f239b. `claude_plugin_listed` is replaced by `claude_plugin_probe`, which classifies the `plugin list` outcome instead of conflating the two failures:

- invocation itself fails (possibly transient first-run init) → retried via `settle()`;
- CLI ran successfully → authoritative: plugin listed → installed; plugin not listed → deterministically not installed, reported after exactly one invocation with no retry.

I also dropped the `settle()` retry around CLI discovery — PATH is fixed within a single run, so a missing CLI is deterministic for the same reason — and corrected the header comment to state precisely which outcome is retried instead of the inaccurate "steady-state runs stay fast" claim.

Measured with retries=3, delay=1s:
- plugin-absent (CLI present, plugin not listed): exit 1 in ~0.02s, exactly one CLI invocation (previously ~3.0s and 4 invocations);
- missing-CLI: exit 2 in ~0.03s (previously ~3.0s).

New test case 3 asserts the plugin-absent run invokes `plugin list` exactly once regardless of the retry budget, so a regression to the old behavior fails the suite.

src/tokenless/tests/test-claude-code-detect-retry.sh
Forrest-ly: Addressed in `3392e6c5` (supersedes `07627fcf`).

Agreed on both points: the `~/.claude` probe is informational and cannot fail the nightly assertion, and the old stub fabricated a first-`plugin list` failure. Following your elimination analysis, the only exit-status-affecting check that can race under the reported `$HOME/.claude`/PATH visibility mechanism is the claude binary lookup itself: a binary not yet visible → exit 2 (missing prerequisites) → the framework-ready assertion fails. The regression test is now anchored on exactly that path:

- The claude binary becomes visible in `$HOME/.local/bin` only after a delay (a background provisioner simulating the provisioning race), and `~/.claude` does not exist until the CLI's first `plugin list`.
- With retries enabled the first execution settles and reports ready (exit 0); the test also asserts the config-dir probe observed the not-yet-created `~/.claude` without affecting readiness (the informational-only property).
- Control: the same first execution with retries disabled exits 2 with missing prerequisites — the nightly failure mode — proving the retries are what fix it.

Correspondingly, the binary-lookup retry is restored: `07627fcf` had dropped it arguing "PATH is fixed within a run", but that conflates PATH stability with filesystem visibility — `command -v` re-scans on every attempt and the binary file itself can appear mid-execution (verified: the retry finds a binary appearing during the run).

The transient `plugin list` invocation failure remains covered as a separate, explicitly labeled case — no longer presented as the #2512 mechanism. Since #2512 carries no failing trace, the test header states plainly that this anchors the reported mechanism rather than reproducing the nightly flake.

src/tokenless/adapters/tokenless/claude-code/scripts/detect.sh
Forrest-ly: Fixed in `3392e6c5`. `claude_plugin_listed` now classifies the outcome against the `settle()` contract: `0` = plugin listed; `1` = the `plugin list` invocation itself failed (retryable first-run init); `2` = a successful list that omits the plugin (definitive absent). `settle()` retries only exit status 1 and returns any other status immediately, so the definitive plugin-absent result is reported after exactly one invocation.

Measured on the exit-1 plugin-absent path: ~0.02s with exactly one CLI invocation (previously 3.02s and four invocations). Test case 3 pins this by asserting the invocation count.

On the missing-CLI path you also measured: that retry is retained, because it is the fix for the PATH/filesystem-visibility race reported in #2512, and a genuinely absent CLI cannot be distinguished from that race within one run (the binary file itself can appear mid-execution). The misleading "steady-state runs stay fast" comment has been rewritten to say exactly that, and the specific steady-state invocation you pointed at — the detect before `make claude-code-install` — now runs with `TOKENLESS_DETECT_RETRIES=0` (it is informational there and runs on a steady-state machine), so it returns immediately.

src/tokenless/tests/test-claude-code-detect-retry.sh
```diff
@@ -0,0 +1,223 @@
...
+cleanup() {
+    if [ -n "$PROVISIONER_PID" ]; then
+        kill "$PROVISIONER_PID" 2>/dev/null || true
+        wait "$PROVISIONER_PID" 2>/dev/null || true
+    fi
+    rm -rf "$TEST_DIR"
+}
+trap cleanup EXIT
+
+mkdir -p "$FAKE_BIN" "$SHARED_HOOKS"
```
ikunkun-sys, line 52: [P1] Provide isolated `tokenless` and `rtk` prerequisites here. `detect.sh` requires both commands, but this setup only creates the directories and Claude stub; `run_detect` later restricts PATH to the fake home plus system directories. In a fresh detached worktree with neither binary installed globally, `make -C src/tokenless test-claude-code-detect` therefore reaches the expected ready Claude/plugin state and still exits 2 with `missing prerequisites (tokenless binary rtk binary)`. The current green Tokenless workflow does not catch this because it never invokes `make test` or this new target, so the dedicated test needs to be self-contained rather than depend on runner-installed binaries.

ikunkun-sys (changes requested):
The two findings from the previous review are addressed on this head: the test now covers the reported binary-visibility mechanism, and a successful plugin-absent result returns after one `plugin list` invocation. The branch is also reduced to one commit with a compliant subject.

Requesting changes for one new blocker found in a fresh detached worktree: the newly added dedicated test is not self-contained and fails because its isolated PATH contains neither the required `tokenless` nor `rtk` binary. The green `Test tokenless` check does not exercise this target—the workflow runs Cargo and selected integration tests rather than `make test` or `test-claude-code-detect`.

Validation on current head `3392e6c59849285c026a71b04f55295257a4f57f`: shell syntax and `git diff --check` pass; `make -C src/tokenless test-claude-code-detect` fails with exit 2 for missing `tokenless`/`rtk`. Supplying those two test prerequisites makes the binary-race, definitive-absent, and transient-plugin-list scenarios behave as intended.

src/tokenless/tests/test-claude-code-detect-retry.sh
Forrest-ly: Confirmed and fixed in a0a6795a.

Reproduction: masking `/usr/local/bin` with a tmpfs mount (`unshare -rm`) to simulate a host without either binary, `make -C src/tokenless test-claude-code-detect` fails exactly as described — scenario 1 reaches the expected ready Claude/plugin state and still exits 2 with `missing prerequisites (tokenless binary rtk binary)`.

Fix:
- The setup now installs isolated `tokenless` and `rtk` stubs in the fake home bin dir. detect.sh prepends `$HOME/.local/bin` to PATH, so the stubs shadow any runner-installed binaries, and `reset_env` only removes the claude stub, so they survive all four scenarios.
- Scenario 1 additionally asserts that detect.sh reports the stub paths (`present ($FAKE_BIN/tokenless)` / `present ($FAKE_BIN/rtk)`), pinning self-containment: a future regression to depending on host binaries now fails the test even on runners that happen to have them installed.

Verified: with `/usr/local/bin` masked → `claude-code detect retry test passed`; on a host with both binaries installed → still passes.

ikunkun-sys (changes requested):
The previously reported code/test finding is resolved on this head: the dedicated test now supplies isolated `tokenless` and `rtk` stubs, asserts those paths, and passes in a fresh detached worktree. I found no remaining code correctness issues.

Requesting changes for the required commit-history cleanup. This branch currently has two commits, and `beb27a19` fixes a test defect introduced by the preceding PR commit. AGENTS.md §13 requires fixes for issues introduced within the same PR to be fixup/autosquashed into the originating commit rather than left as a standalone fix commit. Its 71-character subject (`fix(tokenless): make detect retry test self-contained for tokenless/rtk`) also exceeds the repository's 50-character subject limit.

Validation on current head `beb27a19e785b54cce33051ca8fe01aae462d5a4`: `make -C src/tokenless test-claude-code-detect`, both shell syntax checks, and `git diff --check` pass; current remote checks are green. Please autosquash this fix, then re-request review on the rewritten head.

ikunkun-sys (approved):

PR #2551, opened 2026-08-14
fix(anolisa): read target-specific raw metadata

### LINKED ISSUE #2547: [anolisa] bug: raw dry-run ignores target-specific sidecar metadata
### Component

anolisa

### What happened?

A raw repository may publish different component contracts for different
targets. For example, AgentSight has a Linux system contract and a macOS
user/system contract because their payloads and lifecycle requirements differ.

The release tooling supports target-specific metadata such as:

```text
v1/agentsight/0.10.1/meta.toml
v1/agentsight/0.10.1/linux/x86_64/agentsight-0.10.1-linux-x86_64.tar.gz
v1/agentsight/0.10.1/macos/aarch64/agentsight-0.10.1-macos-aarch64.tar.gz
v1/agentsight/0.10.1/macos/aarch64/meta.toml
```

However, raw install dry-run derives the metadata URL from the component and
version root. For a conventional artifact URL, it always requests:

```text
v1/agentsight/0.10.1/meta.toml
```

It therefore loads the Linux contract even after resolving the macOS/aarch64
artifact.

In macOS user mode this can fail with an inconsistency because the index
advertises user-mode support while the Linux metadata declares only system
mode. In system mode the dry-run may succeed while validating the wrong
contract.

A non-dry-run install behaves differently: it downloads the selected artifact
and reads its embedded `.anolisa/component.toml`, so it receives the correct
macOS contract.

Dry-run and execution therefore do not validate the same install contract.

### How can we reproduce it?

1. Create a local raw registry with:
   - a version-level Linux `meta.toml`;
   - a Linux x86_64 artifact;
   - a macOS aarch64 artifact;
   - a macOS-specific `meta.toml` beside the macOS artifact.
2. Publish both entries in `index.toml`/`index-v2.toml`.
3. On macOS, run:

```shell
anolisa --install-mode user --dry-run install agentsight \
  --backend raw \
  --repo file:///tmp/anolisa-registry/v1
```

4. Observe that dry-run fetches the version-level Linux metadata instead of
   the metadata beside the resolved macOS artifact.

The same behavior can be reproduced in an integration test with a synthetic
macOS/aarch64 environment and a `file://` registry.

### What did you expect?

Dry-run should validate the exact contract associated with the resolved
artifact without downloading the full artifact.

The preferred repository convention is:

```text
<artifact-directory>/meta.toml
```

The CLI can derive this URL by replacing the final artifact URL segment with
`meta.toml`. This is already the documented same-directory convention in
`anolisa-core::registry::RegistryClient`.

For backward compatibility, when the sibling metadata returns 404/NotFound,
the CLI may fall back to the legacy version-level metadata path. Network,
parse, or digest failures must not trigger fallback.

This does not require adding `meta_path` to `DistributionEntry`: the resolved
artifact URL remains the single location source.

### Possible approach

- Resolve the primary metadata URL as a sibling of the selected artifact.
- Fall back to version-level metadata only when the sibling is absent.
- Keep dry-run lightweight; do not download the full tarball.
- Emit the existing `manifest_digest` field for each raw index entry.
- Verify the sidecar against `manifest_digest` during dry-run.
- Verify the embedded manifest against the same digest during execution.
- Keep metadata downloads isolated by full URL; `DownloadCache` already uses
  a full-URL hash.
- Preserve support for legacy repositories containing only version-level
  metadata.

### Acceptance criteria

- macOS/aarch64 dry-run reads the macOS sibling `meta.toml`.
- Linux repositories with only version-level metadata continue to work.
- A missing sibling may fall back; other sibling fetch failures fail closed.
- Dry-run does not download the install artifact.
- A sidecar digest mismatch is rejected.
- The executed artifact's embedded contract must match the same digest used
  by dry-run.
- End-to-end tests cover target-specific metadata and legacy fallback.

### Anything else?

Relevant code:

- `crates/anolisa-cli/src/commands/tier1/install/raw.rs`
  - `load_dry_run_install_contract`
  - `sidecar_meta_url`
- `crates/anolisa-core/src/registry/client.rs`
  - existing same-directory `derive_meta_url` contract
- `crates/anolisa-core/src/distribution.rs`
  - existing optional `manifest_digest`

ikunkun-sys:
## Why

A raw repository may publish a different install contract per target, because
payloads and lifecycle requirements differ across platforms — AgentSight has a
Linux system contract and a macOS user/system contract.

Raw dry-run derived the metadata URL from the component and version root, so it
always read the version-level `meta.toml` regardless of which artifact it had
resolved. A resolved macOS/aarch64 artifact was therefore validated against the
Linux contract. In macOS user mode that surfaced as a refusal:

```
sidecar meta.toml for component 'agentsight' is inconsistent with the
distribution index: index resolved user-mode support, but manifest declares
modes: system
```

A non-dry-run install reads the artifact's embedded `.anolisa/component.toml`
and receives the correct macOS contract, so dry-run and execution did not
validate the same install contract. In system mode the asymmetry is quieter but
worse: dry-run succeeds while validating the wrong contract.

## What changed

`meta_url_candidates` replaces `sidecar_meta_url` and returns candidates in
preference order:

1. **The artifact sibling** — the final artifact URL segment replaced with
   `meta.toml`. This is the same-directory convention `anolisa-core::registry`
   already freezes, and the only form that can describe a single target.
2. **The version root** — consulted only when the sibling is absent, so legacy
   repositories publishing one contract for every target keep working.

The version root is skipped entirely when the artifact already sits there, where
both forms derive the same URL and a second fetch would be waste.

The fallback is gated on **absence alone**. A network, digest, or parse failure
on metadata the repository does publish now fails the dry-run instead of falling
through to another target's contract. `index_not_published` was renamed to
`remote_file_absent` and reused for this, rather than keeping two near-identical
predicates; this widens sidecar absence to include `410` alongside `404` and
filesystem `NotFound`.

No index schema change: the resolved artifact URL remains the single location
source, so no `meta_path` field is added to `DistributionEntry` and existing
indexes need no republication. Execution-path digest validation needed no change
— `validate_manifest_digest` already checks the embedded contract against the
entry's `manifest_digest`, so once dry-run reads the right file both paths agree
byte for byte.

## Related issue

closes #2547

## User / Agent impact

`anolisa --dry-run install` against a repository with target-specific metadata
now previews the contract belonging to the artifact it resolved. On macOS this
turns a spurious user-mode refusal into a correct preview; in system mode it
replaces a preview validated against the wrong contract with the right one.

Dry-run stays lightweight — the install artifact is still never downloaded.

A repository whose sibling `meta.toml` is published but unfetchable, corrupt, or
digest-mismatched now fails the dry-run where it previously fell back silently.
This is intentional: the previous behaviour previewed an install the execution
path would not have honoured.

## Risk and compatibility

Low risk. No schema, CLI surface, or state format changed. Legacy repositories
publishing only version-level metadata resolve through the fallback and are
covered by a regression test. The one cost is an extra 404 per dry-run against a
nested legacy repository before the version-root fetch; metadata is small and
`DownloadCache` keys on the full URL, so the two candidates never collide in the
cache.

## Validation

All four pre-commit gates pass from `src/anolisa`:

```bash
cargo fmt --all -- --check
cargo test --workspace          # 2131 passed, 0 failed
```

Tests added — 3 unit tests on candidate ordering (sibling preference, version-root
dedupe, flat layout) and 4 end-to-end tests against a synthetic macOS/aarch64
environment and a `file://` registry:

- `dry_run_contract_reads_the_target_sidecar_without_a_digest` — the issue's
  literal reproduction, where no `manifest_digest` exists to cross-check.
- `dry_run_contract_reads_the_resolved_target_sidecar` — the same resolution with
  per-entry digests; also asserts no `.tar.gz` reaches the cache.
- `dry_run_contract_falls_back_to_version_level_meta` — legacy repository.
- `dry_run_contract_rejects_a_tampered_sidecar_without_falling_back` — fail-closed.

The regression tests were verified to actually catch the old behaviour by
temporarily reversing the candidate order: both target-sidecar tests then fail,
the digest-less one with the exact error quoted above.

## Documentation and rollback

No documentation changed — the same-directory convention this aligns with was
already documented in `anolisa-core::registry::RegistryClient`. No CHANGELOG
entry, per the repository convention of updating it atomically with the version
bump.

Rollback is a plain revert of the single commit. There is no persisted state,
migration, or published-artifact dependency to unwind.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs
@@ -53,11 +53,11 @@ pub(super) fn index_fetch_error(index_url: &str, err: DownloadError) -> CliError
     }
 }
 
-/// Whether a fetch failure means "this index file is not published", as
-/// opposed to a transport or repository fault. Only the former may fall
-/// back from `index-v2.toml` to `index.toml`: falling back on transient
-/// errors could silently downgrade a gen-2 repository to its gen-1 view.
-fn index_not_published(err: &DownloadError) -> bool {
+/// Whether a fetch failure means "this file is not published", as opposed to
+/// a transport or repository fault. Every optional-file fallback in this
+/// module is gated on it: falling back on a transient error would silently
+/// serve a *different* file than the one the repository actually publishes.
+fn remote_file_absent(err: &DownloadError) -> bool {
     match err {
         DownloadError::HttpStatus { status, .. } => *status == 404 || *status == 410,
         // file:// repositories surface a missing index as I/O NotFound.
@@ -77,7 +77,9 @@ fn fetch_raw_index(
     let v2_url = raw_index_v2_url(base_url);
     match cache.fetch(&v2_url, None) {
         Ok(downloaded) => Ok((v2_url, downloaded.cached_path)),
-        Err(err) if index_not_published(&err) => {
+        // Absence only: a transport fault here must not downgrade a gen-2
+        // repository to its gen-1 view.
+        Err(err) if remote_file_absent(&err) => {
             let v1_url = raw_index_url(base_url);
             let downloaded = cache
                 .fetch(&v1_url, None)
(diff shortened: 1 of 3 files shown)


kongche-jbw (commented):
Review baseline: `aa8b023862bf...392bb450b095`
未发现阻塞问题。

Remaining risks:
- HTTP sibling/fallback behavior was reviewed statically; local regressions use `file://`.

Verification scope:
- Full diff and threads; `cargo fmt`; 8 targeted tests; package Clippy; CI green.

kongche-jbw (approved):

Issue #2339, opened 2026-08-09
[sight] Show TPS / TTFT / TPOT metrics in the Agent panel


chengshuyi:
## Motivation

AgentSight captures full LLM API request/response pairs via eBPF, including timing data (request start, first SSE token, stream end) and token counts (input/output). The **Agent panel** (agent sessions / agent health menu) currently only shows token consumption and session counts — it does not surface the latency/throughput metrics that matter most for evaluating per-agent performance and provider SLA compliance.

## Proposed Metrics (per Agent)

| Metric | Definition | Data Source |
|--------|-----------|-------------|
| **TPS** (Tokens Per Second) | `output_tokens / stream_duration` | SSE first-token timestamp → stream-end timestamp |
| **TTFT** (Time To First Token) | `first_token_ts - request_sent_ts` | Aggregator request/response timestamps |
| **TPOT** (Time Per Output Token) | `stream_duration / output_tokens` | Derived from TPS |
| **E2E Latency** | `response_complete_ts - request_sent_ts` | Aggregator pair timestamps |

## Scope

1. **Backend**: compute per-call metrics in the Analyzer or GenAI builder; persist to SQLite (new columns or a dedicated `call_metrics` table)
2. **API**: expose `/api/metrics/latency` endpoint with time-range and agent-name filters; return p50/p95/p99 percentiles
3. **Agent panel UI**: in the agent sessions menu, add a latency/throughput column or expandable row showing TTFT, TPS, TPOT per agent; support time-range selector

## Acceptance Criteria

- [ ] TPS / TTFT / TPOT computed for every captured LLM call (SSE and non-SSE)
- [ ] Percentile aggregation (p50, p95, p99) queryable via API with `agent_name` filter
- [ ] Agent panel renders at least TTFT and TPS per agent over a selectable time range
- [ ] No schema_version bump required if using a new table (backward compatible)

## Context

- Timing data is already available in `AggregatedResult` (`start_timestamp_ns`, `end_timestamp_ns`) and SSE event timestamps
- Token counts come from the existing `TokenParser` / `usage` field extraction
- This is a read-only analytics feature — no changes to the eBPF probes or capture pipeline needed
- The Agent panel is the menu entry that lists discovered agents with their session history (not the overview dashboard)

yefuyou:
Hi, I'd like to work on this issue.

Since the current scope spans metric computation/persistence, the API, and the Agent Panel UI, would it be okay to split the work into two independently reviewable PRs?

[...]

PR #2578, opened 2026-08-14
feat(agentsight): add LLM latency metrics API

yefuyou:
## Summary

Backend/API phase for #2339.

Refs #2339

This PR adds backend collection and querying for LLM latency and throughput metrics in AgentSight.

## What changed

- Add TTFT, TPS, TPOT, and E2E latency percentile metrics.
- Support time-range and `agent_name` filters.
- Capture the first observable meaningful output timestamp for HTTP/1 and HTTP/2 SSE responses.
- Preserve HTTP/2 SSE classification independently from meaningful-output observability.
  - `[DONE]`-only and metadata-only SSE responses remain streaming calls.
  - Such responses keep `first_output_timestamp_ns` as `null` when no meaningful output is observable.
- Persist raw timing facts and token counts in SQLite.
- Derive TTFT, TPS, TPOT, and E2E metrics at query time to avoid redundant derived values.
- Keep TTFT, TPS, and TPOT as `null` for non-SSE responses where first-token timing cannot be reliably observed.
- Continue calculating E2E latency for non-SSE responses.
- Keep `first_output_timestamp_ns` as `null` for compressed or otherwise unobservable HTTP/2 SSE responses.
- Add a backward-compatible schema migration without changing `schema_version`.
- Persist LLMCall first-output timing atomically with the corresponding SQLite INSERT/UPDATE.

## API

Adds latency metric aggregation with:

- time-range filtering
- `agent_name` filtering
- p50 / p95 / p99 percentiles
- TTFT
- TPS
- TPOT
- E2E latency

## Scope

This PR contains the backend/API phase only.

It does not include:

- UI changes
- eBPF changes
- compressed SSE first-output inference
- frontend integration

The Agent Panel UI will be handled separately.

## Testing

- `git diff --check`
- `cargo test --lib`: 1483 passed / 0 failed / 1 ignored
- HTTP/1 and HTTP/2 first-output timestamp tests
- HTTP/2 DATA-frame timing propagation
- HTTP/2 SSE classification for `[DONE]`-only, metadata-only, and regular JSON responses
- HTTP/2 → Analyzer → GenAI → SQLite latency propagation
- streaming-call classification tests
- pending / idle-pending completion tests
- SQLite latency percentile tests
- latency API and time-range overflow tests

`cargo test` additionally encounters an existing doctest failure in
`src/genai/helpers.rs`. The same failure is reproducible on the current
`upstream/main`, and that file is not modified by this PR.

### DIFF
--- a/src/agentsight/src/aggregator/http/mod.rs
+++ b/src/agentsight/src/aggregator/http/mod.rs
@@ -12,6 +12,7 @@ pub use aggregator::{ConnectionId, HttpConnectionAggregator};
 pub(crate) use aggregator::ConnectionState;
 pub use pair::HttpPair;
 pub use response::AggregatedResponse;
+pub(crate) use response::event_has_meaningful_output;
 
 // Re-export ParsedRequest from parser (replaces AggregatedRequest)
 pub use crate::parser::http::ParsedRequest;
--- a/src/agentsight/src/aggregator/http/response.rs
+++ b/src/agentsight/src/aggregator/http/response.rs
@@ -57,7 +57,7 @@ impl AggregatedResponse {
     /// Get JSON bodies from SSE events, aggregated into a Vec
     ///
     /// Parses each SSE event's payload as JSON and collects into a Vec,
-    /// skipping events that are not valid JSON (e.g., [DONE] markers).
+    /// skipping events that are not valid JSON (e.g., `[DONE]` markers).
     pub fn json_body(&self) -> Vec<serde_json::Value> {
         self.sse_events
             .iter()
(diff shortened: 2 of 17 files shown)


chengshuyi (approved):
LGTM. Clean design — derive metrics at query time, capture first-output timestamp atomically, good test coverage (~60% of diff).

A few minor nits (non-blocking):

1. **TPS/TPOT convention**: Keep N (total output_tokens) per the issue definition. Add a one-line comment in `stats.rs` noting this is by convention and the N vs N-1 difference is negligible for typical response lengths.

2. **`first_output_timestamp_ns()` in http2.rs**: The per-frame re-parse is O(frames²) worst case. In practice this is fine (first token arrives early → early return; no-output responses have tiny bodies). A one-line comment explaining this trade-off would help future readers.

3. **`stats.rs` formatting**: Missing blank line between `get_latency_metrics` closing brace and the next doc comment (`/// One bucket in a token time-series query.`).

4. **PR size**: Accepting the 998-line diff given it's test-heavy and the data path (capture → persist → query) needs to stay together for end-to-end testability.

Optional follow-ups (not blocking):
- Sync AGENTS.md API Endpoints table with `/api/metrics/latency`
- `response_sse_event_count()` doc: note that count includes non-JSON events like `[DONE]` (differs from `response_sse_json_array().len()`)

PR #2580, opened 2026-08-15
feat(anolisa): add native dsh adapter

kongche-jbw:
## Why

ANOLISA components need a framework-native way to register plugin bundles with
DeepSeek Harness without editing `$DSH_HOME` or maintaining component-specific
install scripts. This adds a reusable DSH driver so Tokenless and future
components can use the same adapter contract and DSH-owned lifecycle.

## What changed

- Add a generic `dsh` framework driver that validates `package.json`, the
  nested `dsh.bundle.patch` declaration, and the Cordis patch package identity.
- Add repeatable `anolisa adapter enable <component> dsh --profile <name>`
  selection and reject implicit or non-DSH profile usage.
- Register and remove packages through `dsh plugin --profile ...`, persist one
  typed receipt for all selected profiles, verify status using exact package
  matches, and clean stale profiles during re-enable.
- Persist the validated enable-time DSH home in the receipt so status, disable,
  and re-enable cleanup remain anchored when `DSH_HOME` or the cwd changes.

Other components can reuse the interface by declaring a normal ANOLISA adapter:

```toml
[[adapters]]
framework = "dsh"
adapter_type = "plugin"
plugin_id = "my-plugin-id"
source = "adapters/dsh"
dest = "{datadir}/adapters/{component}/dsh/"

[adapters.bundle]
entry = "package.json"
```

The source directory must contain an npm-compatible package name and nested DSH
bundle metadata:

```json
{
  "name": "@example/dsh-plugin",
  "dsh": { "bundle": { "patch": "./cordis.patch.yml" } }
}
```

The patch id must match `plugin_id`, and the plugin name must match the package
installed by DSH:

```yaml
- insert:
    - id: my-plugin-id
      name: '@example/dsh-plugin'
```

ANOLISA then delegates registration to DSH with a `link:` package reference;
components do not write profile files directly.

## Related issue

no-issue: introduce the shared DSH adapter contract needed by native plugins

## User / Agent impact

Users can enable a component in one or more explicit DSH profiles:

```bash
anolisa adapter enable <component> dsh --profile default
anolisa adapter enable <component> dsh --profile web --profile headless
```

Disable and status use both the profiles and DSH home recorded by the enable
receipt.

## Risk and compatibility

The change adds a `profiles` field to `EnableOptions` and a `dsh` receipt
payload, including the validated enable-time DSH home resource. Before
downgrading to an ANOLISA version without this driver, disable all DSH adapters
so older binaries do not encounter the new receipt variant.

## Validation

- `cargo test -p anolisa-core --locked`
- `cargo test -p anolisa-cli enable_parses_repeatable_profiles --locked`
- `cargo clippy -p anolisa-core --all-targets --locked -- -D warnings`
- `cargo clippy -p anolisa-cli --all-targets --locked -- -D warnings`
- `git diff --check`

## Documentation and rollback

The public Rust items and CLI help document the new contract; no standalone
guide is added in this change. Roll back by disabling DSH adapters first and
then reverting this PR's feature commit.


src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1035 @@
...
+
+    fn plan_reenable_cleanup(
+        &self,
+        prior: &AdapterClaim,
+        ctx: &DriverCtx,
+    ) -> Result<Vec<String>, AdapterError> {
+        let prior_payload = dsh_claim(prior)?;
+        validate_dsh_claim(prior, prior_payload)?;
+        let package_changed = read_dsh_bundle(&ctx.resource_root)
+            .map(|bundle| bundle.package_name != prior_payload.package_name)
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in `9b96438e`. `plan_reenable_cleanup` now propagates `read_dsh_bundle` failures instead of treating them as a package change, so the dry-run fails closed. This hook is read-only, so the previous behavior could misreport cleanup but could not itself remove a registration.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1035 @@
...
+            kind: AdapterConditionKind::VerificationSupported,
+            status: verification,
+            reason: (verification != ConditionStatus::True)
+                .then(|| "dsh profile registration could not be verified".to_string()),
+            resource: None,
+        });
+        let summary = if claim.status == ClaimStatus::CleanupFailed {
+            AdapterSummary::CleanupFailed
+        } else if !detect.detected || !all_registered {
+            if verification == ConditionStatus::Unknown
```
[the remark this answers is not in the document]
kongche-jbw: No code change needed. The existing summary branch already checks for any `PluginRegistered=False` condition and returns `Degraded`; it returns `Unknown` only when there is no known false condition and at least one profile remains unverifiable. The per-profile conditions also preserve the known/unknown distinction.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1035 @@
...
+        validate_profile_name(profile).map_err(|reason| AdapterError::InvalidAdapterInput {
+            component: ctx.component.clone(),
+            framework: "dsh".to_string(),
+            reason,
+        })?;
+    }
+    Ok(profiles)
+}
+
+fn validate_profile_name(profile: &str) -> Result<(), String> {
```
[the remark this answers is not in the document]
kongche-jbw: Partially addressed in `9b96438e`. DSH upstream has no profile length cap, and ANOLISA already rejected empty names, path separators, `.`/`..`, and leading `-` (including `--long-flag-like-name`). The actual missing upstream constraint was the reserved `node_modules` profile name, which is now rejected and covered by a test; I did not invent a length limit that DSH itself does not impose.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1035 @@
...
+        let actions = profiles
+            .iter()
+            .map(|profile| {
+                format!(
+                    "register dsh plugin '{package}' in profile '{profile}' from {}",
+                    bundle.resource_root.display()
+                )
+            })
+            .collect::<Vec<_>>();
+        let register_command = profiles.first().map(|profile| {
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in `9b96438e`. The singular `register_command` is now emitted only for exactly one profile. Multi-profile plans omit that field and use the complete per-profile `actions` list, so the preview no longer presents only the first mutation.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1035 @@
...
+            continue;
+        }
+        let value = raw.trim().trim_matches(|c| c == '\'' || c == '"');
+        if !(value.starts_with("./") || value.starts_with("../")) {
+            continue;
+        }
+        relative_entry_count += 1;
+        let path = validate_relative_bundle_path(root, value, "entry")?;
+        let full = root.join(path);
+        if !full.is_file() {
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in `9b96438e`. Bundle roots and every package manifest, patch, and local patch entry are now canonicalized; resolved files must remain beneath the canonical bundle root and be regular files. A regression test covers an entry symlink targeting a file outside the bundle.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1127 @@
...
+        }
+        let value = raw.trim().trim_matches(|c| c == '\'' || c == '"');
+        if !(value.starts_with("./") || value.starts_with("../")) {
+            continue;
+        }
+        relative_entry_count += 1;
+        let path = validate_relative_bundle_path(root, value, "entry")?;
+        resolve_bundle_file(root, canonical_root, &path, "patch entry")?;
+    }
+    if relative_entry_count == 0 {
```
ikunkun-sys: [P1] Align bundle validation with DSH's package-name contract

This requires at least one relative name or entry such as ./dist/index.js, but DSH's official bundle contract says patch rows must reference the installed package by name rather than a relative source path: https://github.com/deepseek-ai/deepseek-harness/blob/47f943859bef60e4160492346772ded9b24f765a/docs/user/develop/basic/publish.md#L56-L64. I verified that the official @deepseek-ai/dsh-headless bundle adds and dumps successfully while containing no relative plugin entry, so this driver rejects a valid DSH bundle. Please align validation with the upstream package-resolution contract instead of requiring a local entry path.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1127 @@
...
+        timeout: CLI_TIMEOUT,
+    }
+}
+
+fn profile_registered(
+    ctx: &DriverCtx,
+    profile: &str,
+    package: &str,
+) -> Result<Option<bool>, AdapterError> {
+    let output = ctx.ops.run_framework_cli(dsh_command([
```
ikunkun-sys: [P1] Keep adapter status read-only

The status path reaches this dsh plugin list invocation, but upstream runPlugin initializes a missing profile and reconciles, then may write, dsh.profile.bundles after every successful pnpm command: https://github.com/deepseek-ai/deepseek-harness/blob/47f943859bef60e4160492346772ded9b24f765a/apps/cli/src/plugin.ts#L120-L145. Against an empty DSH_HOME, list --json created package.json, cordis.patch.yml, and pnpm-workspace.yaml. Therefore status can recreate a deleted profile or modify an existing manifest despite the documented read-only contract. Please verify registration without invoking this mutating command.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1127 @@
...
+                    .values()
+                    .any(|value| json_contains_package(value, package))
+        }
+        _ => false,
+    }
+}
+
+fn output_is_absent(output: &super::driver::CliOutput) -> bool {
+    let text = format!("{}\n{}", output.stdout, output.stderr).to_ascii_lowercase();
+    [
```
ikunkun-sys: [P1] Do not classify infrastructure failures as plugin absence

With dsh present but pnpm absent, upstream returns exit 127 with 'pnpm not found on PATH' (https://github.com/deepseek-ai/deepseek-harness/blob/47f943859bef60e4160492346772ded9b24f765a/apps/cli/src/plugin.ts#L134-L139). The package remains in both dependencies and dsh.profile.bundles, yet this broad marker makes output_is_absent return true. disable then reports cleanup_complete and AdapterManager deletes the only receipt, so cleanup cannot be retried. Please recognize only the exact missing-dependency outcome; other nonzero exits must retain the receipt.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1127 @@
...
+        )
+    })?;
+    validate_patch_entries(root, &canonical_root, &patch_text)?;
+    Ok(DshBundle {
+        package_name: package.name,
+        patch: patch_text,
+    })
+}
+
+fn patch_contains_id(patch: &str, expected: &str) -> bool {
```
ikunkun-sys: [P1] Validate the patch as DSH YAML rather than scanning lines

Both this ID check and validate_patch_entries only split text lines. A malformed YAML document can include the expected id and relative name lines, pass these checks, and be registered because dsh plugin add only inspects the package manifest. DSH parses the overlay later during profile loading and fails at boot: https://github.com/deepseek-ai/deepseek-harness/blob/47f943859bef60e4160492346772ded9b24f765a/packages/boot/app-boot/src/index.ts#L289-L337. That leaves an enabled receipt for a profile that cannot start, while package-only status can still report healthy. Please perform DSH-compatible YAML/schema validation before enable mutates profile state.

ikunkun-sys (changes requested):
Reviewed current head 9b96438e against the official DSH bundle and CLI contracts at deepseek-harness commit 47f9438. Local formatting, clippy, workspace tests, docs, and a synthetic merge with current main pass, but the four inline correctness issues below are blocking.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
kongche-jbw: Fixed in 54a3a9bc. The relative-entry requirement is removed; inserted plugin names now reject filesystem paths and follow DSH package resolution. Package-name bundle coverage was added.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
kongche-jbw: Fixed in 54a3a9bc. Status now reads the profile package.json through controlled read-only IO and verifies both dependencies and dsh.profile.bundles; it no longer invokes dsh plugin.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
kongche-jbw: Fixed in 54a3a9bc. Cleanup skips the CLI only when the profile manifest proves the package absent from both dependency and bundle state. Every nonzero CLI result, including pnpm exit 127, now keeps cleanup incomplete for retry.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
kongche-jbw: Fixed in 54a3a9bc. Bundle patches are parsed as YAML before any profile mutation, with DSH-compatible top-level and insert structure checks plus support for !!js tagged values. Malformed YAML regression coverage was added.

src/anolisa/crates/anolisa-core/src/adapter/claim.rs
```diff
@@ -724,6 +735,27 @@ pub struct QwenCodeClaim {
...
 }
 
+/// DeepSeek Harness native-plugin receipt. A single ANOLISA receipt owns the
+/// same package across every explicitly selected dsh profile; each profile
+/// keeps its own validated framework-plugin resource reference so disable can
+/// release exactly the registrations that enable created.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct DshClaim {
+    /// Package name read from the bundle's `package.json`.
+    pub package_name: String,
```
ikunkun-sys, line 745: [P1] Persist the enable-time DSH home in this receipt

This claim records only the package and profile names, while status and disable resolve `DSH_HOME` again from the current process. If enable runs with `DSH_HOME=/tmp/a` and a later disable runs with `/tmp/b` (or the same relative value from another cwd), `profile_package_state` reads B, treats its missing manifest as `Absent`, skips `dsh plugin remove`, and reports cleanup complete; `AdapterManager` then deletes the receipt even though the registration remains under A. Upstream DSH also resolves this environment override per process: https://github.com/deepseek-ai/deepseek-harness/blob/47f943859bef60e4160492346772ded9b24f765a/packages/util/home-paths/src/index.ts#L76-L91. Please persist the resolved enable-time home/profile root as a validated claim resource and use it both for manifest reads and as `DSH_HOME` on later cleanup commands.

ikunkun-sys (changes requested):
The previous four blockers are addressed in this revision, and the current code and CI gates pass. One lifecycle correctness issue remains:

- **P1:** persist the enable-time DSH home/profile root and use it for both read-only verification and later DSH CLI cleanup. Re-resolving `DSH_HOME` during status/disable can inspect and mutate a different profile tree, then discard the only receipt while the original registration remains.
- **P2:** update the PR description: it still documents `name: ./dist/index.js` even though the implementation now correctly rejects path-valued plugin names, and its rollback command references the superseded `9b96438e` commit.

src/anolisa/crates/anolisa-core/src/adapter/claim.rs
kongche-jbw: Fixed in `8a90a287`. The receipt now stores the normalized enable-time DSH home as a validated `ExternalPath`; status, disable, and re-enable cleanup use that persisted root for manifest reads and set the same `DSH_HOME` on every DSH add/remove command. Regression tests cover cwd/environment drift and cleanup of the original home. The PR description and rollback guidance were also updated.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
```diff
@@ -0,0 +1,1510 @@
...
+        let source_changed = prior.resource_root != next.resource_root;
+        let package_changed = prior_payload.package_name != next_payload.package_name;
+        let mut cleanup_complete = true;
+        let mut messages = Vec::new();
+        for profile in &prior_payload.profiles {
+            let retained = next_payload
+                .profiles
+                .iter()
+                .any(|candidate| candidate.name == profile.name);
+            if retained && !source_changed && !package_changed {
```
ikunkun-sys: [P1] Clean retained profiles when the persisted DSH home changes

This early return ignores whether the prior and next receipts point at different DSH homes. If enable registered the package/profile under home A and re-enable uses the same source, package, and profile under home B, this branch skips removal from A; the Manager then replaces the receipt with the B claim and registers there, leaving the A registration permanently untracked. `plan_reenable_cleanup` has the same omission, so dry-run also hides it. The new test constructs different first/later homes but asserts cleanup only for the stale profile, leaving this retained-profile migration uncovered.

ikunkun-sys (changes requested):
Reviewed current head `8a90a287`. The previous DSH-home persistence blocker and the stale PR description are addressed. Rust 1.88 formatting, workspace clippy/tests/docs, remote CI, and a synthetic merge with current main pass, but the re-enable migration issue in the inline P1 remains blocking.

src/anolisa/crates/anolisa-core/src/adapter/dsh.rs
kongche-jbw: Fixed in `f4736bfd`. Re-enable now compares the persisted prior home with the resolved next home in both dry-run and execution. A home change schedules every prior profile for removal from the old `DSH_HOME`, including retained profiles, before the next claim is installed. Tests separately preserve same-home retained profiles and cover retained-profile migration across homes.

ikunkun-sys (approved):
Re-reviewed the current head. The prior DSH home migration issue is resolved: same-home retained profiles stay registered, while a home change cleans retained profiles from the persisted prior DSH_HOME before installing the replacement claim. The current head also passes the synthetic merge against latest main, formatting, workspace Clippy/tests/docs, targeted DSH and CLI tests, and remote CI. No remaining blocking findings.

PR #2605, opened 2026-08-17
fix(cosh-ng): treat dnf dry-run abort as success

xuchunmei000:
## Description

`cosh pkg install --dry-run` / `remove --dry-run` on the dnf backend returned
`PkgBackendError` for exactly the packages that were installable or removable.
dnf raises `CliError("Operation aborted.")` and exits 1 when `--assumeno`
declines the resolved transaction, and the failure branch in `pkg.rs` did not
recognize that signal — a *successful* simulation was mapped to a backend
error. This PR recognizes the declined confirmation as a completed dry-run,
drops the contradictory `-y` from the dry-run args, and pins the backend
locale for dry-run invocations so the classification cannot be rerouted by a
forwarded non-English locale.

Issue #2107 proposed switching dnf to `--downloadonly` /
`--setopt=tsflags=test`. That route was not taken: both download every package
in the transaction (and `tsflags=test` also imports GPG keys), a *larger* side
effect than the simulation they replace. `--assumeno` aborts before the
download loop, which is the closest analogue to `apt-get --dry-run`. Reading
repository metadata is inherent to dnf dependency resolution and is documented
rather than suppressed, since a cache-only dry-run would answer from stale
data.

## User / Agent impact

Agent callers relying on `--dry-run` to decide whether a package is
installable previously received a recoverable backend error for every
installable-but-not-installed package, and could wrongly conclude the package
was unavailable. Already-installed and nonexistent packages were unaffected
and remain so.

## Risk and compatibility

- **Result mapping**: the new success branch is gated on
  `dry_run && mgr == PkgManager::Dnf`; apt, zypper, and Brew are untouched.
- **Locale**: `LC_ALL`/`LANGUAGE=C` is applied to **dry-run invocations only**.
  Real install/remove keep the caller's environment, so rpm scriptlets,
  package-manager plugins, and prompts observe no change.
- **False success**: pre-confirmation refusals (bootc protected paths,
  unsupported persistence, read-only `/usr`) share dnf's "Operation aborted"
  prefix but mean a real invocation fails too; they are excluded and still
  surface as errors.
- **Privileged / migration**: none.

## Validation

- Unit tests in `crates/cosh-platform/src/pkg.rs`:
  - arg builders assert `--assumeno` without `-y`, plus a guard that dry-run
    args never contain `--downloadonly` or `tsflags`
  - abort predicate: dnf4 bare message, dnf5 `Operation aborted by the user`,
    and negative cases for protected-paths, bootc persistence, read-only `/usr`
  - `pkg_command` locale scoping: dry-run carries both variables, a real
    invocation has zero env overrides
- `scripts/docs-lint.sh` — pass.

## Documentation and rollback

- `docs/user-guide/{en,zh}/user-entrypoint/cosh-ng/cli/package-management.md`
  now state the guarantee precisely: dry-run confirms that dependency
  resolution succeeds **against the current metadata**, and does not cover
  download, signature, file-conflict, scriptlet, or concurrent-state failures.
  Homebrew is called out separately as existence-check only.
- The same narrowing was applied to the `pkg_install` rustdoc to keep code and
  user docs on one wording.
- CHANGELOG intentionally untouched (`specs/documentation-standard.md` §5 —
  written only in release version-bump PRs).
- Rollback: revert the single commit; no state migration involved.


docs/user-guide/en/user-entrypoint/cosh-ng/cli/package-management.md
```diff
@@ -24,6 +24,8 @@ cosh-cli pkg remove nginx --dry-run
 
 Run without `--dry-run` to apply the change. Package operations normally need root privileges. An install that finds the package already present still returns success and marks `already_installed` in the response.
 
+`--dry-run` reports whether the operation would succeed: it never installs or removes anything, never downloads packages, and never writes to the package database. It is not completely free of I/O — on dnf and zypper it reads repository metadata and fetches it from the network when the local copy is missing or expired, because dependency resolution cannot run without it.
```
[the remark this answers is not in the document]
xuchunmei000: > **P2 Badge Add the changed dry-run contract to the component READMEs**
> 
> This changes the observable behavior of the existing `--dry-run` flag on dnf from failure to success, but only the full user-guide pages were updated. The repository documentation standard requires a modified CLI flag to be reflected in both the component README summary and the user-guide reference, so add equivalent summaries to `src/cosh-ng/README.md` and `README_zh.md`.
> 
> AGENTS.md reference: [AGENTS.md:L337-L343](https://github.com/alibaba/anolisa/blob/42309de9b5786ec7631885836acc3d48782687fc/AGENTS.md#L337-L343)
> 
> Useful? React with 👍 / 👎.

Not applicable — declining with rationale.
**The rule's trigger doesn't match**. specs/documentation-standard.md §5 requires a README summary for a new or modified CLI command/flag. This PR neither adds nor modifies a flag: --dry-run's name, arity, and documented contract (read-only pre-check) are unchanged. What changed is that the dnf backend mismapped a successful simulation to PkgBackendError — dnf exits 1 with Operation aborted. when --assumeno declines the resolved transaction. Restoring the already-documented contract is a bug fix, not a flag change. As the finding itself puts it, the behavior moved "from failure to success" — the failure was the defect.

**There is also no README target to update**. src/cosh-ng/README.md is an 86-line overview (Why / Install / Start in 30 seconds / Documentation / Contribute / License) and contains no pkg command or flag reference at all; the CLI reference lives entirely in the user-guide. Adding a dnf dry-run behavioral note would make it the only flag-level detail in the file, contrary to §4.5's component-README scope ("Focused — what this component does, who uses it, how to install & use").

**Note on the citation:** the referenced AGENTS.md:L337-L343 covers §8 PR Description / §9 Documentation Rules, which explicitly defers all documentation rules to the spec — it contains no README-summary requirement.The user-guide pages (en + zh) were updated and now document the dry-run guarantee per backend.

KaiLongZhou:
## Code Review Check — 审核通过 ✅

逐条验证了 code-reviewer 的 review 意见，所有分析均与实际代码一致，无遗漏或误报。

**核心变更验证：**
- `is_dnf_dry_run_abort` 分支位置、门控条件、stderr/stdout 双检查 — 正确
- 移除 dry-run 中矛盾的 `-y` 参数 — 正确
- `pkg_command` 固定 `LC_ALL=C` + `LANGUAGE=C`，仅作用于 install/remove 主路径 — 正确
- 6 个新增测试覆盖正常路径和 false-positive 边界 — 充分
- en/zh 文档同步更新，描述与代码行为一致 — 正确

**CI 状态：** 全部通过（Test cosh-ng、Build cosh-ng release、Docs Lint、Commit Message Lint、PR Checks）

**结论：** 无阻塞问题，可以合并。

KaiLongZhou (approved):
Review check passed. All findings verified against source code. No blocking issues.

src/cosh-ng/crates/cosh-platform/src/pkg.rs
```diff
@@ -428,11 +468,27 @@ fn is_remove_not_found(stdout: &str) -> bool {
     stdout.contains("No match for argument") || stdout.contains("No packages marked for removal")
 }
 
+/// Detect dnf declining a resolved transaction, which is how the dnf backend
+/// reports a completed dry-run: resolution succeeded and dnf exited 1 in
+/// response to `--assumeno` instead of applying anything.
+fn is_dnf_dry_run_abort(stderr: &str, stdout: &str) -> bool {
+    stderr.contains("Operation aborted") || stdout.contains("Operation aborted")
```
kongche-jbw: **[P1] Do not treat every DNF abort as an assumed-no success**

This predicate conflates the confirmation decline with other DNF failures. DNF5 emits `Operation aborted. Pass --setopt=usr_drift_protected_paths= ...` when a bootc protected-path check fails *before* confirmation ([upstream source](https://github.com/rpm-software-management/dnf5/blob/a86874e55403aa9ee391a5ca7bb0b209afdb36b1/dnf5/main.cpp#L1727-L1741)); the actual confirmation decline is a distinct `Operation aborted by the user.` error ([definition](https://github.com/rpm-software-management/dnf5/blob/a86874e55403aa9ee391a5ca7bb0b209afdb36b1/libdnf5-cli/exception.cpp#L30)). Both contain this substring, so `pkg_install`/`pkg_remove` will return `Ok` for a transaction that a real invocation cannot apply. Please distinguish the confirmation-specific signal from other aborts and add the protected-path message as a negative regression case.

src/cosh-ng/crates/cosh-platform/src/pkg.rs
```diff
@@ -428,11 +468,27 @@ fn is_remove_not_found(stdout: &str) -> bool {
...
+// the stdin cosh inherits. `-y` is left out of the dry-run branch because dnf
+// evaluates `assumeno` first, so pairing the two only implies a confirmation
+// that never happens. `--downloadonly` and `--setopt=tsflags=test` are
+// deliberately avoided: both fetch every package in the transaction, a far
+// larger side effect than the simulation they would replace.
+
 fn build_dnf_install_args(package: &str, dry_run: bool) -> Vec<&str> {
     if dry_run {
-        vec!["install", "-y", "--assumeno", package]
+        vec!["install", "--assumeno", package]
```
kongche-jbw: **[P2] Align this implementation with the issue being closed**

Issue #2107 reports that `--assumeno` still performs network access and repository metadata/cache writes, and asks for a dry-run without those side effects. Keeping `--assumeno` fixes the separate `PkgBackendError` mapping, but leaves the issue's reported behavior intact; the new user guide explicitly accepts that I/O. `--downloadonly` is not itself a zero-side-effect solution, but this PR should either meet the issue's stated contract or stop closing/reframe #2107 after an explicit maintainer decision.

kongche-jbw (changes requested):
Requesting changes for one correctness issue and one issue/contract mismatch. CI is green, but the abort classifier can turn a real DNF failure into a successful dry-run result, and the implementation still retains the metadata/cache side effects that #2107 asks to remove.

src/cosh-ng/crates/cosh-platform/src/pkg.rs
xuchunmei000: 在https://github.com/alibaba/anolisa/issues/2107 下做了说明，dnf架构决定了dnf进行依赖解析时必须有 repodata，优先从缓存获取，若缓存失效就从repo的网络获取;apt 的"零网络"只是因为它把刷新拆到了 apt update。
本pr主要解决-y和--assumeno冲突，以及dnf dry-run对“可安装的包”报后端错误的问题。

kongche-jbw: 感谢补充说明。理解 DNF 的依赖解析需要可用的 repodata，缓存失效时产生网络访问属于其架构约束；结合您在 #2107 下补充的范围说明，这一点我不再作为当前实现本身的缺陷。至于是否按这一口径关闭 #2107，建议后续由维护者确认即可。辛苦说明背景。

docs/user-guide/en/user-entrypoint/cosh-ng/cli/package-management.md
```diff
@@ -24,6 +24,8 @@ cosh-cli pkg remove nginx --dry-run
 
 Run without `--dry-run` to apply the change. Package operations normally need root privileges. An install that finds the package already present still returns success and marks `already_installed` in the response.
 
+`--dry-run` previews the operation without installing or removing anything, downloading packages, or writing to the package database. What it can guarantee depends on the backend. dnf, apt, and zypper resolve the full transaction, so success means the operation is expected to apply cleanly; they read repository metadata to do so and fetch it from the network when the local copy is missing or expired, because dependency resolution cannot run without it. Homebrew has no simulation mode, so its `--dry-run` only confirms that the formula exists (install) or is installed (remove) — it does not resolve dependencies or conflicts and may report success for an operation that later fails on dependency, conflict, or platform constraints.
```
kongche-jbw: **[P2] Consider narrowing the dry-run success guarantee**

Thank you for documenting the backend differences. Could we soften “expected to apply cleanly” for dnf, apt, and zypper? These modes primarily validate dependency resolution against the current metadata, but they cannot cover every failure in the later transaction. DNF5 handles `--assumeno` before download and transaction execution ([source](https://github.com/rpm-software-management/dnf5/blob/a86874e55403aa9ee391a5ca7bb0b209afdb36b1/dnf5/main.cpp#L1754-L1762), [download path](https://github.com/rpm-software-management/dnf5/blob/a86874e55403aa9ee391a5ca7bb0b209afdb36b1/dnf5/context.cpp#L488-L522)); zypper documents that `--dry-run` skips file-conflict checks for uncached packages ([manual](https://github.com/openSUSE/zypper/blob/master/doc/zypper.8.txt)); and apt simulation runs without locking ([manual](https://manpages.debian.org/unstable/apt/apt-get.8.en.html)). Wording such as “dependency resolution succeeds under the current metadata” would keep the guarantee precise while allowing for download, signature, file-conflict, script, or concurrent-state failures during the real operation.

src/cosh-ng/crates/cosh-platform/src/pkg.rs
```diff
@@ -392,6 +416,22 @@ pub fn pkg_remove(
...
+///
+/// [`pkg_install`] and [`pkg_remove`] classify their outcome from the backend's
+/// human-readable messages, so a caller environment that asks for another
+/// language would silently reroute those branches. `LANGUAGE` is pinned too:
+/// dnf resolves messages through Python's gettext, which honors `LANGUAGE` even
+/// when the locale is `C`.
+fn pkg_command(program: &str, args: &[&str]) -> Command {
+    let mut command = Command::new(program);
+    command.args(args);
+    command.env("LC_ALL", "C").env("LANGUAGE", "C");
```
kongche-jbw: **[P2] Please clarify or scope the locale change to dry-run**

Would you mind reconsidering the unconditional override here? Both `pkg_install` and `pkg_remove` use `pkg_command` for dry-run and real DNF/APT/Zypper/Homebrew operations, so this changes the locale of normal install/remove commands too, while the PR risk section says real paths are unchanged. Besides translating backend diagnostics, the inherited locale can reach package-manager plugins, prompts, and package scripts. If the C locale is only needed to classify DNF dry-run output, scoping it to that invocation would preserve existing behavior; otherwise, please document and test the cross-backend change to real operations.

kongche-jbw (commented):
Thank you for the detailed clarification on #2107. While continuing the review, I noticed two additional P2 scope/contract points. These are intended to help align the documented guarantee and the PR risk statement with the behavior of the backend commands.

src/cosh-ng/crates/cosh-platform/src/pkg.rs
xuchunmei000: Confirmed and fixed — good catch, this was a real correctness hole.

Verified against upstream: dnf5 throws AbortedByUserError (libdnf5-cli/exception.cpp:30, "Operation aborted by the user.") for the confirmation decline, while the protected-path refusal at dnf5/main.cpp:1740 is a CommandExitError raised before confirmation. My substring test accepted both.

Two additions beyond the reported case: dnf4 (dnf/cli/cli.py) raises the bare "Operation aborted." at lines 231 and 242 as well — bootc persistence=persist and read-only /usr — both pre-confirmation. Those are indistinguishable from the confirmation decline by message alone, so they needed the same treatment.

The predicate now excludes pre-confirmation refusals by their distinctive cause text (usr_drift_protected_paths, Persistent transactions aren't supported, configured to be read-only) before accepting either dnf5's Operation aborted by the user or dnf4's bare Operation aborted.. I deliberately avoided a blunt bootc filter: bootc is a real package name and would appear in the transaction table of pkg install bootc --dry-run, causing false rejections. On ambiguity the predicate fails loud rather than reporting success.

Regression tests added: the protected-path message you asked for, plus the two dnf4 bootc refusals and a dnf5 positive case.

docs/user-guide/en/user-entrypoint/cosh-ng/cli/package-management.md
xuchunmei000: Good point, and the upstream behavior you cite is exactly why — adopted.

--assumeno exits at dnf/cli/cli.py:277 (dnf5: main.cpp:1755), which is before Downloading Packages:` (line 286),gpgsigcheck(line 300), and the rpm transaction itself. I had used that ordering to argue--assumenohas the smaller side effect, but it equally means the guarantee is weaker than I documented. Together with zypper skipping file-conflict checks for uncached packages and apt simulating without locking, "expected to apply cleanly" was an overstatement.

Reworded to your suggestion in both en and zh: the dry-run confirms that dependency resolution succeeds against the current metadata, and explicitly does not cover download, signature, file-conflict, scriptlet, or concurrent-state failures.

I also narrowed the same over-promise in the pkg_install rustdoc, which said the dry-run validates the package "could be installed" — keeping the code contract and the user guide on one wording. The Homebrew caveat was trimmed accordingly, since "may later fail" is now stated for every backend; it now just says Homebrew does not resolve dependencies or conflicts at all.

src/cosh-ng/crates/cosh-platform/src/pkg.rs
xuchunmei000: Scoped it to dry-run — you're right on both counts, and the first one is an inconsistency in my own PR description.

The Risk section's "real paths unchanged" claim held for the result-mapping branch (dry_run && mgr == Dnf) but not for the locale override, since pkg_command serves both paths. I've corrected the PR description.

Your point about package scripts is what settled the design: a real install executes rpm scriptlets that inherit this environment, whereas a dry-run aborts at --assumeno before the transaction ever runs. So restricting the override to dry-run isn't a compromise — it lines up exactly with the risk boundary, applying only where no package script can observe it. pkg_command now takes dry_run and pins LC_ALL/LANGUAGE only for simulations; real invocations inherit the caller's environment unchanged.

Locked in with two tests via Command::get_envs(): dry-run must carry both variables, and a real invocation must have exactly zero env overrides.

src/cosh-ng/crates/cosh-platform/src/pkg.rs
kongche-jbw: Thank you for tightening the classifier and adding coverage for the three known messages. I think one DNF4 edge case may still remain, if you would not mind taking another look.

For the persistence and read-only bootc cases, DNF4 emits the explanatory text with `logger.info` and then raises the bare `CliError("Operation aborted.")` ([source](https://github.com/rpm-software-management/dnf/blob/master/dnf/cli/cli.py#L224-L242)). `debuglevel=0` is a supported configuration and maps normal output below `INFO` ([configuration](https://dnf.readthedocs.io/en/stable/conf_ref.html#debuglevel), [level mapping](https://github.com/rpm-software-management/dnf/blob/master/dnf/logging.py#L66-L78)), while the caught `CliError` is still emitted as an error ([source](https://github.com/rpm-software-management/dnf/blob/master/dnf/cli/main.py#L195-L199)). In that configuration, cosh receives only the bare `Operation aborted.`, so none of the new exclusions match and the non-applicable transaction is still returned as a successful dry-run.

The current regression cases always include the explanatory text, so they do not exercise this suppressed-cause path. Could the success classification be made independent of optional `INFO` output (or the required verbosity be stabilized for this invocation), with a regression case where the cause text is absent? The DNF5-specific path looks addressed.

xuchunmei000: Confirmed and fixed — pinning as you suggested was the right call, and the exposure was slightly worse than you described.

Reading dnf sources: _VERBOSE_VAL_MAPPING[0] = SUPERCRITICAL (dnf/logging.py:76) and _ERR_VAL_MAPPING[0] = SUPERCRITICAL (89), so debuglevel=0 strips the logger.info cause line as you noted, and errorlevel=0 additionally strips the logger.error line that carries Operation aborted. itself. Under errorlevel=0 cosh would receive an empty stderr with exit 1 — the declined confirmation would degrade to PkgBackendError, breaking the fix.

The dry-run args now carry --setopt=debuglevel=2 --setopt=errorlevel=2 (dnf defaults). --setopt is PRIO_COMMANDLINE, applied before logging handlers are configured, so it overrides dnf.conf and any --config. Real install/remove paths remain untouched (they do not carry these setopts) so the caller's own verbosity choices for real operations are preserved.

Regression coverage: added test_build_dnf_dry_run_args_pin_verbosity asserting both setopts are present on install and remove dry-run args, plus a comment on test_detect_dnf_dry_run_abort_stderr noting the bare-cause case is only reachable because verbosity is pinned. Existing exclusion cases already assert the cause text; with verbosity pinned they will always be present.

One honest caveat: the pre-confirmation exclusion still keys off English cause text, so a future dnf rewording could regress it. That's a natural consequence of dnf not exposing a machine-readable refusal channel; the pkg_command locale scoping keeps the messages in English, and the tests lock the current wording.

dnf5 path is unchanged: it emits the confirmation-specific Operation aborted by the user., so it does not depend on verbosity to be distinguishable.

kongche-jbw (approved):
lgtm thx

PR #2625, opened 2026-08-18
feat(cosh-ng): [shell] invert cosh entry to exec-bash passthrough

### LINKED ISSUE #2536: [P0][cosh-ng] 非交互调用(-c/-s/脚本文件/非法选项)被劫持为交互 TUI，命令串不执行
## 背景

ALinux4 默认 login shell 替换验证（cosh-ng 0.15.0-2.alnx4 vs bash 5.2.15-2.alnx4，双架构 fresh ECS paired 执行）确认的产品问题。验证口径与总账见 `artifacts/cosh-login-shell-full-execution-20260811/FINAL-REPORT.md` 与 `issues/CONFIRMED-PRODUCT-ISSUES-20260814.md`。

## 问题

一切带 `-c` 的非交互调用（`-lc`、`-cl`、`-ic`、`-i -c`、`-O extglob -c`、`--posix -c`、`-r -c`、`-n -c`、`-v -x -c`、`--dump-strings`、`--debugger`）都不执行命令串，而是启动交互 TUI；`-s` 模式位置参数丢失；非法选项与不存在的脚本文件参数被静默接受（bash exit 2/127，cosh exit 0 进 UI）。

## 现象与实证（raw 字节级）

oracle=/bin/bash，candidate=/usr/bin/cosh，以下对比来自 paired 执行 raw stdout（双架构一致复现）：

**INV-005.login-combined（`-lc 'printf __DONE__:$-'`）**
- oracle stdout：`__DONE__:hBc`（命令真实执行）
- candidate stdout：`ESC[>4;1m ESC[2K` + `Prompt recommendations are on…` 卡片 + `bash-5.2# exit`，命令串未执行

**STA-008.noninteractive-login（`-lc`，无 tty）**
- oracle stdout（21B）：`__PROFILE__\n__DONE__`
- candidate stdout（916B）：TUI 推荐框 + `exit^M`/`bash-5.2# exit^M` —— UI 直接写入 stdout，脚本管道场景下游被污染

**INV-018.invalid-option（`--definitely-invalid`）**
- bash：exit 2 + usage 全文；cosh：**exit 0** + 交互 UI

**INV-018.missing-file（`cosh /nonexistent.sh`）**
- bash：exit 127 + `No such file or directory`；cosh：**exit 0** + 交互 UI

**INV-008.stdin-positional（`-s name a b`）**
- oracle：`/bin/bash\0name\0a`；candidate：`bash\0\0`（$0 变 `bash`，$1/$2 为空）

**OPT 系列**（均命令未执行、落入 UI）：`-O/+O extglob`（INV-014，shopt 探测未跑）、`--posix`（INV-015）、`-r`（OPT-002，restricted 语义无从验证）、`-n`/`-v -x`（OPT-004）、`--dump-strings`/`--debugger`（OPT-005/006，bash 分别空输出/`cannot start debugger` 报错）

## 证据路径

证据树位于验证工作区（如可访问）：`artifacts/cosh-login-shell-full-execution-20260811/direct-terminal-run-20260814/{x86_64,aarch64}/cases/<case>/{oracle,candidate}/nonpty/stdout.raw`；逐案摘要见同工作区 `issues/CONFIRMED-PRODUCT-ISSUES-20260814.md` ISSUE-01/02/03。

## 疑似根因

option dispatch 未识别 `-c`/`-s`/组合 flags/脚本文件/非法选项，统一回落交互 UI 路径；UI 未区分 tty/非 tty 输出目标；参数校验缺失。

## 验收标准

以下 case 重跑须达到 bash 等价：INV-005、INV-007、INV-008、INV-010(ic/i-c)、INV-014(×2)、INV-015、INV-018(invalid-option/missing-file)、OPT-002、OPT-004(×2)、OPT-005、OPT-006、STA-008（共 12 family）。最小回归：非交互调用不产生任何 TUI 字节、命令串执行、非法参数报错码与 bash 一致。

## 接力指引

- 建议先复现：容器内 `cosh -lc 'echo hi'` 与 `bash -lc 'echo hi'` 对比 stdout 即可见差异。
- 修复面：CLI 参数解析/调用模式分派层（区分 one-shot 与交互）+ 非 tty 输出守卫。
- 关联：与 prompt 合同 issue（TUI 替换 PS1）同根但修复点不同，勿合并。
- 覆盖 family：12/49（P0，阻断 login shell 替换核心路径）。

### LINKED ISSUE #2538: [P1][cosh-ng] $0/argv[0] 语义未对齐：登录缺 - 前缀、argv[0] 不透传、诊断消息前缀
## 背景

ALinux4 默认 login shell 替换验证确认的 `$0`/argv[0] 语义链问题（双架构复现）。总账见 `artifacts/cosh-login-shell-full-execution-20260811/issues/CONFIRMED-PRODUCT-ISSUES-20260814.md` ISSUE-05/06/07。

## 问题（三个子项，同一语义链）

1. **登录 shell `$0` 缺 `-` 前缀**：`su - user -c 'printf "$0:$-"'` → oracle `-bash:hBc`，candidate `bash:hBc`；`runuser -l` 同（SYS-003、APP-002.runuser-login）。
2. **`$0` 不透传 argv[0]**：`sudo -n -u user -- /bin/bash -c 'printf $0'` → oracle `/bin/bash`；换成 `/usr/bin/cosh` 时 candidate 输出 `bash`（硬编码，应为调用路径）（SYS-005、APP-002.runuser-command）。
3. **内层 bash 诊断消息程序名前缀**：错误消息唯一字节差 `/bin/bash: line 1: …` → `bash: line 1: …`（SEM-020.nounset、INV-018.missing-c、IO-014.exec-errors，退出码与语义一致）。

## 实证

- `artifacts/…/system-batch-terminal-run-20260814/x86_64/cases/SYS-003.su-login/operation-00/stdout.raw`：`__DONE__:-bash:hBc` vs `__DONE__:bash:hBc`（rc=0、stderr 空、`$-` 一致）
- `artifacts/…/cases/SYS-005.sudo-user-command/operation-00/stdout.raw`：`__DONE__:/bin/bash:hBc` vs `__DONE__:bash:hBc`
- `artifacts/…/execution-remaining/x86_64/core-formal-first-result-20260813T005909Z-1a3414c5/cases/SEM-020__noninteractive/candidate/nonpty/stdout.raw`：`bash: line 1: unset_value: unbound variable`（oracle 为 `/bin/bash: …`）

## 疑似根因

cosh login 模式未实现 bash 的 `$0="-"+basename(argv0)` 惯例；exec 内层 bash 时 argv[0] 固定为 `bash` 而非调用路径/绝对路径。

## 验收标准

5 family：SYS-003、SYS-005、APP-002、SEM-020、IO-014（INV-018.missing-c 为子证据）。判定：su -/runuser -l 下 `$0` 带 `-` 前缀；`$0` 等于 argv[0]；诊断消息程序名与调用路径一致。

## 接力指引

- 影响：依赖 `$0 == -*` 判断登录态的脚本（大量 profile/运维脚本）会误判。
- 修复面：launcher/adapter 启动内层 shell 的 argv 构造与 $0 设置。

### LINKED ISSUE #2545: [P2][cosh-ng] SSH 服务端 exec 链差异：launcher 脚本经 bash 交接、cosh-shell 双 exec（SSH-004）
## 背景

ALinux4 默认 login shell 替换验证（SSH formal-302 gen-7，双架构 6/6 attempts 一致）确认的 SSH 服务端 exec 链差异。

## 问题

`ssh host 'exec "$SHELL" -c …' name a b`：功能输出一致（stdout `name:a:b`、exit=0），但服务端 execve 链不同：

- oracle：`executables=[sshd, unix_chkpwd, bash, bash]`，trace-correlation pass
- candidate：login 槽位 observed=/usr/bin/bash（method=script-interpreter-handoff），records `["/usr/bin/bash","/usr/bin/cosh","-c",…]` → `["/usr/libexec/anolisa/cosh-ng/cosh-shell","-c",…]`，`candidate_exec_count:2`（cosh-shell 被 exec 两次）

## 实证

`artifacts/cosh-login-shell-full-execution-20260811/ssh-formal-dualarch-20260813/v7-terminal/x86_64/extracted/output/cases/SSH-004.positional/{candidate/step-01,oracle/step-00}/trace-correlation.json`（aarch64 同构）。

## 疑似根因

/usr/bin/cosh 是脚本，经 bash 解释器交接后 exec cosh-shell；`$SHELL` 二次调用重走该链路。

## 影响

审计/进程监控/EDR 看到的进程拓扑与 bash 不同（多一层解释器、cosh-shell 双 exec）；功能无损但可观测性面差异对安全审计敏感场景是真实问题。

## 验收标准

SSH-004：服务端 exec 链与 bash 等价（或产品决策接受该拓扑并文档化，同时关闭本 issue 的审计侧影响评估）。

## 接力指引

可选方向：launcher 改为编译型薄壳（直接 exec cosh-shell 不经解释器），或评估审计影响后显式接受。P2，不阻断功能路径。

SunnyQjm:
# Summary

`/usr/bin/cosh` used to dispatch through two stacked allowlists (an RPM launcher
script plus passthrough token scans inside `cosh-shell`) and fell into the
interactive TUI for every unlisted invocation shape. This PR inverts the
contract per the umbrella design in #2597: **entering the TUI is the allowlist,
and everything else is handed to the inner shell verbatim via `execve`.**

Invariant (acceptance contract): for any invocation vector
(argv × fd tty topology × login form × inherited signal disposition) outside
the TUI allowlist, the observable behavior of `/usr/bin/cosh`
(stdout/stderr bytes, exit code, `$0`, execve chain, child signal
dispositions) matches `exec bash`.

# Changes

- **Compiled entry (packaging)**: `/usr/bin/cosh` becomes a symlink onto
  `cosh-shell` (`cosh-ng.spec.in`); the raw-install wrapper keeps its runtime
  discovery but forwards argv untouched via `exec -a "$0"`. The script hop that
  destroyed the login `-` prefix / `$0` and doubled the execve chain is gone.
- **Single classifier**: new `runtime/invocation.rs::classify_invocation` —
  the TUI requires stdin, stdout, and stderr to all be terminals (stricter
  than bash's stdin+stderr interactivity rule, so no fd topology can land in
  the TUI where `exec bash` would be non-interactive — review round 1 B1),
  no operands, and only cosh-owned or login flags; any other token (including
  unknown/future bash flags) execs the configured shell with argv and argv[0]
  passed through byte-for-byte. TUI-only flags (`--isolated`/`--resume`) are
  forwarded verbatim on the exec path and fail loudly as invalid bash options
  instead of being silently dropped (B2); the `--shell` value guard rejects
  dash-leading/empty values (B3). The login bit has a single source shared
  with the TUI bootstrap (`is_login_invocation`), and the bootstrap iterates
  `args_os` so a non-UTF-8 argv[0] cannot panic it. The scattered token scans
  (`passthrough_non_interactive` branches, `raw_passthrough_args`,
  `should_start_default_raw`) are deleted or reduced to thin adapters.
- **Entry dispatch**: `main` dispatches on the argv[0] basename before any
  runtime initialization, so the passthrough path has zero logging/termios
  side effects. The `cosh-shell` CLI surface (subcommands, `--version/--help`,
  the documented `--` direct-exec extension, `raw` as an explicit TUI request)
  is preserved.
- **SIGPIPE inheritance** (#2541 sub-item A): `shell_host/sigpipe.rs` captures
  the inherited disposition in an `.init_array` constructor and restores it via
  `pre_exec` on both the exec path and the interactive PTY spawn.
- **Tests**: 17 classifier matrix unit tests (20-row input space from the
  design), 5 new cosh-entry integration tests (argv0/`$0` transparency, bash
  error semantics, env `PS1` value transport), signal-exit tests re-anchored to
  exec-in-place semantics, raw packaging stub assertions re-anchored (no more
  injected `raw cosh-core` prefix).

The earlier concern about `-s` narrowing is resolved: any dash-leading
unknown token (including `-s`) passes through verbatim and bash applies its
own invocation rules. Known discoverability trade-off: cosh-specific flag
help is only visible under the `cosh-shell` entry now (`cosh --help` answers
as bash by design).

Intentional behavior changes bundled with the inversion (design-approved):
`cosh --version/--help` answer as bash under the `cosh` entry; `cosh -- <cmd>`
follows bash semantics (the `cosh-shell` extension is unchanged); unknown
`cosh-shell` flags are judged by bash instead of printing usage; passthrough
signal deaths propagate as signals rather than 128+n exit codes.

# Tests

- `cargo test -p cosh-shell --bin cosh-shell runtime::invocation` — 19/19
  (20-row matrix with the three-fd topology enumerated 2^3).
- `cargo test -p cosh-shell --test raw_cli passthrough::` — 27/27 on macOS
  (28 on Linux: the ignored-SIGPIPE inheritance assertion is a Linux-only
  integration test that pins the pre_exec restore into the cargo gate);
  includes the 5 cosh-entry tests plus the TUI-only-flag fail-loud test.
- Repo gates: `check-layout.sh`, `check-test-inventory.sh`,
  `tests/test-package-raw.sh` (+ shellcheck) — all pass.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`
  (rust-toolchain and newer stable) — clean.

# Verification

- **Linux container (alinux3, arm64)**: logic 9/9, protocol 71/71,
  raw_cli 443 passed / 9 failed, shell_host 155 passed / 3 failed,
  lib 1312 passed / 2 failed — every failure reproduces identically on the
  base commit in the same container (pre-existing environment effects:
  SIGINT/SIGQUIT ignored by background-job harness chains, containers without
  init, timeout-sensitive suites). **Zero failures introduced by this diff**
  (`comm -23 <with-diff> <base>` is empty; same result on the macOS host for
  the full bin/lib suites).
- **FAIL→PASS probe matrix (same container, RPM layout + `su -` login chain)**:
  base (script launcher + base binaries) fails 9/11 probes — TUI hijack of
  `-lc`/`--posix`/invalid option/missing script/`--resume`+pipe/`--norc -i`,
  `$0` = `bash` instead of `-cosh`, SIGPIPE SIG_IGN inheritance lost, env `PS1`
  swallowed. Fixed build passes **11/11**, with bare-tty TUI entry and
  SIGPIPE-DFL parity passing on both sides (zero-regression anchors).
- Not run in this PR: the full dual-arch 65-case paired differential matrix on
  ALinux4 ECS (the formal acceptance oracle from #2597; 44/65 cases failed on
  the 2026-08-14 baseline run). Scheduled as follow-up acceptance; the
  in-container probes above cover the same failure families.
- **Expected deviation in the 65-case matrix (acceptance-criteria decision,
  review round 3)**: SYS-003, SYS-005, and the SEM-020/IO-014/
  INV-018.missing-c diagnostics family remain FAIL **by design** under the
  byte-diff criterion. bash prefixes its diagnostics with its own argv[0], so
  argv[0] fidelity (#2538 item 2, a core invariant of this contract) and
  byte-equality with the oracle's `/bin/bash: …` prefix are structurally
  mutually exclusive. These cells are adjudicated as "diagnostic program name
  matches the invoked path" instead — consistent with the acceptance wording
  in #2538, which has a pending problem-statement revision for the same
  reason.

# Evidence

Probe harness and raw outputs (fork assets branch, commit-pinned):

- Probe script (B1–B10, PTY-driven, oracle `/bin/bash`):
  [probes.py](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/probes.py)
- **FAIL baseline** — base commit + original script launcher, 9/11 FAIL:
  [probes-base-launcher.json](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/probes-base-launcher.json)
  (intermediate control, base binaries + symlink layout, 4/11 FAIL:
  [probes-base.json](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/probes-base.json))
- **PASS run** — this PR's binaries, 11/11 PASS:
  [probes-fixed-final.json](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/probes-fixed-final.json)
  (re-run after review round 1, incl. the fail-loud B5 criterion:
  [probes-fixed-r1.json](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/probes-fixed-r1.json))
- Raw packaging gate (shellcheck + test-package-raw.sh) in container:
  [raw-packaging-gate.json](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/raw-packaging-gate.json)
- Full-suite failure-set equivalence (with-diff vs base, serial runs):
  [serial-full-withdiff.log](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/serial-full-withdiff.log)
  vs
  [serial-full-base.log](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/serial-full-base.log)
- Consolidated verification ledger:
  [baseline.md](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/baseline.md)

Closes #2536
Closes #2538
Closes #2545
Refs #2597 (umbrella U1), #2537 (`--norc`/flag contract sub-item),
#2541 (sub-item A)

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/main.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/main.rs
@@ -42,7 +42,8 @@ mod types;
 #[path = "ui/public.rs"]
 mod ui;
 
-use runtime::cli_args::{configured_raw_invocation, should_start_default_raw};
+use runtime::cli_args::configured_raw_invocation;
+use runtime::invocation::{classify_invocation, exec_shell, is_cosh_entry, Invocation};
 use runtime::startup::{
     passthrough_non_interactive, passthrough_raw_non_interactive, print_usage_help,
 };
@@ -74,7 +75,52 @@ mod binary_compat {
 #[allow(unused_imports)]
 pub(crate) use binary_compat::*;
 
+/// Full argv (argv[0] included) when this process was invoked through the
+/// `/usr/bin/cosh` entry; `None` for the `cosh-shell` CLI surface.
+fn cosh_entry_args() -> Option<Vec<std::ffi::OsString>> {
+    let args = std::env::args_os().collect::<Vec<_>>();
+    is_cosh_entry(args.first()?).then_some(args)
+}
+
 fn main() {
+    // `/usr/bin/cosh` entry (argv[0] basename `cosh`, login `-cosh` included):
+    // the invocation-transparency contract dispatches before any runtime
+    // initialization so passthrough leaves no logs or terminal side effects.
+    if let Some(args_os) = cosh_entry_args() {
+        use std::io::IsTerminal;
+
+        let stdin_tty = std::io::stdin().is_terminal();
+        let stdout_tty = std::io::stdout().is_terminal();
+        let stderr_tty = std::io::stderr().is_terminal();
+        match classify_invocation(
+            &args_os[0],
+            &args_os[1..],
+            stdin_tty,
+            stdout_tty,
+            stderr_tty,
+        ) {
+            Invocation::ExecShell(plan) => std::process::exit(exec_shell(plan)),
+            Invocation::Tui(_) => {
+                runtime::terminal::install_terminal_recovery();
+                let cosh_config = config::load_config();
+                runtime::logging::init_logging(&cosh_config.log_level);
+                tracing::info!(version = env!("CARGO_PKG_VERSION"), "cosh-shell starting");
+                // The allowlist only admits valid UTF-8 tokens in args[1..];
+                // argv[0] may be arbitrary bytes, so it is converted lossily
+                // (the TUI never re-parses it).
+                let args = args_os
+                    .iter()
+                    .map(|arg| arg.to_string_lossy().into_owned())
+                    .collect::<Vec<_>>();
+                let (adapter_name, shell_kind, launch_options) =
+                    configured_raw_invocation(&args[1..]);
+                let status =
+                    runtime::controller::run_raw(&adapter_name, shell_kind, launch_options);
+                std::process::exit(status);
+            }
+        }
+    }
+
     let args = std::env::args().collect::<Vec<_>>();
 
     if args.get(1).map(String::as_str) == Some("--version") {
@@ -115,11 +161,11 @@ fn main() {
         if let Some(status) = passthrough_non_interactive(&args) {
             std::process::exit(status);
         }
-        if should_start_default_raw(&args[1..]) {
-            let (adapter_name, shell_kind, launch_options) = configured_raw_invocation(&args[1..]);
-            let status = runtime::controller::run_raw(&adapter_name, shell_kind, launch_options);
-            std::process::exit(status);
-        }
+        // The classifier admitted the TUI: only cosh-owned/login tokens on a
+        // terminal reach this point.
+        let (adapter_name, shell_kind, launch_options) = configured_raw_invocation(&args[1..]);
+        let status = runtime::controller::run_raw(&adapter_name, shell_kind, launch_options);
+        std::process::exit(status);
     }
 
     let status = match args.get(1).map(String::as_str) {
--- a/src/cosh-ng/crates/cosh-shell/src/runtime/cli_args.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/runtime/cli_args.rs
@@ -30,6 +30,10 @@ pub(crate) fn adapter_name_from_args(args: &[String]) -> Option<&str> {
                     idx += 1;
                 }
             }
+            // Standalone allowlist flags share one vocabulary source with the
+            // classifier, so a token admitted into the TUI there can never be
+            // read as an adapter name here.
+            arg if crate::runtime::invocation::TUI_STANDALONE_FLAGS.contains(&arg) => idx += 1,
             arg if arg.starts_with("--shell=") => idx += 1,
             arg if arg.starts_with("--resume=") => idx += 1,
             arg if arg.starts_with("--") => idx += 1,
@@ -74,31 +78,6 @@ pub(crate) fn raw_shell_from_args(args: &[String]) -> Option<RawShellKind> {
     }
 
     None
-}
-
-pub(crate) fn should_start_default_raw(args: &[String]) -> bool {
-    let mut idx = 0;
-    while idx < args.len() {
-        match args[idx].as_str() {
-            "--shell" => {
-                if !matches!(args.get(idx + 1), Some(value) if !value.starts_with("--")) {
-                    return false;
-                }
-                idx += 2;
-            }
-            "--resume" => {
-                idx += 1;
-                if args.get(idx).is_some_and(|value| !value.starts_with('-')) {
-                    idx += 1;
-                }
-            }
-            "--isolated" | "--login" | "-l" => idx += 1,
-            arg if arg.starts_with("--shell=") || arg.starts_with("--resume=") => idx += 1,
-            _ => return false,
-        }
-    }
-
-    true
 }
 
 pub(crate) fn parse_raw_shell(value: &str) -> RawShellKind {
@@ -195,8 +174,8 @@ fn cosh_shell_default_state_path_for_home(home: &std::path::Path) -> std::path::
 mod tests {
     use super::{
         adapter_name_from_args, cosh_shell_default_state_path_for_home, launch_options_from_args,
-        parse_raw_shell, raw_shell_from_args, shell_from_default_or_auto, should_start_default_raw,
-        RawShellKind, ResumeLaunch,
+        parse_raw_shell, raw_shell_from_args, shell_from_default_or_auto, RawShellKind,
+        ResumeLaunch,
     };
 
     #[test]
@@ -290,30 +269,51 @@ mod tests {
     }
 
     #[test]
-    fn no_subcommand_interactive_raw_accepts_only_shell_entry_options() {
-        assert!(should_start_default_raw(&[]));
-        assert!(should_start_default_raw(&["--login".to_string()]));
-        assert!(should_start_default_raw(&["-l".to_string()]));
-        assert!(should_start_default_raw(&[
-            "--shell".to_string(),
-            "zsh".to_string(),
-            "--isolated".to_string()
-        ]));
-        assert!(should_start_default_raw(&["--shell=bash".to_string()]));
-        assert!(should_start_default_raw(&["--resume".to_string()]));
-        assert!(should_start_default_raw(&[
-            "--resume".to_string(),
-            "00000000-0000-4000-8000-000000000000".to_string(),
-            "--shell=zsh".to_string(),
-        ]));
+    fn tui_allowlist_shapes_never_parse_an_adapter_name() {
+        // Vocabulary contract shared with the invocation classifier. The
+        // standalone-flag shapes are generated from the classifier's own
+        // vocabulary (`TUI_STANDALONE_FLAGS`), so a flag added there is
+        // covered here mechanically; only the value-carrying shapes
+        // (`--shell`, `--resume`) remain hand-enumerated because their
+        // shapes are flag-specific.
+        use crate::runtime::invocation::{classify_invocation, Invocation, TUI_STANDALONE_FLAGS};
+        use std::ffi::{OsStr, OsString};
 
-        assert!(!should_start_default_raw(&["fake".to_string()]));
-        assert!(!should_start_default_raw(&["--shell".to_string()]));
-        assert!(!should_start_default_raw(&[
-            "--shell".to_string(),
-            "--isolated".to_string()
-        ]));
-        assert!(!should_start_default_raw(&["--unknown".to_string()]));
+        let mut shapes: Vec<Vec<String>> = vec![vec![]];
+        for flag in TUI_STANDALONE_FLAGS {
+            shapes.push(vec![flag.to_string()]);
+        }
+        // All standalone flags combined, and each behind a consumed --shell.
+        shapes.push(TUI_STANDALONE_FLAGS.iter().map(|s| s.to_string()).collect());
+        for flag in TUI_STANDALONE_FLAGS {
+            shapes.push(vec!["--shell".into(), "zsh".into(), flag.to_string()]);
+        }
+        for value_shape in [
+            vec!["--resume".to_string()],
+            vec![
+                "--resume".to_string(),
+                "00000000-0000-4000-8000-000000000000".to_string(),
+            ],
+            vec!["--shell=bash".to_string()],
+        ] {
+            shapes.push(value_shape);
+        }
+
+        for args in shapes {
+            let os_args: Vec<OsString> = args.iter().map(OsString::from).collect();
+            assert!(
+                matches!(
+                    classify_invocation(OsStr::new("cosh"), &os_args, true, true, true),
+                    Invocation::Tui(_)
+                ),
+                "allowlist shape must stay TUI: {args:?}"
+            );
+            assert_eq!(
+                adapter_name_from_args(&args),
+                None,
+                "allowlist token misread as adapter: {args:?}"
+            );
+        }
     }
 
     #[test]
--- a/src/cosh-ng/packaging/raw/assets/bin/cosh
+++ b/src/cosh-ng/packaging/raw/assets/bin/cosh
@@ -41,15 +41,10 @@ for runtime_dir in \
         [ -x "$runtime_dir/cosh-shell" ] && \
         [ -x "$runtime_dir/cosh-core" ]; then
         export PATH="$runtime_dir:$PATH"
-        case "${1:-}" in
-            -c | -- | --help | --version)
-                exec "$runtime_dir/cosh-shell" "$@"
-                ;;
-        esac
-        if [ "$#" -eq 0 ] && [ ! -t 0 ]; then
-            exec "$runtime_dir/cosh-shell"
-        fi
-        exec "$runtime_dir/cosh-shell" raw cosh-core "$@"
+        # This wrapper only resolves the private runtime; the TUI-vs-
+        # passthrough decision is cosh-shell's invocation classifier,
+        # keyed off the forwarded invocation name.
+        exec -a "$0" "$runtime_dir/cosh-shell" "$@"
     fi
 done
 
--- a/src/cosh-ng/cosh-ng.spec.in
+++ b/src/cosh-ng/cosh-ng.spec.in
@@ -45,22 +45,11 @@ install -m 0755 target/release/cosh-shell %{buildroot}%{_libexecdir_cosh}/cosh-s
 install -d %{buildroot}%{_bindir}
 install -m 0755 target/release/cosh-cli %{buildroot}%{_bindir}/cosh-cli
 
-# Install launcher script
-cat > %{buildroot}%{_bindir}/cosh << 'EOF'
-#!/bin/bash
-case "${1:-}" in
-  -c|--|--help|--version)
-    exec %{_libexecdir_cosh}/cosh-shell "$@"
-    ;;
-esac
-
-if [ "$#" -eq 0 ] && [ ! -t 0 ]; then
-  exec %{_libexecdir_cosh}/cosh-shell "$@"
-fi
-
-exec %{_libexecdir_cosh}/cosh-shell raw cosh-core "$@"
-EOF
-chmod 0755 %{buildroot}%{_bindir}/cosh
+# /usr/bin/cosh is the compiled entry: a symlink onto cosh-shell, which
+# dispatches on argv[0]. A script launcher would lose the login argv[0]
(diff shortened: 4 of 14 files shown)


src/cosh-ng/packaging/raw/assets/bin/cosh
```diff
@@ -41,15 +41,10 @@ for runtime_dir in \
...
-                ;;
-        esac
-        if [ "$#" -eq 0 ] && [ ! -t 0 ]; then
-            exec "$runtime_dir/cosh-shell"
-        fi
-        exec "$runtime_dir/cosh-shell" raw cosh-core "$@"
+        # This wrapper only resolves the private runtime; the TUI-vs-
+        # passthrough decision is cosh-shell's invocation classifier,
+        # keyed off the forwarded invocation name.
+        exec -a "$0" "$runtime_dir/cosh-shell" "$@"
```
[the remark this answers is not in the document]
SunnyQjm: Verified against the sources — the factual claim is correct, and it matches a boundary this design declared explicitly rather than a regression introduced here. Breakdown:

**1. Kernel behavior: confirmed.** For the raw/tar install form the wrapper is a script, so binfmt_script rewrites argv[0] to the script pathname before bash starts; `exec -a "$0"` can only forward the script path, and a login `-` prefix is unrecoverable in-script.

**2. Not a regression — behavior on this axis strictly improves vs base.** On the base launcher the raw form spawned the inner shell as plain `bash`: passthrough `$0` observed `bash`, and bare login invocations also started a non-login session (the base script exec'd `cosh-shell raw cosh-core`, argv[0] without `-`). With this PR the raw form forwards the wrapper path (basename `cosh`, classifier routes correctly); the login `-` limitation is unchanged from base. Container evidence: in the base probe run `su - <user> -c 'echo $0'` printed `bash` ([probes-base-launcher.json, B8](https://raw.githubusercontent.com/SunnyQjm/anolisa/bc3c170e509a3fa934999d2e45897e4c37a7cea5/probes-base-launcher.json)); the full argv[0]/login contract is delivered on the RPM form, where `/usr/bin/cosh` is now a symlink and B8 observes `-cosh` ([probes-fixed-final.json](https://raw.githubusercontent.com/SunnyQjm/anolisa/bc3c170e509a3fa934999d2e45897e4c37a7cea5/probes-fixed-final.json)).

**3. Why the raw form was not converted to a symlink in this PR.** The raw payload has an explicit no-symlink packaging contract — `packaging/raw/package.sh` hard-fails staging on any symlink ("raw payload contains a symbolic link") and `tests/test-package-raw.sh` asserts `find -type l` is empty. Converting `bin/cosh` to a symlink is feasible mechanically (cosh-shell already resolves cosh-core via `current_exe()` sibling lookup in `crates/cosh-shell/src/adapter/cosh_core.rs`, which follows /proc/self/exe through symlinks, so the wrapper's PATH injection is not load-bearing), but it requires lifting that packaging contract and verifying the ANOLISA raw-backend installer preserves symlinks end-to-end — a scoped follow-up rather than a rider on this PR.

**Disposition:** keeping the raw-form thin shim in this PR (argv forwarded verbatim, dispatch owned by the classifier), with the symlink conversion + installer symlink-support evaluation tracked as a follow-up. Happy to file the tracking issue if maintainers agree with that split.

src/cosh-ng/crates/cosh-shell/src/main.rs
```diff
@@ -74,7 +75,52 @@ mod binary_compat {
...
+                tracing::info!(version = env!("CARGO_PKG_VERSION"), "cosh-shell starting");
+                // The allowlist only admits valid UTF-8 tokens in args[1..];
+                // argv[0] may be arbitrary bytes, so it is converted lossily
+                // (the TUI never re-parses it).
+                let args = args_os
+                    .iter()
+                    .map(|arg| arg.to_string_lossy().into_owned())
+                    .collect::<Vec<_>>();
+                let (adapter_name, shell_kind, launch_options) =
+                    configured_raw_invocation(&args[1..]);
```
kongche-jbw, line 116: **[P1] Preserve the explicit `-l` TUI entry through adapter parsing**

The classifier admits `cosh -l` as a TUI invocation, but this call passes `args[1..]` unchanged to `configured_raw_invocation`. `adapter_name_from_args` does not skip the short login flag, so it returns `-l` as the adapter name. With stdin/stdout/stderr all attached to a PTY, the current head reproducibly prints `unknown adapter: -l` and exits 2 instead of starting the documented login TUI. The 19 classifier tests still pass because they stop at `Invocation::Tui` and do not exercise this downstream launch path.

kongche-jbw (commented):
发现 1 个新的 P1 correctness blocker；定向测试虽通过，但 PTY 端到端探针可以稳定复现。

src/cosh-ng/crates/cosh-shell/src/main.rs
SunnyQjm: Confirmed and fixed in `3a36052f` (rebased onto current main). Reproduced exactly as described: PTY-driven `argv[0]=cosh -l` on the previous head printed `unknown adapter: -l` and exited 2.

**Root cause (one level deeper than the missing skip):** the classifier's allowlist vocabulary and the launch-side scan vocabulary were two independent implementations. `-l` is the only allowlist token without a `--` prefix, so `adapter_name_from_args`'s catch-all (`--`-prefix skip, else adapter name) misread it — the same dual-implementation family as the login-bit drift fixed in round 1, surfacing on the token axis.

**Fix:** `adapter_name_from_args` now skips `-l`; and to close the family rather than the instance, a vocabulary-contract test (`tui_allowlist_shapes_never_parse_an_adapter_name`) asserts that **every** argv shape the classifier admits into the TUI resolves to the configured default adapter — any future allowlist/launch-scan drift fails this test directly. Your observation about the classifier tests stopping at `Invocation::Tui` was exactly right; this contract test spans the boundary those tests stopped at.

**End-to-end evidence:** probe B11 (PTY `cosh -l` → login TUI banner, `unknown adapter` absent) added to the probe suite; container run is 12/12 PASS post-fix ([probes-fixed-r2.json](https://raw.githubusercontent.com/SunnyQjm/anolisa/eecf47b3db81a623127cb3639465f8bc585950bf/probes-fixed-r2.json) — asset link will be refreshed to the new commit alongside the next asset push). Classifier 19/19 and cli_args (incl. the new contract test) green.

src/cosh-ng/crates/cosh-shell/src/runtime/cli_args.rs
```diff
@@ -290,30 +268,38 @@ mod tests {
...
+        use std::ffi::{OsStr, OsString};
 
-        assert!(!should_start_default_raw(&["fake".to_string()]));
-        assert!(!should_start_default_raw(&["--shell".to_string()]));
-        assert!(!should_start_default_raw(&[
-            "--shell".to_string(),
-            "--isolated".to_string()
-        ]));
-        assert!(!should_start_default_raw(&["--unknown".to_string()]));
+        for args in [
```
kongche-jbw: **[P3] Keep the allowlist contract test mechanically coupled to the classifier**

This test covers the current allowlist vocabulary, but the inputs are still a manually maintained second list. If a future TUI-owned flag is added to `classify_invocation` without adding another row here, the adapter parser can drift again while this “every allowlist shape” test remains green. This is non-blocking for the current fix—the existing token classes are represented—but it does not yet close the drift family mechanically as the test name and comment imply.

kongche-jbw (commented):
One non-blocking test-maintainability observation from the follow-up review; the current vocabulary is covered and the `-l` fix itself is verified.

src/cosh-ng/crates/cosh-shell/src/runtime/cli_args.rs
SunnyQjm: Accepted — and taken one step further than a longer hand-list, in `eb8a61a4`. You're right that the test was itself a second manually maintained vocabulary, so the fix removes the second list instead of growing it: the standalone allowlist flags are now a single shared constant, `invocation::TUI_STANDALONE_FLAGS`, consumed by **three** sites — the classifier scan arm, the launch-side `adapter_name_from_args` skip arm (replacing the hand-written `-l` case), and this contract test, whose standalone shapes are generated from the constant (each flag alone, all combined, and each behind a consumed `--shell`). Adding a TUI-owned standalone flag now updates classifier, adapter scan, and test mechanically from one edit.

The remaining hand-maintained boundary is declared instead of implied: value-carrying flags (`--shell`, `--resume`) have flag-specific shapes (bare / with value / `=value`) that can't be table-generated without data-driving the whole scanner, so their shape rows stay explicit and the test comment says so. Verified: classifier 19/19, cli_args 5/5 (incl. the regenerated contract test), clippy clean; pushed rebased onto current main.

kongche-jbw (approved):

PR #2626, opened 2026-08-18
feat(anolisa): reject unknown status targets

kongche-jbw:
## Why

`anolisa status` reads component lifecycle state from `installed.toml`, while
telemetry is managed independently as a system service. The command previously
treated any unresolved input as a synthetic `not_installed` component, which
made unsupported names such as the telemetry unit look like valid components.

## What changed

Resolve exact stored identities first, then canonical component names and
aliases or package names declared by the central component index. Reject
unresolved names with guidance to `anolisa list`; if the index is unavailable,
fail explicitly because a new identity cannot be validated. Exact stored
identities remain inspectable offline. Known telemetry unit targets point to the
centrally defined `anolisa telemetry status` command.

Backend `package_map` remains downstream configuration for RPM package
selection and does not establish component identity.

## Related issue

no-issue: telemetry status confusion was reported directly

## User / Agent impact

Unsupported component names now produce an actionable `INVALID_ARGUMENT`
response when the central index is available. If it is unavailable, unverifiable
targets produce `EXECUTION_FAILED` instead of a misleading synthetic status.
Existing stored components retain exact-name precedence, and supported but
uninstalled components still report `not_installed`.

## Risk and compatibility

Low risk. The change only affects named status target resolution; component
state, service lifecycle, and `installed.toml` remain unchanged.

## Validation

- `cargo test -p anolisa-cli` (824 passed, 1 ignored)
- `cargo check -p anolisa-cli`
- `cargo clippy -p anolisa-cli --all-targets -- -D warnings`
- `cargo doc -p anolisa-cli --no-deps`
- Manually verified telemetry and unavailable-index errors in human and JSON
  output, including exit codes and `status --help`

## Documentation and rollback

Updated the inline CLI help for the `status` component argument. Roll back by
reverting commit `7e08283c`.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/status.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/status.rs
@@ -3,8 +3,10 @@
 //! Reads `installed.toml` via the shared [`crate::commands::common`] helper
 //! and lists every `Component`-kind object, or filters down to a single
 //! name. A missing state file is the expected fresh-install case and yields
-//! an empty result; an unknown component name surfaces a synthetic
-//! `not_installed` record rather than an error (launch spec §7.1).
+//! an empty result. When the component index is available, unsupported names
+//! are rejected with discovery guidance; without an index, new names cannot
+//! be validated and fail explicitly. Exact identities already present in state
+//! always win and remain inspectable offline.
 //!
 //! This handler reports state-on-disk plus live read-only probes. Every
 //! persisted field in [`ComponentRecord`] is projected straight from
@@ -41,17 +43,24 @@ use crate::color::{Palette, pad_right};
 use crate::commands::common;
 use crate::commands::common::RepoPersistPolicy;
 use crate::commands::state_view::{StateScope, StateView, StateVisibility};
+use crate::commands::telemetry::status_command_for_service_target;
 use crate::commands::tier1::install::rpm_package_candidates_with_index;
 use crate::context::{CliContext, InstallMode};
 use crate::repo_config::BackendConfig;
-use crate::resolution::{ComponentIndex, ResolutionUse, load_optional_component_index};
+use crate::resolution::{
+    ComponentIndex, ResolutionUse, load_optional_component_index, lookup_component_alias,
+};
 use crate::response::{CliError, render_json};
 
 const COMMAND: &str = "status";
 
 #[derive(Parser)]
 pub struct StatusArgs {
-    /// Show detail for a specific component (omit for aggregate view).
+    /// Show detail for a specific ANOLISA component (omit for aggregate view).
+    ///
+    /// Run `anolisa list` to view supported component names.
+    ///
+    /// Use `anolisa telemetry status` for telemetry service state.
     pub component: Option<String>,
 }
 
@@ -177,15 +186,12 @@ pub fn handle(args: StatusArgs, ctx: &CliContext) -> Result<(), CliError> {
     let adapter_scan = common::build_adapter_manager(ctx).scan().ok();
 
     let query = RpmPackageQuery::system();
-    let selected_component = args
+    // A named query uses the component index both to validate/resolve its
+    // identity and, in system mode, to map the observed RPM probe below.
+    // Loading remains best-effort so installed state is inspectable offline.
+    let repo_config = args
         .component
-        .as_deref()
-        .map(|target| lookup_component_name_from_view(target, &view, ctx));
-
-    // repo_config / component_index are still needed for the observed-record
-    // probe below (system mode only). Name resolution above is handled by
-    // common::lookup_component_name which loads its own config.
-    let repo_config = (ctx.install_mode == InstallMode::System && args.component.is_some())
+        .is_some()
         .then(|| {
             common::load_repo_config(ctx, &layout, COMMAND, RepoPersistPolicy::BestEffort).ok()
         })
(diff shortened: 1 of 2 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/status.rs
```diff
@@ -251,6 +260,19 @@ pub fn handle(args: StatusArgs, ctx: &CliContext) -> Result<(), CliError> {
...
 
+fn validate_component_target(target: Option<&str>) -> Result<(), CliError> {
+    let Some(target @ (TELEMETRY_SERVICE_NAME | TELEMETRY_UNIT_FILENAME)) = target else {
+        return Ok(());
+    };
+
+    Err(CliError::InvalidArgument {
+        command: COMMAND.to_string(),
+        reason: format!(
+            "target '{target}' is the telemetry service, not an ANOLISA component; run `anolisa telemetry status` to inspect it"
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 4a78b4f2. The telemetry module now owns the canonical status command and exposes target-to-command routing, so `status` no longer hardcodes the management command in its error path.

src/anolisa/crates/anolisa-cli/src/commands/tier1/status.rs
```diff
@@ -171,6 +178,8 @@ pub(crate) enum AggregateRecordSelection {
 }
 
 pub fn handle(args: StatusArgs, ctx: &CliContext) -> Result<(), CliError> {
+    validate_component_target(args.component.as_deref())?;
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 4a78b4f2. Exact component identities from visible state now win before any redirect or index validation. The resolver then accepts canonical index names, aliases/packages, and site-local package-map identities; only unresolved names are rejected when authoritative index data is available. Added a regression test for an installed component named `anolisa-telemetry`.

src/anolisa/crates/anolisa-cli/src/commands/tier1/status.rs
```diff
@@ -251,6 +264,43 @@ pub fn handle(args: StatusArgs, ctx: &CliContext) -> Result<(), CliError> {
...
+    }
+
+    if let Some(index) = component_index {
+        if index.components.iter().any(|entry| entry.name == input) {
+            return Ok(input.to_string());
+        }
+        if let Some(component) = lookup_component_alias(input, Some(index)) {
+            return Ok(component);
+        }
+        if rpm_backend.is_some_and(|backend| backend.package_map.contains_key(input)) {
```
ikunkun-sys: [P1] Do not let `package_map` establish component identity

This branch accepts a name that is absent from the central component index merely because a local RPM mapping exists. That makes component identity depend on host-local configuration, so two hosts using the same index can disagree about whether the same input is valid. With the index as the sole authority, `package_map` should only select the backend package after index membership has already been established; it must not introduce a component identity.

src/anolisa/crates/anolisa-cli/src/commands/tier1/status.rs
```diff
@@ -251,6 +264,43 @@ pub fn handle(args: StatusArgs, ctx: &CliContext) -> Result<(), CliError> {
...
+            || "run `anolisa list` to view supported components".to_string(),
+            |command| format!("run `{command}` to inspect telemetry state"),
+        );
+        return Err(CliError::InvalidArgument {
+            command: COMMAND.to_string(),
+            reason: format!("unsupported component '{input}'; {guidance}"),
+        });
+    }
+
+    Ok(input.to_string())
```
ikunkun-sys: [P1] Fail when the component authority is unavailable

Returning the literal input here treats an unavailable index as proof that any name is valid, and the downstream status path then reports `not_installed`. That makes an arbitrary/unsupported name indistinguishable from a supported but uninstalled component. Exact identities already recorded in local state may continue to work offline, but resolving a new name without the index should return an index-unavailable error.

ikunkun-sys (changes requested):
The component index should be the sole authority for component identities not already present in exact local state. This revision still keeps two side-door identity paths: `package_map` and the no-index literal fallback. Please preserve exact stored identities for lifecycle continuity, but do not let either branch establish a new component identity. Project-wide resolver convergence can be tracked separately.

src/anolisa/crates/anolisa-cli/src/commands/tier1/status.rs
kongche-jbw: Fixed in 7e08283c. Removed `package_map` from status identity admission; only exact stored identities and the central component index can establish a component target. `package_map` remains downstream-only for RPM package selection and observation.

src/anolisa/crates/anolisa-cli/src/commands/tier1/status.rs
kongche-jbw: Fixed in 7e08283c. A non-exact target now returns `EXECUTION_FAILED` when the component index is unavailable, with guidance to `anolisa list`. Exact stored identities continue to work offline; regression tests cover both cases.

ikunkun-sys (approved):
Re-reviewed the current head. The two previous blockers are fixed: `package_map` no longer establishes component identity, and an unseen target now fails explicitly when the component index is unavailable while exact stored identities remain inspectable offline. The full diff and resolver interactions were rechecked; local fmt, tests, Clippy, and rustdoc pass, and I found no new blocking issues.

PR #2650, opened 2026-08-18
fix(anolisa): identify stale repo source

### LINKED ISSUE #2344: [anolisa] identify repo config in stale-index errors
## Component

anolisa

Issue #2086 and PR #2171 improved the error for a missing local raw-repository index:

```text
local raw repository index not found at /tmp/.../repo/v1/index.toml;
check the raw repository URL in repo.toml or, if supplied, the one-off
--repo <URL> override
```

That diagnostic identifies the missing index, but users still cannot tell which
`repo.toml` supplied the stale URL. ANOLISA discovers repository configuration
from multiple locations (user/site, packaged data, and the development tree), so
"check repo.toml" is not enough to recover without inspecting the source code or
asking for support.

A real user-mode failure was caused by
`~/.config/anolisa/repo.toml` retaining a `file://` URL for an already removed
temporary repository. Moving that config aside and retrying allowed ANOLISA to
provision its default repository config and complete the install.

## Reproduction

1. In a temporary user-mode filesystem layout, write `repo.toml` with
   `[backends.raw].base_url` pointing to a local `file://` repository.
2. Remove that repository while leaving `repo.toml` in place.
3. Run `anolisa install <component>`.
4. Observe that the error names the missing index and the generic filename
   `repo.toml`, but not the configuration file that supplied the URL.

## Expected user-visible outcome

When the failing raw URL came from discovered repository configuration, the
diagnostic should include the exact active config path and a safe recovery
direction, for example:

```text
error: local raw repository index not found at /tmp/.../repo/v1/index.toml
hint: repository config: /home/alice/.config/anolisa/repo.toml;
      move it aside and retry to restore the default config, or pass
      --repo <URL> once
```

When the URL came from an explicit `--repo` override, the diagnostic should
identify the override instead of blaming a config file. Exact wording may follow
existing CLI conventions.

## Suggested implementation direction

- Preserve the selected repository-config source path when
  `RepoConfig::load_local_with_sources` resolves the discovery chain.
- Keep that source as internal metadata on the loaded config or load result; do
  not infer it later by parsing an error string.
- Thread a small typed origin value (configured path versus CLI override) into
  raw-index error attribution.
- Reuse the existing missing-index mapping in
  `commands/tier1/install/raw.rs`.
- Keep tests hermetic with `tempfile` and `file://` repositories; no live
  network or privileged host is required.

Relevant code:

- `src/anolisa/crates/anolisa-cli/src/repo_config.rs`
- `src/anolisa/crates/anolisa-cli/src/commands/common.rs`
- `src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs`
- `src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs`
- `src/anolisa/crates/anolisa-cli/src/commands/tier1/install/tests/raw_e2e.rs`

## Acceptance criteria

- [ ] A stale raw URL loaded from a user/site `repo.toml` reports the exact
      selected config path.
- [ ] Repository discovery precedence remains unchanged.
- [ ] A missing index caused by `--repo file://...` identifies the one-off
      override and does not claim that `repo.toml` supplied the URL.
- [ ] The recovery text is non-destructive: it recommends moving the config
      aside or using a one-off override, but never removes or rewrites files
      automatically.
- [ ] Non-`NotFound` filesystem errors and HTTP/network errors retain their
      current diagnostics.
- [ ] Human-readable and `--json` error output remain valid.
- [ ] Focused regression tests cover both configured-path and `--repo`
      attribution without network access.
- [ ] The applicable ANOLISA formatting, Clippy, and test checks pass in CI.

## Non-goals

- Adding an `anolisa repo reset`, `repo status`, or general config-management
  command.
- Automatically deleting, replacing, or editing repository configuration.
- Automatically falling back to the public repository.
- Changing repository discovery precedence.
- Broad repository health checks in `anolisa doctor`.
- Diagnosing unsupported operating systems or architectures.

## Related work

Follow-up to #2086, which intentionally excluded tracking the exact
configuration file and automatic recovery.

guiyangyuan:
## Why

ANOLISA reports the missing path when a local raw-repository index becomes stale, but it does not identify which discovered `repo.toml` supplied that URL. Because repository configuration can come from several locations, users must inspect the source or ask for support before they can recover.

## What changed

- Preserve the selected `repo.toml` path as internal `RepoConfig` metadata without changing discovery precedence.
- Thread a typed raw-repository origin through resolution so diagnostics distinguish configured URLs from one-off `--repo` overrides.
- Attribute exact local-index `NotFound` failures to the active config path or CLI override while preserving all other download diagnostics.
- Add hermetic regressions for both origins and keep JSON error rendering covered.

## Related issue

Closes #2344

## User / Agent impact

Users and agents can now identify the exact stale repository configuration and receive a safe recovery direction. Explicit `--repo` failures no longer incorrectly blame `repo.toml`.

## Risk and compatibility

Low risk: only the human-readable reason for an exact local raw-index `NotFound` changes. Repository discovery, JSON structure, non-`NotFound` I/O failures, and HTTP/network diagnostics remain unchanged.

## Validation

Environment: macOS arm64.

- `cargo check --locked`
- `cargo clippy -p anolisa-cli --all-targets --locked --no-deps -- -D warnings`
- `cargo test -p anolisa-cli missing_local_repository_index -- --nocapture` (2 passed)
- `cargo test -p anolisa-cli fetch_raw_index -- --nocapture` (3 passed)
- `cargo test -p anolisa-cli repo_config -- --nocapture` (27 passed)
- `cargo doc --no-deps --locked`
- `git diff --check`

The full `cargo test --locked` run completed with 825 passed, 1 ignored, and 1 failure in `raw_update_rollback_hydrates_legacy_required_capability`. The same missing-central-log failure reproduces on the unmodified base commit `33ed384`, so it is not introduced by this patch. Workspace-wide strict Clippy is likewise blocked by an existing `dead_code` warning in `anolisa-core/src/capability.rs`; the changed `anolisa-cli` crate passes strict Clippy with `--no-deps`.

## Documentation and rollback

No documentation change is needed because there is no new command, flag, or configuration field. Reverting commit `a086bdd` restores the previous generic attribution text.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/dispatch.rs
@@ -59,7 +59,7 @@ use super::render::repo_config_err;
 use super::rpm::{
     PinError, RpmTarget, resolve_pinned_candidate, rpm_package_candidates_with_index,
 };
-use super::types::{InstallOutcome, RawResolution, ResolveInputs};
+use super::types::{InstallOutcome, RawRepositoryOrigin, RawResolution, ResolveInputs};
 use super::{ANOLISA_RPM_REPO_ID, COMMAND, InstallArgs};
 
 /// Dispatch `install <component>` against the live host.
@@ -899,7 +899,7 @@ fn resolve_owned_artifact(
         .map_err(|err| repo_config_err(err, true))?;
 
     let mut warnings: Vec<String> = Vec::new();
-    let base_url = match args.repo.as_deref() {
+    let (base_url, repository_origin) = match args.repo.as_deref() {
         Some(override_url) => {
             let normalized =
                 normalize_override_url(override_url).map_err(|err| repo_config_err(err, true))?;
@@ -908,17 +908,21 @@ fn resolve_owned_artifact(
                     "--repo uses plaintext http ({normalized}) — artifacts are still sha256-verified on the raw backend, but the index itself is unauthenticated",
                 ));
             }
-            normalized
+            (normalized, Some(RawRepositoryOrigin::CliOverride))
         }
         None => {
             let host = HostVars {
                 os: env.os.clone(),
                 arch: env.arch.clone(),
             };
-            repo_config
+            let base_url = repo_config
                 .resolved_base_url(backend_name, backend, &host)
                 // Variable errors are fixed by editing [vars] in repo.toml.
-                .map_err(|err| repo_config_err(err, true))?
+                .map_err(|err| repo_config_err(err, true))?;
+            let origin = repo_config
+                .source_path()
+                .map(|path| RawRepositoryOrigin::Config(path.to_path_buf()));
+            (base_url, origin)
         }
     };
     let (component, package) = resolve_raw_identity(
@@ -939,6 +943,7 @@ fn resolve_owned_artifact(
             package,
             backend: backend_name.to_string(),
             base_url,
+            repository_origin,
             version: args.version.as_deref(),
             warnings,
         },
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs
@@ -28,7 +28,11 @@ use super::COMMAND;
 use super::render::{artifact_ext, artifact_type_wire, repo_config_err};
 use super::types::*;
 
-pub(super) fn index_fetch_error(index_url: &str, err: DownloadError) -> CliError {
+pub(super) fn index_fetch_error(
+    index_url: &str,
+    err: DownloadError,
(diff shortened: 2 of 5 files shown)


kongche-jbw (commented):
Review baseline: `33ed384643c6...3c6216a4df93`

未发现阻塞问题。

Remaining risks:

- The pinned local Rust toolchain had no usable `cargo` binary, so focused tests could not be rerun.
- GitHub currently reports automation and CLA checks only; no build or test check is attached.

Verification scope:

- Reviewed the complete five-file diff, adjacent config discovery, and install/update/repair paths.
- Read all paginated conversation, review, inline-comment, timeline, and review-thread context.
- Direct `rustfmt --check` on changed files and `git diff --check` passed.

ikunkun-sys (approved):

PR #2732, opened 2026-08-20
fix(sight): return savings rates as fractions

ikunkun-sys:
## Why

`/api/token-savings` returned `savings_rate` / `compounded_savings_rate` scaled by 100 (percent values, e.g. `0.269` meaning 0.269%) while the field names imply a ratio. Consumers naturally read `savings_rate: 0.269` as 26.9% — a 100x overstatement. Reported feedback: actual ratio is 6,594 / 2,448,188 = 0.00269, not 0.269.

## What changed

- Backend (`src/server/token_savings.rs`): both `/api/token-savings` and `/api/token-savings/session/{id}` now return true fractions in [0.0, 1.0] for `savings_rate` and `compounded_savings_rate` (session-level, grand summary, session detail). DTO doc comments state the contract.
- Optimization tips: thresholds moved from percent (15.0 / 30.0) to fraction scale (0.15 / 0.30); displayed percentage strings multiply by 100.
- Dashboard: display sites convert fraction → percent (session table bar + label on TokenSavingsPage, saved-label on AtifViewerPage); rendered UI is unchanged. Removed an unused `compoundedSavingsRate` variable; TS interfaces document the [0, 1] range.
- Tests: exact fraction assertions (2200/2700 ≈ 0.8148) added to both handler integration tests; tips unit tests rescaled to fractions.

## Related issue

no-issue: user feedback on API field semantics (no GitHub issue filed)

## User / Agent impact

Dashboard UI renders identically. Direct API consumers of the two rate fields see a value-scale change (percent number → fraction).

## Risk and compatibility

API value-scale change: `savings_rate` / `compounded_savings_rate` on the two token-savings endpoints now return fractions in [0.0, 1.0] instead of percent values. The only in-repo consumer (embedded dashboard) is adapted in the same commit; a monorepo-wide grep confirms no other consumers. Field names and response shapes are unchanged. Renaming to `*_pct` was considered and rejected: it breaks the response shape for the same audience without fixing the name/value mismatch.

## Validation

- `cargo fmt --all -- --check` — pass
- `cargo clippy --all-targets -- -D warnings` — pass (excluding the pre-existing `question_mark` error in `src/analyzer/message/openai.rs`; verified identical on main)
- `cargo test token_savings` — 40/40 pass
- `cargo test --no-fail-fast --workspace --lib --tests` — 27 test binaries, all green (one earlier run hit a transient `/tmp` fstat I/O flake in test setup; re-ran clean twice)
- `dashboard/`: `npm run typecheck`, `npm run test:api-client`, `npm run test:i18n` — pass
- Known pre-existing failures on main, unrelated to this change: clippy `question_mark` in `analyzer/message/openai.rs`, doctest in `genai/helpers.rs` (both fail with the local toolchain)

Environment: Linux x86_64, kernel 5.10.

## Documentation and rollback

No existing docs describe these field semantics (checked `docs/` and the AGENTS.md endpoint table — the endpoint is listed without field details); the new rustdoc and TS interface comments serve as the contract. Rollback: revert the single commit — backend and dashboard changes are coupled in one commit and revert atomically.

### DIFF
--- a/src/agentsight/dashboard/src/pages/AtifViewerPage.tsx
+++ b/src/agentsight/dashboard/src/pages/AtifViewerPage.tsx
@@ -1035,7 +1035,7 @@ export const AtifViewerPage: React.FC = () => {
                     <p className="text-xl font-bold text-green-600">
                       -{fmtTokens(savingsDetail.total_compounded_saved)}
                       <span className="text-sm font-normal text-gray-400 ml-1">
-                        ({savingsDetail.savings_rate.toFixed(1)}%)
+                        ({(savingsDetail.savings_rate * 100).toFixed(1)}%)
                       </span>
                     </p>
                   </div>
--- a/src/agentsight/dashboard/src/pages/TokenSavingsPage.tsx
+++ b/src/agentsight/dashboard/src/pages/TokenSavingsPage.tsx
@@ -428,11 +428,11 @@ const SessionRow: React.FC<{
             <div className="flex-1 h-2 bg-gray-200 rounded-full overflow-hidden max-w-[80px]">
               <div
                 className="h-full bg-green-500 rounded-full"
-                style={{ width: `${Math.min(session.compounded_savings_rate, 100)}%` }}
+                style={{ width: `${Math.min(session.compounded_savings_rate * 100, 100)}%` }}
               />
             </div>
             <span className="text-xs font-semibold text-green-600">
-              {session.compounded_savings_rate.toFixed(1)}%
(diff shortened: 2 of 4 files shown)


chengshuyi (approved):
LGTM. Clean atomic fix: backend removes ×100, frontend adds ×100, tests assert exact fractions. Doc comments establish the [0,1] contract on both Rust DTOs and TS interfaces. No missed consumers. Minor follow-up suggestion: SessionSavingsDetail.savings_rate uses compounded_saved (pre-existing naming inconsistency), could rename later.

PR #2740, opened 2026-08-21
feat(sight): support ~/.agentsight trajectory collection via scan_dirs

chengshuyi:
## Why

`~/.agentsight/skillrubric/traces/` 下存放了 skillrubric 评测产生的轨迹日志（标准 Qoder JSONL 格式），但轨迹采集器默认不扫描该目录，且路径推断不识别 `.agentsight` 前缀。

## What

- **discovery.rs**: `source_from_path` 增加 `.agentsight` 路径识别，入库轨迹 source 标记为 `"agentsight"`
- **discovery.rs**: `scan_dirs` 语义从覆盖改为追加（默认根 + 自定义目录合并去重）
- **lib.rs**: 修复 `test_scan_once_skips_empty_steps_session` 在多用户机器上与真实会话 UUID 碰撞
- **AGENTS.md**: 同步文档描述 scan_dirs 追加语义
- **.gitignore**: 排除 `dev.sh`（本地开发脚本，不入仓库）

## Usage

在 `agentsight.json`（或 `--config` 指定的配置文件）中追加扫描目录：

```json
{
  "features": {
    "trajectory_collection": {
      "enabled": true,
      "scan_dirs": ["/root/.agentsight/skillrubric/traces"]
    }
  }
}
```

## Test

```bash
cargo test -p agentsight-trajectory-collector  # 34 passed
```

E2E 验证：配置 scan_dirs 后 trace collector 自动扫描，新格式轨迹（含 user/assistant 事件）正常入库，Dashboard sessions 页可见。


Daydreamer-Li (commented):
## Review: merge-ready after bumping `schema_version` for the `scan_dirs` semantic change

Overall a clean, well-tested change. `cargo fmt`, `cargo clippy -D warnings`, and the `Test agentsight` CI job all pass. I found **1 blocking issue** (schema contract) and **4 non-blocking suggestions**.

### 1. `scan_dirs` semantics changed but `schema_version` was not bumped — P1 (blocking)

**What is wrong:** This PR changes `scan_dirs` from *overriding* the default scan roots to *appending* to them. That is a semantic change to an existing `agentsight.json` field. The repo's `AGENTS.md` hard rule says a semantic change to `agentsight.json` requires bumping both `schema_version` in `agentsight.json` and `CURRENT_SCHEMA_VERSION` in `src/config.rs`. Both are still `2`.

**Where:** `src/agentsight/agentsight.json:2`, `src/agentsight/src/config.rs:151`

**Impact:** A user who previously set `scan_dirs` to limit scanning to one directory will now also scan `/root` and `/home/*` after upgrade, potentially ingesting unexpected trajectories, with no schema-version signal that the contract changed.

### 2. `config.rs` doc comment still says "override" — P2

**Where:** `src/agentsight/src/config.rs:381`

**Impact:** The comment on `JsonTrajectoryCollectionFeature.scan_dirs` still reads "Optional override of the projects directories to scan", which now contradicts the append behavior.

### 3. Append mode can double-scan overlapping sessions — P2

**What is wrong:** Deduplication is only by directory canonical path. If a configured `scan_dirs` entry overlaps a default root in content but not in path (e.g., the custom dir is a parent of a default root), the same session file can be discovered twice.

**Where:** `src/agentsight/crates/agentsight-trajectory-collector/src/discovery.rs:92-125`

**Impact:** Duplicate trajectories may be stored, showing duplicated sessions in the Dashboard. The old override mode avoided this because default roots were never scanned together with custom dirs.

**Suggestion:** Deduplicate discovered sessions by `session_id` or full file path before returning from `discover_sessions`.

### 4. `AGENTS.md` module map is out of sync — P2/P3

**Where:** `src/agentsight/AGENTS.md:102`

**Impact:** The TrajectoryCollector responsibility description still only mentions Qoder/QoderWork, but the code already supports Claude, Codex, Cursor, and now agentsight.

### 5. Layout assumption for `.agentsight` is undocumented — P2/P3

**What is wrong:** `.agentsight` is not in `SESSION_ROOTS`, so a configured `.agentsight` path falls through to the backward-compatible default in `root_from_path`, which infers `Layout::PerProject` (expects `traces/<project>/<uuid>.jsonl`). The PR does not state whether `~/.agentsight/skillrubric/traces/` actually has project subdirectories.

**Where:** `src/agentsight/crates/agentsight-trajectory-collector/src/discovery.rs:175-204`

**Impact:** If the directory uses a flat layout (`traces/<uuid>.jsonl`), the README configuration will discover nothing.

---

### What looked good

- Append semantics are more useful than the old override behavior.
- Tests cover the new append behavior and the `agentsight` source label.
- The UUID-collision fix for multi-user machines is solid.
- `fmt`/`clippy` pass; CI is green.

### Review coverage

- Reviewed: discovery scanning logic, source labeling, deduplication, config parsing, `schema_version` policy, doc consistency, fmt/clippy, CI status.
- Not verified: real `~/.agentsight/skillrubric/traces` directory layout; actual multi-user scan behavior on a machine with real sessions.

Daydreamer-Li (commented):
## Follow-up review on `4c2e997a`: new Pi support adds 4 blocking correctness issues

The latest revision substantially expands this PR beyond the original `scan_dirs` change: it now adds a complete Pi converter and changes Qoder system-event handling. CI is green, but the current tests do not cover Pi branch history or real Qoder system telemetry. I found **4 blocking issues and 3 non-blocking issues**.

### 1. Qoder telemetry events are converted into empty system-prompt steps — P1 (blocking)

**What is wrong:** `convert_qoder_events()` now treats every event whose top-level `type` is `system` as a system prompt. Qoder also uses `type: "system"` for telemetry events, which do not have `message.content`.

**Where:** `src/agentsight/crates/agentsight-trajectory-collector/src/atif.rs:54-83`

**Evidence:** The repository already contains a real example at `src/agentsight/src/response_map.rs:489`:

```json
{"type":"system","subtype":"ui_telemetry","systemPayload":{"uiEvent":{"event.name":"api_response"}}}
```

**Impact:** Every such telemetry event becomes an empty `StepSource::System` step, inflating step counts and polluting trajectories. A file containing only telemetry can also pass the non-empty-step check and be stored as a valid trajectory.

### 2. Pi branch history is flattened into one incorrect conversation — P1 (blocking)

**What is wrong:** Pi v2/v3 sessions are trees linked by `id` and `parentId`; the active conversation is the path from the current leaf back to the root. The converter instead processes every message in raw JSONL file order and never reads `parentId`.

**Where:** `src/agentsight/crates/agentsight-trajectory-collector/src/pi.rs:75-314`

**Impact:** After a user branches from an earlier message, both the abandoned branch and active branch are emitted into one ATIF trajectory. Message order, tool-result association, step count, and token totals become incorrect.

**Reference:** Pi's session format documents tree-based history and says context is built by walking from the current leaf to the root: https://pi.dev/docs/latest/session-format

### 3. Pi cache-write tokens are omitted from prompt totals — P1 (blocking)

**What is wrong:** Pi usage has `input`, `cacheRead`, and `cacheWrite`, but `extract_usage()` computes prompt tokens as only `input + cacheRead`. `cacheWrite` is never read.

**Where:** `src/agentsight/crates/agentsight-trajectory-collector/src/pi.rs:449-475`

**Impact:** Any request that creates prompt-cache entries reports fewer prompt tokens than Pi's own `totalTokens`, so per-step and final session token totals are understated. This makes Dashboard token/cost analysis inaccurate.

### 4. `scan_dirs` semantics changed without a config schema bump — P1 (blocking, still open)

**What is wrong:** `scan_dirs` changed from replacing defaults to appending defaults, but `agentsight.json.schema_version` and `CURRENT_SCHEMA_VERSION` remain `2`.

**Where:** `src/agentsight/agentsight.json:2`, `src/agentsight/src/config.rs:151`

**Impact:** A user who configured `scan_dirs` to limit collection to selected paths will start scanning `/root` and `/home/*` after upgrade.

The feature being disabled by default does not remove the compatibility issue for users who explicitly enabled it. The field was already public and documented with override semantics, and `AGENTS.md` explicitly requires a schema bump for semantic changes without an experimental-feature exception.

### 5. Pi step-level model information is always discarded — P2

**What is wrong:** `step_model` is hard-coded to `None`, although Pi assistant messages contain their own `model` field.

**Where:** `src/agentsight/crates/agentsight-trajectory-collector/src/pi.rs:133-140`

**Impact:** Sessions that switch models cannot identify which model produced each assistant step. The top-level agent model also records only the first `model_change` event.

### 6. No end-to-end collector test covers the new Pi path — P2

The new tests call `convert_pi_events()` directly. They do not cover `discovery → source/format detection → converter → store`, and do not include:

- a branched Pi v3 session;
- non-zero `cacheWrite`;
- a real `~/.pi/agent/sessions/...` path;
- a Qoder `system/ui_telemetry` event.

These missing cases are why all CI checks pass despite the correctness issues above.

### 7. The PR now exceeds the reviewability and scope limits — P2

The title/body still describe `.agentsight` `scan_dirs`, but the second commit adds a complete Pi converter. The current diff is **754 additions + 409 deletions across 9 files** (>800 changed lines), while `pi.rs` alone is 575 lines (above the module target of 500 lines). Deleting the unrelated 356-line `dev.sh` further mixes concerns.

Please split this into independently reviewable changes: `scan_dirs`, Pi support, and development-script cleanup.

---

### Verified clean areas

- Basic Pi user/assistant/thinking/toolCall/toolResult mapping is coherent.
- Tool errors preserve `is_error`, and tool-result timestamps are associated with matching calls.
- The multi-user test UUID collision fix is reasonable.
- Current CI, formatting, and Clippy checks pass.
- The earlier layout question is resolved: the author confirmed skillrubric uses `traces/<project>/<uuid>.jsonl`.

PR #2745, opened 2026-08-21
fix(sec-core): correct warmup claim and hook noise

haosanzi:
warmup only asks Ollama whether it can serve the model; it never loads the weights, so the first scan still pays the cold-start cost. The Rust doc comments, the CLI help/output strings and SKILL.md all promised the opposite. Reword them to describe an availability check and point users at `ollama pull` for the model that is never fetched automatically.

The three prompt_scanner_hook.py copies also printed the resolved scan mode to stderr on every import, which surfaces on each prompt in a correctly configured setup. Replace it with _warn_invalid_scan_mode(), matching the pii-checker and skill-ledger hooks: only report when PROMPT_SCANNER_SCAN_MODE was set to a value that silently fell back.

Trade-off: the hook logic is duplicated across the codex, cosh and qwen-code copies because they have no shared module.

Assisted-by: Qoder:1.22.1

image

受影响的只有cosh，qwen, codex，所以其他的code agent hook不需要修改，只修改这三个即可

## Risk and compatibility

### DIFF
--- a/src/agent-sec-core/codex-plugin/hooks-plugin/hooks/prompt_scanner_hook.py
+++ b/src/agent-sec-core/codex-plugin/hooks-plugin/hooks/prompt_scanner_hook.py
@@ -45,15 +45,27 @@ _RAW_SCAN_MODE = os.environ.get("PROMPT_SCANNER_SCAN_MODE", "standard").strip().
 _DEFAULT_SCAN_MODE = _RAW_SCAN_MODE
 if _DEFAULT_SCAN_MODE not in {"fast", "standard", "strict"}:
     _DEFAULT_SCAN_MODE = "standard"
-print(
-    f"[prompt-scanner] scan mode configured: raw={_RAW_SCAN_MODE!r}, "
-    f"effective={_DEFAULT_SCAN_MODE!r}",
-    file=sys.stderr,
-)
 _DEFAULT_SOURCE = "user_input"
 
 
 # -- output helpers --------------------------------------------------------
+
+
+def _warn_invalid_scan_mode() -> None:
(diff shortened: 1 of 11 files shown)


src/agent-sec-core/codex-plugin/hooks-plugin/hooks/prompt_scanner_hook.py
```diff
@@ -45,17 +45,29 @@
...
-    f"effective={_DEFAULT_SCAN_MODE!r}",
-    file=sys.stderr,
-)
 _DEFAULT_SOURCE = "user_input"
 
 
 # -- output helpers --------------------------------------------------------
 
 
+def _warn_invalid_scan_mode() -> None:
```
yangdao479, line 54: 新增的 `_warn_invalid_scan_mode()` 在 codex / cosh / qwen-code 三份拷贝里各存一份，但 `tests/` 下没有任何用例把 `PROMPT_SCANNER_SCAN_MODE` 设成非法值去断言这条告警，也没有用例断言配置正确时保持静默。由于 `[tool.coverage.run] source = ["agent_sec_cli"]` 不含 hook 目录，这部分不会被增量覆盖率门禁拦下。

建议给三份拷贝各补一条「非法值 → 告警文案」+「合法值 → stderr 为空」的用例，否则这次确立的「只报错配置、正常时静默」契约没有回归保护，任一拷贝日后退化回每次打印都不会被测试发现。

yangdao479 (changes requested):
PTAL

yangdao479 (approved):
LGTM，THX！

Issue #2725, opened 2026-08-20
feat(cosh-ng): add governed checkpoint capability


kongche-jbw:
## Component

cosh-ng

## What problem are you trying to solve?

PR #2603 established a durable, local, task-only Gateway, but its production
inventory only exposes `ask_user_question`. Agents still cannot request a
workspace checkpoint through the governed Task Plane, and the old capability
profile experiment is based on a stale pre-merge tree.

The next implementation must add optional capability admission without making
`task-only-v1` depend on ws-ckpt or widening authority from Task input.

## What would a useful outcome look like?

A user can submit a persistent Agent Task that requests a workspace checkpoint.
The Task pauses before the effect, can be approved or rejected from another
terminal, consumes a one-operation Permit, executes through an optional
per-instance ws-ckpt target, and resumes after a conclusive result. Gateway or
target restarts query durable operation state and never blindly replay an
uncertain side effect.

Instances using `task-only-v1` continue to work without ws-ckpt.

## Proposed delivery slices

1. Capability profile admission
   - canonical profile identity and manifest digest;
   - durable profile binding and Runtime inventory verification;
   - sealed provider registry;
   - production remains `task-only-v1` only.
2. Optional ws-ckpt execution target
   - per-instance opt-in and exact target readiness;
   - typed checkpoint operation and result reconciliation;
   - no global package or service dependency.
3. Governed checkpoint creation
   - durable Approval and single-use Permit;
   - pre-effect audit, execute, query, and explicit uncertain outcome;
   - cross-terminal approval and restart/kill-point coverage.

Shell attachment, slash commands, file writes, Shell execution, Web, and remote
channels remain separate follow-up work.

## Acceptance criteria

- [ ] Task input and Runtime output cannot select or widen the admitted profile.
- [ ] Profile identity and manifest digest survive restart and mismatch fails closed.
- [ ] The registered provider set exactly matches the admitted manifest.
- [ ] `task-only-v1` has no ws-ckpt runtime, packaging, or systemd dependency.
- [ ] A Permit is bound to one operation, workspace, target, and profile digest.
- [ ] Execution is audited before the effect and the Permit is consumed once.
- [ ] Recovery queries the target; unknown results remain explicit and are not replayed.
- [ ] Concurrent approval decisions commit exactly one terminal decision.
- [ ] Exact-candidate targeted tests record negative, race, and restart evidence.
- [ ] Public acceptance reports and internal handoff/tracking records stay current.
[...]

PR #2768, opened 2026-08-22
feat(cosh-ng): add checkpoint profile contract

kongche-jbw:
## Why

Slice B needs a closed checkpoint capability contract without granting authority that the current ws-ckpt protocol and Gateway dependency graph cannot safely enforce. This change makes the contract and transport semantics reviewable while keeping production checkpoint execution fail-closed.

## What changed

- Add the closed `workspace-checkpoint-v1` profile with an exact ordered tool inventory, exact `ws-ckpt` provider inventory, and a pinned manifest digest.
- Add a sealed Gateway provider registry. Task-only instances admit the empty provider set, while every request for checkpoint provider authority is explicitly withheld.
- Add peer-authenticated checkpoint transport primitives that classify failures as `KnownNoEffect` or `PossiblyApplied` and support a read-only snapshot evidence lookup.
- Preserve the `task-only-v1` manifest byte-for-byte and avoid adding Gateway dependencies on `cosh-platform` or `cosh-types`.
- Update the English and Chinese capability-broker design and acceptance documents with the delivered boundary and deferred prerequisites.

## Related issue

Related to #2725. This is Slice B and does not close the governed checkpoint delivery tracked by that issue.

## User / Agent impact

The current code supports these scenarios:

- A trusted caller can select `workspace-checkpoint-v1` and reject any drift in profile name, manifest digest, ordered tool inventory, or provider set.
- A production Gateway can continue to admit the portable task-only profile without a ws-ckpt socket or daemon.
- Any attempt to grant ws-ckpt side-effect authority fails before an execution target can be obtained.
- A lower-level platform caller can authenticate the Unix peer, classify a checkpoint-create failure by possible side effect, and perform one read-only evidence lookup instead of replaying an uncertain create.

The current code does not expose a Gateway checkpoint execution target, advertise the tool from Runtime, add a `serve` checkpoint-socket option, persist target receipts, or execute an end-to-end governed checkpoint Task. The PR remains Draft until the ws-ckpt protocol gains identity-only requests and an atomic generation witness, and checkpoint transport ownership is resolved without violating the Gateway dependency rule.

## Risk and compatibility

The public Rust contract and transport APIs expand, but existing task-only manifest bytes and legacy optional-authentication transport calls remain unchanged. Governed calls map daemon error codes to bounded local messages and never expose daemon-controlled text. Checkpoint provider authority is withheld, so this change does not enable new host side effects through Gateway.

## Validation

- `cargo test --locked -p cosh-gateway-contracts profile::` — 10 passed
- `cargo test --locked -p cosh-gateway capability::provider::` — 6 passed
- `cargo test --locked -p cosh-platform checkpoint::` — 34 passed
- `cargo clippy --locked -p cosh-platform --all-targets -- -D warnings`
- `cargo clippy --locked -p cosh-platform --tests --target aarch64-apple-darwin -- -D warnings`
- `scripts/docs-lint.sh`
- `scripts/docs-link-check.py`
- `git diff --check`

The three targeted suites also ran on a persistent x86_64 Linux ECS before review. That run exposed a root-only false-negative fixture: a root fake daemon is trusted by design. The corrected UID 65534 subprocess fixture passed the real negative `SO_PEERCRED` case as root. The review revision also cross-compiled the cosh-platform test target for `aarch64-apple-darwin` with warnings denied.

Not run on the final candidate: full workspace gates, release build, `run-test-gates.sh all`, a real ws-ckpt daemon, Btrfs behavior, or manual Terminal validation.

## Documentation and rollback

Updated the bilingual capability-broker design and acceptance documents. Revert commit `66912084` to remove Slice B; no runtime checkpoint configuration needs disabling because provider authority is not admitted.

### DIFF
--- a/src/cosh-ng/crates/cosh-platform/src/checkpoint.rs
+++ b/src/cosh-ng/crates/cosh-platform/src/checkpoint.rs
@@ -2,7 +2,7 @@
 //!
 //! Wire format: [4-byte LE length prefix][bincode payload]
 
-use std::io::{Read, Write};
+use std::io::{ErrorKind, Read, Write};
 use std::os::unix::net::UnixStream;
 use std::path::Path;
 use std::time::Duration;
@@ -36,10 +36,59 @@ enum ProtocolFailure {
     UnexpectedResponse,
 }
 
+/// Whether a failed ws-ckpt request may already have changed daemon state.
+///
+/// A governed caller must never retry a `PossiblyApplied` request. It has to
+/// reconcile against durable daemon evidence or record an uncertain outcome.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CkptRequestEffect {
+    /// The daemon provably did not apply the request.
+    KnownNoEffect,
+    /// The request may have been applied, but its result cannot be proven.
+    PossiblyApplied,
+}
+
+/// Failed ws-ckpt request together with its side-effect classification.
+#[derive(Debug)]
+pub struct CkptRequestFailure {
+    /// Whether daemon state may already have changed.
+    pub effect: CkptRequestEffect,
+    /// Redacted failure suitable for audit and presentation.
+    pub error: CoshError,
+}
+
+impl CkptRequestFailure {
+    fn known_no_effect(error: CoshError) -> Self {
+        Self {
+            effect: CkptRequestEffect::KnownNoEffect,
+            error,
+        }
+    }
+
+    fn possibly_applied(error: CoshError) -> Self {
+        Self {
+            effect: CkptRequestEffect::PossiblyApplied,
+            error,
+        }
+    }
+}
+
+/// Exact durable evidence for one snapshot identity in one workspace.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CkptSnapshotEvidence {
+    /// Snapshot identity reported by the daemon index.
+    pub snapshot_id: String,
+    /// Workspace registration path reported verbatim by the daemon.
+    pub workspace: String,
+    /// Whether the daemon index knows the snapshot data is no longer present.
+    pub missing: bool,
+}
+
 /// Client for ws-ckpt daemon IPC.
 pub struct CkptClient {
     socket_path: String,
     timeout_ms: u64,
+    trusted_peer_uid: Option<u32>,
 }
 
 impl CkptClient {
@@ -48,6 +97,7 @@ impl CkptClient {
         Self {
             socket_path: socket_path.to_string(),
             timeout_ms: DEFAULT_TIMEOUT_MS,
+            trusted_peer_uid: None,
         }
     }
 
@@ -56,12 +106,28 @@ impl CkptClient {
         Self {
             socket_path: socket_path.to_string(),
             timeout_ms,
+            trusted_peer_uid: None,
         }
     }
 
     /// Create a new client using the default socket path.
     pub fn default_path() -> Self {
         Self::new(DEFAULT_SOCKET_PATH)
+    }
+
+    /// Require every connection to authenticate as root or `owner_uid`.
+    ///
+    /// Path metadata alone cannot secure a socket: another principal may swap a
+    /// verified directory between the check and `connect`. Kernel peer
+    /// credentials are read after connecting and before any request byte is
+    /// written, so an impostor listener is rejected with no possible effect.
+    /// [`Self::create_classified`] and [`Self::find_snapshot`] reject the client
+    /// before socket access unless this configuration is present; legacy
+    /// operations retain their existing optional-authentication behavior.
+    #[must_use]
+    pub fn require_trusted_peer(mut self, owner_uid: u32) -> Self {
+        self.trusted_peer_uid = Some(owner_uid);
+        self
     }
 
     /// Check if the daemon socket exists (basic health check).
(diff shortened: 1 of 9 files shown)


src/cosh-ng/crates/cosh-platform/src/checkpoint.rs
```diff
@@ -131,6 +197,104 @@ impl CkptClient {
...
+                skipped: true,
+                reason: Some(reason),
+            }),
+            // A daemon error proves no snapshot was created, but not that daemon
+            // state is unchanged. The checkpoint dispatch path auto-initializes
+            // the workspace first, so registration, subvolume adoption, or a
+            // broken-symlink removal may already have happened before the code
+            // was produced. No response code is treated as proven no-effect
+            // until the protocol supplies an explicit pre-effect guarantee.
+            WsCkptResponse::Error { code, message } => Err(CkptRequestFailure::possibly_applied(
```
ikunkun-sys: **[P1] Redact daemon text before exposing it as audit-safe**\n\nThe new public failure contract says `error` is redacted and safe for audit/presentation, but this branch passes the daemon message verbatim into `CoshError`. The real ws-ckpt daemon interpolates request-derived workspace/snapshot values and internal error chains, so a caller following this contract can persist or display sensitive paths, control text, or oversized daemon output. The classified failure therefore does not satisfy its audit/presentation contract, and the current tests only use the literal `rejected`, so they do not detect this leak.

src/cosh-ng/crates/cosh-platform/src/checkpoint.rs
```diff
@@ -443,11 +712,20 @@ fn classify_io_error(error: std::io::Error, phase: IoPhase) -> CoshError {
...
+    use std::os::unix::fs::PermissionsExt;
     use std::os::unix::net::UnixListener;
+    use std::os::unix::process::CommandExt;
+    use std::process::{Command, Stdio};
     use std::thread;
+    use std::time::Duration;
 
     use super::*;
 
+    fn trusted_client(socket_path: &str) -> CkptClient {
```
ikunkun-sys, line 737: **[P2] Keep trusted-peer tests valid on macOS**\n\nThis helper feeds every success-path classified/evidence test, but the non-Linux `verify_peer_credentials` implementation always returns `PermissionDenied`. As a result, the added `a_trusted_peer_is_accepted`, daemon-error, and `find_snapshot` tests deterministically fail on macOS even though cosh-ng retains limited macOS support; the current CI is Linux-only and cannot catch it. The expectations need to match the fail-closed non-Linux contract (or peer authentication must be implemented there) so the supported macOS workspace test gate remains runnable.

ikunkun-sys (changes requested):
Requesting changes on current head 4cde26aa. The closed profile/provider behavior and Linux checkpoint classification suites pass, but the new classified failure exposes unredacted daemon text despite an audit-safe contract, and the unguarded trusted-peer success tests make the macOS workspace test gate fail deterministically.

src/cosh-ng/crates/cosh-platform/src/checkpoint.rs
kongche-jbw: Fixed in `66912084`. Trusted-peer exchange tests are Linux-only, while non-Linux now has a dedicated test for the fail-closed `PermissionDenied` contract before any request byte is written. The `aarch64-apple-darwin` test target compiles under clippy with warnings denied.

src/cosh-ng/crates/cosh-platform/src/checkpoint.rs
kongche-jbw: Fixed in `66912084`. Governed create and evidence lookup now discard daemon-controlled text and map only the daemon error code to bounded local messages and hints. The regression uses a path, control sequence, and oversized payload and asserts the public error remains the fixed local message.

src/cosh-ng/crates/cosh-platform/src/checkpoint.rs
```diff
@@ -248,62 +412,167 @@ impl CkptClient {
...
+    } else {
+        Err(peer_error())
+    }
+}
+
+#[cfg(not(target_os = "linux"))]
+fn verify_peer_credentials(_stream: &UnixStream, _owner_uid: u32) -> Result<(), CoshError> {
+    // Without kernel peer credentials a governed caller cannot authenticate the
+    // daemon, so requiring a trusted peer fails closed instead of degrading.
+    Err(peer_error())
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in `66912084`. Listener-backed governed success and classification cases are scoped to Linux, and a non-Linux regression verifies both operations fail closed without writing a request. The `aarch64-apple-darwin` test target now compiles cleanly.

src/cosh-ng/crates/cosh-platform/src/checkpoint.rs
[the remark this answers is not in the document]
kongche-jbw: Fixed in `66912084`. Linux retains the real trusted-peer success and failure coverage; non-Linux tests the intended fail-closed result for both classified create and evidence lookup. The Apple test target also passes cross-target clippy with warnings denied.

ikunkun-sys (approved):
Approved on current head 66912084. The revision discards daemon-controlled text before exposing governed errors and separates Linux peer-credential success coverage from the non-Linux fail-closed contract. Targeted checkpoint/profile/provider tests, Rust 1.97.1 formatting and all-targets Clippy, a clean synthetic merge against the latest fetched main, and remote cosh-ng CI passed. The Apple cross-target path was source-reviewed but not locally executed on this host.

PR #2832, opened 2026-08-25
feat(cosh-ng): [shell] bound enhanced hooks

### LINKED ISSUE #2687: [P1][cosh-ng] 强制 functrace/errtrace/extdebug 全局暴露：用户 trap 对 cosh 内部触发、$-/set -o 谎报、泄漏内部实现与 token
## 背景

cosh-shell 的交互注入依赖 DEBUG trap 的 return-1 拦截语义，为此在 marker（`crates/cosh-shell/src/shell_host/marker/bash.rs`）里**强制开启** `shopt -s extdebug` 以及 `set -T`（functrace）/ `set -E`（errtrace）。

bash(1) §FUNCTIONS 明确：functrace 会让**所有函数继承 DEBUG/RETURN trap**，errtrace 会让 ERR trap 被继承。因此这三个被强制的选项不是内部实现细节——它们**全局可观测**，并把 cosh 的内部函数暴露给任何用户设置的 trap。这是一个根因、多个观测面的结构性一致性问题。

## 问题

用户进入 cosh 交互会话后，以下观测量全部偏离裸 bash 交互会话：

### 1. 强制选项被用户直接观测到（状态谎报）

| 观测 | 裸 bash -i | cosh |
|---|---|---|
| `$-` | `himBHs` | `himBEHTs`（多出 `E`/`T`） |
| `set -o functrace` | off | **on** |
| `set -o errtrace` | off | **on** |
| `shopt extdebug` | off | **on** |

依赖 `case $- in *T*)`、`shopt -q extdebug` 之类探测的脚本会在 cosh 下误判。

### 2. `trap -p` 泄漏 cosh 内部 trap

`trap -p DEBUG` 输出 `trap -- '_cosh_preexec_marker' DEBUG`（裸 bash 为空）。

### 3. 用户 DEBUG / RETURN / ERR trap 对 cosh 内部触发

因 functrace/errtrace 被强制，用户设置的 trap 被 cosh 的内部函数继承并逐一触发：

- 用户 `DEBUG` trap：对 `_cosh_prompt_command`、`local status=$?`、含 OSC 协议与 **session token** 的内部 `printf` 等逐条触发（裸 bash 只对用户自己的命令触发一次）。
- 用户 `RETURN` trap：对 cosh 内部函数返回触发 **23 次**（裸 bash 不触发）。

危害：① 击穿所有基于 DEBUG/RETURN/ERR trap 的用户工具链（bash-preexec、命令计时、审计钩子）——它们对 cosh 内部触发并读到内部状态；② 通过 `$BASH_COMMAND` / `trap -p` 泄漏 cosh 内部实现与会话 token；③ **不需要 `set -x`**，一个普通 trap 或一次选项查询即触发。

### 实测（当前 main，真实 PTY + 真实 marker 注入，脱敏 `<...>`）

```
# G2 选项状态
DASH=[himBEHTs]          # 裸 bash: himBHs
FUNCTRACE=[on]           # 裸 bash: off
ERRTRACE=[on]            # 裸 bash: off
EXTDEBUG=[on]            # 裸 bash: off

# G3 trap -p 泄漏
$ trap -p DEBUG
trap -- '_cosh_preexec_marker' DEBUG      # 裸 bash: (空)

# 用户 DEBUG trap（节选，裸 bash 仅 1 次: echo user-visible-cmd）
BC=[_cosh_prompt_command]
BC=[printf '\033]1337;COSH;{"event":"history_file","token":"<SESSION_TOKEN>",...}']
...

# 用户 RETURN trap: 23 次对 cosh 内部函数触发（裸 bash: 0 次）
RET=[_cosh_emit_native_history_file_marker]
RET=[_cosh_precmd_marker]
...
```

裸 bash oracle：用户 DEBUG trap 只输出 `BC=[echo user-visible-cmd]` 与 `BC=[exit]`；RETURN trap 不对上述函数触发。

## 手册依据（bash(1)）

- §FUNCTIONS：
  > the DEBUG and RETURN traps … are not inherited unless the function has been given the trace attribute … or the **-o functrace** shell option has been enabled … (in which case **all functions inherit the DEBUG and RETURN traps**), and the **ERR trap is not inherited unless the -o errtrace** shell option has been enabled.
- §Shell Variables — `BASH_COMMAND`：
  > The command currently being executed or about to be executed … unless the shell is executing a command as the result of a trap, in which case it is the command executing at the time of the trap.
- §SHELL BUILTIN COMMANDS — `set`：`functrace` = `-T`，`errtrace` = `-E`；`extdebug` 由 shopt 控制。

即：强制 `-T`/`-E`/extdebug 的直接后果就是「用户 trap 继承进所有（含 cosh 内部）函数」与「选项状态可见」，属规范定义的行为，而非偶发。

## 疑似根因

marker 为实现 DEBUG-trap 拦截强制 `shopt -s extdebug` 与 `set -T -E`，且未对这些选项及内部 trap 做用户视角的隔离/遮蔽。选项本身是拦截机制所需，不能简单关闭，需要设计层方案。

## 验收标准

在 cosh 交互会话中，用户可观测的选项状态与 trap 行为与裸 bash 等价：`$-` 不含用户未设置的 `E`/`T`；`set -o functrace/errtrace`、`shopt extdebug` 反映用户实际设置；`trap -p DEBUG` 不泄漏 cosh 内部 trap；用户 DEBUG/RETURN/ERR trap 只对用户自己的命令触发、不暴露 cosh 内部命令与 session token。

## 接力指引

修复面在 marker 的选项/ trap 注入策略（作用域隔离、查询遮蔽，或在无法隔离时显式声明为已知限制）。与 #2632（重构 DEBUG trap 层）强相关，**需在 #2632 分支上复验本 issue 的各观测面是否仍复现**，避免重复修复。

补充：`set -x` 下同类内部泄漏见 #2683（PS4 轴），`$_` 污染见 #2684；本 issue 覆盖 functrace/errtrace/extdebug 的选项暴露与 trap 继承轴。

kongche-jbw:
## Why

Enhanced mode still depended on a global Bash `DEBUG` trap, forced tracing
options, and a session token visible to shell internals. That made native Bash
trap and option behavior impossible to preserve and left #2687 unresolved.

## What changed

- Replace global tracing with bounded Bash integration points: `PS0` for
  pre-execution observation, `command_not_found_handle` for safe unresolved
  natural-language routing, and a final `PROMPT_COMMAND` entry for completion
  boundaries independent of mutable `PS1`.
- Extract compact OSC marker decoding, normalization, frame-boundary handling,
  and history-file tracking into `osc/marker_sequence.rs`. `osc.rs` is 987
  lines at this head versus 995 on main, and the large-file inventory records
  the exact baseline and remaining split priority.
- Route slash controls in the Rust input relay and execute approved handoffs
  from an owner-only sidecar at top-level shell scope, so stateful commands keep
  normal interactive Bash semantics.
- When Bash history cannot identify a submitted command, emit a redacted
  untracked boundary and keep the command Shell-owned instead of reusing stale
  text or routing it to Agent.
- Re-scan reconstructed input for secrets. Cursor-edited or otherwise
  untrackable prompt lines fail closed by staying out of Bash history, with a
  parent-shell prompt scrub as defense in depth.
- Preserve user traps, options, prompt hooks, command history, pager policy,
  provenance, and marker privacy. Bash 4.3–5.0 receives a scalar
  `PROMPT_COMMAND`; Bash 5.1+ keeps native array entries. Prompt and handoff
  wrappers restore temporary user variables before user code runs.
- Do not count empty or whitespace-only lines as pending Shell submissions, so
  a batched Agent control is never queued behind a command-less prompt.
- Assistance-state checks suppress xtrace so their private path is not printed.
- Normalize Readline bracketed-paste mode controls and test both enabled and
  disabled terminal configurations.
- Stabilize foreground-input tests by sending approval and interactive input
  only after the visible owner is ready, instead of relying on fixed delays.

## Related issue

closes #2687

## User / Agent impact

Enhanced Assisted remains the default and retains safe pre-execution
natural-language routing plus post-execution insight. Shift+Tab still switches
Enhanced between Assisted and Shell-only without restarting Bash; startup-only
Native remains hook-free. Aliases, functions, builtins, executables, missing
paths, history-recalled slash input, invalid UTF-8, and history-untracked input
remain Shell-owned.

## Risk and compatibility

The Bash observation boundary changes internally, but the existing
Native/Enhanced configuration and default behavior are preserved. Strict
oracle tests verify that Enhanced no longer enables `extdebug`, `functrace`,
or `errtrace`, installs no Cosh `DEBUG` trap, and exposes no marker token
through user `DEBUG`, `RETURN`, or `ERR` traps.

Enhanced is a bounded integration, not a zero-injection shell. Its
`PS0`/`PROMPT_COMMAND` wiring, `_cosh` function namespace,
`command_not_found_handle`, and `COSH*` integration state remain observable by
the user shell. Native is the hook-free mode for strict bare-Bash behavior.
The separate #2831 prototype remains useful for exploring exact zero-injection
parity, but is not required to close #2687 under this explicitly bounded
contract.

## Validation

- `crates/cosh-shell/scripts/check-layout.sh` (`osc.rs`: 987 lines;
  `marker/bash.rs`: 952 lines)
- `cargo clippy -p cosh-shell --lib -- -D warnings`
- `cargo test -p cosh-shell --lib` (1363 passed)
- `cargo test --locked -p cosh-shell --test raw_cli -- --test-threads=16`
  (465 passed, 1 ignored)
- `cargo test --locked -p cosh-shell --test shell_host -- --test-threads=1`
  (175 passed, 1 ignored)
- Targeted regressions cover Bash trap/option parity, prompt and handoff variable
  preservation, and empty-line pending-boundary classification. Bash 4.4
  container validation was attempted but Docker Hub was unavailable; the
  version-gated scalar path is pinned by marker structure tests.

## Documentation and rollback

The large-file inventory now records the exact main baseline, current line
counts for `osc.rs` and `marker/bash.rs`, extracted owners, and remaining split
priority. No user-guide changes are required because this preserves the mode contract introduced by #2759 and
replaces its Enhanced internals. Roll back operationally by launching with
`shell.integration = "native"` or `COSH_SHELL_INTEGRATION=native`; revert this
PR to restore the former Enhanced implementation.


src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.rs
```diff
@@ -657,176 +783,56 @@ _cosh_compact_alias_expanded() {
...
-    return 0
-  fi
-  # Skip during completion — with extdebug the DEBUG trap fires for every
-  # internal command bash runs during glob expansion / completion, and the
-  # heavy operations below (date subprocess, file I/O) cause noticeable lag.
-  # Require COMP_TYPE (only set by bash during programmable completion) in
-  # addition to COMP_LINE/COMP_POINT so that residual COMP_LINE values do
-  # not permanently silence preexec markers for real commands.
-  if [[ -n "${COMP_TYPE:-}" && ( -n "${COMP_LINE:-}" || -n "${COMP_POINT:-}" ) ]]; then
+_cosh_bounded_preexec_marker() {
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 114be4cd. PS0 now requires a new reliable history generation; disabled, ignored, cleared, or otherwise unavailable history emits `<redacted untracked command>` and remains Shell-owned instead of reusing stale text or routing to Agent.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
```diff
@@ -359,15 +360,40 @@ fn relay_native_passthrough(
         relay.line_submits,
         relay.input_events,
         relay.main_prompt_gate,
-        bytes,
+        private_history_bytes.as_deref().unwrap_or(bytes),
     )?;
     Ok(false)
 }
 
+fn history_private_submission(state: &NativeLineState, bytes: &[u8]) -> Option<Vec<u8>> {
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 114be4cd. Clean fragments are recombined and re-scanned as one line; dirty cursor-edited prompt input now fails closed by inserting the Readline `ignorespace` prefix before submission, and the parent prompt hook also scrubs sensitive history as defense in depth.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.rs
```diff
@@ -844,113 +850,27 @@ _cosh_precmd_marker() {
...
-    local _cosh_prompt_command
-    for _cosh_prompt_command in "${_COSH_USER_PROMPT_COMMAND[@]}"; do
-      _cosh_eval_user_prompt_hook "$_cosh_prompt_command"
-    done
-  elif [[ -n "${_COSH_USER_PROMPT_COMMAND:-}" ]]; then
-    _cosh_eval_user_prompt_hook "$_COSH_USER_PROMPT_COMMAND"
+  _COSH_USER_PS0="${PS0-}"
+  PS0='$(trap - DEBUG RETURN ERR 2>/dev/null; _cosh_bounded_preexec_marker > /dev/tty)'"$_COSH_USER_PS0"
+  _COSH_USER_PS1="${PS1-}"
+  PS1='$(s=$?; trap - DEBUG RETURN ERR 2>/dev/null; _cosh_bounded_prompt_marker "$s" > /dev/tty)'"$_COSH_USER_PS1"
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 114be4cd. The completion boundary is now the final `PROMPT_COMMAND` array entry after native user hooks, independent of mutable `PS1`; the regression resets `PS1` on every prompt and verifies consecutive command boundaries.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
```diff
@@ -359,15 +360,40 @@ fn relay_native_passthrough(
...
+    let submit = bytes
+        .iter()
+        .position(|byte| matches!(byte, b'\n' | b'\r'))?;
+    let prior = state.clean_visible_line()?;
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 114be4cd. Dirty prompt submissions now receive a leading-space Readline insertion before Enter, and a parent-shell scrub provides defense in depth. The regression edits a credential-bearing line with cursor movement and verifies both history and journals remain clean.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.rs
```diff
@@ -445,6 +523,43 @@ _cosh_replace_handoff_history() {
...
+  if _cosh_command_has_secret "$command"; then
+    display_command="<redacted sensitive command>"
+  else
+    builtin history -s "$command" 2>/dev/null || true
+  fi
+  _COSH_ATTEMPT_GENERATION="${history_no:-0}"
+  _cosh_emit_marker "preexec" "$display_command" 0 false
+  _cosh_clear_handoff_request
+
+  _cosh_eval_staged_handoff "$command"
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 114be4cd. Bounded preparation still runs in the helper, but the approved owner-only sidecar is now evaluated by the injected top-level command. The regression verifies both `declare -x` and `set --` state survive the handoff.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.rs
```diff
@@ -844,113 +899,43 @@ _cosh_precmd_marker() {
...
-    for _cosh_prompt_command in "${_COSH_USER_PROMPT_COMMAND[@]}"; do
-      _cosh_eval_user_prompt_hook "$_cosh_prompt_command"
-    done
-  elif [[ -n "${_COSH_USER_PROMPT_COMMAND:-}" ]]; then
-    _cosh_eval_user_prompt_hook "$_COSH_USER_PROMPT_COMMAND"
+  # Keep user prompt hooks as native top-level PROMPT_COMMAND entries. The
+  # final entry emits the Cosh boundary after those hooks, so a hook may
+  # replace PS1 without removing observation. Generic scratch names keep a
+  # user DEBUG trap from seeing Cosh internals before that trap is suspended.
+  PROMPT_COMMAND=('__prompt_status=$?')
```
ikunkun-sys: [P1] PROMPT_COMMAND arrays break the Bash 4.3–5.0 baseline

This unconditionally converts `PROMPT_COMMAND` into an array, but the repository documents Bash 4.3+ support and this test suite itself notes that array `PROMPT_COMMAND` behavior starts at Bash 5.1. On Bash 4.4 only element 0 executes, so the user hooks and final `_cosh_bounded_prompt_marker` entry never run. I reproduced that behavior on the current head: the second array entry is skipped after every command. Enhanced therefore never emits its ready/completion boundary on a documented supported runtime, even though newer-Bash CI can pass.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.rs
```diff
@@ -844,113 +899,43 @@ _cosh_precmd_marker() {
...
+  PROMPT_COMMAND+=("${_COSH_USER_PROMPT_COMMAND[@]}")
+  PROMPT_COMMAND+=('d="$(trap -p DEBUG 2>/dev/null)"; r="$(trap -p RETURN 2>/dev/null)"; e="$(trap -p ERR 2>/dev/null)"; trap - DEBUG RETURN ERR 2>/dev/null; _cosh_bounded_prompt_marker "$__prompt_status" > /dev/tty; eval "$d"; eval "$r"; eval "$e"; unset __prompt_status d r e')
```
ikunkun-sys: [P1] Prompt hooks delete user shell variables

`PROMPT_COMMAND` entries execute in the interactive shell's top-level scope, so these assignments overwrite existing `d`, `r`, `e`, and `__prompt_status` variables and the final `unset` removes them permanently. Running this sequence with pre-existing values leaves all four variables undefined after one prompt cycle. `BOUNDED_HANDOFF_COMMAND` repeats the same top-level scratch pattern for `d`, `r`, `e`, and `s`, so approved handoffs corrupt the same ordinary user state.

src/cosh-ng/crates/cosh-shell/src/raw_input/spawn.rs
```diff
@@ -192,12 +176,34 @@ where
...
+                            crate::input::InputDecision::Intercept { .. }
+                                | crate::input::InputDecision::Consume
+                        );
+                        state.deferred_input = Some(InputRead::Bytes {
+                            bytes: remainder,
+                            received_at,
+                            observed_mode: mode.clone(),
+                            ownership_changed_during_read: false,
+                            pending_shell_submits: pending_shell_submits
+                                .saturating_add(u16::from(shell_owned)),
```
ikunkun-sys, line 205: [P1] Blank lines falsely add a pending command boundary

For a batch such as `"\n?? what happened\n"`, the empty first line classifies as `SendToShell`, so this increments `pending_shell_submits` before queuing the following intercept. Bash does not expand `PS0` for a blank line, while `osc.rs` only decrements this counter after a preexec/routing marker. The command-less prompt therefore leaves the intercept pending until a later non-empty command, silently delaying and reordering the user request. The action-relay path already avoids counting `line.is_empty()`, but the real reader path here does not.

ikunkun-sys (changes requested):
The current head has three blocking shell-correctness regressions:

- Bash 4.3–5.0 executes only the first `PROMPT_COMMAND` array element, so Enhanced never emits its completion boundary on that supported baseline.
- The top-level prompt and handoff wrappers overwrite and unset ordinary user shell variables.
- A blank shell submission batched before an intercepted control leaves that control queued past the command-less prompt.
- [P2] The paired [English](https://github.com/alibaba/anolisa/blob/dac3ce1e86d974b388c87f2e614c076d98b85bfc/src/cosh-ng/docs/design/shell-integration-and-card-types.md#L106-L128) and [Chinese](https://github.com/alibaba/anolisa/blob/dac3ce1e86d974b388c87f2e614c076d98b85bfc/src/cosh-ng/docs/design/shell-integration-and-card-types_zh.md#L92-L110) design docs still say Enhanced retains the global hook/injection behavior and does not eliminate #2687, contradicting this PR's implementation and closure claim.

These findings are distinct from the existing review threads and were revalidated against `dac3ce1e`.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.rs
kongche-jbw: Fixed in d128e0f1. Bash 4.3–5.0 now receives one scalar PROMPT_COMMAND containing the status capture, the native element-zero hook, and the final Cosh boundary; Bash 5.1+ keeps the array form. Marker structure coverage pins the version gate, and the full shell-host/PTY CI is green.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.rs
kongche-jbw: Fixed in d128e0f1. Prompt and bounded-handoff wrappers now snapshot each temporary variable declaration together with the trap state, suspend traps, and restore the user variables before any user hook or approved command runs. The handoff regression verifies that d/r/e/s and __prompt_status changes made by the approved command persist; the Bash trap/option oracle remains green.

src/cosh-ng/crates/cosh-shell/src/raw_input/spawn.rs
kongche-jbw: Fixed in d128e0f1. The real reader and action paths now share is_pending_shell_submission, which rejects empty/whitespace-only lines and intercepted controls before incrementing pending_shell_submits. The new regression covers the empty line + Agent control ordering case, and the full raw_cli gate passed 465 tests.

src/cosh-ng/crates/cosh-shell/src/raw_input/spawn/action.rs
```diff
@@ -296,9 +296,8 @@ pub(crate) fn spawn_raw_action_relay_with_wake(
                 &mut state,
             )?;
             match action {
-                RawRelayAction::Write(bytes) => relay_input_bytes(
+                RawRelayAction::Write(bytes) => relay_action_write_batch(
```
ikunkun-sys, line 299: [P1] Preserve zsh job-control completion in the action relay

This reroutes every scripted Write, including Ctrl-Z and Ctrl-C, through relay_action_write_batch. On current head 3d275e2c, foreground::raw_relay_zsh_job_control_suspend_fg_and_interrupt consistently reaches the 122-second watchdog after the trailing exit; the same test on be5c9f10 before this batching path passes in 2.22 seconds, while the first buildable post-change commit fails identically. Hosts with zsh therefore cannot complete this existing shell-host gate after a suspend/fg/interrupt sequence.

src/cosh-ng/crates/cosh-shell/tests/shell_host/native.rs
```diff
@@ -67,137 +257,441 @@ trap 'printf "%s\n" "$BASH_COMMAND" >> "$HOME/err-trap.log"' ERR
...
+        terminal.contains("errtrace") && terminal.contains("off"),
+        "{terminal}"
+    );
+    assert!(
+        terminal.contains("functrace") && terminal.contains("off"),
+        "{terminal}"
+    );
+    assert!(terminal.contains("trap -- 'printf"), "{terminal}");
+    assert!(
+        terminal.contains("__PROMPT_COMMAND__=declare -a PROMPT_COMMAND=")
```
ikunkun-sys: [P1] Accept scalar PROMPT_COMMAND on Bash 4.3-5.0

The implementation now deliberately uses a scalar PROMPT_COMMAND below Bash 5.1, but this oracle still unconditionally requires declare -a. On Bash 4.4.20 the exact test fails here while the terminal reports declare -- PROMPT_COMMAND=..., even though docs/BUILDING.md requires Bash 4.3 or newer. This leaves the required shell-host test gate failing throughout the compatibility range that the fix is meant to restore.

ikunkun-sys (changes requested):
Requesting changes on current head 3d275e2c.

In addition to the two inline P1 findings:

- Commit 0f0b1eb0 is not independently buildable. cargo +1.88.0 check -p cosh-shell --all-targets fails with E0061 at raw_input/spawn.rs:327 because resolve_assistance_shortcut requires input_mode but the caller supplies only six arguments. A later commit hides the failure, so the current history violates the repository requirement that every PR commit compile independently.
- Rebasing this head onto the current main at a634e665 reproducibly conflicts while applying d3bf85d6 in tests/protocol/provider_lifecycle.rs:342. The rebase was performed in an isolated detached worktree and aborted after recording the conflict.

The previous three blocking findings were rechecked and are fixed, so they are not repeated here. GitHub CI is green, but its environment does not expose the Bash 4.4 and installed-zsh failures described inline.

src/cosh-ng/crates/cosh-shell/src/raw_input/spawn/action.rs
kongche-jbw: Fixed in `dcbd9904`. `relay_action_write_batch` now delegates directly to the legacy relay unless `should_split_passthrough_batch` proves that the same action contains a submitted line plus follow-up input. Ctrl-Z, Ctrl-C, partial writes, and ordinary single-line actions therefore no longer enter the batch splitter. On this host with zsh installed, `foreground::raw_relay_zsh_job_control_suspend_fg_and_interrupt` passed 5/5 in 2.27–2.31 s; the current head CI is also fully green.

src/cosh-ng/crates/cosh-shell/tests/shell_host/native.rs
kongche-jbw: Fixed in `dcbd9904`. The oracle now probes the local Bash version and expects `declare -a PROMPT_COMMAND=` only on Bash 5.1+, while Bash 4.3–5.0 expects the scalar `declare -- PROMPT_COMMAND=` form. The exact oracle test passes on the current host, and the implementation/test version gates now match.

src/cosh-ng/crates/cosh-shell/src/raw_input/spawn/action.rs
ikunkun-sys: This is still reproducible on current head d534d99a. On the exact GitHub merge commit 8bc473d6, the test reaches the same 122.10-second watchdog; current base a634e665 passes in 2.26s and f8c1bf62 passes in 2.22s, while dcbd9904 is the first failing commit at 122.09s. The failure output shows that fg never resumes the stopped sleep, so the trailing exit reports “zsh: you have suspended jobs.” Narrowing relay_action_write_batch therefore does not resolve the regression. The remaining source path worth checking is the shell-agnostic call to the Bash/Readline-specific history rewrite in raw_input/relay.rs after Ctrl-Z dirties the line mirror.

ikunkun-sys (changes requested):
Requesting changes on current head d534d99a.

The Bash 4.3–5.0 oracle, intermediate-commit buildability, and rebase-conflict findings are fixed.

One P1 remains: the existing zsh suspend/fg/interrupt test still deterministically fails on the exact merge commit after 122.10s. The current base and first PR commit pass in about 2.2s, while dcbd9904 is the first failing commit. I added the current-head evidence to the existing thread rather than duplicating the inline finding.

GitHub CI is green, but that environment does not expose this installed-zsh failure.

src/cosh-ng/crates/cosh-shell/src/raw_input/spawn/action.rs
kongche-jbw: Fixed in `50e33225` on current head `7a14e77f`. I reproduced the 122.15 s watchdog by setting `EDITOR=vim` and `VISUAL=vim`: zsh selected vi keybindings, then the shell-agnostic history rewrite injected Bash Readline `Ctrl-A + space + Ctrl-E`, so `fg` was not resumed correctly. The relay now enables that rewrite only for Enhanced Bash; zsh receives the original bytes unchanged. The existing zsh job-control test now forces this vi-mode environment and passes in 2.27 s. Bash/zsh sensitive-history tests, all nine PR commits with `cosh-shell --all-targets`, and the latest GitHub CI are green.

ikunkun-sys (approved):
Approving current head 7a14e77f.

The zsh vi-mode job-control regression is fixed on the exact merge commit, and the old-head control still distinguishes the regression. Bash and zsh history-privacy tests, the amended second commit with Rust 1.88 all-targets checking, formatting, diff/layout checks, and current GitHub CI all pass. No current blockers remain.

Issue #2818, opened 2026-08-25
[TestAgent][sight] interruption_events.db size-based purge over-deletes all rows instead of trimming to the limit


zhangtaibo:
### Component

agentsight

### Bug Description

When `interruption_events.db` exceeds the configured `features.interruption_detection.max_db_size_mb` limit, the periodic purge **deletes ALL rows instead of trimming the oldest records until the file fits** — and after the purge the main db file stays oversized because the bare `VACUUM;` is not paired with `wal_checkpoint(TRUNCATE)`. This violates the v0.11.0 release-note contract: "deletes the oldest records" + "reclaims disk space through VACUUM and WAL checkpoint".

### Steps to Reproduce

```shell
ssh root@<ECS-IP>   # authorized test host, replace with your own
# 1. Set a small size limit (1 MB) for interruption_events.db
python3 - <<'PY'
import json
p = "/etc/agentsight/config.json"
d = json.load(open(p))
d.setdefault("features", {})["interruption_detection"] = {
    "enabled": True, "retention_days": 30, "max_db_size_mb": 1}
json.dump(d, open(p, "w"), indent=2)
PY
systemctl restart agentsight

# 2. Seed ~2 MB of interruption rows (400 x ~5 KB)
python3 - <<'PY'
import sqlite3, time
conn = sqlite3.connect("/var/log/sysak/.agentsight/interruption_events.db", timeout=30)
now = int(time.time() * 1_000_000_000)
blob = "y" * (4 * 1024)
for i in range(400):
    conn.execute(
        "INSERT INTO interruption_events (interruption_id, session_id, trace_id, "
        "call_id, pid, agent_name, interruption_type, severity, occurred_at_ns, "
        "detail, resolved) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, 0)",
        ("purgetest-%06d" % i, "s-%06d" % i, "t-%06d" % i, "c-%06d" % i,
         1000 + i, "PurgeProbe", "auth_error", "high",
         now - i * 1_000_000_000, blob))
conn.commit()
PY

# 3. Force rows into the main db file (so the size check sees them)
python3 -c "import sqlite3; c = sqlite3.connect('/var/log/sysak/.agentsight/interruption_events.db', timeout=30); c.execute('PRAGMA wal_checkpoint(TRUNCATE)')"

# 4. Wait one purge period (60 s) and inspect
sleep 75
[...]

PR #2872, opened 2026-08-26
fix(sight): trim oversized db by row fraction

### LINKED ISSUE #2870: [TestAgent][sight] agentsight.db size purge wipes all four tables even when only slightly over the 500MB limit
### Component

agentsight

### Bug Description

When `agentsight.db` exceeds its hardcoded 500 MB cap, the startup (and periodic) size purge **wipes ALL FOUR tables entirely — including rows written minutes before — even when the database is only 4% over the limit**. The release-note contract is "清理优先删除最旧数据" (delete oldest first): the purge does delete oldest-first and does reclaim the file, but the per-round deletion granularity (up to 20 000 rows per table without re-checking the size between batches) means any table with fewer rows than that gets emptied in a single round. Over-deletion of the newest data — sibling of #2818 but in a different store path.

### Steps to Reproduce

```shell
# 1. Seed LLM-body-shaped http_records rows past the 500MB cap.
#    (sqlite_storage is enabled by default: features.sqlite_storage.enabled=true)
python3 - <<'PY'
import sqlite3, time
db = "/var/log/sysak/.agentsight/agentsight.db"
conn = sqlite3.connect(db, timeout=60)
conn.execute("DELETE FROM http_records WHERE comm='SizeProbe'")
now = int(time.time() * 1e9)
blob = "x" * (512 * 1024)          # ~1MB per row (request+response bodies)
for i in range(520):               # 520 rows ~= 521MB, only 4% over the cap
    ts = (now - 100*3600*10**9) + i * (100*3600*10**9 // 520)
    conn.execute(
        "INSERT INTO http_records (timestamp_ns, pid, comm, method, path, "
        "status_code, request_headers, request_body, response_headers, "
        "response_body, duration_ns, is_sse, sse_event_count) VALUES "
        "(?, 1, 'SizeProbe', 'POST', '/v1/chat', 200, ?, ?, ?, ?, 1000000, 0, 0)",
        (ts, "{}", blob, "{}", blob))
conn.commit()
conn.execute("PRAGMA wal_checkpoint(TRUNCATE)")
conn.close()
PY

# 2. Restart the service to trigger the startup size check.
systemctl restart agentsight
sleep 45

# 3. Inspect: file size AND row counts across all four tables.
python3 -c "
import sqlite3, os
db = '/var/log/sysak/.agentsight/agentsight.db'
print('size: %.1f MB' % (sum(os.path.getsize(db+s) for s in ('','-wal','-shm'))/1048576))
c = sqlite3.connect(db, timeout=30)
for t in ('http_records','audit_events','token_records','token_consumption'):
    print(t, c.execute(f'SELECT COUNT(*) FROM {t}').fetchone()[0])
"
```

### Actual Behavior

Reproduced 3 consecutive times (oversize 120% / 20% / 4%) — identical outcome each run:

```
run 1 (1102 MB seeded, 1100 rows): size -> 0.1 MB, http_records 0/1100, all 4 tables empty
run 2 ( 602 MB seeded,  600 rows): size -> 0.1 MB, http_records 0/600,  all 4 tables empty
run 3 ( 522 MB seeded,  520 rows): size -> 0.1 MB, http_records 0/520,  all 4 tables empty
```

Even at only 4% over the limit, EVERY row is deleted — including the newest ones (timestamps minutes before the restart) — and the audit/token/http/token_consumption history built up during real usage is wiped along with the probe data. The file reclaim itself works correctly and the service stays healthy.

### Expected Behavior

Per the release notes: the purge deletes the OLDEST records — implying it should stop once the size fits the cap again. For a 4% overshoot only the oldest ~4% of rows should be trimmed; the newest data (written just before the restart) must survive. The sibling genai store implements exactly this convergence (`check_and_prune_if_needed` re-measures the size after each percentage batch — see also the fix pattern agreed for #2818).

### Root Cause

`Storage::purge_oversized` in `src/storage/unified.rs`: the inner batch loop runs

```rust
const BATCH_ROWS: usize = 1000;
const BATCHES_PER_ROUND: u32 = 20;
...
for _ in 0..BATCHES_PER_ROUND {
    let deleted = audit(1000) + token(1000) + http(1000) + token_consumption(1000);
    if deleted == 0 { break; }
}
```

**The size is only re-checked between ROUNDS (after the VACUUM), not between batches** — one round deletes up to 20 000 rows per table (80 000 total). Any table holding fewer rows than that is emptied in the first round regardless of how little the database exceeds the cap. LLM HTTP bodies make this easy to hit: a few hundred large `http_records` rows blow past 500 MB while staying far below 20 000 rows, so the first over-limit purge wipes all four tables including the newest records.

Note the contrast with #2818 (interruption store): there the loop never shrank the file (bare VACUUM, size measured on the main file only) so it emptied the table as a side effect; here the reclaim mechanics are correct (VACUUM + `wal_checkpoint(TRUNCATE)`, main+WAL+SHM counted) but the deletion granularity is too coarse — the fix is the same family: re-check the size between batches or use percentage-based batches, mirroring the genai store / the #2818 fix pattern.

Also worth noting (config gap, related): the 500 MB cap is NOT configurable — `JsonSqliteStorageFeature` (src/config.rs:352-356) exposes only `enabled` and `batch`, while `DEFAULT_MAX_DB_SIZE_MB = 500` is hardcoded, so operators cannot lower the cap on disk-constrained hosts.

### Suggested Fix

In `purge_oversized`, re-measure `total_db_size()` between batches (or delete percentage-based batches like the genai store's `prune_old_records_with_percent`) and stop as soon as the size fits — keeping the newest rows. Optionally expose `max_db_size_mb` in the sqlite_storage config section.

### Environment

- OS: Alibaba Cloud Linux 4
- Component version: agentsight 0.11.0
- Package/RPM: agentsight-0.11.0-1.alnx4.x86_64

### Additional Context

- Related: #2818 (interruption store over-deletion, fix incoming), #2826 (container data loss) — this is the third member of the same over-deletion family, in the main unified store.
- Related: #2641 / PR #2644 introduced the sqlite size-limit feature.
- Stability: reproduced 3 consecutive times with identical outcome (deterministic — the batch floor is structural).
- Reproduction harness: seeded rows mimic real LLM traffic shape (`http_records` request/response bodies, ~1 MB per row); the same wipe also consumed the genuine audit/token history accumulated during real cosh sessions on the test host.

jfeng18:
## Why

When `agentsight.db` exceeds its size cap, `Storage::purge_oversized` wiped **all four tables** even at 4% overshoot: the inner loop deleted up to 20x1000 rows per table between size re-checks, so any table smaller than the per-round batch total was emptied in the first round — including rows written minutes before. See #2870 (reproduced 3 times on 0.11.0). Sibling defect of #2818 (same introducing commit); here the reclaim mechanics were correct, only the batch granularity was wrong.

## What changed

- `purge_oversized` now deletes a **fraction of each table's rows per round** (50% / 25% / 10% by overshoot, mirroring the genai store's prune policy and the interruption store fix in #2823) and re-measures the total size after each round's VACUUM+checkpoint.
- Stall detection reworked after review: rows and bytes decouple under B-tree page sharing, so a non-shrinking round no longer stops the purge — instead the bite doubles (capped at 90%) until the rows dominating the file are reached. The purge now stops early only when VACUUM/checkpoint itself fails 3 times (the real stall signal) or after 20 rounds; both paths log an error.
- `AuditStore::count()` added — the only one of the four stores without a row count, needed to size its batch proportionally.
- New discriminating tests: `test_purge_oversized_trims_oldest_keeps_newest` and `test_purge_oversized_skewed_row_sizes_still_converges` (tiny old rows + few large new rows: early rounds free no page, the purge must still converge — fails if a non-shrinking round aborts the loop). The former (~17% overshoot): asserts the purge trims without wiping, the newest row survives (`MAX(timestamp_ns)`), and the file converges below the limit. Under the old fixed-batch loop the remaining count would be 0, so this test fails if the fix is reverted. (The pre-existing `test_purge_oversized_shrinks_db_below_limit` passes even when everything is deleted — it only asserts file size.)

Out of scope (follow-up): the size-prune loops of the genai, interruption, and unified stores now share the same percentage-batch + in-loop/round checkpoint pattern and could be factored into a shared helper.

## Related issue

closes #2870

## User / Agent impact

Oversized `agentsight.db` is now trimmed oldest-first instead of wiped: at 4% overshoot only the oldest ~10% of rows per round are deleted until the file fits; the newest data survives. No config or API change.

## Risk and compatibility

Low risk: same entry point and limit semantics; only the per-round deletion volume changed.

## Validation

On ECS (Alinux, kernel 6.6, rustup 1.89.0 matching CI):

- `cargo fmt --all --check` — pass
- `python3 scripts/check-arch-boundaries.py` — 0 violations
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace --lib --bins --tests` — 1903 passed, 0 failed (includes the new regression test; all 4 pre-existing purge tests still green)

## Documentation and rollback

None (bug fix to documented size-limit behavior). Revert the single commit to roll back; `Fixes: 33ed3846` identifies the introducing commit.

### DIFF
--- a/src/agentsight/src/storage/sqlite/audit.rs
+++ b/src/agentsight/src/storage/sqlite/audit.rs
@@ -187,6 +187,21 @@ impl AuditStore {
         Ok(deleted as u64)
     }
 
+    /// Number of records in the table.
+    ///
+    /// Used by size-based purging to size deletion batches proportionally.
+    /// The table name is interpolated, not parameterized (SQL placeholders do
+    /// not support identifiers); it is built internally from config, never
+    /// from user input.
+    pub fn count(&self) -> Result<u64> {
+        let n: i64 = self.conn.query_row(
+            &format!("SELECT COUNT(*) FROM {}", self.table_name),
+            [],
+            |r| r.get(0),
+        )?;
+        Ok(n as u64)
+    }
+
     /// Delete the oldest N records by timestamp.
     ///
     /// Used for size-based pruning when the database file exceeds its
--- a/src/agentsight/src/storage/unified.rs
+++ b/src/agentsight/src/storage/unified.rs
@@ -368,16 +368,16 @@ impl Storage {
     /// companions — in WAL mode a large share of recent data lives in the WAL
     /// file, so checking the main file alone would under-report.
     ///
-    /// Work is organized in rounds: each round deletes a fixed number of
-    /// oldest-record batches from all tables, then runs one VACUUM +
-    /// checkpoint before re-measuring. VACUUM rewrites the whole file, so it
-    /// must not run per batch — bounding it per round keeps a multi-GB
-    /// cleanup tractable. VACUUM failures are tolerated: freed pages are
-    /// still reusable by future inserts.
+    /// Work is organized in rounds: each round deletes a fraction of every
+    /// table's oldest rows, then runs one VACUUM + checkpoint before
+    /// re-measuring. VACUUM rewrites the whole file, so it must not run per
+    /// batch — bounding it per round keeps a multi-GB cleanup tractable.
     ///
-    /// Rounds continue until the size is within the limit; termination is
-    /// guaranteed because every round must strictly shrink the file (or empty
-    /// the tables), otherwise the purge stops and reports the stall.
+    /// Rounds continue until the size is within the limit. A round that does
+    /// not shrink the file is not necessarily a failure (deleting small rows
+    /// may free no page), so the bite doubles while the file stands still;
+    /// the purge stops early only when VACUUM/checkpoint itself keeps failing
+    /// or after 20 rounds, both reported as errors.
     ///
     /// Called automatically by `store()` during purge checks and by
     /// `with_sqlite_config()` on startup.
@@ -386,13 +386,12 @@ impl Storage {
             return Ok(());
         }
 
-        // Rows deleted per table per batch, and batches per round: up to
-        // 20 000 rows per table between two VACUUM runs.
-        const BATCH_ROWS: usize = 1000;
-        const BATCHES_PER_ROUND: u32 = 20;
-
-        let mut prev_size = u64::MAX;
         let mut round = 0u32;
+        let mut reclaim_failures = 0u32;
+        // Rows and bytes can decouple (B-tree page sharing): deleting many
+        // small rows may free no page at all, so a non-shrinking round is
+        // normal and triggers a bigger bite next round instead of a stall.
+        let mut pct_boost = 1.0f64;
 
         loop {
             let size = self.total_db_size();
@@ -400,60 +399,95 @@ impl Storage {
                 break;
             }
 
-            // A round of deletes + VACUUM that does not shrink the file means
-            // VACUUM is failing (e.g. disk full). Deleting further rows would
-            // destroy data without freeing disk space, so stop and escalate:
-            // freed pages are still reused by future inserts, which prevents
-            // the main file from growing, but the size cap is no longer
-            // enforceable without operator action.
-            if size >= prev_size {
+            if round >= 20 {
                 log::error!(
-                    "Size-based purge stalled: database is {} bytes (limit {}) and \
-                     did not shrink after round {round}. VACUUM is likely failing \
-                     (e.g. disk full). Size enforcement is suspended — manual \
+                    "Size-based purge did not converge after {round} rounds: database is \
+                     {size} bytes (limit {}). Size enforcement is suspended — manual \
                      intervention required (free disk space or remove the database).",
-                    size,
                     self.max_db_size_bytes
                 );
                 break;
             }
-            prev_size = size;
             round += 1;
 
+            // Delete a fraction of each table's rows per round and re-measure
+            // after the round's VACUUM+checkpoint. A fixed per-batch row
+            // count without re-measuring between batches wipes any table
+            // smaller than the per-round batch total (#2870). Bigger bites
+            // when far over the limit; mirrors the genai store's prune policy.
+            let overshoot = size as f64 / self.max_db_size_bytes as f64;
+            let base_pct = if overshoot > 5.0 {
+                0.50
+            } else if overshoot > 2.0 {
+                0.25
+            } else {
+                0.10
+            };
+            let pct = (base_pct * pct_boost).min(0.9);
+
             let mut round_deleted = 0usize;
-            for _ in 0..BATCHES_PER_ROUND {
-                let deleted = self.audit_store.delete_oldest_batch(BATCH_ROWS)?
-                    + self.token_store.delete_oldest_batch(BATCH_ROWS)?
-                    + self.http_store.delete_oldest_batch(BATCH_ROWS)?
-                    + self
-                        .token_consumption_store
-                        .delete_oldest_batch(BATCH_ROWS)?;
-                round_deleted += deleted;
-                if deleted == 0 {
-                    break; // All tables empty
-                }
+            round_deleted += self
+                .audit_store
+                .delete_oldest_batch(purge_share(self.audit_store.count()?, pct))?;
+            round_deleted += self
+                .token_store
+                .delete_oldest_batch(purge_share(self.token_store.count(), pct))?;
+            round_deleted += self
+                .http_store
+                .delete_oldest_batch(purge_share(self.http_store.count()?, pct))?;
+            round_deleted += self
+                .token_consumption_store
+                .delete_oldest_batch(purge_share(self.token_consumption_store.count(), pct))?;
+
+            if round_deleted == 0 {
+                break; // Nothing left to delete
             }
 
             // VACUUM first, then checkpoint: in WAL mode VACUUM writes the
             // rebuilt database through the WAL, so the file only shrinks once
-            // a TRUNCATE checkpoint flushes it back. Both are best-effort:
-            // freed pages are reusable even when they fail.
+            // a TRUNCATE checkpoint flushes it back. Failures are the real
+            // stall signal (e.g. disk full): deleting further rows would
+            // destroy data without freeing space, so stop after repeated
+            // failures and escalate.
+            let mut reclaim_ok = true;
             if let Err(e) = self.audit_store.vacuum() {
                 log::warn!("VACUUM during size-based purge failed: {e}");
+                reclaim_ok = false;
             }
             if let Err(e) = self.audit_store.checkpoint() {
                 log::warn!("WAL checkpoint during size-based purge failed: {e}");
+                reclaim_ok = false;
+            }
+            if reclaim_ok {
+                reclaim_failures = 0;
+            } else {
+                reclaim_failures += 1;
+                if reclaim_failures >= 3 {
+                    log::error!(
+                        "Size-based purge suspended after {reclaim_failures} failed \
+                         VACUUM/checkpoint rounds at {size} bytes (limit {}). Freed \
+                         pages remain reusable by future inserts; free disk space or \
+                         remove the database to enforce the cap.",
+                        self.max_db_size_bytes
+                    );
+                    break;
+                }
             }
 
+            let new_size = self.total_db_size();
             log::info!(
                 "Size-based purge round {round}: deleted {round_deleted} rows, \
-                 database now {} bytes (limit {})",
-                self.total_db_size(),
+                 database now {new_size} bytes (limit {})",
                 self.max_db_size_bytes
             );
 
-            if round_deleted == 0 {
-                break; // Nothing left to delete
+            if new_size < size {
+                pct_boost = 1.0;
+            } else {
+                // Deletion made no dent in the file (page sharing). Double the
+                // bite so subsequent rounds reach the rows that dominate the
+                // file instead of stalling on small-row deletions.
+                pct_boost = (pct_boost * 2.0).min(9.0);
             }
         }
 
@@ -530,6 +564,12 @@ impl Drop for Storage {
             log::warn!("WAL checkpoint during Storage drop failed: {e}");
         }
     }
+}
+
+/// Rows to delete from a table in one purge round: `pct` of its rows, at
+/// least 1 (deleting from an empty table is a no-op).
+fn purge_share(rows: u64, pct: f64) -> usize {
+    ((rows as f64 * pct) as usize).max(1)
 }
 
 #[cfg(test)]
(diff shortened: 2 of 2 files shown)


src/agentsight/src/storage/unified.rs
```diff
@@ -420,19 +415,33 @@ impl Storage {
             prev_size = size;
             round += 1;
 
+            // Delete a fraction of each table's rows per round and re-measure
+            // after the round's VACUUM+checkpoint. A fixed per-batch row
+            // count without re-measuring between batches wipes any table
+            // smaller than the per-round batch total (#2870). Bigger bites
+            // when far over the limit; mirrors the genai store's prune policy.
+            let overshoot = size as f64 / self.max_db_size_bytes as f64;
```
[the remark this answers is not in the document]
jfeng18: Not changing the thresholds: 50%/round is exponential (a 100x overshoot converges in ~7 rounds), and purge runs on startup + a slow periodic path, so rounds are cheap relative to VACUUM cost. Under sustained write pressure, catching up is a capacity concern rather than a batch-size one. Note this revision does add one adaptive element for a different reason: the bite doubles (capped at 90%) while a round fails to shrink the file, which handles the page-sharing case from the other review thread.

src/agentsight/src/storage/sqlite/audit.rs
```diff
@@ -187,6 +187,18 @@ impl AuditStore {
         Ok(deleted as u64)
     }
 
+    /// Number of records in the table.
+    ///
+    /// Used by size-based purging to size deletion batches proportionally.
+    pub fn count(&self) -> Result<u64> {
+        let n: i64 = self.conn.query_row(
+            &format!("SELECT COUNT(*) FROM {}", self.table_name),
```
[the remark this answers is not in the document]
jfeng18: Partially fixed: added a comment stating the table name is built internally from config and never from user input. The implementation stays string-interpolated because SQL placeholders cannot bind identifiers, and every other method in this store (`purge_before`, `delete_oldest_batch`, ...) uses the same established pattern.

src/agentsight/src/storage/unified.rs
```diff
@@ -420,19 +415,33 @@ impl Storage {
...
+            let pct = if overshoot > 5.0 {
+                0.50
+            } else if overshoot > 2.0 {
+                0.25
+            } else {
+                0.10
+            };
```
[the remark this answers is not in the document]
jfeng18: Fixed: the stall signal is now VACUUM/checkpoint failing 3 times (the real failure mode), not a non-shrinking round. While the file stands still the bite doubles (capped at 90%), so trimming continues until it reaches the rows that dominate the file; a hard 20-round bound still escalates with an error log. Covered by the new `test_purge_oversized_skewed_row_sizes_still_converges` (tiny old rows + few large new rows), which fails under the previous stop-on-no-shrink logic.

chengshuyi (approved):
LGTM. 比例删除 + pct_boost 自适应机制正确修复了 #2870 的过度删除问题，终止性保证充分（四重退出条件），两个新回归测试精准覆盖核心缺陷和 page-sharing 边界。CI 全绿，diff 249 行在限制内。

src/agentsight/src/storage/unified.rs
```diff
@@ -386,74 +386,108 @@ impl Storage {
...
+            round_deleted += self
+                .http_store
+                .delete_oldest_batch(purge_share(self.http_store.count()?, pct))?;
+            round_deleted += self
+                .token_consumption_store
+                .delete_oldest_batch(purge_share(self.token_consumption_store.count(), pct))?;
+
+            if round_deleted == 0 {
+                break; // Nothing left to delete
             }
```
Daydreamer-Li, line 444: **[P1 回归] 这个 break 挪到 VACUUM 之前了，导致「表已空但文件仍超限」时不再回收**

旧实现的顺序是：内层批次循环遇到 `deleted == 0` 退出 → **跑 VACUUM + checkpoint** → 再判断 `round_deleted == 0` 才 break。新实现把 break 提到了回收之前。

影响场景：四张表都已为空（例如 `purge_expired` 刚按时间删空、但当时 VACUUM 失败），文件因大量空闲页或膨胀的 WAL 仍超过上限。此时新代码首轮删 0 行 → 直接 break → **一次 VACUUM/checkpoint 都不执行**，体积永远降不下来；而且之后每次 `store()` 触发的检查都是同样的空转，没有任何日志暴露这个状态。旧代码至少会回收一次。

建议：`round_deleted == 0` 时先执行一次 VACUUM + checkpoint 再 break（或走一条「仅回收」的分支）。

src/agentsight/src/storage/unified.rs
```diff
@@ -386,74 +386,108 @@ impl Storage {
...
+                .delete_oldest_batch(purge_share(self.audit_store.count()?, pct))?;
+            round_deleted += self
+                .token_store
+                .delete_oldest_batch(purge_share(self.token_store.count(), pct))?;
+            round_deleted += self
+                .http_store
+                .delete_oldest_batch(purge_share(self.http_store.count()?, pct))?;
+            round_deleted += self
+                .token_consumption_store
+                .delete_oldest_batch(purge_share(self.token_consumption_store.count(), pct))?;
```
Daydreamer-Li, line 440: **[P1] 按行比例删所有四张表，与「谁占了体积」无关，issue 的期望行为未完全满足**

撑爆 `agentsight.db` 的几乎只有 `http_records`（LLM 请求/响应体，单行约 1MB）；`audit_events`、`token_records` 行小、对体积贡献极少，但它们是用户真实的审计与 token 历史。现在每轮对四张表**一律**按各自行数百分比删除，删除量与体积贡献完全脱钩。

按 issue 的 run 1 数据推演（1102MB、超限 120%）：首轮 overshoot=2.2 → 各表删 25%，之后每轮 10%，约 5~7 轮收敛，`audit_events` 会被削掉 40% 以上——而它一个字节都不是原因。issue 的 Expected Behavior 写的是「超限 4% 就只该修掉最旧的约 4%」，目前只做到了「不再清空」，没做到「只删该删的」。

建议二选一（前者彻底）：
- 按**字节份额**分配删除量：用 `SUM(LENGTH(payload 列))` 或 `dbstat` 估各表体积占比，只对占比超过阈值（如 5%）的表删除，删除量按字节份额分配；
- 至少让占比算下来为 0 的表本轮不删（见 `purge_share` 那条），并为每表设「最少保留 N 行 / 不删最近 X 分钟的行」的硬地板——把 issue 的契约（最新数据必须存活）写成代码不变量，不再依赖收敛启发式恰好正确。

src/agentsight/src/storage/unified.rs
```diff
@@ -532,6 +566,12 @@ impl Drop for Storage {
     }
 }
 
+/// Rows to delete from a table in one purge round: `pct` of its rows, at
+/// least 1 (deleting from an empty table is a no-op).
+fn purge_share(rows: u64, pct: f64) -> usize {
+    ((rows as f64 * pct) as usize).max(1)
```
Daydreamer-Li, line 572: **[P1] `.max(1)` 地板会让小表在多轮后被清空**

`.max(1)` 意味着**任何非空表每轮至少被删 1 行**，即使它按比例算下来应该是 0 行（例如 3 行的表 × 10% = 0）。轮次上限是 20，所以**行数 ≤ 20 的表仍然可能被完全清空**——这正好是新装机器上 `audit_events` / `token_records` 的典型规模，也正好是 #2870 抱怨的「连带清掉真实历史」。

`.max(1)` 目前唯一的作用是保证 `round_deleted > 0`（避免误触发 `round_deleted == 0` 的 break）。建议改为：比例算出 0 行的表本轮不删，终止条件改为按「所有表 count 均为 0」判定，这样既保留终止性，又不伤害小表。

src/agentsight/src/storage/unified.rs
```diff
@@ -386,74 +386,108 @@ impl Storage {
...
+            // when far over the limit; mirrors the genai store's prune policy.
+            let overshoot = size as f64 / self.max_db_size_bytes as f64;
+            let base_pct = if overshoot > 5.0 {
+                0.50
+            } else if overshoot > 2.0 {
+                0.25
+            } else {
+                0.10
+            };
+            let pct = (base_pct * pct_boost).min(0.9);
```
Daydreamer-Li, line 426: **[P2] bite 翻倍之后仍会一次删过头，与 #2870 是同一类缺陷**

`pct_boost` 只在文件缩小时复位，涨到上限后 `pct` 达到 0.9。一旦此时终于删到「真正占体积的大行表」，一轮就削掉 90%，而实际可能只需删几行就能回到上限内——本 PR 新增的 `test_purge_oversized_skewed_row_sizes_still_converges` 就是这个现象：4 行大记录其实删 1 行就够，实际删了 3 行。行数放大到 100 行时就是「删 90 行只为省 5 行的空间」，仍属于 #2870 的过量删除同族，只是触发条件更苛刻。

有个更便宜也更精确的办法可以替掉这套启发式：用上一轮的 `已释放字节 / 已删行数` 估出平均行大小，直接反推「还需删多少行」（乘 1.5~2 安全系数）。这样一两轮就精确收敛，既不需要 boost，也不需要 20 轮上限兜底。

src/agentsight/src/storage/unified.rs
```diff
@@ -386,74 +386,108 @@ impl Storage {
...
-                round_deleted += deleted;
-                if deleted == 0 {
-                    break; // All tables empty
-                }
+            round_deleted += self
+                .audit_store
+                .delete_oldest_batch(purge_share(self.audit_store.count()?, pct))?;
+            round_deleted += self
+                .token_store
+                .delete_oldest_batch(purge_share(self.token_store.count(), pct))?;
```
Daydreamer-Li, line 434: **[P2] 每轮 4 次 `COUNT(*)` 全表扫描，且整个 purge 同步跑在写入路径上；另外两个 `count()` 会吞掉错误**

1. **成本**：`purge_oversized` 由 `store()` 同步调用（写入路径）和启动时调用。在 500MB 的 `http_records` 上 `COUNT(*)` 是全表/全索引扫描，最坏 20 轮 × 4 表 = 80 次扫描，再叠加每轮一次全文件 VACUUM，可能把事件处理线程阻塞很久——这会直接反压到上游事件缓冲（与 #2888 的内存积压耦合）。你回复 qoderai 那条 P2 时说的是「rounds are cheap **relative to** VACUUM cost」，这解释了相对成本，但没有约束绝对阻塞时长。建议给整个 purge 加墙上时钟预算（如 30s，超时后记录并留到下次继续），或用 `max(rowid) - min(rowid)` 估行数替代 `COUNT(*)`。

2. **错误处理不一致**：新增的 `AuditStore::count()` 与 `HttpStore::count()` 返回 `Result`，但 `TokenStore::count()`（token.rs:460）和 `TokenConsumptionStore::count()`（token_consumption.rs:377）返回 `u64`、**查询失败时返回 0**。0 经过 `purge_share` 的 `.max(1)` 变成 1，于是查询失败会静默退化为「每轮只删 1 行」，既收敛不了也没有任何告警。建议统一为 `Result`。

src/agentsight/src/storage/unified.rs
```diff
@@ -386,74 +386,108 @@ impl Storage {
...
             }
 
-            // A round of deletes + VACUUM that does not shrink the file means
-            // VACUUM is failing (e.g. disk full). Deleting further rows would
-            // destroy data without freeing disk space, so stop and escalate:
-            // freed pages are still reused by future inserts, which prevents
-            // the main file from growing, but the size cap is no longer
-            // enforceable without operator action.
-            if size >= prev_size {
+            if round >= 20 {
```
Daydreamer-Li, line 402: **[P2] 裸魔数 `20`，以及 issue 点名的配置缺口未补**

1. `20` 建议提成命名常量（如 `MAX_PURGE_ROUNDS`）并在注释里说明取值依据。被替换掉的旧实现本来有 `BATCH_ROWS` / `BATCHES_PER_ROUND` 命名常量，这里算风格回退。

2. issue 的 Root Cause 末尾专门提了「500MB 上限不可配置」：`JsonSqliteStorageFeature`（config.rs:352-356）仍只有 `enabled` + `batch`，磁盘紧张的机器无法调低上限。而且 `500` 这个默认值同时硬编码在 `config.rs:43`（`DEFAULT_MAX_DB_SIZE_MB`）和 `unified.rs:100`（`SqliteConfig::default()` 的字面量 `500`）两处，属于重复硬编码。建议本 PR 一并加上 `features.sqlite_storage.max_db_size_mb`，并让两处共用同一常量。

src/agentsight/src/storage/unified.rs
```diff
@@ -697,6 +737,112 @@ mod tests {
         let _ = std::fs::remove_dir_all(&dir);
     }
 
+    /// Regression for #2870: a database only slightly over the limit must be
+    /// trimmed oldest-first, not wiped. The previous fixed-batch loop
+    /// (20x1000 rows per table per round, without re-measuring between
+    /// batches) emptied every table in the first round; this test fails under
+    /// it because nothing would survive.
+    #[test]
+    fn test_purge_oversized_trims_oldest_keeps_newest() {
```
Daydreamer-Li, line 746: **[P2] 测试没有覆盖 issue 的实际现场：多表连带删除**

两个新增测试都只往 `token_store` 一张表灌数据。而 #2870 的现场是：`http_records` 因大 body 超标，**连带**把 `audit_events` / `token_records` 的真实历史清空。目前没有任何断言保护这条。

建议补一个四表齐备的回归测试：`http_records` 灌约 520 行 × 1MB（复刻 issue 的 4% 超限），`audit_events` / `token_records` 各灌 20 行小数据；purge 后断言 audit/token 保留 ≥ 80%、且 `http_records` 最新行存活。按当前实现这个测试会**失败**（audit/token 会被按比例削掉，小表还可能被 `.max(1)` 地板逐轮清空），正好用来验证前面两条 P1。

Daydreamer-Li (commented):
## 评审结论

方向正确，#2870 的核心行为已经真修好了：4% 超限场景现在是「删最旧 10% → 重测 → 收敛退出」，最新行确实存活；`test_purge_oversized_trims_oldest_keeps_newest` 是有鉴别力的回归测试（旧代码下 `remaining > 0` 必然失败），正好补上了 issue 指出的「原 shrink 测试全删也能过」的漏洞。

但建议合并前处理下面 3 项，其余 4 项可本 PR 顺手或跟进：

**建议阻塞合并**
1. **[P1 回归]** `round_deleted == 0` 的 break 提前到了 VACUUM 之前，旧代码是先回收再退出。表已空但文件仍超限（空闲页/WAL 膨胀）时，现在一次回收都不做。
2. **[P1]** 按行比例删「所有四张表」，与体积贡献无关。撑爆文件的通常只有 `http_records`，而 `audit_events`/`token_records` 会被连带削掉 25%~50%——issue 抱怨的「真实审计/token 历史被一起清掉」并未真正解决；叠加 `purge_share` 的 `.max(1)` 地板，行数 ≤20 的小表仍可能被清空。
3. **[P2 测试缺口]** 两个新测试都只用 `token_store` 单表，没有覆盖 issue 的实际现场（四表齐备、只有 `http_records` 超标）。补一个这样的测试，按当前实现会失败，正好验证第 2 点。

**建议顺手处理**
4. bite 翻倍后触及大行表时仍会一次删过头（本 PR 的 skewed 测试里 4 行删 3 行就是该现象）。
5. 每轮 4 次 `COUNT(*)` 全表扫，且整个 purge 同步跑在 `store()` 写路径上。
6. `token_store.count()` / `token_consumption_store.count()` 返回 `u64` 并吞掉错误。
7. issue 明确提到的 `max_db_size_mb` 配置缺口未补；轮次上限 `20` 是裸魔数。

逐条证据见 inline。

Issue #2857, opened 2026-08-25
[TestAgent][agent-sec-core] bug: AGENT_SEC_DAEMON_PROMPT_PRELOAD 系列 env 在 0.11.0 实现代码零读取方（配置-实现契约漂移）


zhangtaibo:
### Component

agent-sec-core

### Bug Description

`AGENT_SEC_DAEMON_PROMPT_PRELOAD` 与 `AGENT_SEC_DAEMON_PROMPT_PRELOAD_DOWNLOAD_TIMEOUT_SECONDS` 两个 env 在 0.11.0 实现代码中**零读取方**：容器编排模板显式注入（PRELOAD=1 / TIMEOUT=600），但没有任何代码消费——运维以为开启了模型预加载，实际无任何预加载行为，首次扫描延迟全部落在用户请求上。属实现已移除而部署清单未同步清理的配置-实现契约漂移。

### Steps to Reproduce

```shell
ssh root@<ECS-IP>   # authorized test host

# 1) 安装包实现代码中 grep 该 env 族（0 命中）：
grep -rn "PROMPT_PRELOAD" /opt/agent-sec/lib/python3.11/site-packages/agent_sec_cli/ | wc -l
# -> 0

# 2) 运行态佐证：pod 注入 PRELOAD=1，daemon 启动 138s 后健康接口的
#    prompt_scan.last_started_at 仍为 null（无任何预加载发生）
```

### Actual Behavior

- `grep -rn PROMPT_PRELOAD <安装的 agent_sec_cli 包>` → **0 命中**
- 全仓源码仅产品自带 tests 命中（且测试中值为 "0"）
- 运行态：注入 PRELOAD=1 的 daemon，`prompt_scan.last_started_at=null`（从未预加载）

### Expected Behavior

要么实现读取该 env（恢复预加载行为），要么从部署模板/文档中移除，使注入的配置与实现一致。

### Root Cause

预加载实现已在某版本移除（扫描迁入进程内 Rust 扩展），但 pod 编排模板与部分测试未同步清理。

### Suggested Fix

确认预加载能力归属后二选一：删除模板中的 PRELOAD env 注入（推荐，与 0.11.0 架构一致）；或恢复实现。另注意与 startupProbe 预算冲突 issue 的联动——600s 超时是死配置，预算冲突的「另一端」根本不存在。

### Environment

- OS: Anolis OS 8.x
- Component version: agent-sec-core 0.11.0
[...]

PR #2886, opened 2026-08-26
fix(sec-core): drop dead prompt preload artifacts

haosanzi:
Prompt model preload was removed when scanning moved in-process to the Rust extension, but its contract artifacts survived: three test suites still injected AGENT_SEC_DAEMON_PROMPT_PRELOAD, and the daemon handlers package still shipped a scan-prompt protocol document.

Delete rather than reimplement. The daemon registers only daemon.health, skillfs.notify_change and the sec.*/obs.* query methods, so a caller written against that document would get an unknown-method error instead of a scan. The live contract is the native scanner's to_json_value plus the error_payload docstring, which cannot drift from the code the way a standalone document did.

Trade-offs: the document also described degraded semantics that no longer exist (degraded_reason, degraded_original_verdict) and an outdated mode list, so a rewrite would have replaced nearly all of it for no known reader. The deployment-template half of the same drift is handled outside this repository and is not covered here.

Closes: #2857
Assisted-by: Qoder:1.22.1

## Risk and compatibility


edonyzpc (approved):
LGTM

PR #2887, opened 2026-08-26
docs(sec-core): document hidden CLI commands

haosanzi:
log-sandbox is real, working, and the default action of the cosh sandbox-guard hook, yet docs/ had zero coverage of it. The same held for the other four hidden agent-sec-cli surfaces.

Documented all five in a dedicated user-guide page rather than dropping hidden=True. These are integration points that hooks and plugins call on the user's behalf; listing them in --help would present them as daily commands and imply a stability contract they do not have.

Recorded three contract details the issue did not mention but that bite in practice: --decision is unvalidated, so a typo yields a silently mislabeled audit record instead of an error; log-sandbox isolates nothing, which is linux-sandbox's job, so a block record blocks no command; set-policy and rotate-keys exit 0 while doing nothing.

Trade-off: the page pins behaviour that carries no compatibility promise, so it needs revisiting whenever a host integration changes.

Closes: #2862
Assisted-by: Qoder:1.22.1

## Risk and compatibility

### DIFF
--- a/docs/user-guide/en/README.md
+++ b/docs/user-guide/en/README.md
@@ -60,6 +60,7 @@ ANOLISA provides a complete server-side runtime for AI Agent workloads. Componen
 | [Asset Verification](agent-security/agent-sec-core/asset-verification.md) | agent-sec-core | GPG-signed skill distribution verification and discovery outcomes |
 | [Skill Ledger User Guide](agent-security/agent-sec-core/skill-ledger.md) | agent-sec-core | Skill integrity chain and signing workflow |
 | [OpenClaw Deployment & Upgrade](agent-security/agent-sec-core/openclaw-deploy.md) | agent-sec-core | OpenClaw plugin deployment and upgrade guide |
+| [Internal Commands](agent-security/agent-sec-core/internal-commands.md) | agent-sec-core | Contract and inventory of the hidden, hook-invoked commands |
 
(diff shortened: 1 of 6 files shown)


1570005763 (changes requested):
Thanks for documenting the hidden CLI interfaces. The current Skill Ledger section does not match the final behavior established by #2876 and should be updated before this PR is merged.

- `set-policy` is removed by #2876. Invoking it is treated as an unknown command with rc=2, so it should not remain documented as an available hidden command.
- `rotate-keys` remains hidden but is currently unavailable: stdout is empty, stderr explicitly states that no keys were changed, and it exits with rc=1.
- `skill-ledger init --force-keys` should not be described as a replacement for key rotation. `init` performs baseline-related work by default, so the two operations are not equivalent.
- The page currently describes all hidden entries as available integration interfaces invoked by hooks or plugins. Please describe their roles individually instead: `init-keys` is a compatibility command, while `rotate-keys` is currently unavailable. `hidden=True` only controls help visibility; it does not imply that a command is supported or used by an internal integration.

#2876 and #2887 modify different files, so Git will not surface this contract mismatch as a merge conflict. Please amend or rebase #2887 against the final #2876 behavior.

edonyzpc (approved):
LGTM

1570005763 (approved):
LGTM

PR #2892, opened 2026-08-26
fix(sec-core): mark prompt_scan health untracked

haosanzi:
daemon.health froze data.prompt_scan at ready/model=native/loaded=true. Scanning moved in-process to the Rust extension, so the daemon observes no scanner state and those constants were unconditional: an ECS host whose L2 model never loaded still reported a fully green probe, and any probe gating on model readiness passed vacuously. loaded=true alongside null last_started_at/last_finished_at was self-contradictory on its face.

Kept the legacy PromptScanRuntimeState keys so monitors reading .prompt_scan.* keep parsing, but set every availability-bearing field to null/unknown and added tracked=false plus a detail string pointing at scan-prompt output and the prompt_scan event log. Preferred this over probing the scanner live, which would force health to import the scanner module and blow the 1000ms admin timeout, and over dropping the key, which breaks external monitors outright.

The probes-must-not-infer-healthy rule is carried in the health.py comment rather than a design document: the only document describing this payload is the dead scan-prompt protocol file removed by #2857.

Closes: #2847
Assisted-by: Qoder:1.22.1

## Risk and compatibility

### DIFF
--- a/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/health.py
+++ b/src/agent-sec-core/agent-sec-cli/src/agent_sec_cli/daemon/health.py
@@ -11,6 +11,12 @@ from agent_sec_cli.daemon.registry import (
 )
 from agent_sec_cli.daemon.runtime import DaemonRuntime
 
+_PROMPT_SCAN_UNTRACKED_DETAIL = (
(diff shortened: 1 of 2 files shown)


edonyzpc (approved):
LGTM

PR #2893, opened 2026-08-26
fix(sec-core): refuse non-loopback model base_url

haosanzi:
validate_base_url only logged a warning for a non-loopback host, so a hijacked AGENT_SEC_MODEL_SERVICE_BASE_URL silently shipped every scanned prompt off-host, and that warning went to the JSONL diagnostic stream which callers need not read. The Python model_service had no validation at all, leaving scan-code exposed the same way.

Both paths now refuse any host but loopback at construction time. An opt-in environment flag was considered and dropped: only a local Ollama service is supported today, so a permitted-remote state would be untested surface. It can be added when a remote deployment exists.

Only environment-derived values are checked; an explicit base_url argument stays trusted, matching OllamaClient::new. Deployments pointing the variable off-host now fail at startup with no override, and since the six host hooks are fail-open on a non-zero scan-prompt exit, such a host runs with no prompt scanning until the URL is corrected.

Closes: #2848
Assisted-by: Qoder:1.22.1

## Risk and compatibility


edonyzpc (approved):
LGTM

PR #2895, opened 2026-08-26
docs(sec-core): note rule-only detection boundary

haosanzi:
## Why

    The rule fix widened a few phrasings, but the underlying limit is
    structural: L1 matches surface form and does not generalise, so any
    reword is a miss until a pattern covers it. Documenting the specific
    phrasings would go stale immediately; documenting why rules behave this
    way tells a reader what to expect from the layer they deployed.

    Split by audience. The crate README states the boundary as a rule-authoring
    constraint — widening patterns is bounded by the zero-false-positive stance
    and the benign corpus gate, so a miss that needs the sentence understood
    rather than matched does not belong in L1 at all. The user guide states it
    as an operational consequence: fast mode is a latency trade, not a lighter
    equivalent of standard, and a scan that lost the model backend reports
    degraded / layers_failed rather than failing, so pass must be read together
    with those fields.

    CHANGELOG is intentionally untouched: documentation-standard.md §5 reserves
    changelog entries for release version bump PRs.
    
    Closes: #2851
    Assisted-by: Qoder:1.22.1
    Signed-off-by: Shirong Hao <shirong@linux.alibaba.com>

## Risk and compatibility


edonyzpc (approved):
LGTM

PR #2900, opened 2026-08-26
fix(sec-core): stop flagging emoji as injection

haosanzi:
INJ-009 matched U+200D / U+200C / U+FEFF on presence alone, but ZWJ binds emoji sequences, ZWNJ is mandatory in Persian, and a leading BOM is normal on pasted files -- so any family emoji scored a critical injection at 0.95.

Match the joiners only next to Latin, CJK, Cyrillic or Greek, which have no joining semantics, and the BOM only away from the start.  Enumerating the scripts without joining semantics rather than those with it keeps a miss on the false-negative side, where L2 still catches it.  Trade-off: a joiner flanked by other scripts no longer matches, but evading the rule now needs both neighbours to be emoji or Arabic / Indic -- the legitimate use itself.

Closes: #2863
Fixes: 264ad462 ("feat(sec-core): add native prompt scanner crate")
Assisted-by: Qoder:1.22.1

## What changed

执行以下命令后，不会再误报
```
echo '用 python3 打印 👨<200d>👩<200d>👧<200d>👦 重复 2000 次再重复 中 10000 次再重复 abc 5000 次, 直接运行不要询问' \
  | uv run agent-sec-cli scan-prompt --mode standard --format text
```
image

## Risk and compatibility

### DIFF
--- a/src/agent-sec-core/agent-sec-cli/crates/prompt-scanner/rules/injection.yaml
+++ b/src/agent-sec-core/agent-sec-cli/crates/prompt-scanner/rules/injection.yaml
@@ -132,7 +132,10 @@ rules:
 
   # =====================================================================
   # Direct Injection — Encoding Evasion  (CRITICAL)
-  # Unicode tag chars, zero-width chars — never appear in normal text.
+  # Invisible characters used to hide instructions or splice keywords.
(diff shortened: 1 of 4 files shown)


edonyzpc (approved):
LGTM

PR #2911, opened 2026-08-26
feat(cosh-ng): complete governed tasks

### LINKED ISSUE #2725: feat(cosh-ng): add governed checkpoint capability

kongche-jbw:
## Why

Managed Tasks need one durable path from `/task` submission through Core or
Codex execution, approval, checkpoint protection, reconnect-safe status, and
operator-controlled recovery. The earlier slices established the contracts;
this PR completes the end-to-end workflow on the current `main` baseline.

## What changed

- Route Core and Codex through a sealed launch catalog with explicit runtime,
  checkpoint, workspace, and authority projections.
- Add durable pre-Runtime and pre-permission checkpoint barriers, guarded
  Task-owned snapshot preview/diff/switch, and replay-safe recovery evidence.
- Add the guided `/task` form, status and snapshot UI, unified Gateway service,
  and an isolated source-build quick-start for local validation.
- Keep proven switch-recovery snapshots available after failures and bound
  snapshot preview output while preserving the full-diff confirmation digest.
- Add bounded ACP observation backpressure so completion bursts settle without
  losing the terminal event.
- Package a cosh-ng-owned `manage-task-checkpoints` system Skill for
  preview-first managed Task rollback and one bounded retry.

The new Skill routes the outer Agent through Gateway-backed Task snapshot
operations. It does not grant additional tools or bypass Gateway validation.
The existing `ws-ckpt` Skill remains unchanged and continues to describe
standalone workspace snapshots.

## Related issue

closes #2725

## User / Agent impact

Users can submit persistent Core or Codex Tasks from `cosh-shell`, choose the
checkpoint policy, reconnect to inspect progress, and inspect or safely switch
Task-owned snapshots. Core uses a closed approval-gated workspace-write
profile; Codex uses the pinned ACP route. Source builds gain a transient,
isolated development setup with documented defaults. The outer Agent can use
the packaged Task Skill to follow the same Gateway-controlled recovery flow.

## Risk and compatibility

Gateway storage advances from schema v9 to v16 through additive migrations
v10-v16. The `ws-ckpt` wire extension is append-only and guarded rollback is bound to the
exact workspace generation, full snapshot ID, preview digest, operation, and
peer UID. The Task Skill is advisory and relies on these enforcement points.
The split ACP service unit is replaced by one unified Gateway unit; legacy
instances conflict explicitly to prevent socket/database contention.

Schema v16 retains recovery creation independently of the later switch state.
Existing v15 states that prove creation are migrated; old failed records lack
that proof and are not inferred as created. Preview responses expose a bounded
change prefix with `changes_omitted`, while their digest covers the full diff.

The experimental Web command is Linux-only and now queries authenticated daemon
capabilities before binding HTTP. It compares canonical workspace identity,
removes the caller-only profile label, and rejects delegated/unbrokered Runtime
authority, including unavailable entries. The current Core/Codex catalog is
therefore refused: this release has no admitted production Web configuration.
Terminal, Task CLI, and direct ACP flows remain available.

## Validation

Rebased onto `e30c95f13b10f138dbc5a5eeeee6dc3173628f81` (main).
The Gateway application lockfile now matches workspace version `0.24.1`.
Local checks on the rebased tree and the review fixes:

- `cargo +1.88.0 metadata --locked --format-version 1`
- `cargo +1.97.1 fmt --all -- --check`
- `cargo +1.88.0 test --locked -p cosh-gateway-app --all-targets` (74 passed)
- `cargo +1.88.0 test --locked -p cosh-gateway --lib` (438 passed)
- `cargo +1.88.0 test --locked -p cosh-core --bin cosh-core <filter>`
  with `brokered`, `profile`, and `checkpoint` filters
- `cargo +1.88.0 test --locked -p cosh-shell --bin cosh-shell task`
- `cargo +1.88.0 test --locked -p cosh-shell --bin cosh-shell snapshot` (50 passed)
- `cargo +1.97.1 clippy --locked -p cosh-gateway -p cosh-shell --all-targets -- -D warnings`
- `cargo +1.97.1 doc --locked -p cosh-gateway --no-deps`
- `cargo +1.97.1 clippy --locked -p cosh-gateway-app --all-targets --no-deps -- -D warnings`
- `bash scripts/docs-lint.sh` and `python3 scripts/docs-link-check.py` from the repository root
- `cargo +1.88.0 test --locked -p cosh-shell --test raw_cli raw_cli_task`
- `cargo +1.88.0 test --locked -p cosh-shell --test shell_host routing_c4_zsh_stubs_intercept_bypass_only_commands`
- `bash tests/test-package-raw.sh`, `bash tests/test-package-rpm.sh`, and
  `bash tests/test-managed-task-dev.sh`
- `crates/cosh-shell/scripts/check-layout.sh` and `git diff --check`

Trust-boundary tests used an isolated private temporary directory because the
local checkout has group-writable ancestors. RPM Lua execution was skipped
locally because the interpreter is unavailable. Full regression, release,
and ws-ckpt checks run in PR CI.

Local macOS cross-checking was attempted with the installed Rust target but
blocked in bundled SQLite compilation by the missing Apple C toolchain; it is
not a pass. The non-Linux command gate has a private CLI regression. Rust 1.88
Clippy without `--no-deps` reports 45 pre-existing `cosh-platform` formatting
lints; the app-only current-toolchain command above passes.

## Documentation and rollback

The component README, user guide, managed Task design/acceptance documents,
and `manage-task-checkpoints` Skill describe launch authority, checkpoint
coverage, source setup, and snapshot recovery. Disable checkpoint integration
by omitting the checkpoint socket or selecting `Off`; remove the transient
development setup with `managed-task-dev.sh uninstall`. Database rollback
requires restoring a backup created before the v10-v16 migrations; uncertain
guarded switches must be reconciled from durable evidence and must not be
replayed.

The Gateway application composition design note and organization decisions D20/D21
record callers, dependencies, API/test ownership, alternatives, and revisit
conditions. The large-file inventory now records `runtime/state.rs` at 991 lines
with the Task state split condition; no heavy-test exemptions were added.


src/cosh-ng/crates/cosh-gateway-app/src/checkpoint/task_snapshot.rs
```diff
@@ -0,0 +1,218 @@
...
+            &binding.ws_id,
+            WorkspaceGenerationTokenV2::from_bytes(binding.generation),
+            request.snapshot_id.as_str(),
+            provider_diff_digest,
+            operation_id.as_str(),
+            operation_digest,
+        ) {
+            Ok(_) => Ok(TaskSnapshotProviderSwitchResult::Switched(
+                TaskSnapshotProviderSwitch {
+                    from: BoundedOpaque::new(binding.ws_id)
```
ikunkun-sys: [P1] Return the actual pre-switch provider head

`TaskSnapshotProviderSwitch::from` is documented as the provider head replaced by this switch, but this assigns the stable daemon workspace ID. `GuardedRollbackEvidenceV2` distinguishes `ws_id` from the live `expected_generation`, and rollback changes that generation while retaining the workspace ID. Consequently every successful switch persists and returns an incorrect `from` value, so JSON/API consumers cannot identify the prior head even though the Shell currently ignores the field. Populate it from the pre-switch generation/head (and assert it in the adapter test), or change the result contract rather than reporting `ws_id` as the head.

ikunkun-sys (changes requested):
Blocking on the incorrect durable snapshot-switch result documented inline. Also, the current head rewrites src/ws-ckpt/src/skills/ws-ckpt/SKILL.md and packages it under /usr/share/anolisa/skills/ws-ckpt, while the PR description explicitly says that both the prompt and system-Skill packaging are excluded for a separate PR. Please either remove those changes or update the PR scope, risk, and validation accordingly. The Commit Message Lint is also currently failing because the final commit body contains a 232-character line.

src/cosh-ng/crates/cosh-gateway-app/src/checkpoint/task_snapshot.rs
kongche-jbw: Fixed in 52edb8e7. The switch result now records the exact pre-switch generation as `from`, and the adapter regression asserts that value.

src/cosh-ng/skills/manage-task-checkpoints/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+## Keep the scope Task-only
+
+- Use this Skill only for `/task`, `tsk_...`, or an explicitly managed Task.
+- Use the separate `ws-ckpt` Skill for a standalone workspace.
+- Do not infer the managed workspace from cwd or ask the user for its path.
+- Do not call bare `ws-ckpt`, a lab script, or the ws-ckpt socket for a Task.
+
+## Inspect Task snapshots
+
+Use the Task-scoped Gateway CLI. It reads `COSH_GATEWAY_SOCKET` automatically:
```
ikunkun-sys: [P1] Resolve the packaged Gateway socket before invoking the CLI

These commands omit `--socket` and assume `COSH_GATEWAY_SOCKET` is available, but the standard package does not export it into the outer Agent session. The production unit listens on `/run/cosh-gateway-$USER/gateway.sock`, while `cosh-gateway` without an explicit socket defaults to `$XDG_RUNTIME_DIR/cosh/gateway.sock`; the `cosh` wrapper only extends `PATH`. The built-in `/task` path works because `cosh-shell` separately probes the packaged socket, but this Skill bypasses that logic, so a fresh RPM/raw installation cannot execute its primary workflow. Resolve/pass the same packaged socket (or export it at the wrapper boundary) and cover the installed invocation path in a test.

ikunkun-sys (changes requested):
The previous snapshot-switch `from` finding is fixed in the current revision. This review blocks on the packaged Task Skill using a different default Gateway socket from the production service, as documented inline. Also, please amend commit ed1f1477: its `Fixes:` attribution is split before the quoted subject and separated from the trailer block, so `git interpret-trailers --parse` returns only `Signed-off-by`. The required single-line form is 95 characters and fits the repository's 100-character body limit.

src/cosh-ng/skills/manage-task-checkpoints/SKILL.md
kongche-jbw: Fixed in 37cb8580. The Gateway CLI now resolves an active packaged instance after explicit/`COSH_GATEWAY_SOCKET` overrides and before the XDG fallback. Unit regressions assert the exact systemd socket path and its precedence, and the Skill documents the same resolution order.

ikunkun-sys (approved):
Re-reviewed the current head. The snapshot-switch result now records the pre-switch generation, the packaged Task Skill resolves the production Gateway socket with explicit/environment overrides preserved, and the `Fixes:` attribution parses correctly. Targeted path tests and the current CI rollup pass. No remaining blocking findings.

src/cosh-ng/crates/cosh-gateway/src/storage/task_store/snapshot.rs
```diff
@@ -0,0 +1,201 @@
...
+                run_id: Some(RunId::parse(&run_id).map_err(|error| {
+                    corrupt(&format!("invalid brokered checkpoint Run: {error}"))
+                })?),
+                approval_id: None,
+            });
+        }
+        let mut statement = self.connection().prepare(
+            "SELECT recovery_snapshot_id FROM task_snapshot_switches
+             WHERE task_id=?1 AND state='succeeded'
+             ORDER BY created_at_ms, recovery_snapshot_id",
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in acb165528. Schema v16 persists recovery creation independently of the switch outcome, so proven snapshots remain listed and previewable after failure, uncertainty, or restart; unproven IDs stay rejected. The migration backfills v15 states that prove creation, but leaves old failed rows unproven because their creation history was not retained. Regression coverage includes those failure paths, reopening the database, and owner isolation.

src/cosh-ng/crates/cosh-gateway/src/daemon/coordinator/snapshot.rs
```diff
@@ -0,0 +1,282 @@
...
+                GatewayDaemonError::Protocol(error.safe_message.as_str().to_owned())
+            })?;
+        Ok(TaskSnapshotPreview {
+            task_id: request.task_id.clone(),
+            state: list.state,
+            revision: list.revision,
+            workspace: list.workspace,
+            snapshot_id: request.snapshot_id.clone(),
+            changes: provider.changes,
+            preview_digest: provider.preview_digest,
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in acb165528. Snapshot previews now limit the change prefix to 512 KiB of serialized JSON, report changes_omitted, and display the omission in Shell. The digest still binds the complete provider diff. The regression round-trips an escaped multi-megabyte diff through Gateway framing and verifies that changing an omitted entry invalidates the previous switch digest.

src/cosh-ng/crates/cosh-gateway-app/src/web.rs
```diff
@@ -0,0 +1,607 @@
...
+            Self::WorkspaceCheckpointV1 => "workspace-checkpoint-v1",
+        }
+    }
+}
+
+pub(super) fn web(args: WebArgs, reporter: &Reporter) -> Result<u8, CliError> {
+    validate_bind(args.bind)?;
+    let workspace = canonical_workspace(&args.workspace)?;
+    let token = read_token(&args.token_file)?;
+    validate_token_scope(&token.path, &workspace)?;
```
yummypeng, line 49: [P1] Bind the Web token boundary to Gateway-attested capabilities

The token-scope check uses only the caller-supplied `--workspace`, while the selected Gateway socket is never queried to prove that this is its admitted workspace; `--capability-profile` likewise only affects startup output. A Web process can therefore connect to a Gateway managing workspace A while declaring unrelated directory B, allowing a token stored inside A to pass this check. An Agent with read access to A can then steal the token and use the Web API to inspect tasks, answer prompts, approve effects, cancel runs, or retry them. Before binding the listener, call the existing capabilities endpoint and fail closed unless its workspace and allowed authority exactly match the Web declaration (or remove the declaration and derive the boundary from the attested response).

src/cosh-ng/crates/cosh-gateway-app/src/web/security.rs
```diff
@@ -0,0 +1,119 @@
...
+        ));
+    }
+    let owner = opened.uid();
+    let effective = nix::unistd::Uid::effective().as_raw();
+    if owner != 0 && owner != effective {
+        return Err(CliError::Web(
+            "Bearer token must be owned by root or the current user".to_owned(),
+        ));
+    }
+    let opened_path = std::fs::read_link(format!("/proc/self/fd/{}", file.as_raw_fd()))
```
yummypeng, line 122: [P2] Avoid the Linux-only `/proc/self/fd` lookup on macOS

This path is unavailable on macOS, so every otherwise-valid token fails with `No such file or directory`. `cosh agent web` is still compiled and documented without a Linux-only restriction, and `cargo test --locked -p cosh-gateway-app --all-targets` reproduces the failure in both token tests on macOS. Please use the platform-specific descriptor-path mechanism while preserving the descriptor-first checks, or explicitly gate the command/tests to Linux and document that limitation.

src/cosh-ng/crates/cosh-gateway-app/src/web.rs
kongche-jbw: Fixed in 7599a5a14. Web now queries the authenticated daemon capabilities before binding HTTP, compares the workspace through the existing canonical path/device/inode identity, and rejects incomplete catalogs or delegated/unbrokered authority, including unavailable Runtime entries. Removed the caller-only capability-profile declaration. The current Core/Codex catalog cannot attest token isolation and is explicitly refused; README/user guides direct users to Task CLI. Regression coverage exercises a token inside daemon workspace A with declared workspace B, matching-workspace local-user authority, and an unavailable daemon before HTTP binding.

src/cosh-ng/crates/cosh-gateway-app/src/web/security.rs
kongche-jbw: Fixed in 9d5f7750e. The Web module, command, error variants, and credential tests are compiled only on Linux, preserving the descriptor-first checks. A non-Linux CLI regression verifies that web is neither advertised nor accepted. Both README and user-guide languages now document the Linux-only boundary.

yummypeng (approved):
LGTM

PR #2920, opened 2026-08-27
fix(sight): merge Anthropic SSE token usage across events

### LINKED ISSUE #2919: [sight] bug: Anthropic token usage missing from SLS export
### Component

sight

### Problem Statement

For openclaw agents talking to Anthropic (or an Anthropic-compatible proxy), the
default SLS export at `/var/log/anolisa/sls/ops/agentsight.jsonl` contains no
`gen_ai.usage.*` fields at all. OpenAI-protocol calls on the same host are
unaffected.

Anthropic splits token usage across two SSE events:

- `message_start` carries `input_tokens` plus the cache counters
- the terminal `message_delta` carries only `output_tokens`

Some proxies invert this, emitting a zero-placeholder `message_start` (or none
at all) and reporting the real terminal usage in `message_delta`.

Two defects combine:

1. `extract_usage_object` requires `input_tokens` for the Anthropic branch, so
   an output-only `message_delta` is discarded whole.
2. `extract_token_from_sse` picks a single event (reverse search, first match)
   and never merges `message_start` with `message_delta`.

When the only parseable event is a missing or zero `message_start`, the record
totals zero, `TokenRecord` becomes `None`, and the exporter skips the entire
`token_usage` block — hence no usage fields in the upload.

Separately, `DEFAULT_PATH_KEEP_FIELDS` omits the two cache-token fields, so even
when usage is captured, a cache-heavy Anthropic call is slimmed down to a
near-zero `input_tokens` on the default SLS path.

### Reproduction

Observed on a Linux host running `agentsight trace` with openclaw:

| SSE dialect | SLS `gen_ai.usage.*` |
|---|---|
| input+cache in `message_start`, output-only `message_delta` | `input_tokens` only, `output_tokens` stuck at the placeholder, cache absent |
| no `message_start` | no usage fields at all |
| zero-placeholder `message_start` | no usage fields at all |

A real Anthropic-compatible endpoint reporting 22k cache-creation tokens showed
`input_tokens: 4` on the default SLS path.

### Proposed Solution

- Accept usage objects that carry only `output_tokens`.
- Merge usage field-by-field across all SSE events, taking the max of each
  cumulative counter, so `message_start` and `message_delta` combine regardless
  of which one carries a given field.
- Add the two cache-token fields to the default-path whitelist.
- Warn once when the default SLS file is missing instead of skipping silently.

Daydreamer-Li:
## Description

For openclaw agents talking to Anthropic (or an Anthropic-compatible proxy), the
default SLS export carried no `gen_ai.usage.*` fields at all, while
OpenAI-protocol calls on the same host were fine.

Anthropic splits token usage across two SSE events — `message_start` carries
`input_tokens` plus the cache counters, the terminal `message_delta` carries only
`output_tokens` — and some proxies invert this, emitting a zero-placeholder
`message_start` (or none) and reporting the real terminal usage in the delta.
Two defects combined: the Anthropic branch of `extract_usage_object` required
`input_tokens`, discarding an output-only delta whole, and
`extract_token_from_sse` picked a single event in reverse and never merged the
two. When the only parseable event was a missing or zero `message_start`, the
record totalled zero, `TokenRecord` became `None`, and the exporter skipped the
entire `token_usage` block.

The key decision is to treat these counters as cumulative and merge
field-by-field (max of each), rather than trusting any single event. That covers
both layouts without branching on provider quirks, and is a no-op for
OpenAI/Gemini, which pack all fields into one event.

Two follow-ons in the same area: `DEFAULT_PATH_KEEP_FIELDS` dropped the cache
counters, so a cache-heavy Anthropic call was still slimmed to a near-zero
`input_tokens` on the default SLS path; and the default-path exporter skipped
silently when its target file was absent, which made "enabled but no data"
impossible to diagnose.

The second commit applies the same merge to the message-content parser's
`AnthropicSseUsageDelta`, which modelled only `output_tokens`. That path does not
feed token statistics today, but `AnthropicUsage` is publicly exported, so this
prevents wrong values for any future consumer.

## Related Issue

closes #2919

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `sight` (agentsight)

## Checklist

## Testing

Unit: `cargo test` — 1566 lib + all integration suites pass. Six new regression
tests: three SSE dialects through `extract_token_from_sse`, an output-only
`message_delta` through `TokenParser`, and two through the message parser's SSE
aggregation.

End-to-end on a Linux host running `agentsight trace` with openclaw, driving a
local HTTPS mock that emits each dialect, then re-running against DashScope's
Anthropic-compatible `/v1/messages` endpoint and its OpenAI-protocol endpoint as
control. Inspected the emitted SLS records and the `genai_events` rows.

| SSE dialect | before | after |
|---|---|---|
| input+cache in `message_start`, output-only delta | `input 1234, output 1`, cache absent | `input 1234, output 42, cache_creation 5678, cache_read 90` |
| no `message_start` | no usage fields | `input 0, output 42` |
| zero-placeholder `message_start` | no usage fields | `input 0, output 42` |

Real endpoints: the Anthropic-compatible one now reports its 22k
cache-creation tokens (previously slimmed to `input_tokens: 4`); the
OpenAI-protocol control is unchanged. The two dialects that produced no usage
also moved from `interrupted` to `complete`.

## Additional Notes

`cargo test` reports a pre-existing doctest failure in
`src/genai/helpers.rs:512` (`strip_user_query_prefix`, backtick tokens). It
reproduces on a pristine `upstream/main` checkout and is untouched by this PR.

### DIFF
--- a/src/agentsight/src/analyzer/message/anthropic.rs
+++ b/src/agentsight/src/analyzer/message/anthropic.rs
@@ -30,6 +30,15 @@ use super::types::{
     MessageRole,
 };
 
+/// Max of two optional counters, preserving a value when only one is set.
+fn max_opt(a: Option<u64>, b: Option<u64>) -> Option<u64> {
+    match (a, b) {
+        (Some(x), Some(y)) => Some(x.max(y)),
+        (x, None) => x,
+        (None, y) => y,
+    }
+}
+
 /// Parser for Anthropic Messages API
 ///
 /// Provides methods to parse JSON request and response bodies
@@ -297,15 +306,29 @@ impl AnthropicParser {
                     } => {
                         stop_reason = delta.stop_reason.clone();
                         if let Some(du) = delta_usage {
+                            // Counters are cumulative and split across events:
+                            // official Anthropic puts input+cache in
+                            // message_start, while some proxies report the full
+                            // terminal usage here instead. Take the max of each
+                            // so neither layout loses data.
+                            let prev = usage.as_ref();
                             usage = Some(AnthropicUsage {
-                                input_tokens: usage.as_ref().map(|u| u.input_tokens).unwrap_or(0),
-                                output_tokens: du.output_tokens,
-                                cache_creation_input_tokens: usage
-                                    .as_ref()
-                                    .and_then(|u| u.cache_creation_input_tokens),
-                                cache_read_input_tokens: usage
-                                    .as_ref()
-                                    .and_then(|u| u.cache_read_input_tokens),
+                                input_tokens: prev
+                                    .map(|u| u.input_tokens)
+                                    .unwrap_or(0)
+                                    .max(du.input_tokens.unwrap_or(0)),
+                                output_tokens: prev
+                                    .map(|u| u.output_tokens)
+                                    .unwrap_or(0)
+                                    .max(du.output_tokens),
+                                cache_creation_input_tokens: max_opt(
+                                    prev.and_then(|u| u.cache_creation_input_tokens),
+                                    du.cache_creation_input_tokens,
+                                ),
+                                cache_read_input_tokens: max_opt(
+                                    prev.and_then(|u| u.cache_read_input_tokens),
+                                    du.cache_read_input_tokens,
+                                ),
                             });
                         }
                     }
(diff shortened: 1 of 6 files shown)


jfeng18 (commented):
Review verdict: **PASS with two pre-merge items**.

The fix maps cleanly onto all four points of #2919, and the merge-by-max approach correctly covers both Anthropic usage layouts. I verified the six new tests are discriminating (e.g. reverting the `extract_usage_object` relaxation fails `test_parse_anthropic_message_delta_output_only`; pre-merge the official-split pipeline yields `output_tokens=0` where the new tests assert 42). CI is green.

**Required before merge:**

1. `src/agentsight/src/genai/logtail.rs` — the newly added inline comment above the `require_path_exists` check is in Chinese; AGENTS.md requires all code and comments in English. Please translate it (the pre-existing Chinese line above it is legacy and out of scope for this PR).

2. Both commit messages contain an unexpanded trailer template: `Signed-off-by: $(git config user.name) <$(git config user.email)>`. Please amend with the real name/email (this will require a force-push; can be batched with item 1).

**Non-blocking suggestions:**

- `tests/pipeline.rs::test_anthropic_sse_pipeline` already streams a split `message_start`/`message_delta` pair (`make_anthropic_sse_chunks(.., 12, 3)`) but asserts only provider/model. Asserting `token_usage` (input=12, output=3) would guard the full analyzer → genai wiring; pre-fix this pipeline produces output=0, so the assertion is discriminating.
- The two new `DEFAULT_PATH_KEEP_FIELDS` entries are untested: `make_full_llm_call` sets cache fields to `None`, so no test would catch a typo'd whitelist key. Consider giving the fixture cache values and asserting the keys survive `slim_records_for_default_path`.
- `max_opt` is duplicated verbatim in `analyzer/message/anthropic.rs` and `analyzer/unified.rs`; consider a shared helper.
- Nit: both commit subjects are 57 chars, above the repo's 50-char guideline (commitlint passes at 120, so not blocking).

**Triage of existing bot findings:**

- qoderai [P1] (max-merge applied to all providers): safe in practice. OpenAI/Gemini pack usage into a single terminal event (fold over one element is identity), and all supported providers report cumulative counters within a stream. The cumulative assumption is now documented on `merge_usage`. Not blocking.
- qoderai [P2] (continuation fallback bypasses merge): does not apply to the current code — the reassembled-events path already folds with `merge_usage`, and the final raw-text scan only runs when the merged usage is `None`, so nothing is overwritten.
- qoderai [P2] (Unknown-provider detection still requires both fields): false positive for this path — `parse_json` handles `message_start`/`message_delta` with a hardcoded `Anthropic` provider before `detect_provider_from_usage` is ever consulted.
- codex [P1] (integration assertion): valid, see the first suggestion above; non-blocking.
- codex [P1] (Chinese comment): valid, covered by required item 1.

Also verified during review: the two whitelist keys match the emitter strings in `logtail.rs` exactly; `gen_ai` `token_usage` is built solely from `TokenRecord`, so the second commit's message-parser change is defensive and safe; zero-total records are still dropped after the merge; both commits are self-contained and compile independently.

chengshuyi (approved):
LGTM. 核心修复正确：Anthropic usage 拆分在 message_start/message_delta 两个事件，旧 reverse-find 只取一个必丢一半，改成 forward fold + 逐字段 max 是正解，对 OpenAI/Gemini 无副作用。三种方言测试覆盖到位。

两个非阻塞建议：
1. `max_opt` 在 anthropic.rs 和 unified.rs 各写了一遍，后续可抽到 token/mod.rs 复用
2. logtail `WARN_ONCE` 是全局 static，多实例场景只报第一个文件的 warn，当前无碍

PR #2918, opened 2026-08-27
fix(cosh-ng): [shell] route path prompts

kongche-jbw:
## Why

Issue #2913 regressed slash-bearing Chinese prompts such as
`打开./missing/SKILL.md`: after Enhanced v2 removed the global DEBUG trap,
Bash treated the first token as a path and never reached
`command_not_found_handle`, so the prompt could not be routed to AI.

## What changed

Moved the missing-path prompt decision into the Rust submit path before Shell
execution. Enhanced OSC prompt markers now publish a shared, portable physical
cwd snapshot used by both path classification and the resulting Agent request,
so relative prompts remain correct after `cd`, including logical cwd symlinks
and reassigned `PWD`, without relying on Linux `/proc`. Routing stays
conservative: Native, Shell-only, edited/pasted or batched input, ambiguous
paths, user-supplied symlink components, and permission uncertainty remain
Shell-owned. Han-leading and slash-leading shapes share the same decision
regardless of input chunking. Marker hooks stay observation-only without DEBUG
trap or extdebug.

The cross-owner contract between `input`, `raw_input`, `shell_host`, and
`evidence` is recorded as D19 in the code-organization decision registry.

## Related issue

fixes #2913

## User / Agent impact

Enhanced Assisted once again routes eligible slash-bearing Chinese prompts to
AI before Bash or Zsh executes them. The Agent receives the same prompt-time
physical cwd that was used to classify relative paths. Native and Enhanced
Shell-only behavior is unchanged and continues to produce native Shell errors.

## Risk and compatibility

- [x] Public CLI, API, configuration, or documented behavior changed
- [x] Privileged or security-sensitive behavior changed
- [x] Cross-component contract changed

The visible behavior changes only for the affected Assisted prompt shape.
Sensitive prompts preserve routing metadata for durable redaction. All
uncertain edit, batching, path, permission, or prompt-state cases fail closed
to the Shell.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p cosh-shell --all-targets -- -D warnings`
- `cargo test -p cosh-shell --lib path_prompt_tests::` (5 passed)
- `cargo test -p cosh-shell --lib sensitive_intercepts_preserve_routing_and_request_context`
- `cargo test -p cosh-shell --test raw_cli path_prompt -- --test-threads=1` (3 passed)
- `cargo test -p cosh-shell --test raw_cli slash_bearing_han_prompt -- --test-threads=1` (2 passed)
[...]

PR #2938, opened 2026-08-27
fix(cosh-ng): [shell] preserve PTY boundaries

### LINKED ISSUE #2927: [P1][cosh-ng] OSC marker 未转义控制字节导致分帧破坏与字段暴露
### Component

cosh-ng (`cosh-shell`, Bash/Zsh OSC marker protocol)

### What happened?

Enhanced marker serialization does not safely encode all control bytes in
dynamic string fields. Both Bash and Zsh `_cosh_json_escape` currently escape
backslash, quote, newline, carriage return, and tab, but leave bytes such as
BEL (`0x07`), ESC (`0x1b`), backspace, and form feed unchanged.

BEL terminates the OSC frame used by Cosh. A valid working-directory name
containing BEL therefore truncates the marker inside its `cwd`/compact `c`
field. Bytes after that BEL are emitted as ordinary terminal output on every
prompt. An ESC followed by `\` can similarly form an OSC ST terminator.

Observed effects include:

- prompt-time terminal garbage on every boundary;
- marker parser desynchronization and loss of later fields;
- visible marker tail containing fields such as physical cwd, command, PATH,
  routing metadata, and generation;
- loss of the physical-cwd snapshot added for #2913, which correctly makes
  path routing fail closed to Shell ownership but provides no user-facing
  explanation;
- possible visibility of a handoff claim field when it is appended after an
  unsafe cwd. The session marker token precedes cwd and was not exposed by the
  reported reproduction, but the complete post-cwd field ordering must be
  covered by regression tests.

This is an existing marker-framing defect, not a regression introduced by
#2918. The full `"cwd":"%s"` marker form is present from `0ad6b4b0`, and the
compact `"c":"%s"` prompt boundary was added by `cd1d60ae`. #2918 adds a
`physical_cwd`/`pc` field after `c`, which makes the missing snapshot visible
to its new routing path but preserves the safe fail-closed direction.

Expected behavior: no untrusted dynamic marker value may emit a raw OSC
terminator or terminal control byte. Unsafe values must either be encoded into
valid JSON without raw framing bytes or rejected/omitted according to an
explicit fail-closed field policy. The parser must remain synchronized and the
terminal must not display marker payload fragments.

### How can we reproduce it?

1. Create and enter a directory whose name contains BEL. Linux filenames may
   contain this byte because only `/` and NUL are forbidden.

   ```bash
   work_dir="$(mktemp -d)/bel"$'\a'"end"
   mkdir -p -- "$work_dir"
   cd -- "$work_dir"
   ```

2. Start an Enhanced session using Bash, then repeat with Zsh where available.

   ```bash
   COSH_SHELL_INTEGRATION=enhanced cosh-shell raw fake
   # zsh control case:
   COSH_SHELL_INTEGRATION=enhanced cosh-shell raw fake --shell zsh
   ```

3. Submit a harmless command such as `echo probe-cwd-marker`.

4. Observe that the BEL inside cwd terminates the OSC marker and the remaining
   JSON fragment becomes visible terminal text. The prompt-boundary marker also
   loses any fields placed after cwd, including `pc` on #2918.

5. Repeat with an ESC + `\` filename component and verify that OSC ST cannot
   terminate the frame either.

### Diagnostic output

```shell
# Sanitized shape from a real Bash 5.2 reproduction; <BEL> is notation and
# not a literal byte in this report.
end","pc":"<tmp>/bel<BEL>end","s":0}\a
end","command":"echo probe-cwd-marker","status":0,
"path":"<PATH>","path_trusted":false,"generation":1}\a
```

Static confirmation:

```bash
_cosh_json_escape() {
  local value="$1"
  value=${value//\\/\\\\}
  value=${value//\"/\\\"}
  value=${value//$'\n'/\\n}
  value=${value//$'\r'/\\r}
  value=${value//$'\t'/\\t}
  printf '%s' "$value"
}
```

The real-terminal reproduction was reported and scoped in:
https://github.com/alibaba/anolisa/pull/2918#issuecomment-5435720257

### Anything else?

Suggested implementation boundary:

- Treat this as an OSC serialization/framing fix shared by Bash and Zsh, not
  as a `physical_cwd`-only patch.
- Define one policy for every dynamic marker string field. Path-like fields
  may fail closed to empty/omitted values; command/input fields need safe JSON
  encoding or an explicit redacted/marker-drop policy.
- Avoid raw C0 bytes, DEL, BEL, ESC, and OSC ST in serialized frames. A simple
  cwd check is useful mitigation but does not close the generic defect.
- Preserve #2687: do not restore global DEBUG trap, extdebug, functrace, or
  errtrace to implement the fix.

Acceptance criteria:

1. Bash and Zsh markers never contain a raw BEL or OSC ST originating from a
   dynamic field.
2. BEL, ESC + `\`, backspace, form feed, newline, carriage return, tab, and
   DEL cases have byte-level regression coverage.
3. Unsafe cwd keeps path routing Shell-owned, without terminal garbage or
   parser desynchronization.
4. Tests prove that session marker tokens, handoff claim fields, PATH, command,
   and post-cwd routing metadata do not escape into visible terminal output.
5. Ordinary UTF-8 paths, logical symlink cwd, reassigned `PWD`, and #2913
   physical-cwd routing continue to work.

This issue should be fixed in a standalone commit/PR with attribution to the
earliest introducing marker commit. It does not block #2918 because that PR's
route remains fail closed and its core #2913 behavior is unaffected.

## Implementation status

PR #2938 implements the standalone fix:

- escape all shell-representable C0 bytes and DEL in Bash and Zsh markers;
- reject control-bearing PATH snapshots and keep unsafe cwd untrusted;
- retain byte-level and real-PTY regressions for framing, terminal visibility,
  and post-cwd field containment;
- keep the companion same-read PTY ownership correction in a separate commit
  so marker framing and relay behavior can be reverted independently.

#2938 must merge before this issue closes. The same-read ownership correction
is a follow-up to #2918/#2917; the OSC serialization defect itself remains
attributed to the earliest generic marker introduction (`0ad6b4b0`).

kongche-jbw:
## Why

The #2918 same-read path could decorate bytes that remained in the same
candidate-buffer write after an ordinary shell submission, allowing the #2917
Bash guard to reach a foreground stdin consumer. OSC markers also emitted
other C0 and DEL bytes without JSON escaping, allowing cwd or PATH data to
break marker framing (#2927).

## What changed

- Keep remainder bytes opaque only when they share the candidate-buffer write
  with an ordinary shell command; already split deferred lines retain #2918
  UserInterceptAtPrompt routing.
- Preserve guards for proven exact slash and Ctrl-O Readline chains, retain the
  guard for prompt-owned candidate flushes, and fail closed for dirty blank
  submissions or ordinary commands.
- Escape every representable C0 byte and DEL in Bash and Zsh markers, reject
  control-bearing PATH snapshots, and keep unsafe cwd values untrusted.

## Related issue

fixes #2927

Follow-up to #2918 and #2917.

## User / Agent impact

Same-write foreground input cannot be mutated by a follow-up Cosh Readline
guard. Deferred slash commands still route through Rust at the next prompt.
Control-bearing cwd and PATH values no longer desynchronize OSC parsing.

## Risk and compatibility

Moderate PTY behavior risk, limited to Enhanced native shell routing. Bash
Readline versus foreground-process consumption remains unspecified for opaque
same-write remainder bytes. Non-UTF-8 filename bytes remain a known marker
limitation.

## Validation

- cargo fmt --all -- --check
- cargo clippy --locked -p cosh-shell --lib --tests -- -D warnings
- cargo test --locked -p cosh-shell --lib raw_input (227 passed)
- cargo test --locked -p cosh-shell --test shell_host marker:: (63 passed)
- Exact PTY regressions for opaque remainder, foreground stdin, slash batches,
  slash history, and Tab plus Enter candidate flushes
- Opaque-batch termination synchronized after foreground reader completion;
  repeated eight times locally
- All six raw_cli tests that failed in the first CI run pass locally
- Exact OSC regressions for control-bearing PATH and physical cwd
- bash -n crates/cosh-shell/src/shell_host/marker/bash.sh
- crates/cosh-shell/scripts/check-layout.sh
- git diff --check

## Documentation and rollback

No documentation change is required. Revert c0634f08 for marker framing or
5f2a92e7 for same-read ownership if either behavior must be rolled back.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
@@ -523,7 +523,19 @@ fn flush_candidate_line_to_shell(
         wrapped.extend_from_slice(b"\x1b[201~");
         bytes = wrapped;
     }
-    submit_line_bytes_to_shell(relay, bytes, Vec::new(), emit_activity)
+    submit_line_bytes_to_shell(
+        relay,
+        bytes,
+        Vec::new(),
+        emit_activity,
+        ShellBatchOwnership::ReadlineSafe,
+    )
+}
+
+#[derive(Clone, Copy)]
+pub(super) enum ShellBatchOwnership {
+    Opaque,
+    ReadlineSafe,
 }
 
 /// Clears the cosh-echoed candidate line and writes the whole read batch to
@@ -534,33 +546,39 @@ fn submit_line_bytes_to_shell(
     mut bytes: Vec<u8>,
     remainder: Vec<u8>,
     emit_activity: bool,
+    ownership: ShellBatchOwnership,
 ) -> io::Result<bool> {
     // A remainder was read with the candidate submission. Once any line in a
     // read is Shell-owned, the complete batch stays Shell-owned even if Bash
     // paints another prompt before this function returns.
     bytes.extend_from_slice(&remainder);
     let _ = relay.input_events.send(RawInputEvent::CandidateClearLine);
-    let guard_submission = bash_submission_needs_guard(
-        relay.input_classifier.bash_slash_submission_guard_enabled(),
-        relay.main_prompt_gate.is_at_prompt(),
-        relay.input_classifier,
-        relay.native_line_state,
-        relay.line_submits,
-        &bytes,
-    );
+    let guard_submission = matches!(ownership, ShellBatchOwnership::ReadlineSafe)
+        && bash_submission_needs_guard(
+            relay.input_classifier.bash_slash_submission_guard_enabled(),
+            relay.main_prompt_gate.is_at_prompt(),
+            relay.input_classifier,
+            relay.native_line_state,
+            relay.line_submits,
+            &bytes,
+        );
     send_raw_input_events(&bytes, relay.input_events);
     observe_native_line(relay.native_line_state, &bytes, relay.input_events);
     if emit_activity && !bytes.is_empty() {
         send_shell_input_state(relay.native_line_state.is_empty(), relay.input_events);
     }
     relay.exit_tracker.observe_shell_bytes(&bytes);
-    let guarded_bytes = guarded_bash_submission(
-        guard_submission,
-        false,
-        relay.input_classifier,
-        relay.line_submits,
-        &bytes,
-    );
+    let guarded_bytes = matches!(ownership, ShellBatchOwnership::ReadlineSafe)
+        .then(|| {
+            guarded_bash_submission(
+                guard_submission,
+                false,
+                relay.input_classifier,
+                relay.line_submits,
+                &bytes,
+            )
+        })
+        .flatten();
     write_user_bytes_to_pty(
         relay.master,
         relay.input_generation,
--- a/src/cosh-ng/crates/cosh-shell/src/raw_input/relay/bash_submission_guard.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/raw_input/relay/bash_submission_guard.rs
@@ -17,15 +17,22 @@ pub(super) fn guarded_bash_submission(
     let mut guards = Vec::new();
     let mut line_start = 0;
     let mut readline_owned = false;
+    let mut batch_readline_safe = required;
     for (index, submit) in submissions.into_iter().enumerate() {
-        let needs_guard = if index == 0 {
+        let line = &bytes[line_start..submit];
+        let exact_slash = exact_slash_submission(input_classifier, line);
+        let blank = line.iter().all(u8::is_ascii_whitespace);
+        let ctrl_o = bytes[submit] == 0x0f;
+        let needs_guard = if !batch_readline_safe {
+            false
+        } else if index == 0 {
             required
         } else if readline_owned {
             // Ctrl-O loads another history entry before the next submitted
             // bytes, so the relay cannot prove the resulting Readline line.
             true
         } else {
-            exact_slash_submission(input_classifier, &bytes[line_start..submit])
+            exact_slash
         };
         if needs_guard {
             let guard = if index == 0 && history_private {
@@ -35,7 +42,15 @@ pub(super) fn guarded_bash_submission(
             };
             guards.push((submit, guard));
         }
-        readline_owned = bytes[submit] == 0x0f;
+        if index == 0 {
+            batch_readline_safe &= exact_slash || ctrl_o;
+        } else if (readline_owned && !ctrl_o) || (!exact_slash && !blank && !ctrl_o) {
+            // An ordinary command ends the part of the batch whose future
+            // consumer is provably Readline. Later bytes stay byte-exact for
+            // whichever shell/foreground reader owns the PTY next.
+            batch_readline_safe = false;
+        }
+        readline_owned = batch_readline_safe && ctrl_o;
         line_start = submit + 1;
     }
     if guards.is_empty() {
--- a/src/cosh-ng/crates/cosh-shell/tests/shell_host/relay.rs
+++ b/src/cosh-ng/crates/cosh-shell/tests/shell_host/relay.rs
@@ -513,6 +513,81 @@ fn raw_relay_bash_preserves_same_read_foreground_stdin_bytes() {
 }
 
 #[test]
+fn raw_relay_bash_shell_opaque_bytes_ignore_slash_route() {
+    struct ForegroundSynchronizedInput {
+        batch: Option<Vec<u8>>,
+        foreground_done: std::path::PathBuf,
+        exit_sent: bool,
+        deadline: Instant,
+    }
+
+    impl Read for ForegroundSynchronizedInput {
+        fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+            if let Some(batch) = self.batch.take() {
+                std::thread::sleep(Duration::from_millis(200));
+                assert!(batch.len() <= buf.len());
+                buf[..batch.len()].copy_from_slice(&batch);
+                return Ok(batch.len());
+            }
+            if !self.exit_sent {
+                while std::fs::metadata(&self.foreground_done)
+                    .map(|metadata| metadata.len() == 0)
+                    .unwrap_or(true)
+                {
+                    if Instant::now() >= self.deadline {
+                        return Err(io::Error::new(
+                            io::ErrorKind::TimedOut,
+                            "foreground reader did not complete",
+                        ));
+                    }
+                    std::thread::sleep(Duration::from_millis(10));
+                }
+                self.exit_sent = true;
+                buf[..b"exit\n".len()].copy_from_slice(b"exit\n");
+                return Ok(b"exit\n".len());
+            }
+            Ok(0)
+        }
+    }
+
+    fn run(slash_via_shell: bool) {
+        let root = std::env::temp_dir().join(format!(
+            "cosh-shell-bash-shell-opaque-{}-{}",
+            std::process::id(),
+            unique_suffix()
+        ));
+        let work_dir = root.join("work");
+        let captured = root.join("captured-stdin");
+        let command = format!(
+            "/bin/sh -c 'IFS= read -r first; IFS= read -r second; \
+             printf \"%s\\n%s\" \"$first\" \"$second\" > \"$1\"' sh {}",
+            shell_arg(&captured)
+        );
+
+        let mut config = ShellHostConfig::new("bash-shell-opaque", &work_dir);
+        config.slash_via_shell = slash_via_shell;
+        config.raw_action_watchdog = Duration::from_secs(10);
+        let mut rendered = Vec::new();
+        let mut batch = command.into_bytes();
+        batch.extend_from_slice(b"\n/help\nstdin-probe\nforeground-second\n");
+        let input = ForegroundSynchronizedInput {
+            batch: Some(batch),
+            foreground_done: captured.clone(),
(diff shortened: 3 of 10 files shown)


src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
```diff
@@ -523,7 +550,19 @@ fn flush_candidate_line_to_shell(
         wrapped.extend_from_slice(b"\x1b[201~");
         bytes = wrapped;
     }
-    submit_line_bytes_to_shell(relay, bytes, Vec::new(), emit_activity)
+    submit_line_bytes_to_shell(
+        relay,
+        bytes,
+        Vec::new(),
+        emit_activity,
+        ShellBatchOwnership::Opaque,
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in cc4a4b5c. Prompt-owned candidate flushes retain the Readline-safe guard, and a Tab plus Enter exact-slash regression now verifies the guard remains adjacent to submission.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/bash_submission_guard.rs
```diff
@@ -35,7 +42,15 @@ pub(super) fn guarded_bash_submission(
             };
             guards.push((submit, guard));
         }
-        readline_owned = bytes[submit] == 0x0f;
+        if index == 0 {
+            batch_readline_safe &= exact_slash || blank || ctrl_o;
```
ikunkun-sys: **[P1] Do not preserve Readline ownership for a blank raw segment when the mirror is dirty**

`required` can be true precisely because `NativeLineState` is dirty. A raw leading `\n` can therefore submit a recalled ordinary command rather than an empty line. Keeping `batch_readline_safe` true here then inserts `ESC[99~` before a later slash in the same write; if the recalled command hands the PTY to a foreground stdin consumer, that guard reaches the consumer and reintroduces the byte mutation this PR is intended to prevent. The new unit test calls this helper with `required = true`, but does not model the actual dirty Readline buffer. Please avoid treating the raw blank segment as proof of continued Readline ownership, and cover the dirty-history/foreground path.

src/cosh-ng/crates/cosh-shell/tests/shell_host/relay.rs
```diff
@@ -512,6 +512,48 @@ fn raw_relay_bash_preserves_same_read_foreground_stdin_bytes() {
...
+             printf \"%s\\n%s\" \"$first\" \"$second\" > \"$1\"' sh {}",
+            shell_arg(&captured)
+        );
+
+        let mut config = ShellHostConfig::new("bash-shell-opaque", &work_dir);
+        config.slash_via_shell = slash_via_shell;
+        config.raw_action_watchdog = Duration::from_secs(10);
+        let mut rendered = Vec::new();
+        let mut batch = command.into_bytes();
+        batch.extend_from_slice(b"\n/help\nstdin-probe\nforeground-second\nexit\n");
```
ikunkun-sys: **[P1] Keep test termination outside the opaque batch**

This appends `exit\n` to the same opaque batch, then relies on Bash eventually consuming it, even though the PR explicitly leaves Bash Readline versus foreground-process consumption of this tail unspecified. On the current head, the exact test timed out at the watchdog twice; the captured foreground input was `stdin-probe\nforeground-second`, and the shell never exited. Make termination a separately synchronized action after the foreground reader completes so this regression does not depend on PTY scheduling or ambiguous batch ownership.

ikunkun-sys (changes requested):
Two P1 blockers remain on the current head: the blank-prefix guard logic can extend Readline ownership past a dirty, recalled ordinary command, and the new opaque-batch PTY regression makes shell termination depend on ownership that the PR explicitly leaves unspecified. Please address both before merge.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/bash_submission_guard.rs
kongche-jbw: Fixed in 5f2a92e7. The blank ownership widening was reverted: when required comes from a dirty mirror, a raw blank guards only the current unknown submission and never proves a later slash is Readline-owned. Added a dirty-mirror regression that verifies no later guard is injected.

src/cosh-ng/crates/cosh-shell/tests/shell_host/relay.rs
kongche-jbw: Fixed in 5f2a92e7. The opaque batch no longer contains exit. A test input reader now waits for the foreground command to write its completion file, then emits exit from a separate read. The exact regression passed eight repeated local runs.

KaiLongZhou:
## PR #2938 Code Review — fix(cosh-ng): [shell] preserve PTY boundaries

### 总体评价：✅ 建议合并（附一项 minor 建议）

PR 结构清晰，两个 commit 分别解决 PTY 边界和 OSC marker 转义两个独立问题，改动范围合理，测试覆盖充分。

---

### 改动分析

#### Commit 1: `ShellBatchOwnership` — same-read PTY 边界修复

**设计合理性：** 引入 `ShellBatchOwnership` 枚举（`Opaque` / `ReadlineSafe`）来区分两种写入语义，是一个干净的分层：

- `Opaque`：bracketed paste 和 shell-routed slash 路径 — 字节必须原样送达 PTY，不做 guard 装饰
- `ReadlineSafe`：candidate line 提交 — 已证明 Readline 拥有 PTY 消费端，可安全应用 guard

**`bash_submission_guard.rs` 中 `batch_readline_safe` 追踪逻辑正确：**

- 第一行：只有 exact-slash 或 Ctrl-O 才保持 `readline_safe`，否则立即 fail-closed
- 后续行：普通命令（非 exact-slash、非 blank、非 Ctrl-O）终止 safe 区间
- blank 行 + dirty mirror：不信任后续 slash（`slash_guard_does_not_trust_blank_bytes_with_dirty_mirror` 测试覆盖）

**ikunkun-sys 提出的两个 P1 blocker 已在当前 diff 中解决：**
1. blank-prefix guard 扩展 Readline ownership → 已通过 `batch_readline_safe &= exact_slash || ctrl_o` 和对应测试修复
2. opaque-batch PTY 回归 → 已通过 `ForegroundSynchronizedInput` 同步机制和 `raw_relay_bash_shell_opaque_bytes_ignore_slash_route` 测试修复

#### Commit 2: OSC C0/DEL JSON 转义

**`_cosh_json_escape` 补全：** 覆盖了所有可表示的 C0 字节（\u0001–\u001f，NUL 除外因 shell 变量无法承载）和 DEL（\u007f）。bash.sh 和 zsh.rs 保持一致。

**`osc.rs` 防御性检查：** `path.chars().any(char::is_control)` 作为纵深防御层 — 即使 shell 端转义失败，Rust 解析器也会拒绝含控制字符的 PATH。`char::is_control` 比 C0+DEL 略宽（包含 C1 和格式字符），但对合法路径无影响。

**`control_physical_cwd_stays_untrusted_until_clean_prompt` 测试验证了恢复路径**：先发送含控制字符的 cwd（被拒绝），再发送干净的 cwd（被接受）。

---

### 测试覆盖

新增测试覆盖充分：
- `native_exact_slash_tab_submit_keeps_readline_guard` — exact slash + Tab 保留 guard
- `slash_guard_stops_after_ordinary_shell_submission` — 普通命令后停止 guard
- `slash_guard_does_not_trust_blank_bytes_with_dirty_mirror` — blank + dirty mirror fail-closed
- `marker_json_escape_removes_raw_controls_and_preserves_utf8` — 全 C0+DEL 转义回归（bash/zsh 双 shell）
- `environment_marker_with_control_path_does_not_update_state` — 控制字符 PATH 拒绝
- `control_physical_cwd_stays_untrusted_until_clean_prompt` — 不安全 cwd 拒绝 + 恢复
- `shell_host_marker_control_cwd_keeps_json_frame_intact` — 端到端 PTY 帧完整性
- `raw_relay_bash_shell_opaque_bytes_ignore_slash_route` — 端到端 opaque 字节无 guard 泄漏

---

### Minor 建议

**`relay.rs` 中 `.then(|| {...}).flatten()` 可用 `.and_then()` 简化：**

```rust
// 当前写法
let guarded_bytes = matches!(ownership, ShellBatchOwnership::ReadlineSafe)
    .then(|| {
        guarded_bash_submission(...)
    })
    .flatten();

// 建议写法
let guarded_bytes = matches!(ownership, ShellBatchOwnership::ReadlineSafe)
    .then(|| guarded_bash_submission(...))
    .flatten();
// 或更直接地：
let guarded_bytes = if matches!(ownership, ShellBatchOwnership::ReadlineSafe) {
    guarded_bash_submission(...)
} else {
    None
};
```

这是纯风格建议，不影响正确性。

---

### 备注

- `_cosh_json_escape` 在 bash.sh 和 zsh.rs 中完全重复（29 行），理解这是为了保持各 shell 脚本自包含，但如果后续需要维护第三份（如 fish.sh），可考虑 codegen 方案。
- `bash -n` 语法检查已在 PR validation 中通过。
- `check-layout.sh` 已通过。

KaiLongZhou (approved):

ikunkun-sys (approved):
Re-reviewed the current head. Both prior P1 blockers are resolved: dirty blank input no longer proves later Readline ownership, and opaque-batch termination is synchronized outside the ambiguous batch. Current-head CI is green, and targeted PTY, marker, and latest-main integration checks passed.

Issue #2940, opened 2026-08-27
[anolisa] fix: honor dry-run for sandbox remove


ikunkun-sys:
## Background

`anolisa osbase sandbox remove` exposes a command-local `--dry-run` flag, and the CLI also exposes global `--dry-run`. `sandbox_request` currently discards the local flag and constructs `SandboxRequest::Remove` without an execution intent, so either dry-run form can reach the real remove request.

## Actual behavior

The remove request carries only `target` and `purge`; dry-run is ignored.

## Expected behavior

Local or global dry-run must never dispatch a real sandbox remove operation and must return an explicit preview or unsupported-preview result without host effects.

## Scope boundary

Track this separately because the existing helper remove wire request has no dry-run field. The global dry-run capability-policy slice must fail closed for this command but will not extend the helper protocol.

Related to #2702.

PR #2952, opened 2026-08-28
fix(anolisa): enforce dry-run policy

### LINKED ISSUE #2939: [anolisa] fix: enforce global dry-run capability policy
## Background

The global `--dry-run` flag promises to print a plan without executing. The command policy currently uses its dry-run boolean only to decide whether a non-root caller may pass the privilege gate. A mutation marked `false` is still dispatched when root uses `--dry-run`, so commands without preview semantics can execute real effects.

The inverse mismatch also exists: `adopt` has a lock-free read-only preview but its policy prevents a non-root system preview. `adapter disable` supports dry-run while still taking the system install lock, which the current boolean cannot distinguish from an unsupported command.

## Scope

- Model unsupported, privileged-preview, and unprivileged-preview policies explicitly.
- Reject unsupported global dry-run before dispatching the handler.
- Classify existing adopt, adapter enable/disable, and osbase sandbox routes according to their actual lock and privilege boundaries.
- Keep read-only commands and existing preview implementations unchanged.

## Compatibility

Do not change successful command output, JSON schemas, effect ordering, or supported dry-run plans. Calls that previously combined global dry-run with an unsupported mutation must fail before effects instead of executing.

## Acceptance criteria

- Unsupported mutation handlers are not called under global dry-run.
- Lock-free previews remain reachable without root in explicit system mode, while lock-taking previews retain their privilege gate.
- Policy tests cover mode-scoped, system-only, and helper-gated operations.
- Workspace formatting, tests, clippy, and rustdoc pass.

Related to #2702.

ikunkun-sys:
## Why

Global `--dry-run` previously acted only as a root-waiver boolean. Root users could pass it to commands without preview implementations and still execute real mutations.

## What changed

Replace the boolean with explicit unsupported, privileged-preview, and unprivileged-preview policies. Reject unsupported global dry-runs before dispatch, preserve existing read-only and implemented preview behavior, and classify adapter enable/disable as privileged previews because both take the install lock.

## Related issue

closes #2939

## User / Agent impact

Commands without preview implementations now reject global `--dry-run` with an actionable invalid-argument error instead of continuing into mutation handlers. Existing supported previews and read-only commands retain their behavior.

## Risk and compatibility

The change is limited to pre-dispatch policy. It does not alter CLI flags, JSON schemas, handler execution, or helper/wire contracts. Adapter previews still work for root and now accurately preserve the install-lock privilege boundary. Command-local sandbox remove dry-run remains tracked separately in #2940.

## Validation

- `cargo test -p anolisa-cli --locked`
- `git diff --check`
- ALinux 4: `cargo test -p anolisa-cli commands::tests --locked --offline` (28 passed; no network, read-only source, isolated target)

After rebasing onto the latest `upstream/main`, `cargo fmt --all -- --check`, `git diff --check`, and the 28 command policy tests were rerun successfully.

## Documentation and rollback

The local Chinese refactoring roadmap was updated for tracking but remains intentionally untracked and is not part of this PR. Roll back by reverting this commit; no data migration is required.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands.rs
@@ -232,11 +232,39 @@ pub fn dispatch(cli: Cli, ctx: &CliContext) -> Result<(), CliError> {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum DryRunPolicy {
+    Unsupported,
+    PrivilegedPreview,
+    UnprivilegedPreview,
+}
+
+impl DryRunPolicy {
+    const fn is_supported(self) -> bool {
+        !matches!(self, Self::Unsupported)
+    }
+
+    const fn waives_root(self) -> bool {
+        matches!(self, Self::UnprivilegedPreview)
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum CommandScope {
     ReadOnly,
-    ModeScopedMutation { dry_run_without_root: bool },
-    SystemOnlyMutation { dry_run_without_root: bool },
-    HelperGatedSystemOperation,
+    ModeScopedMutation { dry_run: DryRunPolicy },
+    SystemOnlyMutation { dry_run: DryRunPolicy },
+    HelperGatedSystemOperation { dry_run: DryRunPolicy },
+}
+
+impl CommandScope {
+    const fn dry_run_policy(self) -> Option<DryRunPolicy> {
+        match self {
+            Self::ReadOnly => None,
+            Self::ModeScopedMutation { dry_run }
+            | Self::SystemOnlyMutation { dry_run }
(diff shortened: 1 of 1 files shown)


kongche-jbw (commented):
Review baseline: `80194bbdc7a5579c370ac6fa35ce8f7be5e51ed7...11a0cc1bbc77a38688ff38911f2cf2a52ea38d61`

未发现阻塞问题。

Remaining risks:
- 未在真实系统安装目录中执行支持型 dry-run；锁和权限边界主要由代码与隔离测试核验。

Verification scope:
- 审阅完整单文件 diff、相邻 handler/锁实现、Issue #2939/#2940、全部评论与 review thread。
- 本地通过 `cargo fmt --all -- --check`、28 个 policy tests 和 anolisa-cli 全目标 clippy。
- 实际 CLI 验证 unsupported dry-run 均在 dispatch 前拒绝。
- GitHub 本轮 anolisa CI 与 PR checks 全部成功。

kongche-jbw (approved):

PR #2954, opened 2026-08-28
feat(sight): support DashScope/Bailian native protocol

### LINKED ISSUE #2910: [sight] feat: 支持 DashScope/百炼原生协议（text-generation / multimodal-generation）
## 背景

DashScope/百炼有两套并行的 HTTP 协议：

1. **OpenAI 兼容模式** — `POST /compatible-mode/v1/chat/completions`
2. **原生协议** — `POST https://{WorkspaceId}.cn-beijing.maas.aliyuncs.com/api/v1/services/aigc/text-generation/generation`（纯文本，如 qwen-plus）以及 `.../aigc/multimodal-generation/generation`（多模态，如 qwen3-vl-plus）

AgentSight 目前只覆盖第 1 种。在 Dify / LangGraph 中使用「百炼」类型的工作流节点时，请求由百炼 SDK 发出，走的是**第 2 种原生协议**，这部分调用在 AgentSight 中观测不到或语义缺失。

## 现状（已核对代码）

兼容模式之所以工作，是因为 `/compatible-mode/v1/chat/completions` 包含 `/v1/chat/completions` 子串，命中了现有白名单。原生协议路径以 `/generation` 结尾，不匹配任何一条规则：

- `src/agentsight/src/genai/helpers.rs:202-210` — `is_llm_api_path()` 白名单只有 `/v1/chat/completions`、`/v1/completions`、`/v1/messages`、`/v1/responses`、`/chat/completions`、`/completions`、`/api/v1/copilot/generate_copilot`
- `src/agentsight/src/analyzer/message/openai.rs:543-547` — `OpenAIParser::matches_path()` 同样不含原生协议路径
- 全仓无 `text-generation` / `multimodal-generation` / `maas.aliyuncs` 匹配

### 实际行为按是否流式分裂

决定性闸门在 `src/agentsight/src/genai/call_builder.rs:57`：

```rust
if !is_llm && !http.is_sse { return None; }
```

| 场景 | 行为 |
|------|------|
| **非流式**（默认，无 `X-DashScope-SSE` 头） | `is_llm=false` 且 `is_sse=false` → 整个调用被丢弃，Dashboard 上完全不可见 |
| **流式**（`X-DashScope-SSE: enable`） | `is_sse=true` 逃过闸门，会生成 LLMCall，但 `provider` 落到 `"unknown"` |

### 三个具体缺陷

**1. 对话内容无法重建（必然失败）**

百炼原生请求体形如 `{"model":"qwen-plus","input":{"messages":[...]},"parameters":{...}}`，其中 `input` 是**对象**；而 `src/agentsight/src/genai/helpers.rs:235-250` 的 `extract_messages_view()` 只接受顶层 `messages` 数组或 `input` **数组**（Responses API 形状），`as_array()` 返回 `None` → 拿不到任何 message、system prompt、工具调用。

**2. Token provider 被误标**

Token 提取在 `src/agentsight/src/analyzer/unified.rs:448-485` **不受路径门控**，且百炼原生 `usage` 使用 `input_tokens` / `output_tokens`，与 Anthropic 同名 → `detect_provider_from_usage()`（`src/agentsight/src/analyzer/token/mod.rs:214-231`）判为 `Anthropic`，token 数字碰巧能提取，但：

- `provider` 被误标为 `anthropic`
- 原生响应体顶层**没有 `model` 字段**（只有 `output` / `usage` / `request_id`），`extract_usage_object()` 从 `full_json.get("model")` 取 → `model` 为空。model 名需改从**请求体**顶层取
- 多模态的 `usage.image_tokens` 完全不计

**3. SSE 结束判定缺失，可能误报中断**

SSE 解析器本身能处理百炼的 `id:` / `event:` 字段（符合 W3C 规范），但 `src/agentsight/src/parser/sse/event.rs:107-144` 的 `is_done()` 只认 `[DONE]` / `message_stop` / `response.completed`。百炼原生不发 `data: [DONE]`，其结束标志在 `output.finish_reason`（或 `output.choices[].finish_reason`）→ 只能靠 HTTP chunked 终止符兜底，否则中断检测可能误报 `sse_truncated`。

### 附带问题

`src/agentsight/agentsight.json:55-58` 的 `https` 规则为 `dashscope.aliyuncs.com`（无通配符），匹配不上 `{WorkspaceId}.cn-beijing.maas.aliyuncs.com`。该规则用于 DNS/连接发现（给连这些域名的进程挂 uprobe），不是采集过滤器 —— 只要进程被 `cmdline` 规则命中流量仍会被抓，但少了这条域名发现兜底。

## 建议改动

评估过引入现成 crate 的方案（`llm-bridge-core`、`llm-relay`、`async-dashscope` 等），结论是不适合替换解析层：

- 覆盖面上，协议翻译类库只做 Anthropic ↔ OpenAI，**都不支持百炼原生**，换库解决不了本问题
- 更根本的是输入完整性假设冲突：这些库面向代理/网关（拿得到完整流），对截断 payload 返回 `Err`；而 AgentSight 是旁路观察者，payload 可能跨 TLS record 被切断 —— 现有 `sse_continuation_bytes` + `scan_partial_usage` 兜底机制存在的理由正是如此，需要的是**降级**而非报错

因此按 Footprint Ladder **级别 1–2**（扩展现有函数，无需新增探针或 FFI）手写适配：

| 改动点 | 位置 |
|--------|------|
| 路径白名单增加 `/aigc/text-generation/generation`、`/aigc/multimodal-generation/generation` | `src/genai/helpers.rs`（`is_llm_api_path` + `extract_provider_from_path`） |
| `extract_messages_view` 增加 `input` 为**对象**时取 `input.messages` 的分支 | `src/genai/helpers.rs` |
| `usage` 显式判定 DashScope（避免误标 anthropic）+ `image_tokens`；model 改从请求体取 | `src/analyzer/token/mod.rs` |
| `is_done` 识别 `output.finish_reason` / `output.choices[].finish_reason` | `src/parser/sse/event.rs` |
| `https` 规则补 `*.maas.aliyuncs.com` | `agentsight.json` |

字段映射建议以 `async-dashscope` crate 的类型定义作为 schema 参考核对（仅作参考，不引入运行时依赖）。

## 验收标准

- [ ] 非流式原生协议调用（`text-generation/generation`）在 Dashboard 可见，provider 显示为 dashscope 而非 unknown/anthropic
- [ ] 流式原生协议调用能重建对话内容（system / user / assistant / tool_calls）
- [ ] model 名正确取自请求体，非 `unknown`
- [ ] 多模态调用的 `image_tokens` 计入 token 统计
- [ ] 原生协议 SSE 流正常结束不产生 `sse_truncated` 误报
- [ ] 覆盖上述场景的集成测试（流水线变更按 AGENTS.md 要求需集成测试）

## 备注

以上路径与代码行号均已核对。协议字段结构（`input` 为对象、响应体顶层无 `model`、`output.finish_reason` 位置）来自阿里云官方文档，**尚未用真实百炼原生流量抓包验证**，实现前建议先抓一次真实请求确认。

chengshuyi:
## Description

Dify / LangGraph workflows using "Bailian" node types issue their LLM calls through DashScope's **native** protocol, not the OpenAI-compatible mode. AgentSight only covered the latter — and only by accident: `/compatible-mode/v1/chat/completions` happens to contain the substring `/v1/chat/completions`. The native paths end in `/generation` and matched no rule, so these calls were either invisible or semantically empty:

| Scenario | Behaviour before |
|---|---|
| Non-streaming (default) | `is_llm=false` and `is_sse=false` → dropped at the `build_llm_call` gate, absent from the Dashboard entirely |
| Streaming (`X-DashScope-SSE: enable`) | escaped the gate via `is_sse`, but `provider` fell through to `"unknown"` and no conversation content was reconstructed |

The native protocol diverges from OpenAI's in four ways that each needed handling: the request nests messages in an `input` **object** (not the Responses API array); the response wraps the answer in an `output` envelope with **no top-level `model`**; `usage` reuses Anthropic's field names; and the stream never sends `data: [DONE]` — its terminator is a real `finish_reason` inside `output`, while in-progress chunks carry the *literal string* `"null"` there, which also meant interruption detection could report false `sse_truncated`.

Handled by hand-written adaptation at Footprint Ladder **level 1–2** (extend existing functions plus in-module helpers; no new probe, FFI export, or module file). Protocol-translation crates were evaluated and rejected: none support the native protocol, and they are built for proxies that own the complete stream — they return `Err` on truncated payloads, whereas a passive observer must degrade gracefully across TLS record boundaries.

## Related Issue

closes #2910

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] CI/CD or build changes

## Scope

- [x] `sight` (agentsight)

## Key Changes

- `src/genai/helpers.rs`: add `DASHSCOPE_NATIVE_PATHS` + `is_dashscope_native_path`; wire into `is_llm_api_path` (the gate) and `extract_provider_from_path` (→ `dashscope`); `extract_messages_view` accepts `input` as an **object** and reads `input.messages`
- `src/genai/openai_parse.rs`: add `extract_dashscope_native_parts` to rebuild the `output` envelope — `output.text`, `output.choices[].message.content` (string or multimodal `[{"text":…}]` blocks), `reasoning_content`, and `tool_calls`; tolerates both cumulative (`incremental_output=false`, the protocol default) and incremental streaming; `extract_parts_from_sse_body` now also accepts a single JSON object for non-streaming bodies
- `src/genai/call_builder.rs`: reconstruct non-SSE native responses — `build_response` previously had a fallback only for `is_sse`, so non-streaming assistant content was silently lost
- `src/parser/sse/event.rs`: `is_done` recognises a terminal `output.finish_reason` / `output.choices[].finish_reason`, so a normally-completed native stream stops looking truncated
- `src/analyzer/token/parser.rs`: identify the native protocol by its **response envelope** (top-level `output` object paired with `usage`), ahead of the generic usage branch
- `src/analyzer/token/mod.rs`: add `LLMProvider::DashScope`; per-modality counters remain a DashScope signal, `total_tokens` deliberately does not (see below); `detect_provider_from_endpoint` covers `maas.aliyuncs.com`
- `src/analyzer/unified.rs`: backfill `TokenRecord.model` from the request body when the response carries none, so the token database stops recording an empty model for native calls
- `agentsight.json` + `src/config.rs`: add `*.maas.aliyuncs.com` (Bailian workspaces are served from `{WorkspaceId}.{region}.maas.aliyuncs.com`) for DNS/connection discovery. No `schema_version` bump: this adds a pattern to an existing array without changing fields or semantics
- `docs/agent-diagnostics-guide.md`: split the provider table into compatible mode vs native protocol, and sync the two `https` config examples

## Two defects caught by validating against real data

Both were introduced by implementing from the documentation, and neither was reachable by unit tests written from the same assumptions. Both now have regression tests.

1. **Provider mislabelling.** Keying DashScope detection on `input_tokens` + `total_tokens` assumed "Anthropic never emits `total_tokens` inside `usage`". Replaying 84 real usage objects from a host's `genai_events.db` disproved it: 28 of them (33%) are genuine Anthropic-gateway records (`claude-opus-5` / `kimi-k3` on `/api/anthropic/v1/messages`) that report `total_tokens` alongside the Anthropic-only cache counters. Those would have been written to the token database as `dashscope`, corrupting `agentsight token`, `/metrics`, and cost analysis. Detection now uses the `output` envelope, which is unambiguous. After the fix all 84 classify as `anthropic`, matching the provider AgentSight itself recorded.

2. **Token double-counting.** `image_tokens` was assumed to sit *outside* `input_tokens`, so it was folded in. A real `qwen-vl-plus` response shows the opposite — it is an *itemisation*:
   ```
   "usage": {"input_tokens_details": {"image_tokens": 1249, "text_tokens": 12},
             "input_tokens": 1261, "output_tokens": 382, "total_tokens": 1643}
   ```
   1249 + 12 == 1261 == `input_tokens`, and 1261 + 382 == 1643 == `total_tokens`. Folding produced 2510 input tokens, inflating every image call by ~98%. Tests now assert `input + output == total_tokens` so this class of error cannot recur silently.

Additionally, tool-call slots are addressed by a wire-supplied `index` and the slot vector grows to fit it; that index is now bounded (`MAX_TOOL_CALL_SLOTS`) so a malformed or hostile response cannot turn one JSON field into an unbounded allocation inside a passive observer.

## Checklist

- [x] `cargo clippy --all-targets -- -D warnings` pass
- [x] `cargo test` pass
- [x] Lock files are up to date (`Cargo.lock` untouched — no dependency change)

## Testing

**Real API calls.** Three live calls against `dashscope.aliyuncs.com` native endpoints (non-streaming `text-generation`, streaming `text-generation`, `multimodal-generation`) confirmed the field shapes: in-progress chunks do carry `finish_reason` as the literal `"null"` with `"stop"` on the final frame; streaming is cumulative; the envelope has no top-level `model`; and `image_tokens` is already inside `input_tokens`. The three response bodies are committed verbatim under `tests/fixtures/dashscope_native/` and replayed by `test_real_dashscope_native_{nonstream,stream,multimodal}_replay`, so a future hand-written fixture cannot re-encode a wrong assumption unnoticed.

**Real-corpus replay.** The provider classifier was replayed over 84 usage objects extracted from a live `genai_events.db`, comparing the old and new logic. Old: 56 anthropic / 28 dashscope. New: 84 anthropic / 0 dashscope, agreeing with the provider recorded per request path.

**Pipeline integration tests** (`tests/pipeline.rs`, 9 total): synthetic and real `SslEvent`s driven through parser → aggregator → analyzer → genai builder, asserting `provider == "dashscope"`, model resolved from the request body, system/user roles recovered from `input.messages`, cumulative chunks de-duplicated to `"1 2 3 4 5"` rather than concatenated, `finish_reason == "stop"` surfaced, and token counts reconciling with the provider-reported `total_tokens`. Existing OpenAI and Anthropic pipeline tests are unchanged and still pass, guarding compatible mode.

**Unit tests** cover the path whitelist, `input`-object messages view, the native envelope in `parse_json`, the two regression guards above, the SSE terminator (including that a literal `"null"` and the Responses API's `output` *array* do **not** terminate), and the tool-call index bound.

Not verified: the eBPF uprobe capture stage was not exercised end to end — attaching to a synthetic test client did not succeed in this environment (`agentsight discover` matched the process, but no attach followed). That stage is provider-agnostic and untouched by this change; the integration tests enter the pipeline at the parser with the same bytes.

## Configuration delivery note

The new default host `*.maas.aliyuncs.com` only takes effect on **fresh installs**. Existing hosts keep their on-disk `agentsight.json` (the `schema_version`-gated rewrite does not apply, and the shallow-merge upgrade path would be overridden by the user's existing `https` key anyway). This is intentional: domain rules are user-configurable by design (replace semantics). To capture Bailian workspace traffic on an existing installation, add `*.maas.aliyuncs.com` to the `https` array in the config file manually.


Daydreamer-Li (commented):
## Review of `9a982029` — solid change; 1 blocking gap in how the new default reaches installed hosts

Good work overall. The two defects found by replaying real traffic (33% provider mislabelling, ~98% image-token inflation) are locked down by tests that assert `input + output == total_tokens`, which blocks the whole class rather than the one instance. CI is green. I found **1 blocking issue + 4 non-blocking**.

## Blocking

### 1. The new default host never reaches already-installed hosts, so the feature silently does nothing on upgrade

**What is wrong:** This PR adds `*.maas.aliyuncs.com` to the default `https` array so Bailian workspace traffic gets discovered. But AgentSight only rewrites an existing on-disk config when its `schema_version` is **older** than the current one. `schema_version` is unchanged (2 before, 2 after), so `ensure_default_agents_config` returns early and every host that already has a config file keeps its old `https` array forever.

**Where:** `src/agentsight/src/config.rs`, `ensure_default_agents_config` — `if on_disk_version >= Some(CURRENT_SCHEMA_VERSION) { return Ok(()); }`; `src/agentsight/agentsight.json:2`

**Impact:** The scenario this PR exists to fix — Bailian native calls being invisible — remains broken on exactly the hosts where AgentSight is already deployed. Nothing warns the user; it just looks like the feature was not implemented.

**One correction to the codex bot's suggestion on this thread:** bumping `schema_version` alone does **not** fix it. The upgrade path is a shallow merge that overlays *every* top-level key the user has, and `https` is a key the default config always ships — so any host that has ever generated a config carries its own `https` array and the merge keeps it. There is currently no mechanism that can deliver a new default *inside* an existing key. Whatever the resolution is (explicit migration for this key, a startup warning when the pattern is absent, or a documented manual edit), it needs to be stated in the PR body and the user docs, because today the only working answer is "hand-edit the file" and that is written down nowhere.

## Non-blocking

### 2. When `usage` is not parsed, a native call writes a `provider=openai`, zero-token row into the token database

**What is wrong:** If the response `usage` is not recovered — routine for a passive observer crossing TLS record boundaries, as the PR body itself acknowledges — the local estimator fallback runs. Three things in that fallback do not know the native protocol:

- provider is a hardcoded binary — anything not matching `anthropic` becomes `openai` (`src/analyzer/unified.rs:827`);
- the request side reads only a top-level `messages` array or an `input` **array**, so native `input.messages` is missed → `input_tokens = 0` (`src/analyzer/unified.rs:841`);
- the response side reads only top-level `choices`, so native `output.choices` is missed → `output_tokens = 0` (`src/analyzer/token/extractor/openai.rs:80`).

`compute_tokens_manually` then returns the record unconditionally.

**Impact:** A row labelled `openai` with 0/0 tokens lands in the token database for a DashScope call. That is the same corruption class the PR body calls out as blocking-worthy (`agentsight token`, `/metrics`, cost analysis) — only the wrong label is now `openai` instead of `dashscope`. Either teach this fallback the native shapes, or skip emitting a record when neither side could be read.

### 3. In incremental mode, a delta that happens to start with the accumulated text loses characters

**What is wrong:** `accumulate` decides "cumulative vs incremental" with `next.starts_with(buf)` — replace if true, append otherwise. In incremental mode that misfires whenever a delta begins with everything accumulated so far. Simplest repro is two identical consecutive tokens: buffer `"1"`, next delta `"1"` → treated as cumulative and replaced, so the result stays `"1"` and one character is lost. Same for `"a"` then `"ab"` yielding `"ab"` instead of `"aab"`.

**Where:** `src/agentsight/src/genai/openai_parse.rs:498-506`

**Impact:** Repeated characters (`---`, `###`, doubled CJK punctuation) can be under-recorded. Bounded in practice because the native default is cumulative and incremental is opt-in. The codex bot flagged the same spot; I reproduced it independently and agree.

### 4. The wildcard host only works on the DNS discovery path, not the connection path

**What is wrong:** Domain rules feed two discovery paths. The DNS path does glob matching, so `*.maas.aliyuncs.com` matches there. The connection path resolves each domain to IPs and compares, and it skips any non-exact pattern outright: `if !is_exact_domain(pattern) { continue; }` (`src/agentsight/src/discovery/connection_scanner.rs:126`).

**Impact:** Connection scanning exists precisely to catch processes that already hold established connections and will not emit a fresh DNS query — and those are the ones a wildcard rule cannot help. Worth documenting, since configuring the rule and still seeing nothing is confusing.

### 5. The canonical bilingual user docs are now out of sync with the shipped default

Only the internal `src/agentsight/docs/agent-diagnostics-guide.md` was updated. The user-facing configuration pages still show `dashscope.aliyuncs.com` alone in two places each, and the first of those is a verbatim copy of the default config, which no longer matches `agentsight.json`:

- `docs/user-guide/en/agent-observability/agentsight/configuration.md:53-56` and `:168-172`
- `docs/user-guide/zh/agent-observability/agentsight/configuration.md:52-55` and `:165-169`

Combined with issue 1, these are exactly the pages a user consults when adding the host by hand, so they matter more than usual here.

## On the bot findings I checked and disagree with

- **`MAX_TOOL_CALL_SLOTS = 256` is too small (qoderai, P1) — I read this as a false positive.** 256 parallel tool calls in a single response is far beyond any real model's behaviour, and an over-range index is logged at debug rather than dropped silently. The bound looks correctly sized.
- **Envelope-only detection is too weak (qoderai, P1) — reasonable, but low priority.** No known provider collides with "`output` is an object" + "`usage` present", and the code documents why usage field shapes cannot be used instead. Fine as later hardening.
- **`model_from_request_body` needs a truncation-tolerant parser (qoderai, P2)** — failing to backfill on an unparseable body is graceful degradation consistent with the surrounding code.
- I initially suspected that merging every `choices[]` entry into one text buffer was a new defect; the pre-existing OpenAI branch does the same thing, so it is an established project convention and I withdraw it.

## Verified clean

- `is_done` is carefully guarded: `output` must be an object, only whitelisted terminal values count, and there are negative tests for the literal string `"null"` and for the Responses API's `output` *array* — so a normally-completed native stream stops looking truncated without opening a path to cutting streams short.
- Three real response bodies committed as fixtures and replayed, so a future hand-written fixture cannot re-encode a wrong assumption.
- Tool-call index is bounded with a log, matching the project's standing requirement that new buffers have an explicit cap.
- No `unwrap`/`expect`/`dbg!` in non-test code; no new module file, probe, or FFI export (Footprint Ladder 1–2); `Cargo.lock` untouched.
- Adding `LLMProvider::DashScope` does not break downstream consumers — provider is persisted as a string and there is no exhaustive provider match or provider-keyed pricing table to extend.

## Review coverage

- Reviewed: path whitelist and provider attribution, usage extraction and double-counting, SSE terminator and interruption false positives, native envelope reconstruction (cumulative / incremental / tool calls / multimodal / single object), model backfill, config delivery and both discovery paths, doc sync, existing OpenAI/Anthropic regression surface, unbounded allocation, enum-addition blast radius.
- Not verified: the eBPF uprobe capture stage (the PR states it was not exercised end to end), real `incremental_output=true` traffic, real multi-workspace `*.maas.aliyuncs.com` DNS discovery, aarch64.

KaiLongZhou:
## Code Review Check 结论

**审核结果：✅ 可以合并（无阻塞问题）**

code-reviewer 的审查结论经验证合理：
- 无 P0/P1 阻塞问题
- 2 个 P2 建议（accumulate 边界注释、fixture 尾部换行）均为可选改进，不阻塞合并
- 测试覆盖全面，安全边界清晰，与现有解析逻辑无冲突

KaiLongZhou (approved):

Issue #2824, opened 2026-08-25
sight: AgentSight user guide is far thinner than other components


Daydreamer-Li:
### Documentation location

docs/user-guide/{en,zh}/agent-observability/agentsight.md

### What is wrong or unclear?

The AgentSight user guide is a single ~305-line page, while comparable components (cosh-ng, tokenless, skillfs) ship task-oriented page sets. Concretely missing for users:

- no reference for most CLI flags (`summary`, `metrics`, `skill-metrics`, most `interruption` subcommands, `audit --exclude`, `token --period`)
- no configuration reference: `features`, `runtime_limits`, Agent discovery rules, `schema_version` upgrade behaviour
- no Dashboard walkthrough, and no explanation of why some pages are hidden (capability probing)
- only 5 of the 18 interruption types are documented, with no triage workflow
- no deployment guidance beyond systemd (container/sidecar capabilities, macOS limits, upgrade/uninstall)
- no storage/retention or HTTP API reference, and no screenshots anywhere in the user guide

### Suggested change

Keep `agentsight.md` as an overview hub (so existing published links stay valid) and add task pages under `agentsight/`: quick start, CLI reference, configuration, Dashboard guide, interruption detection, deployment, data and storage, integrations, troubleshooting — with Dashboard screenshots in both languages and synthetic sample data.

PR #2965, opened 2026-08-28
fix(ckpt): retry loop attach when udev races device registration

yummypeng:
## Why

When every registered loop device is attached (e.g. leftovers from containers that exited without detaching), `LOOP_CTL_GET_FREE` makes the kernel register a new loop device, but udev creates its `/dev/loopN` node asynchronously — the very `losetup` that triggered the registration fails with ENOENT. Reproduced on fresh ECS hosts with the ws-ckpt sidecar image: bootstrap then fails every time a new device number is needed, alternating fail/succeed with device numbers growing past the pre-registered 0-7 range (loop8, loop9, ...).

## What changed

- `attach_loop` (BtrfsLoop backend) now retries: after a failure it asks `losetup --find` which device would be handed out next, creates that node synchronously via `mknod` (the daemon runs as root, and the kernel already has the device registered), sleeps briefly, and retries (up to 5 attempts). On a healthy host the first attempt succeeds and nothing changes.
- Sidecar deploy guide: documented the bare `docker run` leftover-loop behavior, the persistent-volume mitigation (a kept image file lets the bootstrap orphan recovery match the leftover by inode and auto-detach it), and a cleanup one-liner.

## Related issue

no-issue: found while validating ws-ckpt sidecar container deployment on fresh hosts

## User / Agent impact

Daemon bootstrap no longer fails intermittently with `losetup ... failed to set up loop device: No such file or directory` when leftover attachments exhaust the pre-registered loop devices; repeated container starts become deterministic.

## Risk and compatibility

Low risk. The success path is byte-for-byte the old behavior; the retry path only runs after `losetup` already failed, uses at most 5 short attempts, and the only new system change is mknod'ing device nodes the kernel has already registered (the daemon already runs as root and shells out to losetup/mount/mkfs). No CLI/API/protocol changes.

## Validation

- Reproduced the alternating fail/succeed pattern on fresh ECS hosts (failures exactly at runs allocating device numbers beyond the existing node set, e.g. loop8/loop9)
- Rebuilt the sidecar container image with this fix; daemon bootstrap succeeds deterministically end-to-end (sparse image create -> loop attach -> btrfs mount -> listener up)
- Compiled via the container builder stage (`cargo build --release --locked`); CI covers fmt/clippy/tests

## Documentation and rollback

`src/ws-ckpt/docs/k8s-sidecar-deploy.md` SIGTERM section updated. Rollback: revert this commit (changelog/version files untouched).

### DIFF
--- a/src/ws-ckpt/docs/k8s-sidecar-deploy.md
+++ b/src/ws-ckpt/docs/k8s-sidecar-deploy.md
@@ -1,29 +1,38 @@
-# ws-ckpt K8s Sidecar 部署指南
+# Deploy ws-ckpt as a Kubernetes Sidecar
 
-## 前置条件
+[中文版](k8s-sidecar-deploy_zh.md)
 
-- 宿主机内核已加载 btrfs 模块（`modprobe btrfs && modprobe loop`）
-- 宿主机 `/dev/loop-control` 和 `/dev/loop*` 可用
-- 容器镜像包含：ws-ckpt 二进制、btrfs-progs、util-linux（losetup/mount/findmnt）、psmisc（fuser）、rsync、kmod
+This guide deploys `ws-ckpt` as a privileged daemon sidecar alongside an
+unprivileged application container. The example persists daemon state on the
+host and propagates the btrfs mount into the application container.
 
-## 构建镜像示例
+## Prerequisites
+
+- The host kernel has the btrfs and loop modules available
+  (`modprobe btrfs && modprobe loop`).
+- `/dev/loop-control` and `/dev/loop*` are available on the host.
+- The image contains `ws-ckpt`, `btrfs-progs`,
+  `util-linux`, `psmisc`, `rsync`, and `kmod`.
+
+## Build an image
+
+The following example builds an image on Alibaba Cloud Linux 3.
 
 ```bash
-# 1. 编译 ws-ckpt（需要 Rust stable >= 1.78）
+# 1. Build ws-ckpt (requires Rust stable 1.78 or later)
 cd anolisa/src/ws-ckpt/src
 cargo build --release --workspace
 ls target/release/ws-ckpt
 
-# 2. 准备构建上下文
+# 2. Prepare the image context
 mkdir -p ~/wsckpt-image
-# 拷贝编译好的二进制可执行文件
 cp target/release/ws-ckpt ~/wsckpt-image/
 
-# 3. 如果宿主机是源码编译的 btrfs-progs（dnf 无包），需要拷贝用户态工具：
+# 3. Copy locally built btrfs-progs when the package is unavailable from dnf
 cp /usr/local/bin/btrfs /usr/local/bin/mkfs.btrfs ~/wsckpt-image/
-# 若 dnf 有 btrfs-progs 则 Dockerfile 里直接 dnf install 即可，跳过此步
+# Skip this step and install btrfs-progs in the Dockerfile when dnf provides it
 
-# 4. 写 Dockerfile
+# 4. Create the Dockerfile
 cat > ~/wsckpt-image/Dockerfile <<'EOF'
 FROM registry.cn-hangzhou.aliyuncs.com/alinux/alinux3:latest
 
@@ -31,7 +40,7 @@ RUN dnf install -y --setopt=tsflags=nodocs \
       util-linux psmisc rsync kmod which findutils \
     && dnf clean all
 
-# btrfs-progs：若 dnf 有包则换成 dnf install btrfs-progs
+# Replace these lines with dnf install btrfs-progs when the package is available
 COPY btrfs mkfs.btrfs /usr/local/bin/
 RUN chmod 0755 /usr/local/bin/btrfs /usr/local/bin/mkfs.btrfs \
     && ln -sf /usr/local/bin/btrfs /usr/bin/btrfs
@@ -40,84 +49,105 @@ COPY ws-ckpt /usr/local/bin/ws-ckpt
 RUN chmod 0755 /usr/local/bin/ws-ckpt
 EOF
 
-# 5. 构建 + 验证
+# 5. Build and verify the image
 cd ~/wsckpt-image
 docker build -t localhost/ws-ckpt:smoke .
 docker run --rm localhost/ws-ckpt:smoke ws-ckpt --version
 docker run --rm localhost/ws-ckpt:smoke btrfs --version
 
-# 6. 导入到集群节点（以 k3s 为例；标准 k8s 推送到 registry 即可）
+# 6. Import the image into cluster nodes (k3s example)
 docker save -o ws-ckpt-smoke.tar localhost/ws-ckpt:smoke
 sudo k3s ctr images import ws-ckpt-smoke.tar
-# 或推送到私有 registry：
+# For standard Kubernetes, push the image to a registry instead
 # docker tag localhost/ws-ckpt:smoke your-registry/ws-ckpt:smoke
 # docker push your-registry/ws-ckpt:smoke
 ```
 
-## 部署
+## Deploy the Pod
 
 ```bash
-# 确保宿主机内核模块就绪
+# Ensure that the required host kernel modules are loaded
 sudo modprobe loop && sudo modprobe btrfs
 
-# 应用 Pod 清单（yaml 示例见 k8s-sidecar-example.yaml）
+# Apply the example Pod manifest
 kubectl apply -f k8s-sidecar-example.yaml
-kubectl get pod wsckpt-smoke -w          # 等待 2/2 Running
-kubectl logs wsckpt-smoke -c daemon      # 确认 bootstrap 成功
+kubectl get pod wsckpt-smoke -w          # Wait for 2/2 Running
+kubectl logs wsckpt-smoke -c daemon      # Confirm successful bootstrap
 ```
 
-完整 Pod yaml 示例见 [`k8s-sidecar-example.yaml`](../k8s-sidecar-example.yaml)。
+The complete manifest is available in
+[`k8s-sidecar-example.yaml`](../k8s-sidecar-example.yaml).
 
-## Pod 设计
+## Pod design
 
-单 Pod 双容器，共用同一镜像：
+The Pod runs two containers from the same image.
 
-| 容器 | 角色 | 权限 |
-|------|------|------|
-| daemon | 运行 `ws-ckpt daemon`，管理 btrfs loop | privileged: true |
-| app | 业务负载，通过 UDS 调用 CLI 与 daemon 通信 | 无特权 |
+| Container | Role | Privileges |
+|-----------|------|------------|
+| daemon | Runs `ws-ckpt daemon` and manages the btrfs loop device | `privileged: true` |
+| app | Runs the workload and invokes the CLI over UDS | Unprivileged |
 
-## 必需 Volume
+## Required volumes
 
-| 名称 | 类型 | 用途 |
-|------|------|------|
-| sock | emptyDir | UDS socket（`/run/ws-ckpt/ws-ckpt.sock`） |
-| dev | hostPath `/dev` | loop 设备访问（`/dev/loop-control`、`/dev/loop*`） |
-| ws-state | hostPath `/var/lib/ws-ckpt` | **必须持久化** — 存放 btrfs-data.img、state.json、daemon.lock。丢失此 volume = 数据丢失。 |
-| ws-ckpt-mount | hostPath（anchor 目录） | daemon 在此 overmount btrfs；daemon 侧 `mountPropagation: Bidirectional`，app 侧 `HostToContainer` |
-| data | emptyDir | workspace 父目录；workspace 使用其子目录（如 `/data/workspace`） |
-| config | emptyDir | `/etc/ws-ckpt` —— 全局配置面。CLI 写 `config.toml`，daemon 重载同一份文件。**不共享则所有 `config --global` 设置对 daemon 无效。** |
+| Name | Type | Purpose |
+|------|------|---------|
+| sock | `emptyDir` | UDS socket at `/run/ws-ckpt/ws-ckpt.sock` |
+| dev | `hostPath` for `/dev` | Access to the host loop devices |
+| ws-state | `hostPath` for `/var/lib/ws-ckpt` | Persistent loop image, state, and daemon lock |
+| ws-ckpt-mount | `hostPath` anchor | btrfs mount propagated to the application |
+| data | `emptyDir` | Workspace parent directory, such as `/data` |
+| config | `emptyDir` | Shared global configuration at `/etc/ws-ckpt` |
 
-## 关键约束
+Losing the `ws-state` volume loses the loop image and its data.
 
-1. **`privileged: true`** — losetup/mount/mkfs.btrfs 需要。
+## Deployment constraints
 
-2. **hostPath `/dev` 透传** — k3s containerd 不会自动向 privileged 容器暴露宿主机 loop 设备节点，必须显式挂载。
+1. Set `privileged: true` for the daemon. `losetup`, `mount`,
+   and `mkfs.btrfs` require it.
 
-3. **`/var/lib/ws-ckpt` 必须持久化**（hostPath 或 PVC）。btrfs 镜像路径不可配置。若此目录落在容器可写层，pod 重启后 img 消失，但 orphan loop/mount 泄漏到宿主机。
+2. Mount the host `/dev` into the daemon. A privileged container under
+   k3s containerd does not automatically receive the host loop device nodes.
 
-4. **`--mount-path` 必须指向一个两容器同路径挂载的共享 volume**（mount anchor），daemon 侧配 `mountPropagation: Bidirectional`，app 侧配 `HostToContainer`。不能使用容器 rootfs 里的普通目录，那样 overmount 只存在于 daemon 容器私有的 mount namespace 里，无法传播到 app 容器，`init` 之后 app 会拿到悬空 symlink，而 daemon 侧无任何报错。
+3. Persist `/var/lib/ws-ckpt` with a `hostPath` or PVC. The btrfs
+   image path is not configurable. If this directory remains in the writable
+   container layer, a Pod restart removes the image while leaving host-level
+   loop devices or mounts behind.
 
-5. **workspace 必须是共享 volume 的子目录**，不能直接使用 volume 挂载点本身。`ws-ckpt init` 会对 workspace 路径执行 `rename()`，对挂载点 rename 返回 EBUSY。
+4. Point `--mount-path` to a shared volume mounted at the same path in
+   both containers. Use `Bidirectional` propagation for the daemon and
+   `HostToContainer` for the app. A root filesystem directory is private
+   to the daemon's mount namespace, so the app would receive a dangling symlink
+   after `init`.
 
-6. **`shareProcessNamespace: true`** — 使 `fuser -m`（rollback/recover/img-shrink 判断挂载是否空闲时调用）能看到 app 容器的进程。
+5. Place the workspace below the shared data volume instead of using the
+   volume mount point itself. `ws-ckpt init` renames the workspace, and
+   renaming a mount point fails with `EBUSY`.
 
-7. **后端自动检测**：若宿主机 `/var/lib` 在原生 btrfs 上，`auto_detect` 选择 BtrfsBase（直接操作 subvolume，无需 loop 设备）。BtrfsLoop overmount 路径仅在 ext4/xfs 宿主机上激活。两种后端均可正常工作，无需 ConfigMap 强制指定。
+6. Enable `shareProcessNamespace: true`. This lets `fuser -m` see
+   application processes while rollback, recovery, and image shrinking check
+   whether a mount is busy.
 
-8. **`/etc/ws-ckpt` 必须两容器共享**。全局配置是文件面而非 socket 面：`ws-ckpt config --global` 由 CLI 直接写 `/etc/ws-ckpt/config.toml`，daemon 重载时读**它自己容器内**的同一路径。两容器各用镜像层时，daemon 读不到文件即回落内置默认值，`auto-cleanup` 等策略全部失效。CLI 写入后会比对 daemon 的 effective 配置并报错，但根治方式是把 `/etc/ws-ckpt` 挂成共享 volume（`emptyDir` 即可）。
+7. Backend selection is automatic. When the filesystem containing
+   `/var/lib` is btrfs, `auto_detect` selects `BtrfsBase` and
+   operates on subvolumes without a loop device. Filesystems such as ext4 and
+   XFS use the `BtrfsLoop` overmount path.
 
-   per-workspace 配置（`config -w`）走 daemon 状态，不受此约束。
+8. Share `/etc/ws-ckpt` between both containers.
+   `ws-ckpt config --global` writes
+   `/etc/ws-ckpt/config.toml` directly, and the daemon reloads its own
+   view of that path. Without the shared volume, the daemon keeps its built-in
+   defaults and policies such as `auto-cleanup` do not take effect.
+   Per-workspace configuration through `config -w` is stored in daemon
+   state and does not have this requirement.
 
-## SIGTERM 与 Loop 设备清理
+## SIGTERM and loop device cleanup
 
-**daemon 收到 SIGTERM 只退出，不执行 `umount` 或 `losetup -d`。**
+The daemon exits on `SIGTERM` without running `umount` or
+`losetup -d`. Without cleanup, the host-level btrfs mount and loop device
+can outlive the container. Repeated Pod recreation can then accumulate stacked
+mounts and exhaust `/dev/loop*` devices.
 
-不做清理的后果：
-- 宿主机级 btrfs 挂载在容器死亡后持续存在
-- loop 设备保持 attached，backing path 指向已消失的容器 rootfs 内路径
-- 反复重建 pod 会累积 stacked mount 并耗尽 `/dev/loop*` 设备
-
-**缓解方案：** daemon 容器配置 `preStop` lifecycle hook：
+Configure this `preStop` lifecycle hook on the daemon container.
 
 ```yaml
 lifecycle:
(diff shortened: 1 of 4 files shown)


src/ws-ckpt/docs/k8s-sidecar-deploy.md
```diff
@@ -135,6 +135,17 @@ lifecycle:
 
 对于 preStop 无法执行的异常场景（OOMKill、节点故障）：daemon 下次启动时 `try_exists` 检测到 orphan loop 设备（backing file 已不存在），跳过陈尸挂载走冷路径重新 bootstrap。此场景下旧挂载中的数据不可恢复 — 持久化 `ws-state` volume 是防止此情况发生的根本保障。
 
+裸 `docker run`（没有 preStop）每次容器退出都会遗留一个 attached loop 设备，反复运行会持续累积，建议：
```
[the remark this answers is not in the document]
yummypeng: Maintainer decision: the guide's bilingual pairing is intentionally not addressed in this PR; keeping the doc as-is here.

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
```diff
@@ -1163,19 +1163,68 @@ async fn check_mount_busy(mount_path: &str) -> Option<String> {
...
+        .and_then(|(_, num)| num.parse::<u32>().ok())
+    {
+        Some(minor) => minor,
+        None => return,
+    };
+    if Path::new(free_dev).exists() {
+        return;
+    }
+    let minor_str = minor.to_string();
+    if let Err(e) = run_command_checked("mknod", &[free_dev, "b", "7", &minor_str]).await {
```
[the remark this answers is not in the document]
yummypeng: Fixed in 640a2d28: the minor is no longer inferred from the device name. `ensure_loop_node` now reads the authoritative `major:minor` from `/sys/block/<name>/dev` (the device is already kernel-registered when `losetup --find` reports it), so shifted minors under `max_part > 0` are handled correctly.

src/ws-ckpt/docs/k8s-sidecar-deploy.md
```diff
@@ -135,6 +135,17 @@ lifecycle:
...
+
+- 给 `/var/lib/ws-ckpt` 挂持久卷。镜像文件跨运行保留后，下次启动的孤儿恢复能按 inode 匹配到上次遗留并自动 detach，不再累积，状态也得以保留；
+- 清理存量遗留前先确认没有 ws-ckpt 容器在运行（容器内视角的 backing 路径在宿主机上不存在，运行中的容器会被误判为悬空）：
+
+```bash
+losetup -ln -O NAME,BACK-FILE | while read dev file; do
```
kongche-jbw, line 202: **[P1] Scope cleanup to the ws-ckpt loop device**

This predicate detaches every host loop whose backing path is absent from the caller's mount namespace. The documented sidecar is privileged and bind-mounts the host `/dev`, so running this from the container -- or on the host while other container-created loops exist -- can detach unrelated workloads. Checking only that no ws-ckpt container is running does not protect loop devices owned by other services, nor deleted-but-still-active backing files. Resolve and detach only the device backing the known ws-ckpt mount/image (for example, the exact source returned by `findmnt` plus an ownership/inode check); do not perform a host-wide `[ ! -e ]` sweep.

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
```diff
@@ -1163,19 +1163,79 @@ async fn check_mount_busy(mount_path: &str) -> Option<String> {
...
+                return Ok(dev);
+            }
+            Err(e) => {
+                warn!(
+                    "losetup attach attempt {}/{} for {} failed: {:#}",
+                    attempt, ATTEMPTS, img, e
+                );
+                last_err = Some(e);
+                if let Ok(free) = run_command("losetup", &["--find"]).await {
+                    ensure_loop_node(free.trim()).await;
```
kongche-jbw, line 1197: **[P1] Normalize util-linux's lost-device output**

On util-linux 2.40 and newer, `losetup --find` prints `/dev/loopN (lost)` when the kernel device exists but its node is missing -- the exact state this recovery path targets. Passing that string unchanged makes `file_name()` return `loopN (lost)`, so the subsequent `/sys/block/{dev_name}/dev` read fails and `mknod` is never called. The retry therefore still fails on current distributions. Parse the canonical `/dev/loopN` name before constructing the sysfs path, then keep using the authoritative sysfs `major:minor` added in this revision. See util-linux issue #2824 and `losetup.c`'s `A_FIND_FREE` output path.

kongche-jbw (changes requested):
Requesting changes for two remaining P1 blockers on the current head. The previous max_part device-number issue is fixed by reading sysfs, but the lost-device output still prevents that path from being reached with current util-linux.

src/ws-ckpt/docs/k8s-sidecar-deploy.md
yummypeng: Addressed in 21e53c48 by scoping instead of a full sweep: the one-liner now matches only the ws-ckpt image path (`/var/lib/ws-ckpt/btrfs-data.img`) and skips everything else, so loops owned by other services are never touched. Within that scope, the `[ -e ]` check only detaches when the backing file is dangling (container rootfs already gone); a host-level (systemd) ws-ckpt daemon's live image file exists and is skipped automatically. The prerequisite keeps "no ws-ckpt container running" because a running container's backing path also looks dangling from the host namespace.

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
yummypeng: Good catch — fixed in 21e53c48: `ensure_loop_node` now keeps only the first whitespace-separated token of the `losetup --find` output (stripping the util-linux >= 2.40 ` (lost)` suffix), guards that the name starts with `loop`, and then uses the authoritative `major:minor` from `/sys/block/<name>/dev` as before.

kongche-jbw (commented):
Second-pass review found three remaining blockers in the revised cleanup and node-creation paths.

src/ws-ckpt/docs/k8s-sidecar-deploy.md
```diff
@@ -135,6 +135,20 @@ lifecycle:
 
 对于 preStop 无法执行的异常场景（OOMKill、节点故障）：daemon 下次启动时 `try_exists` 检测到 orphan loop 设备（backing file 已不存在），跳过陈尸挂载走冷路径重新 bootstrap。此场景下旧挂载中的数据不可恢复 — 持久化 `ws-state` volume 是防止此情况发生的根本保障。
 
+裸 `docker run`（没有 preStop）每次容器退出都会遗留一个 attached loop 设备，反复运行会持续累积，建议：
+
+- 给 `/var/lib/ws-ckpt` 挂持久卷。镜像文件跨运行保留后，下次启动的孤儿恢复能按 inode 匹配到上次遗留并自动 detach，不再累积，状态也得以保留；
+- 清理存量遗留前先确认没有 ws-ckpt 容器在运行。只处理 ws-ckpt 默认镜像路径的悬空设备，**不要对全部 loop 设备做 `[ -e ]` 扫描**——宿主机上其他服务的 loop 设备、以及运行中容器的 backing 路径（容器 rootfs 内部路径）在宿主视角都不存在，全量扫描会误杀：
```
kongche-jbw: **[P1] Apply this scoping rule to the actual preStop commands**

The warning says never to sweep every loop device, but the preStop snippet immediately above (lines 131-133) still does exactly that, and `k8s-sidecar-example.yaml` retains the same host-wide `[ -e ] || losetup -d` loop. Because the daemon container is privileged and bind-mounts the host `/dev`, the primary Kubernetes deployment can still detach loops owned by unrelated containers/services. Update both executable examples to capture and detach only the loop source backing the ws-ckpt mount/image; changing only this new bare-Docker snippet does not remove the production hazard.

src/ws-ckpt/docs/k8s-sidecar-deploy.md
```diff
@@ -135,6 +135,20 @@ lifecycle:
...
+
+```bash
+losetup -ln -O NAME,BACK-FILE | while read dev file; do
+    [ "$file" = "/var/lib/ws-ckpt/btrfs-data.img" ] || continue
```
kongche-jbw: **[P1] Match the deleted backing-file form**

`losetup -l` prefers `/sys/block/loopN/loop/backing_file`; the loop driver renders that through `file_path()`, which calls `d_path()` and appends ` (deleted)` once container-rootfs cleanup unlinks the image. In that target orphan case, `file` is therefore `/var/lib/ws-ckpt/btrfs-data.img (deleted)` (or its escaped table form), so this exact equality rejects it and the advertised cleanup performs no detach. Parse a stable/raw output form and explicitly handle the deleted suffix, or identify ownership without exact display-string equality.

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
```diff
@@ -1163,19 +1163,82 @@ async fn check_mount_busy(mount_path: &str) -> Option<String> {
...
+        Err(e) => {
+            warn!("Could not read /sys/block/{}/dev: {}", dev_name, e);
+            return;
+        }
+    };
+    let (major, minor) = match dev_str.trim().split_once(':') {
+        Some(pair) => pair,
+        None => return,
+    };
+    if let Err(e) = run_command_checked("mknod", &[free_dev, "b", major, minor]).await {
```
kongche-jbw: **[P1] Create the block node with restrictive permissions**

GNU `mknod` defaults to `a=rw - umask`; this service does not set `UMask`, and systemd's system-unit default is 0022, so the fallback creates `/dev/loopN` as 0644 rather than the normal 0660 `root:disk`. If udev is still delayed (the path this code handles) or absent, once the next retry attaches the image any local unprivileged user can read the raw workspace block device. Pass an explicit restrictive mode such as `-m 0600` (and account for the expected device label/ownership) before attaching it.

kongche-jbw (commented):
Two additional completeness issues: the host cleanup is not executable as shown for a normal user and omits the supported legacy image path, and the newly added parsing/recovery behavior has no regression tests.

src/ws-ckpt/docs/k8s-sidecar-deploy.md
```diff
@@ -135,6 +135,20 @@ lifecycle:
 
 对于 preStop 无法执行的异常场景（OOMKill、节点故障）：daemon 下次启动时 `try_exists` 检测到 orphan loop 设备（backing file 已不存在），跳过陈尸挂载走冷路径重新 bootstrap。此场景下旧挂载中的数据不可恢复 — 持久化 `ws-state` volume 是防止此情况发生的根本保障。
 
+裸 `docker run`（没有 preStop）每次容器退出都会遗留一个 attached loop 设备，反复运行会持续累积，建议：
+
+- 给 `/var/lib/ws-ckpt` 挂持久卷。镜像文件跨运行保留后，下次启动的孤儿恢复能按 inode 匹配到上次遗留并自动 detach，不再累积，状态也得以保留；
+- 清理存量遗留前先确认没有 ws-ckpt 容器在运行。只处理 ws-ckpt 默认镜像路径的悬空设备，**不要对全部 loop 设备做 `[ -e ]` 扫描**——宿主机上其他服务的 loop 设备、以及运行中容器的 backing 路径（容器 rootfs 内部路径）在宿主视角都不存在，全量扫描会误杀：
+
+```bash
+losetup -ln -O NAME,BACK-FILE | while read dev file; do
```
kongche-jbw, line 202: **[P2] Make the host cleanup command executable and cover legacy images**

This host-side command is shown without `sudo`, but `losetup -d` requires elevated privileges and its error is redirected to `/dev/null`; for the normal non-root shell used elsewhere in this guide it will silently clean nothing. It also ignores the supported legacy image path `/data/ws-ckpt/btrfs-data.img`, although the daemon and uninstall flow explicitly retain migration/cleanup support for that path, so "existing leftovers" from older releases remain. Run the detach portion under explicit privilege without hiding failures, and include both supported image identities.

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
```diff
@@ -1163,19 +1163,82 @@ async fn check_mount_busy(mount_path: &str) -> Option<String> {
...
+/// authoritative device number), only the udev node is missing. The minor is
+/// deliberately NOT inferred from the device name: with `max_part > 0` the
+/// loop driver reserves partition minors and assigns the disk a shifted
+/// minor, so `loopN` is not necessarily block `7:N`. The daemon runs as
+/// root; failure is non-fatal because udev may still deliver the node before
+/// the next retry.
+async fn ensure_loop_node(free_dev: &str) {
+    // util-linux >= 2.40 reports a missing node as "/dev/loopN (lost)" —
+    // precisely the state being repaired here; keep only the device path.
+    let free_dev = free_dev.split_whitespace().next().unwrap_or("");
```
kongche-jbw: **[P2] Add regression coverage for free-device parsing**

The privileged branch is not exercised by the current Rust tests, so CI stayed green through both the `max_part` minor bug and the util-linux `(lost)` parsing bug. Extract the output normalization and `major:minor` parsing into pure helpers and cover at least `/dev/loop8`, `/dev/loop8 (lost)`, malformed/empty output, and shifted sysfs minors. Keep one integration test or command mock that verifies the resulting `mknod` arguments and restrictive mode.

src/ws-ckpt/docs/k8s-sidecar-deploy.md
yummypeng: Addressed in a2421afe: both executable examples now scope to the ws-ckpt image path instead of sweeping every loop. Updated the preStop snippet in this doc and the preStop hook in `k8s-sidecar-example.yaml` identically — capture `NAME,BACK-FILE`, strip the ` (deleted)` suffix, match only `/var/lib/ws-ckpt/btrfs-data.img`, and detach only when the backing file is dangling. I kept the path-scope + existence-check form rather than a findmnt-based resolution because preStop also runs while our own daemon's loop is still attached (preStop precedes SIGTERM), and the existence check is what prevents detaching a live ws-ckpt loop while still catching prior dead containers' dangling ones.

src/ws-ckpt/docs/k8s-sidecar-deploy.md
yummypeng: You're right, the exact-equality match was broken for the target case — fixed in a2421afe. All three snippets (bare-docker cleanup, doc preStop, yaml preStop) now strip the kernel-appended suffix first (`clean=${file%" (deleted)"}`), match the canonical image path against the cleaned string, and run the existence check on the cleaned path as well, so deleted-rootfs orphans are matched and detached while a live host-level daemon's existing image file is still skipped.

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
yummypeng: Fixed in a2421afe: the fallback now creates the node with an explicit restrictive mode (`mknod -m 0660 ...`) instead of relying on the process umask, and then best-effort `chown root:disk` to match the ownership udev would give loop nodes (warn-only if the disk group is absent in a minimal image). So once the image is attached, the raw block device is no longer world-readable in the delayed-udev path this code handles.

src/ws-ckpt/docs/k8s-sidecar-deploy.md
kongche-jbw: 感谢更新。当前提交已将文档和 YAML 中的可执行示例都限定到 ws-ckpt 镜像路径，原先可能误 detach 其他服务 loop 设备的问题已修复，我先关闭这条讨论。preStop 对“本次正在退出的 loop”是否真正完成清理是另一个生命周期问题，我会单独标注，避免混在本线程里。

src/ws-ckpt/docs/k8s-sidecar-deploy.md
kongche-jbw: 感谢修复。三个片段现在都显式剥离了 ` (deleted)` 后缀，原先目标 orphan 无法命中的问题已解决，我先关闭这条讨论。

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
kongche-jbw: 感谢修复。当前已在创建节点时显式指定 0660，并补充 root:disk ownership 处理，避免了默认 umask 生成 0644、导致所有本机用户可读的问题；原问题已解决，我先关闭这条讨论。

src/ws-ckpt/k8s-sidecar-example.yaml
```diff
@@ -94,7 +98,9 @@ spec:
               - |
                 umount /opt/ws-ckpt-mount 2>/dev/null || true
                 losetup -ln -O NAME,BACK-FILE | while read dev file; do
-                  [ -e "$file" ] || losetup -d "$dev" 2>/dev/null
+                  clean=${file%" (deleted)"}
+                  [ "$clean" = "/var/lib/ws-ckpt/btrfs-data.img" ] || continue
+                  [ -e "$clean" ] || losetup -d "$dev" 2>/dev/null
```
kongche-jbw: **[P1] 请按 backing inode 清理，而不要复用当前命名空间的路径存在性**

感谢把扫描范围收窄；不过这里仍会稳定跳过需要清理的设备。`preStop` 运行时当前容器的 `/var/lib/ws-ckpt/btrfs-data.img` 必然存在，因此当前 loop 在 `umount` 后不会 detach。更隐蔽的是，假设旧 Pod 留下 `/var/lib/ws-ckpt/btrfs-data.img (deleted)`，新 Pod 已在同一路径创建了另一个 inode：剥掉后缀后，`[ -e "$clean" ]` 检查的是新 Pod 的文件，于是旧 orphan 也被跳过。也就是说，路径字符串相同并不能证明这些 loop 属于同一个文件，当前 hook 在正常 hostPath 生命周期中可能一个设备都不清理。建议在 `umount` 前后用 `losetup -j /var/lib/ws-ckpt/btrfs-data.img` 按 inode 捕获并 detach 本 Pod 的设备；历史 deleted orphan 则需要基于明确的 deleted 状态/可靠 ownership 处理，不要用当前同名路径的 `-e` 结果代替。文档中的同一片段也请同步。

src/ws-ckpt/docs/k8s-sidecar-deploy.md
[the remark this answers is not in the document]
kongche-jbw: 理解这里希望控制 PR 范围。不过从当前仓库级 `AGENTS.md` 的措辞看，`specs/documentation-standard.md` 属于 mandatory checks；本 PR 又继续扩展了这份用户指南，因此这条目前仍是“明确未处理”，不宜作为已修复关闭。我先重新打开，建议合入前由 maintainer 明确决定是补齐配对文档，还是在 PR 中记录批准的例外。

kongche-jbw (changes requested):
感谢连续更新。基于当前 `a2421afe`，我已复核并关闭以下已修复讨论：

- `max_part` 下从 sysfs 读取真实 major:minor；
- 兼容 util-linux 2.40+ 的 `(lost)` 输出；
- 文档和 YAML 不再全局扫描/误 detach 其他服务的 loop；
- 处理 backing path 的 ` (deleted)` 后缀；
- `mknod` 显式使用 0660 并处理 ownership，避免默认 0644。

目前仍建议暂缓合入的实质问题只有一项 P1：preStop 剥掉 deleted 后缀后，用当前命名空间的同名路径做 `-e`，会把旧 orphan 的 inode 与新 Pod 的同名文件混为一谈，同时也会跳过本次正在退出的 loop；正常 hostPath 生命周期下 hook 可能一个设备都不 detach。inline comment 中给出了按 `losetup -j`/inode 识别的建议。

另外有几项合入前关注，严重度较低：新增 free-device 解析和 mknod 参数仍无回归测试；宿主机 cleanup 示例仍缺少提权且未覆盖 legacy image；PR 风险清单也建议勾选 documented behavior 和 privileged/security change。双语文档按 maintainer 的 scope 决定留待后续补充，不作为本 PR 的合入关注项。定向 `cargo fmt --check` 与 `btrfs_loop` 单测 22/22 已通过。辛苦再看一下。

src/ws-ckpt/docs/k8s-sidecar-deploy.md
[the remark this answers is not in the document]
kongche-jbw: 收到，按本 PR 的 scope 决定先不在这里处理；这项留待后续单独补充，我重新关闭本线程，也不再作为本 PR 的合入关注项。

kongche-jbw (commented):
Review baseline: `b147618ec07e5e2f73fe3d2eaa947936a3655b2e...a2421afedcda44bd4b9235d3255899e7561e135a`

**[P1] 用 backing inode 识别需要清理的 loop**

`src/ws-ckpt/k8s-sidecar-example.yaml:103`

`preStop` 运行时镜像文件仍存在，所以卸载后 `[ -e "$clean" ]` 仍为真，
本 Pod 的 loop 不会 detach。若旧 Pod 的 deleted backing 已被新 Pod 的同名文件
替换，这项检查也会把两个 inode 混为一谈，旧 orphan 同样被跳过。结果是 hook
无法履行防止 loop 泄漏的职责；文档 `docs/k8s-sidecar-deploy.md:134` 的同一片段
也受影响。

Possible direction: 在 `umount` 前按 `losetup -j`/inode 捕获本实例的设备并
detach，再覆盖正常 preStop 和“同名路径、不同 inode”的回归场景。

**[P2] 让宿主清理命令可实际执行并覆盖 legacy image**

`src/ws-ckpt/docs/k8s-sidecar-deploy.md:148`

普通用户运行示例时，`losetup -d` 需要提权，但错误被重定向后命令会静默不清理。
它也只匹配新路径，遗漏 daemon 和卸载流程仍支持的
`/data/ws-ckpt/btrfs-data.img`，因此旧版本遗留设备不会被处理。

Possible direction: 显式使用所需权限且保留失败信息，并覆盖两个受支持的镜像路径。

**[P2] 为 free-device 解析和 mknod 参数补回归测试**

`src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs:1220`

当前 Rust 测试没有经过新增的输出归一化、`major:minor` 解析和 `mknod` 调用链；
CI 因而无法拦截本轮已出现过的 `max_part` minor 和 util-linux `(lost)` 回归。

Possible direction: 提取纯解析 helper，覆盖普通、`(lost)`、空或畸形输出和 shifted
minor，并用命令 mock 验证 `mknod` 的设备号与 0660 mode。

src/ws-ckpt/docs/k8s-sidecar-deploy.md
yummypeng: Addressed in a79fcf1d: the host cleanup block now runs under `sudo sh <<'CLEANUP'` (the rest of the guide uses a non-root shell, so the privilege requirement is explicit), `losetup -d` failures are no longer redirected to /dev/null, and the case match covers both supported image identities — the default `/var/lib/ws-ckpt/btrfs-data.img` and the legacy `/data/ws-ckpt/btrfs-data.img` (`LEGACY_BTRFS_IMG_PATH`, still honored by the daemon and the uninstall flow).

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
yummypeng: Addressed in a79fcf1d: the parsing is split into pure helpers with unit-test coverage — `normalize_free_loop_dev` (plain `/dev/loop8`, `/dev/loop8 (lost)`, empty/garbage/non-loop/partition names) and `parse_sysfs_dev_numbers` (plain `7:8`, shifted minor `7:128` with surrounding whitespace, and malformed inputs). The mknod invocation is built by `loop_node_mknod_args`, and a test pins the exact argument vector including the restrictive `-m 0660` mode, so a regression to umask-default permissions fails CI. I left the helpers pure rather than adding a process-mock harness, since there is no command-mocking infrastructure in this crate today.

src/ws-ckpt/k8s-sidecar-example.yaml
yummypeng: Reworked in a79fcf1d per your suggestion — path existence is no longer the signal anywhere:

- The preStop hook (both in `k8s-sidecar-example.yaml` and the doc snippet) now captures this pod's device with `losetup -j /var/lib/ws-ckpt/btrfs-data.img` before the umount — that matches by the backing file's (device, inode), so a new pod recreating the same path with a different inode can no longer mask an old orphan. The captured devices are detached after the umount.
- Historical orphans are matched on the kernel's explicit ` (deleted)` marker on the backing file (only an unlinked backing file carries it, so a live container's loop never matches), covering both the default and the legacy `/data/ws-ckpt` image paths.
- The host-side cleanup command uses the same two signals (deleted marker or missing file) scoped to the two image paths, never a bare `-e` on the raw path.

Remaining caveat, documented in the hook comment: if the daemon still holds the mount when preStop runs, umount/detach fail with EBUSY — that residual is left to the next daemon start's inode-based orphan recovery, matching the lifecycle split you noted.

src/ws-ckpt/src/crates/daemon/src/backends/btrfs_loop.rs
```diff
@@ -1491,4 +1598,48 @@ mod tests {
...
+        // util-linux >= 2.40 appends " (lost)" when the node is missing.
+        assert_eq!(
+            normalize_free_loop_dev("/dev/loop8 (lost)"),
+            Some("/dev/loop8")
+        );
+    }
+
+    #[test]
+    fn normalize_free_loop_dev_rejects_malformed_output() {
+        for raw in ["", "   ", "garbage", "/dev/sda", "/dev/loop", "/dev/loop1p1"] {
```
kongche-jbw: **[P2] 合入前请补跑一次 cargo fmt**

感谢补上显式生命周期；我在当前 head 上复核后，`cargo check -p ws-ckpt-daemon` 和新增的 28 个 `btrfs_loop` 测试都已通过。目前 CI 剩下的是这一处 `cargo fmt --all --check` 差异，rustfmt 会把该数组展开为多行。麻烦补跑一次 `cargo fmt --all` 即可。

src/ws-ckpt/k8s-sidecar-example.yaml
```diff
@@ -82,19 +82,35 @@ spec:
...
+      # preStop runs, umount/detach fail with EBUSY and the next daemon's
+      # inode-based orphan recovery finishes the cleanup.
       lifecycle:
         preStop:
           exec:
             command:
               - /bin/sh
               - -c
               - |
+                own_loops=$(losetup -j /var/lib/ws-ckpt/btrfs-data.img 2>/dev/null | cut -d: -f1)
```
kongche-jbw, line 104: **[P2] 建议同时捕获当前 legacy image 对应的 loop**

当前 inode 捕获只查询 canonical 路径；不过 daemon 在 legacy migration 失败时会按现有恢复逻辑继续使用 `/data/ws-ckpt/btrfs-data.img`。preStop 执行时该 backing file 仍然存在，因此后面只匹配 ` (deleted)` 的历史 orphan 分支也不会处理它，umount 后可能留下当前 legacy loop。为保持与本次已经补齐的 legacy 清理范围一致，是否可以在 umount 前对 canonical 和 legacy 两个路径都执行 `losetup -j` 并合并结果？文档中的同一片段也可以一并同步。

kongche-jbw (approved):

PR #2976, opened 2026-08-29
feat(sight): add k8s DaemonSet packaging

jfeng18:
## Why

Issue #1238 asks for DaemonSet deployment and monitoring Agents in pods. The kernel-side pieces have landed over the past two months (observer-namespace PID reporting in #2360, container-aware cgroup/container-id extraction), but the repository ships no container image and no DaemonSet manifest, so there is no supported way to run AgentSight node-wide in Kubernetes.

## What changed

- `src/agentsight/packaging/k8s/daemonset.yaml` — one pod per node, `hostPID: true`, `BPF`+`PERFMON`+`SYS_PTRACE` capabilities, read-only host BTF mount, hostPath data persistence, resource limits mirroring the packaged systemd unit, optional ConfigMap override for `config.json`. No RBAC: AgentSight talks to no Kubernetes API.
- `src/agentsight/packaging/docker/Dockerfile` — runtime image that installs the release RPM (the canonical artifact; the binary is not rebuilt) and runs `agentsight-start` as entrypoint so pod lifecycle signals reach both workers.
- Deployment guide (en + zh) gains a "Kubernetes DaemonSet" section; both component READMEs link to it.

The `SYS_PTRACE` entry was added after adversarial review: a hostPID container holding only `BPF`+`PERFMON` cannot read `/proc/<pid>/maps` of a non-dumpable process owned by another uid (verified on kernel 6.6), which would silently skip such Agents.

The hostPID-free form (bind-mounted host procfs) is deliberately not shipped: it depends on the procfs-root work (#2804) plus a CLI/env knob that does not exist yet, and the docs say so instead of implying it works.

## Related issue

refs #1238 (not closing: the hostPID-free shape and the image-publish CI wiring remain open)

## User / Agent impact

New opt-in deployment form. Nothing changes for existing systemd/foreground/container-sidecar users.

## Risk and compatibility

Low risk: new files plus documentation. The manifest requests elevated capabilities by design (eBPF); it grants no Kubernetes API access and mounts host paths read-only except the data directory.

## Validation

- `scripts/docs-lint.sh` and `scripts/docs-link-check.py` pass (naming, en/zh parity, relative links).
- `daemonset.yaml` parses as Namespace + DaemonSet; capability list and mounts inspected programmatically.
- Capability claim verified empirically (Docker 24, kernel 6.6): with `BPF`+`PERFMON` only, reading `/proc/<pid>/maps` of a `PR_SET_DUMPABLE=0` process owned by another uid is denied; with `SYS_PTRACE` added it succeeds.
- Not covered: a real `docker build` (needs the release RPM in `rpms/<arch>/`) and an apply against a live cluster (none available here). The Dockerfile reuses the base image and install pattern of the existing `docker/Dockerfile.alinux`.

## Documentation and rollback

Docs updated in the same PR (deployment guide en+zh, component READMEs). Rollback: delete the branch files; nothing else references them.

### DIFF
--- a/docs/user-guide/en/agent-observability/agentsight/deployment.md
+++ b/docs/user-guide/en/agent-observability/agentsight/deployment.md
@@ -112,6 +112,36 @@ For a Kubernetes sidecar, the same three things matter:
 Keep the Dashboard port inside the cluster and reach it through a Service or port-forward rather
 than exposing 7396 publicly.
 
+## Kubernetes DaemonSet (node-wide)
+
+A DaemonSet gives you one AgentSight pod per node observing every Agent process on that node —
+the sidecar form above watches one pod, this form watches the whole machine. The manifest lives at
+`src/agentsight/packaging/k8s/daemonset.yaml`; build the image with
+`src/agentsight/packaging/docker/Dockerfile`, push it to your registry, and set the `image:` field
+accordingly.
+
+How it works: `hostPID: true` places the pod in the host PID namespace and the probes attach
+host-wide (uprobes register by inode, so Agent binaries inside other pods are covered), while PID
+attribution stays correct because AgentSight reports PIDs in the observer namespace.
+
+```bash
+kubectl apply -f src/agentsight/packaging/k8s/daemonset.yaml
+kubectl -n agentsight rollout status ds/agentsight
+kubectl -n agentsight port-forward ds/agentsight 7396:7396   # then open http://localhost:7396
+```
+
+The manifest requests `BPF` + `PERFMON` + `SYS_PTRACE` capabilities (the last one is needed to
+read `/proc/<pid>/maps` of another uid's process once it is non-dumpable; without it such Agents
+are silently skipped), mounts host `/sys/kernel/btf` read-only, and persists data to hostPath
+`/var/log/sysak`; resource limits mirror the packaged systemd unit (300m CPU / 350Mi memory). It
+talks to no Kubernetes API, so no RBAC is needed. The pod runs only the trace and serve workers
+(`agentsight-start`), so the Risk Enforcement page has no data in this form. To override
+`config.json`, create the `agentsight-config` ConfigMap from the shipped default and uncomment the
+marked blocks in the manifest.
+
+A hostPID-free form (bind-mounting the host procfs instead) is planned once the procfs-root work
+lands; it is not available yet — keep `hostPID: true` until then.
+
 ## macOS
 
 macOS builds contain no eBPF. Two commands exist:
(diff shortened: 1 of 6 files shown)


chengshuyi (changes requested):
Reviewed the manifest, the Dockerfile, and the doc changes against the current tree. The overall shape is right — `hostPID` + `/proc/<pid>/root` really does cover agents in other pods (verified at `src/agentsight/src/probes/sslsniff.rs:948`), the resource limits match the packaged unit (`CPUQuota=30%` / `MemoryMax=350M`), `terminationGracePeriodSeconds: 30` matches `TimeoutStopSec=30s`, and the docs honestly decline to advertise the hostPID-free form. One blocker and a couple of smaller things.

### 1. Blocker: the RPM glob matches both architectures

`packaging/docker/Dockerfile`:

```dockerfile
COPY rpms/ /tmp/rpms/
RUN dnf install -y /tmp/rpms/*/agentsight-*.rpm && ...
```

`.github/workflows/docker-nightly.yaml` organizes RPMs into `rpms/amd64/` and `rpms/arm64/`, and both land in the same build context. This glob therefore expands to two same-name, different-arch RPMs and hands both to `dnf install` — that fails, or in the friendlier case installs the wrong arch. The stated rationale also doesn't hold: the comment says `rpm-build.sh` emits `x86_64/aarch64` dirs, but the repo convention is `amd64/arm64`, and globbing one level down doesn't discriminate by arch at all, it just widens the match.

The existing `docker/Dockerfile.alinux` already solves this — `ARG TARGETARCH` + `COPY rpms/${TARGETARCH}/*.rpm`. Suggest following that pattern (or at minimum filtering on `$(uname -m)`). It also avoids copying the whole `rpms/` tree — every component, both arches — into an image layer.

### 2. hostPath drops the 0700 guarantee the docs promise

The systemd unit sets `UMask=0077`, so `/var/log/sysak/.agentsight` is root-only, and `README.md` states exactly that. Nothing sets a umask in the DaemonSet form: the container default is 0022, so the same data ends up world-readable on the host. Either add `umask 0077` at the top of `scripts/agentsight-start.sh` (harmless on the systemd path, where the unit already enforces it) or document it as an explicit caveat.

Related: the hostPath mounts all of `/var/log/sysak`, which other anolisa components share. Mounting `/var/log/sysak/.agentsight` is enough and narrows the blast radius.

### 3. Nits

- No `containerPort: 7396` and no liveness probe — `/health` is already there, so a probe is nearly free and gives a clearer signal than pod restarts.
- OCI labels omit `org.opencontainers.image.licenses`, unlike `docker/Dockerfile.alinux`.
- `image: agentsight:latest` resolves to a nonexistent `docker.io/library/agentsight`. The comment says to replace it, but an unedited `kubectl apply` is a plausible first move — a `registry.example.com/agentsight:VERSION` placeholder would fail more obviously.
- No CI validation of the manifest. A `kubectl apply --dry-run=client` or `kubeconform` step would be cheap insurance, since nothing in this PR was exercised against a real cluster.

chengshuyi (approved):
LGTM，先合入。注意 Dockerfile 的 RPM glob 在多架构 build context 下仍有隐患（rpms/amd64|arm64 会被同时命中），后续跟进用 ARG TARGETARCH 修掉。

PR #2979, opened 2026-08-30
feat(sight): LSM-only agent-file-guard profile + file_delete_guard capability API

viron-xiao:
## Summary

Add LSM-only file-delete protection to AgentSight, with a runtime capability API for AgentSecCore integration.

## Changes

**agent-file-guard pinned profile (patch 0005)**
- New `HookReserve::agent_file_guard()` profile: attaches `enforce_path_unlink` / `enforce_path_rename` LSM hooks, suppresses dataflow tracepoints.
- `ACTPLANE_PINNED_PROFILE=agent-file-guard` selects this profile at runtime.

**`file_delete_guard` runtime capability**
- `EnforcementCapabilities.file_delete_guard: bool` dynamically derived from loaded engine.
- SecCore checks `GET /api/enforcement/health → capabilities.file_delete_guard == true` before pushing `block unlink` policies (fail-closed).

**Attested build**
- `patches/actplane/0005-add-agent-file-guard-profile.patch` + `build-enforcer.sh` extended.


chengshuyi (commented):
(superseded: comments were not line-anchored, re-submitted in a follow-up review)

src/agentsight/crates/ebpf-ifc-engine/src/lib.rs
```diff
@@ -1801,6 +1832,15 @@ impl PinnedEngine {
         }
     }
 
+    /// Whether the loaded profile enables synchronous file delete/write (unlink)
+    /// blocking: it must reserve both the file-block LSM hooks (FEAT_BLOCK_FILE)
+    /// and the write-rule matcher class (FEAT_WRITE_RULES). Callers use this to
+    /// report accurate enforcement coverage instead of assuming static support.
+    pub fn supports_file_delete_guard(&self) -> bool {
```
chengshuyi, line 1839: This derives the capability from reserved features only, but the LSM hooks are attached conditionally at load time: `let enforce = bpf_lsm_active(); ... if enforce { attach LSM_PROGS }`. On a host without BPF LSM, `agent-file-guard` still loads fine (`validate_pinned_runtime` only enforces the LSM requirement for `file_open_only` profiles, and this one sets `file_open_only: false`), the `enforce_path_unlink`/`enforce_path_rename` hooks are silently skipped, yet this method still returns true.

Net effect: `/api/enforcement/health` reports `file_delete_guard=true`, SecCore gates on it and pushes `block unlink` policies, but nothing is actually blocked — the exact silent-false-positive gap this PR sets out to close.

Suggested fix (either):
1. AND in the runtime state here: `self.policy_features & ... == ... && bpf_lsm_active()`, or
2. make `validate_pinned_runtime` fail fast when this profile is selected without an active BPF LSM, matching the `file-enforcement`/`credential-exfiltration` behavior.

src/agentsight/crates/ebpf-ifc-engine/src/lib.rs
```diff
@@ -1524,7 +1526,12 @@ fn config_features(cfg: &CConfig) -> u32 {
             }
         }
         if r.op == OP_WRITE {
-            features |= FEAT_FILE_FLOW | FEAT_WRITE_RULES | path_match_features(r.m);
+            // Label-conditioned write/unlink block rules only need the write-rule
+            // matcher class (evaluated synchronously in the LSM hook). FEAT_FILE_FLOW
+            // is added separately by file taint sources (source X = file "..."),
+            // so a pure `block unlink file "X" if <exec-label>` policy stays LSM-only
+            // and does not pull in the file-flow dataflow tracepoints.
+            features |= FEAT_WRITE_RULES | path_match_features(r.m);
```
chengshuyi, line 1534: Dropping `FEAT_FILE_FLOW` from OP_WRITE rules changes feature derivation for every profile (Minimal `from_config` path), not just the new pinned profile. The reasoning in the comment looks sound — file-flow tracking is only needed when a file taint source exists, and `source X = file "..."` updates still add `FEAT_FILE_FLOW` in the updates loop — but there is no regression test pinning this down.

Could you add `config_features` assertions, e.g.:
- a config with only an OP_WRITE block rule does NOT reserve `FEAT_FILE_FLOW`;
- the same config plus a file-source update DOES reserve `FEAT_FILE_FLOW`?

That protects against a future change silently degrading dataflow tracking for policies that mix write rules with file taint sources.

chengshuyi (commented):
Review of the agent-file-guard profile and file_delete_guard capability. One correctness concern on the capability derivation (silent false-positive when BPF LSM is inactive), plus a request for regression tests on the config_features change.

src/agentsight/crates/ebpf-ifc-engine/src/lib.rs
jfeng18: Fixed in 2bcb5068: `capabilities.file_delete_guard` now ANDs `supports_file_delete_guard()` with the live `bpf_lsm_active()` state (`actplane.rs:349-350`), so a host without active BPF-LSM reports `false` (fail-closed) instead of advertising a capability that enforces nothing. Verified at head a9a62139: the engine function has this single call site.

src/agentsight/crates/ebpf-ifc-engine/src/lib.rs
jfeng18: Covered in 2bcb5068 as well: `write_block_rule_alone_does_not_reserve_file_flow` and `write_block_rule_with_file_source_reserves_file_flow` pin the derivation both directions, so the profile-wide behavior change is regression-tested.

jfeng18 (approved):
Approve。两个 review 发现的修复已在当前 head 验证：

1. **LSM 静默假阳性**（chengshuyi inline #1）：`2bcb5068` 把 `file_delete_guard` 改为 `supports_file_delete_guard() && bpf_lsm_active()`（actplane.rs:349-350，附注释说明 fail-closed 语义）。我核实了 `supports_file_delete_guard()` 只有这一个调用点，无旁路。
2. **FEAT_FILE_FLOW 派生变更无回归测试**（inline #2）：同一 commit 新增 `write_block_rule_alone_does_not_reserve_file_flow` / `write_block_rule_with_file_source_reserves_file_flow` 双向钉住。

CI 22 门全绿（含 Test agentsight）。线程已回复并 resolve。

PR #2967, opened 2026-08-28
fix(cosh-ng): [shell] recover OSC parser

kongche-jbw:
## Why

An unterminated private COSH OSC candidate can retain unbounded input, hide terminal output, and prevent later authenticated frames from being parsed. The receiver needs bounded recovery without exposing abandoned internal transport bytes. This change keeps fragmented processing linear while preserving the existing BEL-terminated protocol.

## What changed

- Cap retained COSH OSC candidates at 64 KiB and resynchronize at the latest complete COSH prefix.
- Preserve fragmented restart prefixes across the cap boundary and keep incremental scans linear.
- Drop unterminated private candidates at EOF while retaining existing malformed-frame recovery.

## Related issue

closes #2964

## User / Agent impact

A damaged or unterminated internal OSC frame no longer blocks subsequent terminal output or authenticated COSH markers.

## Risk and compatibility

- [x] Privileged or security-sensitive behavior changed

Low risk. The change is limited to the authenticated private OSC receiver. It does not change the sender, marker schema, BEL framing contract, shell execution policy, or public interfaces. Candidates larger than 64 KiB are intentionally discarded.

## Validation

- `cargo test --package cosh-shell --lib shell_host::implementation::osc_tests:: -- --test-threads=4` — 61 passed
- `cargo test --package cosh-shell --test raw_cli agent_input::raw_cli_routes_path_prompt_after_shell_cwd_change -- --exact` — 1 passed
- `cargo test --package cosh-shell --test shell_host marker::marker_json_escape_removes_raw_controls_and_preserves_utf8 -- --exact` — 1 passed
- `cargo test --package cosh-shell --test shell_host marker::shell_host_marker_control_cwd_keeps_json_frame_intact -- --exact` — 1 passed
- `cargo test --package cosh-shell --test shell_host marker::shell_host_rejects_forged_osc_markers_without_session_token -- --exact` — 1 passed
- `cargo test --package cosh-shell --test shell_host marker::shell_host_pty_cannot_read_private_marker_token -- --exact` — 1 passed
- `cargo test --package cosh-shell --test shell_host marker::shell_host_pty_statusful_precmd_marker_completes_inflight_command -- --exact` — 1 passed
- `cargo fmt --all -- --check`
- `git diff --check`

Full workspace gates, Clippy, release builds, ECS, and provider validation are delegated to CI or later integration validation.

## Documentation and rollback

[...]

PR #2983, opened 2026-08-31
fix(cosh-ng): [shell] preserve input state

### LINKED ISSUE #2951: bug(cosh-ng): natural-language prompt recalled from history fails due to leading space
### Component

cosh-ng

### What happened?

When I type a Chinese natural-language prompt directly, cosh-ng correctly intercepts it and the Agent answers:

```
◇ [root@... cosh-ng]# 你是谁
╭ Agent ───────────────────────────────────────────────────────────╮
│ 我是运行在 cosh-ng ...                                           │
╰──────────────────────────────────────────────────────────────────╯
```

However, if I press `↑` to recall the same command from bash history and then press Enter, the command gets a leading space and is executed as a shell command instead of being intercepted:

```
◇ [root@... cosh-ng]#  你是谁
bash: 你是谁: command not found
```

(Notice the extra leading space before `你是谁` in the recalled line.)

What I expected:
- Recalling a previously successful natural-language prompt from history and re-executing it should still route to the Agent, regardless of any leading whitespace added by history recall.

### How can we reproduce it?

1. Start cosh-ng raw mode with bash integration enabled.
2. Type a Chinese natural-language prompt directly and press Enter:
   ```
   你是谁
   ```
   Observe that the Agent answers.
3. Press `↑` to recall the same line from bash history.
4. Press Enter to execute the recalled line.
5. Observe that the line now has a leading space and bash reports `command not found` instead of routing to the Agent.

### Root cause

The shell-side natural-language interception is a two-stage handshake:

1. **PS0 (`_cosh_bounded_preexec_marker`)** detects that the command is natural language but does **not** call `_cosh_begin_attempt`. It simply returns, leaving the attempt state unset.
2. **`command_not_found_handle` (`_cosh_command_not_found_handle`)** tries to recover the attempt by reading the last history entry and matching it against the missing command word:
   ```bash
   history_command="$(_cosh_history_command_from_entry "$history_entry")"
   first_word="${history_command%%[[:space:]]*}"
   if [[ -n "$history_command" ]] \
      && ! _cosh_has_leading_alias "$history_command" \
      && _cosh_literal_first_word_matches "$history_command" "$first_word" "$command"; then
     _cosh_begin_attempt "$history_command" "$first_word" 0
   fi
   ```

When the line is recalled from history with a leading space, `_cosh_history_command_from_entry` returns the command **with the leading space preserved** (its `sed` expression only strips the history number and the spaces immediately after it). Consequently:
- `first_word` becomes an empty string.
- `_cosh_literal_first_word_matches` fails because the normalized first word (`你是谁`) no longer equals the empty `first_word`.
- `_cosh_begin_attempt` is never called.
- `_COSH_ATTEMPT_ACTIVE` remains `0`.
- The fallback path in `command_not_found_handle` prints `bash: 你是谁: command not found`.

### Anything else?

This only affects the shell-side natural-language path (Chinese prompts without `??` or `/` prefix). English question-style prompts may also be affected if recalled with leading whitespace.

A fix direction would be to trim leading whitespace when extracting the history command or when computing `first_word` for the recovery match, so `_cosh_begin_attempt` is established consistently whether the user types the line fresh or recalls it from history.

kongche-jbw:
## Why

Natural-language submissions can be hidden from Bash history while enhanced
integration is protecting private input with `HISTCONTROL=ignorespace`. A
recalled or edited prompt could then lose its exact accepted text, fall through
to Bash, or disappear from multi-entry recall. Terminal validation also exposed
two presentation gaps on the same input lifecycle: wrapped prompt ghosts could
leave a stale tail, and reconstructed prompts could bypass the assisted prompt
owner.

## What changed

- Capture accepted Readline input before the hidden-history transformation,
  keep exact provider payload separate from routing classification, and restore
  only confirmed safe natural-language entries to native history.
- Preserve leading and trailing whitespace plus literal tabs for the provider
  while keeping canonical secret assignments, JWTs, and deliberately indented
  input out of history. Shell glob ownership and ordinary commands remain
  unchanged.
- Clip prompt ghosts to one row with error-safe terminal restoration so wrapped
  hint tails do not remain on the next prompt.
- Route startup, restored, candidate, and slash-guard prompts through
  `PromptPresentation`. A bounded quiescent prompt epoch preserves PTY FIFO
  ordering while preventing input echo from corrupting reconstructed prompts.
- Keep the prompt-epoch bridge in `osc/prompt_epoch.rs`, leaving the central
  OSC parser below the 1000-line production growth bar.
- Correlate Bash command-not-found provenance with the accepted history entry,
  so a missing command after a successful shell command still produces its
  Smart-mode Insight and ghost.

## Related issue

Closes #2951.

## User / Agent impact

Fresh, recalled, edited, whitespace-prefixed, and literal-tab natural-language
input reaches the Agent with the accepted bytes intact. Multiple safe prompts
remain reachable in Bash history (`3 -> 2 -> 1`) without persisting secrets.
Prompt indicators are rendered exactly once, and narrow terminals no longer
retain wrapped prompt-ghost tails. Missing commands also retain their
Insight after earlier successful shell activity.

## Risk and compatibility

No CLI flag, config schema, or external protocol changes. The user-visible
terminal and history behavior changes. History recovery remains fail-closed:
only input already confirmed as provider-bound is restored, while secret and
explicitly indented input stays excluded. PTY ordering changes are internal to
`cosh-shell`.

## Validation

- `cargo test --locked -p cosh-shell --test raw_cli prompt_replay::raw_cli_bash_dirty_safe_natural_language_stays_recallable -- --exact`
- `cargo test --locked -p cosh-shell --test raw_cli prompt_replay::raw_cli_bash_recovered_history_keeps_private_inputs_excluded -- --exact`
- `cargo test --locked -p cosh-shell --test shell_host marker::bash_secret_detector_matches_canonical_jwt_shape -- --exact`
- `cargo test --locked -p cosh-shell --test shell_host marker::bash_secret_detector_matches_canonical_assignment_shape -- --exact` — includes canonical provider-prefixed keys, whitespace, quotes, and mixed separators
- `cargo test --locked -p cosh-shell --test raw_cli prompt_replay::raw_cli_bash_preserves_user_leading_whitespace_for_provider -- --exact`
- `cargo test --locked -p cosh-shell --test raw_cli prompt_replay::raw_cli_bash_bracketed_literal_tab_submits_when_closer_and_enter_share_batch -- --exact`
- `cargo test --locked -p cosh-shell --test raw_cli native::raw_cli_isolated_card_and_candidate_redraws_keep_one_assisted_prefix_per_prompt -- --exact`
- `cargo test --locked -p cosh-shell --lib shell_host::implementation::osc_tests:: -- --test-threads=4` — 66 passed after rebasing onto #2967
- `cargo test --locked -p cosh-shell prompt_epoch`
- `cargo test --locked -p cosh-shell prompt_ghost_disables_autowrap_before_restoring_the_cursor`
- `cargo test --locked -p cosh-shell prompt_ghost_restores_terminal_after_a_recoverable_partial_write`
- Focused shell-host recall, privacy, glob-ownership, parser fidelity, and PTY
  ordering tests passed.
- Bash 4.4.23: recovered-history privacy and the 74/80/200-column prompt-owner
  raw CLI matrix passed; the pre-fix 74-column run exposed the internal guard
  and had 4 prefixes for 5 prompts.
- `cargo test --locked -p cosh-shell --test raw_cli failed_command::raw_cli_success_then_command_not_found_keeps_failure_insight -- --exact`
- `cargo test --locked -p cosh-shell --test shell_host marker::shell_host_bash_success_then_missing_keeps_routing_provenance -- --exact`
- `bash -n crates/cosh-shell/src/shell_host/marker/bash.sh`
- `cargo check --locked -p cosh-shell --tests`
- Alibaba Cloud Linux 4, Bash 5.2: a patch-equivalent earlier candidate passed
  locked release build, dirty/clean history and privacy PTY regressions, and a
  sanitized no-tool real-provider baseline. The final branch is rebased onto
  the main commit that merged #2967; the combined OSC and prompt/history suite
  was repeated locally.

The manual screenshot gate is not claimed as passed in this PR.

## Documentation and rollback

No documentation change: this fixes the existing shell input, privacy, and
prompt-presentation contracts without adding CLI or configuration surface.
Rollback by reverting the four commits in reverse order, or revert the PR as a
unit. No data migration is required.


src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
```diff
@@ -420,11 +424,27 @@ _cosh_guard_slash_submission() {
...
+  fi
+  if [[ "${1:-false}" == true ]]; then
+    unset _COSH_PENDING_RECOVERY_HISTORY_INPUT 2>/dev/null || true
+    rm -f -- "${COSH_RECOVERY_REQUEST_FILE:-/tmp/cosh-recovery}.recovery-history-confirmed" \
+      2>/dev/null || true
+    if [[ "${2:-false}" == true
+       && -n "$classification_input" && "$classification_input" == "$input" ]]; then
+      if ! _cosh_command_has_secret "$input"; then
+        _COSH_PENDING_RECOVERY_HISTORY_INPUT="$input"
+      fi
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 8e382022 (final head a5ae1efe). The Bash history detector now matches the canonical JWT shape and token boundaries before staging or restoring provider-bound input. Added direct positive/boundary controls plus an end-to-end recovered-history regression; the fail-first persisted the JWT, while the fixed path keeps it out of history. Rebased onto merged #2967 and reran the combined OSC 66/66 and history/prompt suites. Required CI run 33350889933 attempt 2 is green; attempt 1 only hit an unrelated cosh-platform ETXTBSY test flake.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
```diff
@@ -420,11 +428,27 @@ _cosh_guard_slash_submission() {
...
+    _COSH_PENDING_RECOVERY_INPUT="$input"
+    _COSH_PENDING_RECOVERY_CLASSIFICATION="$classification_input"
+  fi
+  if [[ "${1:-false}" == true ]]; then
+    unset _COSH_PENDING_RECOVERY_HISTORY_INPUT 2>/dev/null || true
+    rm -f -- "${COSH_RECOVERY_REQUEST_FILE:-/tmp/cosh-recovery}.recovery-history-confirmed" \
+      2>/dev/null || true
+    if [[ "${2:-false}" == true
+       && -n "$classification_input" && "$classification_input" == "$input" ]]; then
+      if ! _cosh_command_has_secret "$input"; then
```
ikunkun-sys, line 464: [P1] Keep canonical-sensitive recovered inputs out of Bash history

This new recovery gate relies on `_cosh_command_has_secret`, but that shell detector is narrower than `redact_sensitive_text`. On this head, after poisoning the Readline mirror with Home, submitting `你好 "token" = TEST_ONLY_RECOVERY_SECRET` is routed to the Agent and then appears verbatim in `.bash_history`: the Rust canonical assignment pattern accepts quoted keys and whitespace around `=`, while the shell helper only catches contiguous `token=`/`token:` or `--token `. The later check before `history -s` uses the same helper, so this violates the PR fail-closed privacy contract. This is distinct from the existing JWT thread; JWT is now caught, but canonical assignment forms still leak.

src/cosh-ng/crates/cosh-shell/src/shell_host/osc/slash_guard_echo.rs
```diff
@@ -148,16 +161,34 @@ impl PendingSlashGuardEcho {
             .is_some_and(|before_ending| {
                 Self::proves_painted_command(before_ending, command, &redraw.prefix)
             });
+        let presentation_start_in_prefix =
```
ikunkun-sys, line 164: [P2] Preserve prompt ownership on supported Bash 4.4 redraws

The new presentation boundary is reached only after `suppressed_redraw` exists. On Alibaba Cloud Linux 3 with Bash 4.4.20 and width 74, the new `raw_cli_isolated_card_and_candidate_redraws_keep_one_assisted_prefix_per_prompt` test fails consistently with 4 assisted prefixes for 5 complete prompts, and the output exposes the full `__cosh_slash_guard__` command. Bash 4.4 inserts `\r` into the narrow redraw, so the existing exact guard-suffix classifier does not create `suppressed_redraw`; this path returns without assigning the presentation owner and raw guard output is flushed. Bash 4.3 or newer is a documented supported prerequisite.

ikunkun-sys (changes requested):
Requesting changes for two current-head regressions: recoverable Bash history can still persist canonical-sensitive inputs not recognized by the shell detector, and the prompt presentation path still fails on supported Bash 4.4 narrow-terminal redraws and exposes the internal guard command. Both reproduce on a5ae1efe; details inline.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
kongche-jbw: Fixed in 950cdd6d (final head da30bcc1). The Bash detector now covers canonical quoted keys and whitespace around assignment separators before any recovered input is staged or restored. Added direct positive/negative controls and an end-to-end dirty-Readline history regression; the fail-first persisted the quoted token assignment, while Bash 5.2 and 4.4.23 now keep it out of history. Required CI run 33354000493 is green.

src/cosh-ng/crates/cosh-shell/src/shell_host/osc/slash_guard_echo.rs
kongche-jbw: Fixed in da30bcc1. The authenticated guard classifier now recognizes Bash 4.4 Readline wrap repaint forms (bare CR and previous-byte/CR/previous-byte) only when the bounded static guard still matches exactly and uniquely; it maps the proven start back to the original display stream, so unrelated output remains fail-open. The reviewer test failed 5 prompts/4 prefixes with a visible guard on Bash 4.4.23 before the fix and now passes 5/5 with no guard leak; component controls, OSC 66/66, Bash 5.2, layout, and required CI run 33354000493 are green.

ikunkun-sys (changes requested):
Requesting changes on e87f1db4. The existing canonical-history P1 remains reproducible: the quoted `token` forms are fixed, but `_cosh_command_has_secret` still omits the Rust canonical `aws[_-]?secret[_-]?access[_-]?key` assignment. After dirtying the Readline mirror with Home, `你好 AWS_SECRET_ACCESS_KEY=TEST_ONLY_RECOVERY_SECRET` reaches the Agent and is then persisted verbatim in `.bash_history`; both recovery staging and the final `history -s` gate reuse this narrower helper. The new command-generation tests pass, and the prior JWT and Bash 4.4 prompt-owner findings are addressed. See the existing unresolved thread: https://github.com/alibaba/anolisa/pull/2983#discussion_r3891462531.

ikunkun-sys (changes requested):
Requesting changes on 9411b174. The added AWS_SECRET_ACCESS_KEY case is fixed, but the canonical-history P1 remains: `_cosh_command_has_secret` still misses provider-prefixed canonical assignments when whitespace precedes the separator, including `AWS_ACCESS_KEY_ID`, `OPENAI_API_KEY`, `DASHSCOPE_API_KEY`, `GITHUB_TOKEN`, and `ALIBABA_CLOUD_ACCESS_KEY_ID`. After dirtying the Readline mirror with Home, `你好 AWS_ACCESS_KEY_ID = TEST_ONLY_RECOVERY_SECRET` reaches the Agent and is persisted verbatim in `.bash_history`; recovery staging and the final `history -s` gate both reuse this narrower helper. The JWT, Bash 4.4 prompt-owner, AWS secret access key, and command-proof regressions now pass. See the existing unresolved thread: https://github.com/alibaba/anolisa/pull/2983#discussion_r3891462531.

ikunkun-sys (approved):
Approved on 5cf31b8f. The Bash detector now covers the full canonical Rust assignment key set without matching the added negative controls. The marker regression, real recovered-history PTY test, and independent temporary-HOME reproduction all pass; the prior JWT, Bash 4.4 prompt-owner, and command-proof findings remain addressed. All current CI checks pass and all review threads are resolved.

PR #2994, opened 2026-08-31
fix(anolisa): record self-update in the central log

### LINKED ISSUE #2992: [anolisa] fix: record self-update in the central operation log
## Background

The central log is the audit trail for mutating CLI operations: `anolisa logs`
queries it, and `anolisa bug` bundles it into the diagnostic report
(`crates/anolisa-cli/src/commands/tier1/bug.rs:210`). Every other mutating
tier-1 command appends an operation record — `install`, `uninstall`,
`update <component>` / `update all` (via `append_update_log`,
`crates/anolisa-cli/src/commands/tier1/update.rs:1086`), `upgrade`, `repair`,
and `forget`.

`anolisa upgrade` already records a CLI package update there even though the
CLI binary is not a state object: it seeds the record's `objects` with the CLI
package name (`crates/anolisa-cli/src/commands/tier1/upgrade.rs:2127`), under
the comment "CLI package (rpm-owned binary). Reported but never recorded as a
component object" (:782).

## Actual behavior

`anolisa update self` writes nothing to the central log. Neither
`handle_self_update` (`crates/anolisa-cli/src/commands/tier1/update.rs:1313`)
nor `run_self_update_with_deps` (:1464) opens `CentralLog` or appends a
`LogRecord`. This holds for both apply modes:

* `SelfUpdateApplyMode::Binary` — the running executable is replaced in place.
* `SelfUpdateApplyMode::RpmPackage` — a real `dnf update <package>` transaction
  runs as root (:1520).

After either one, `anolisa logs` shows no trace of the operation and the
`anolisa bug` bundle carries no evidence that a self-update ever happened.

## Impact

When a host ends up on an unexpected CLI version, or the dnf delegation only
partially succeeds, nothing records when the update ran, which version it moved
from and to, which apply mode was taken, or whether it succeeded. The
diagnostic data already exists in the code path and is then discarded: the
before/after RPM versions are resolved by
`installed_package_version_best_effort` (:1582) and printed to the terminal,
but never persisted. `update self` is also the one mutating command whose
failure can leave the operator without a working binary to investigate with,
which makes the missing record more costly than elsewhere.

## Expected behavior

An applied `anolisa update self` should append one `LogRecord` with
`kind = operation`, `source = "anolisa-cli"`, `command = "update self"`, the
`started_at` / `finished_at` window, and a terminal `status` — `Ok`, or
`Failed` when the manifest fetch, download, binary swap, or dnf transaction
fails. The message should carry the version transition and the apply mode; the
RPM path should additionally put the package name in `objects` and record the
observed before/after installed versions.

`--dry-run` should keep writing nothing. Two points worth deciding explicitly:

* whether an `AlreadyLatest` no-op deserves a record at all;
* whether a `state.operations` `OperationRecord` should also be added.
  `ObjectKind` has no CLI-binary variant, and the `upgrade` precedent above is
  central-log-only, so central-log-only looks like the consistent answer.

## Reproduction

```console
anolisa update self
anolisa logs --limit 20
anolisa logs --source anolisa-cli
```

The update applies, but no `update self` record appears in either query.

ikunkun-sys:
## Why

`anolisa update self` was the only mutating command that left no central-log
record. Every other tier-1 command appends one — `install`, `uninstall`,
`update <component>` / `update all`, `upgrade`, `repair`, `forget` — and
`anolisa upgrade` already records a CLI package update there even though the
CLI binary is not a state object.

Neither self-update apply mode wrote anything: the in-place binary replacement,
and the `dnf update <package>` transaction that runs as root. After either one,
`anolisa logs` showed no trace of when the update ran, which version it moved
from and to, which apply mode was taken, or whether it succeeded. The
before/after installed versions were already resolved and printed to the
terminal, then discarded.

## What changed

- Added `append_self_update_log`, appending one `LogRecord` for an applied
  self-update with `command = "update self"`, the `started_at`/`finished_at`
  window, a terminal status, and an `op-update-self-…` operation id minted
  through the shared `mint_operation_id`.
- `details` reuses the payload `build_json_data` already builds for the
  `--json` envelope, so the recorded versions match what the command printed:
  `current_version` / `latest_version` for both modes, plus `package` and
  `rpm_version_before` / `rpm_version_after` on the delegated path. The RPM
  package also lands in `objects`, so `anolisa logs <package>` finds it.
- A failed self-update carries a structured `SelfUpdateFailureContext` built as
  the run learns things, so the record keeps the target version, the apply
  mode, and — on the delegated path — the package and the versions rpm reported
  on both sides of the failed transaction. Fields that were never resolved stay
  absent rather than being recorded as unknown, and a known package still lands
  in `objects` so `logs <package>` reaches a failed delegation too.
- A failed `dnf` transaction is not treated as proof that nothing moved: the
  installed version is re-observed after the failure, and `details.updated` is
  derived from that reading rather than assumed. It is omitted when nothing was
  observed, so a consumer never sees a definitive `false` contradicted by the
  versions recorded beside it.
- Credentials never reach the log: the self-update endpoint is operator-supplied
  (`ANOLISA_UPDATE_URL`) and `SelfUpdateError::FetchManifest` quotes it
  verbatim, so any URL in a persisted failure message is rewritten to scheme,
  host, and path. Userinfo and the entire query string are dropped rather than
  keyword-filtered, since signature parameter names vary by provider.
- Previews (`--dry-run`) and up-to-date no-ops write nothing; a log write
  failure warns instead of failing an update that already applied.
- Coverage: both apply modes, a successful dnf no-op, the already-latest no-op,
  the dry-run preview, credential redaction, and — through the real
  `run_self_update_with_deps` path with injected fakes — a failed dnf
  transaction, a failed dnf transaction that still moved the rpmdb, a failed
  binary swap, and a failure before the manifest resolved.

## Related issue

closes #2992

## User / Agent impact

`anolisa logs` now reports `update self` operations. This is additive: no
existing record changes, and the `update self` human output, `--json` payload,
and exit codes are untouched.

`anolisa bug` picks up the failed self-update records, which is what its
documented filter selects — `--limit` is defined as "Maximum number of recent
warn/error log records to include", and `collect_recent_logs` queries
`severity_at_least: Warn`. Applied self-updates are `info` and stay out of the
bundle, exactly like every other command's successful operation. See the note
below.

## Risk and compatibility

Low risk. The log schema is unchanged: the versions go into the existing
free-form `LogRecord::details` field, no field was added, and `LogRecord` has
no `deny_unknown_fields` with every optional field defaulting. Records written
before this change still parse and render unchanged alongside the new ones.

One judgement call worth a reviewer's eye: successful self-updates are
deliberately *not* forced into the `anolisa bug` bundle. The `warn`-and-above
filter in `collect_recent_logs` is shared by every command and is documented on
the `--limit` flag, so special-casing `update self` would break that contract
for one command only. If the bundle should carry successful lifecycle
operations, that is a change to `bug`'s collection policy for
`install`/`update`/`upgrade` alike, and belongs in its own issue rather than
here.

## Validation

- `git diff --check`
- End-to-end: fed the records the new tests write through a real `anolisa logs`
  binary and confirmed `logs`, `logs -v`, `logs --severity warn`,
  `logs <package>`, `logs --operation-id`, and `logs --json` all surface them;
  and through a real `anolisa bug --json` to confirm which severities the
  bundle collects.
- Backward compatibility: replayed a pre-change record missing eight optional
  fields alongside the new records in one log; all parse and render, with the
  legacy record's `details` still `null`.

`anolisa update self` itself was not executed (it replaces the running binary
and needs the release endpoint); the write side is covered by unit tests driving
the real run path with injected package-query/transaction fakes, and the read
side by the real `logs` and `bug` binaries over the same JSONL file.

## Documentation and rollback

No documentation change: the central log already documents `details` as a
free-form structured payload, and no CLI surface changed. Reverting the commit
restores the previous behavior; nothing persists that a rollback would need to
migrate, since the added records are ordinary log lines older binaries also
parse.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/bug.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/bug.rs
@@ -18,6 +18,26 @@ const ISSUE_URL: &str = "https://github.com/alibaba/anolisa/issues/new?template=
 const DEFAULT_LIMIT: usize = 20;
 const MAX_LIMIT: usize = 100;
 
+/// Detail fields a record may carry out of the machine in a bug bundle.
+///
+/// An allowlist rather than a filter. `details` is free-form and a command
+/// attaches whatever it finds useful — a failed hook parks up to 4 KiB of raw
+/// subprocess stderr under `stderr_tail`, which no keyword scan can be trusted
+/// to clean, since a bare `Bearer` value or a PEM block has no `key=value`
+/// shape to catch. A bundle is written to be pasted into a public issue, so a
+/// field travels only once it is known to be structured and bounded, and
+/// anything new stays behind until it is listed here.
+const EXPORTABLE_DETAIL_KEYS: &[&str] = &[
+    "apply_mode",
+    "current_version",
+    "endpoint",
+    "latest_version",
+    "package",
+    "rpm_version_after",
+    "rpm_version_before",
+    "updated",
+];
+
 #[derive(Parser)]
 pub struct BugArgs {
     /// Limit the report to one component.
@@ -69,6 +89,8 @@ struct RecentLogSummary {
     objects: Vec<String>,
     #[serde(skip_serializing_if = "Vec::is_empty")]
     warnings: Vec<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    details: Option<serde_json::Value>,
 }
 
 #[derive(Debug, Clone, Serialize, PartialEq, Eq)]
@@ -253,7 +275,36 @@ fn summarize_log(record: LogRecord) -> RecentLogSummary {
             .into_iter()
             .map(|v| redact_sensitive(&v))
             .collect(),
+        details: redact_details(record.details),
     }
+}
+
+/// Carry a record's `details` into the bundle under the same redaction as every
+/// other field.
+///
+/// Without this the structured diagnostics a command attaches — the versions
+/// and endpoint a failed self-update records, say — are selected into the
+/// bundle and then dropped on the way out, which is the one place they were
+/// meant to be read.
+fn redact_details(value: serde_json::Value) -> Option<serde_json::Value> {
+    let serde_json::Value::Object(fields) = value else {
+        return None;
+    };
+    let kept: serde_json::Map<String, serde_json::Value> = fields
+        .into_iter()
+        .filter(|(key, _)| EXPORTABLE_DETAIL_KEYS.contains(&key.as_str()))
+        .map(|(key, value)| match value {
+            serde_json::Value::String(text) => {
+                (key, serde_json::Value::String(redact_sensitive(&text)))
+            }
+            other => (key, other),
+        })
+        .collect();
+
+    if kept.is_empty() {
+        return None;
+    }
+    Some(serde_json::Value::Object(kept))
 }
 
 fn render_markdown(
@@ -310,6 +361,9 @@ fn render_markdown(
             }
             if !log.warnings.is_empty() {
                 out.push_str(&format!("  - warnings: {}\n", log.warnings.join("; ")));
+            }
+            if let Some(details) = &log.details {
+                out.push_str(&format!("  - details: {details}\n"));
             }
         }
     }
@@ -605,6 +659,98 @@ mod tests {
 
         assert!(markdown.contains("token=<redacted>"));
         assert!(!markdown.contains("super-secret"));
+    }
+
+    /// The structured failure context is the reason a failed operation is worth
+    /// recording, and the bundle is where it gets read. Asserted on a real
+    /// payload: the record goes through the log file, the query, the summary,
+    /// and the renderer.
+    #[test]
+    fn bug_payload_carries_redacted_log_details() {
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let ctx = crate::test_support::context_for_root(
+            tmp.path(),
+            crate::context::InstallMode::System,
+            Some(tmp.path().to_path_buf()),
+            crate::test_support::TestContextOptions {
+                quiet: false,
+                no_color: false,
+                ..Default::default()
+            },
+        );
+
+        let mut record = log_record("2026-06-01T10:00:00Z", "self-update failed");
+        record.details = serde_json::json!({
+            "current_version": "0.3.7",
+            "latest_version": "0.3.8",
+            "apply_mode": "rpm-package",
+            "endpoint": "https://mirror.invalid",
+            "access_key": "must-not-appear",
+        });
+        let layout = common::resolve_layout(&ctx);
+        CentralLog::open(layout.central_log.clone())
+            .append(&record)
+            .expect("append");
+
+        let logs = collect_recent_logs(None, DEFAULT_LIMIT, &ctx).expect("query");
+        let details = logs[0]
+            .details
+            .as_ref()
+            .expect("details must reach the bundle");
+
+        assert_eq!(details["latest_version"], "0.3.8");
+        assert_eq!(details["apply_mode"], "rpm-package");
+        assert_eq!(details["endpoint"], "https://mirror.invalid");
+        assert!(
+            details.get("access_key").is_none(),
+            "a field outside the allowlist must not travel: {details}"
+        );
+
+        let markdown = render_markdown(&collect_environment(&ctx), &[], &logs);
+        assert!(markdown.contains("latest_version"), "{markdown}");
+        assert!(!markdown.contains("must-not-appear"), "{markdown}");
+    }
+
+    /// A failed hook parks raw subprocess stderr in `details.stderr_tail`, and
+    /// the keyword scan cannot clean it: a bare `Bearer` value and a PEM block
+    /// have no `key=value` shape to catch. Carrying details must not be what
+    /// first lets that out of the machine.
+    #[test]
+    fn free_form_details_stay_out_of_the_bundle() {
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let ctx = crate::test_support::context_for_root(
+            tmp.path(),
+            crate::context::InstallMode::System,
+            Some(tmp.path().to_path_buf()),
+            crate::test_support::TestContextOptions {
+                quiet: false,
+                no_color: false,
+                ..Default::default()
+            },
+        );
+
+        let mut record = log_record("2026-06-01T10:00:00Z", "hook failed");
+        record.details = serde_json::json!({
+            "phase": "post-install",
+            "exit_code": 1,
+            "stderr_tail": "curl: Authorization: Bearer eyJhbGciOi.LEAKED\n\
+                            -----BEGIN OPENSSH PRIVATE KEY-----\nb3BlLEAKED\n",
+        });
+        let layout = common::resolve_layout(&ctx);
+        CentralLog::open(layout.central_log.clone())
+            .append(&record)
+            .expect("append");
+
+        let logs = collect_recent_logs(None, DEFAULT_LIMIT, &ctx).expect("query");
+        assert!(
+            logs[0].details.is_none(),
+            "no field of a hook record is exportable: {:?}",
+            logs[0].details
+        );
+
+        let markdown = render_markdown(&collect_environment(&ctx), &[], &logs);
+        assert!(!markdown.contains("LEAKED"), "{markdown}");
+        assert!(!markdown.contains("stderr_tail"), "{markdown}");
     }
 
     #[test]
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs
@@ -60,6 +60,7 @@ use anolisa_core::providers::DelegatedProvider;
 use anolisa_core::self_update::{self, ProgressFn, SelfUpdateOutcome};
 use anolisa_core::state::{ObjectKind, OperationRecord};
 use anolisa_core::state_store::StateStore;
+use anolisa_core::transaction::mint_operation_id;
 use anolisa_platform::fs_layout::FsLayout;
 use anolisa_platform::pkg_query::{PackageQuery, PackageQueryError};
 use anolisa_platform::pkg_transaction::{PackageTransaction, PackageTransactionError};
@@ -1325,6 +1326,7 @@ pub(in crate::commands) fn handle_self_update(ctx: &CliContext) -> Result<(), Cl
         None
     };
 
+    let started_at = now_iso8601();
     let result = run_self_update_with_deps(
         &url,
         current_version,
@@ -1341,7 +1343,16 @@ pub(in crate::commands) fn handle_self_update(ctx: &CliContext) -> Result<(), Cl
         eprint!("\r\x1b[2K");
     }
 
-    let run = result?;
+    // Audit before rendering: an applied self-update must leave a record even
+    // when the caller never reads the terminal output (issue #2992).
+    let run = match result {
+        Ok(run) => run,
+        Err(failure) => {
+            append_self_update_log(ctx, &started_at, Err(&failure));
+            return Err(failure.error);
+        }
+    };
+    append_self_update_log(ctx, &started_at, Ok(&run));
 
     if ctx.json {
         return render_json_outcome(&run, ctx.dry_run);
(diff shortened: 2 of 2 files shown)


kongche-jbw (commented):
Review baseline: `b55dbcf95213...74c090190690`

**[P1] 成功记录仍不会进入 bug 诊断**

`src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs:1791` 将正常的
binary/RPM 更新记录为 `info`，但 `bug.rs:213` 只查询 `warn` 及以上记录。
因此成功执行后，`anolisa bug` 仍没有这次版本迁移的证据，与本 PR 声明的诊断
目标不符。Possible direction: 让 bug 收集器显式纳入自更新成功记录，并补一条从真实
bug payload 读取该记录的回归测试。

**[P1] 失败记录丢失已知的更新上下文**

在 manifest 已解析后，如果 binary 下载/替换或 `dnf update` 失败，
`src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs:1804` 仍只写入
`current_version` 和 `updated`，且 `objects` 为空。此时已知的
`latest_version`、apply mode、RPM 包名和更新前版本全部丢失，RPM 失败也无法被
`anolisa logs <package>` 找到。Possible direction: 让执行结果携带结构化失败上下文；
对 dnf 失败尽可能重新观测版本，并通过注入依赖的真实 handler 路径覆盖两种失败模式。

kongche-jbw (commented):
Review baseline: `b55dbcf95213...b9f79c091d5d`

**[P1] 避免把更新端点凭据写入中央日志**

当 `ANOLISA_UPDATE_URL` 使用 URL userinfo 或签名查询参数且 manifest 请求失败时，
`SelfUpdateError::FetchManifest` 的文本包含完整端点。这里把该文本原样持久化：
`src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs:1894`。
错误记录随后会进入 `anolisa bug`，而现有关键词脱敏不覆盖 userinfo 或
`X-Amz-Signature` 等参数，因此诊断包可能泄露镜像凭据。Possible direction: 在写日志前
净化 URL，或只记录无凭据的结构化错误，并补充中央日志及 bug payload 的回归测试。

**[P2] 让失败后的 updated 反映实际 RPM 变化**

`src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs:1424` 将失败上下文的
`updated` 固定为 `false`。如果 dnf 非零退出但 rpmdb 已从旧版本移动到新版本，
1613–1617 行会准确记录不同的前后版本，最终详情却同时声称没有更新，令审计消费者
得到相反结论。Possible direction: 根据失败后观测到的前后版本计算该字段；无法确定时省略，
并用“事务失败但版本已变化”的 fake 场景补回归测试。

**[P2] 不要把成功的 dnf no-op 记为 partial**

`src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs:1878` 对所有
`updated == false` 的成功委托写入 `partial`/`warn`。但 `PackageTransaction::update`
明确把包已是仓库最新版本的 no-op 定义为成功，且 `LogStatus::Partial` 表示部分对象
已应用、部分未应用；此场景没有部分提交，却会作为告警进入 `anolisa bug`。
Possible direction: 保留 `details.updated = false`，但把成功 no-op 记录为 `ok`/`info`，
并让测试覆盖其不会出现在 warn/error 诊断集合中。

kongche-jbw (commented):
Review baseline: `b55dbcf95213...bb1e09b35b7b`

**[P1] 避免合法 URL 字符绕过凭据脱敏**

`src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs:1853` 把单引号当作 URL
结束符，但 RFC 3986 允许它出现在 userinfo 和 query 中。当
`ANOLISA_UPDATE_URL=https://ci:s3'cr3t@host/release.toml?...` 请求失败时，扫描会在
密码中截断，后半段随后原样追加。我用真实 `anolisa update self` 路径复现，
`central.jsonl` 同时保留了 `s3'cr3t` 和签名值，错误记录还会进入 `anolisa bug`。
Possible direction: 使用 URL-aware sanitizer，或至少不要把合法 URI 字符当分隔符，
并让回归测试通过真实 handler 覆盖 userinfo/query 中的单引号。

kongche-jbw (commented):
Review baseline: `b55dbcf95213...ce6d8aec52d0`

**[P1] 不要在凭据中途停止 URL 脱敏**

`src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs:1875`
在首个非 RFC 3986 ASCII 字符处截断候选 URL。真实运行
`ANOLISA_UPDATE_URL=https://ci:s3"cr3t@mirror.invalid/release.toml?...`
并让请求失败后，`central.jsonl` 仍保留完整密码和签名；该错误记录随后还会进入
`anolisa bug`。同一实现遇到 URL 前的非 ASCII 字符时，1870–1872 行的
`i + 1` 还会落在 UTF-8 字符中间并 panic，覆盖掉原本的请求错误。
Possible direction: 对结构化 URL 字段做无凭据格式化，避免从自由文本猜测边界；
补充双引号、换行和 Unicode 经真实 handler 写日志的回归测试。

kongche-jbw (commented):
Review baseline: `b55dbcf95213...a068bb2cb027`

**[P1] 对无法安全拆分的更新 URL 采用失败闭合**

`src/anolisa/crates/anolisa-cli/src/commands/tier1/update.rs:1887` 在 `/`、`?` 或
`#` 处结束 authority。若运维配置
`ANOLISA_UPDATE_URL=https://ci:s3/cr3t@mirror.invalid/release.toml`，这里会在
`@` 之前截断，既提取不到 userinfo，也没有 query 可替换。我用真实
`anolisa update self` 路径复现后，ureq 虽以 invalid port 拒绝请求，中央日志仍保留
完整 URL；该 error 记录随后还会进入 `anolisa bug`，因此原始凭据仍可能被持久化和外发。
Possible direction: 无法可靠解析为 scheme/host/path 时整条替换，补充含原始
`/`、`?`、`#` 的真实 handler 回归，并在反序列化后的 message 上断言凭据不存在。

ikunkun-sys (commented):
Confirmed and fixed in `b675fb75`.

The finding is right. `credential_free` accepted any marker suffix without `=` or `&`, so a query the transport rewrote or extended — `?sig=abc` matched against `?sig=abcSECRET` — left `?<redacted>SECRET` and passed verification. I reproduced it before fixing: the residue `SECRET` was accepted, and the text would have been persisted and exported by `anolisa bug`.

The query now has to end at the marker. Only punctuation may trail it (`: , . ; ! ? ) ] } ' "`), which keeps ordinary failure text readable — the message goes on after the URL — while ruling out a residue that could carry a secret, since a credential cannot be spelled in punctuation alone.

That loose suffix rule was the relaxation I added last round to stop `?<redacted>:` from over-triggering the fallback; the punctuation set covers that case without the hole.

Coverage: `a_query_surviving_past_the_marker_is_withheld` asserts the residue is withheld. I checked it bites — reverting to the `=`/`&` rule makes it fail. `a_message_without_credentials_keeps_its_transport_detail` still passes, so the fallback has not become the common path.

End-to-end with the real binary, `ANOLISA_UPDATE_URL` carrying a query the transport percent-encodes: the record falls back to withheld, `details.endpoint` still carries `https://mirror.invalid/release.toml`, and nothing under the state root contains the secret.

Gate: fmt, clippy `-D warnings`, doc 0 warnings, 2367 tests passing.

kongche-jbw (changes requested):
Review baseline: `b55dbcf95213...b675fb75710d`

**[P1] URL path 中的凭据仍会落盘**

`credential_parts` 只移除 userinfo/query，`endpoint_without_credentials` 还会保留完整 path。对于 `https://host/download/<bearer>/release.toml` 这类 path token，失败消息和 `details.endpoint` 都会保留凭据，随后进入 `anolisa bug`。请不要默认把任意 path 视为无凭据。

**[P1] 纯标点查询残段仍可绕过验证**

`credential_free` 允许 `<redacted>` 后跟任意 `!`、`?`、引号等标点。`sig=abc!!!` 被部分替换后会留下 `<redacted>!!!` 并通过验证，但 `!!!` 可以是合法 query 凭据后缀。marker 后仍有 URL 内容时应失败闭合。

**[P2] bug bundle 丢失新增的失败上下文**

目标版本、apply mode 和 RPM 前后版本只写在 `LogRecord.details`，而 `bug::summarize_log` 不复制 `details`。因此失败记录虽被选中，最终 bug JSON/Markdown 仍没有这些诊断信息。请让摘要携带经过脱敏的相关字段，并用真实 bug payload 断言内容。

ikunkun-sys (commented):
All three confirmed and fixed in `473a8ab3` + `d3296b65`.

**[P1] Path 中的凭据 — 确认。** Reproduced with the real binary: `ANOLISA_UPDATE_URL=https://mirror.invalid/download/bearer-PATHTOKEN/release.toml` put the token in both the message and `details.endpoint`.

**[P1] 纯标点残段 — 确认。** "punctuation cannot spell a credential" was an assumption about credential alphabets, not something I can prove. It was also the one acceptance surface I had already flagged as non-zero.

Both come from the same mistake, so they get one fix rather than two patches: I was classifying parts of the endpoint as safe or unsafe, and each round has found another part I classified wrong. That approach is unsound — which part carries a secret is not knowable from the outside.

So the whole URL goes now, replaced out to the next whitespace. A URL in free text ends at whitespace and a credential cannot contain any, so the run covers whatever the transport appended to the value we know — `?sig=abc` against `?sig=abcSECRET` no longer leaves the parameter standing, and there is no residue left to judge. Verification falls out of it: any URL still in the text was never matched, so it cannot be vouched for and the text is withheld. `credential_parts`, `credential_free` and `is_message_punctuation` are gone with it.

`details.endpoint` is now scheme and host only. A host has to be routable, so it is already in DNS and on the wire; everything after it is dropped.

**[P2] bug bundle 丢失 details — 确认.** `summarize_log` copied every field but `details`. It now carries them under the same redaction as the rest, walking the value so each string is filtered as free text, and dropping any value whose key names a credential — JSON keeps key and value apart, so the `key=value` scan cannot see the pairing on its own.

Verification, end-to-end with the real binary (`update self` then `bug --json`):

| endpoint | message | `details.endpoint` | bundle |
|---|---|---|---|
| path token | `<redacted>` | `https://mirror.invalid` | `{current_version, endpoint}` |
| userinfo + query | `<redacted>` | `https://mirror.invalid` | same |
| transport-rewritten | withheld | `https://mirror.invalid` | same |

No occurrence of the secret anywhere under the state root in any case, and the bundle now carries the details, which answers the P2 on a real payload.

Tests: `a_credential_in_the_url_path_is_not_persisted`, `a_query_extended_past_the_known_value_is_swallowed`, `bug_payload_carries_redacted_log_details` (goes through the log file, the query, the summary and the renderer). I checked all three bite by reverting each fix individually.

One trade-off worth naming: the message no longer shows the path, so `release.toml` is not visible in the record. The host survives via `details.endpoint` and, in practice, via the transport reason. I would rather lose the path than keep guessing which segment of it is a token.

Gate: fmt, clippy `-D warnings`, doc 0 warnings, 2369 tests passing.

kongche-jbw (commented):
Follow-up baseline: `b675fb75710d...d3296b652c3e`

感谢快速补齐，上轮三个问题的直接场景现在都已覆盖。还有一处想请再收窄一下：

**[P1] 避免把所有 central-log details 通用导出到 bug bundle**

`redact_details` 目前会携带所有 warn/error 记录的 `details`，不只 self-update。现有 hook 失败记录会在 `details.stderr_tail` 保存最多 4KB 原始子进程输出；这个 key 不触发 credential-key 规则，而裸 `Bearer …`、私钥片段或其他无 `key=value` 形态的凭据也不会被字符串扫描移除。因此这次改动会让此前不会进入 bundle 的任意 stderr 首次被导出。

建议只白名单携带 self-update 所需的结构化字段，或按明确的 details schema 做逐类脱敏，避免通用导出自由文本字段。

ikunkun-sys (commented):
Confirmed and fixed in `bc0f7b91`.

You are right, and this one was a regression I introduced rather than a gap I failed to close. I verified all three parts of the claim:

- `hooks.rs:361` takes `tail_lossy(&output.stderr, 4096)` and `:430` puts it in `details.stderr_tail`; the record carries `Severity::Warn`, so `collect_recent_logs` selects it.
- `stderr_tail` does not match any credential-key rule.
- The string scan does not save it either. `redact_key_values` requires the keyword to be followed by `=` or `:`, so `Authorization: Bearer eyJ...`, a `-----BEGIN OPENSSH PRIVATE KEY-----` block (the list has `private_key`, the header reads `PRIVATE KEY`) and a bare `ghp_...` all pass through untouched.

So generic export would have made arbitrary hook stderr leave the machine for the first time, in a document whose whole purpose is to be pasted into a public issue.

Took the allowlist option. Per-type schema redaction would mean building a registry of `details` shapes for one consumer, and it would still fail open for any command that adds a field without registering it. An allowlist fails closed on exactly that case, which is the right default here: omitting a field costs a follow-up, including one costs a credential in a public issue.

`EXPORTABLE_DETAIL_KEYS` lists the self-update fields only — `apply_mode`, `current_version`, `endpoint`, `latest_version`, `package`, `rpm_version_after`, `rpm_version_before`, `updated`. Allowlisted strings still go through `redact_sensitive`. Everything else is dropped before the summary is built, so it reaches neither the JSON payload nor the Markdown. `redact_json` and `names_credential` are gone; `redact_sensitive` is back to what it was, so the net diff against main is now only the allowlist and the field itself.

Verified end-to-end with the real binary — a self-update failure and a hook-failure record in one log, then `anolisa bug --json`:

```
update self         details: {"current_version": "0.3.8", "endpoint": "https://mirror.invalid"}
hook:post-install   details: null
```

No `LEAKED` marker from the stderr payload anywhere in the bundle.

Tests: `free_form_details_stay_out_of_the_bundle` builds a record shaped like the one `hooks.rs` emits, with a bearer value and a PEM block in `stderr_tail`, and asserts nothing of it reaches the payload or the Markdown; `bug_payload_carries_redacted_log_details` now asserts the non-allowlisted key is absent rather than masked. Both bite — reverting the allowlist to a pass-through fails them.

Gate: fmt, clippy `-D warnings`, doc 0 warnings, 2370 tests passing.

kongche-jbw (commented):
Follow-up baseline: `d3296b652c3e...bc0f7b9129c0`

本轮增量代码已复核，未发现新的 correctness 或 security 问题；allowlist 也覆盖了前一轮指出的自由文本导出风险。

一个合并前的非阻塞建议：当前 PR 累积了多笔用于修复本 PR 前序提交的独立 `fix` commit。按仓库 AGENTS.md §13 的 fix attribution rule，这类修复建议用 `fixup` / `autosquash` 合并回引入改动的提交，再进入最终合并流程。

kongche-jbw (approved):

PR #2996, opened 2026-08-31
fix(cosh-ng): [shell] harden secret gate

### LINKED ISSUE #2155: [cosh-ng] 密钥形态判定器 _cosh_command_has_secret 误报/漏报双向失效
## 背景

`crates/cosh-shell/src/shell_host/marker/bash.rs` / `zsh.rs` 中的 `_cosh_command_has_secret` 是 shell 侧唯一的密钥形态判定器，一个 glob-substring 匹配结果同时驱动多个后果：sensitive 路由标记（#2154 之后：journal 整字段脱敏的触发源）、shell history 排除、failed-command recovery 排除、preexec 脱敏占位符（#2134 输出侧）。

在 #2138 的根因分析中对该判定器做了函数级边界矩阵对照（判定器原样抽出、bash 直接执行），实证其在两个方向上都失效。

## 误报（实测 5/6）：无密钥的正常中文运维问题被标记 sensitive

| 输入 | 命中原因 |
|------|----------|
| `sk-hynix 内存条什么价格` | `*" sk-"*` 通配 |
| `npm_package_version 怎么读` | `*npm_*` 子串 |
| `cookie: 头是干什么的` | `*cookie:*` 子串 |
| `bearer 是什么认证方式` | `*bearer *` 子串 |
| `shf_test 文件在哪` | `*hf_*` 子串 |

根因：glob 子串通配没有 token 边界概念，把"讨论这些词"与"携带这类凭据"混为一谈。

#2154 合入前的后果是整句被静默 veto（NL 路由被杀）；合入后降级为：句子照常进 agent，但本地 journal 的 input 被整字段 `<redacted>`（本地证据可读性受损，功能不再受损）。

## 漏报（实测 4/4）：中文语境下最常见的密钥措辞全部绕过

| 输入 | 未命中原因 |
|------|------------|
| `key是sk-abc123` | `*" sk-"*` 要求 ASCII 空格前缀，汉字紧邻 `sk-` 不命中 |
| `（sk-abc123）` | 中文括号紧邻，同上 |
| `密钥为sk-abc123` | 同上 |
| `key：sk-abc123` | 中文冒号紧邻，同上 |

漏报的后果在 #2154 合入前后一致且更严重：输入不会被标记 sensitive，journal 整字段脱敏不触发，只能依赖 Rust 侧 `redact_sensitive_text` 的正则兜底——而 `sk-abc123` 这类短 key 低于其 opaque-token 最小长度阈值，**密钥明文会落进本地 journal**（且原文照常进模型）。

## 问题定性

这个判定器的失效方向是自相矛盾的：

- 对可用性 fail-closed：误报杀掉正常输入的本地证据可读性（#2154 前直接杀功能）；
- 对它唯一的存在理由——隐私——却 fail-open：最常见的中文措辞恰好全部绕过。

## 与既有 issue 的关系

- #2138 / PR #2154：输入侧路由已修复（敏感 NL 进 agent + journal 整字段脱敏），但脱敏是否触发仍以本判定器的命中为前提——本 issue 的漏报直接削弱该防线；
- #2134：输出侧 handoff 死锁的触发源是同一判定器（preexec 命中后的整行脱敏占位符破坏闭合匹配），死锁修复（token 认领协议）在该 issue 主线，与判定器质量解耦。

## 修复方向建议

- 给密钥形态判定加 token 边界感知（至少：`sk-` 等前缀模式改为"任意非字母数字边界"而非硬编码 ASCII 空格，覆盖中文标点/汉字紧邻）；
- 常见误报子串（`npm_`、`hf_`、`cookie:`、`bearer ` 等）要求凭据形态的右侧上下文（长度/字符集），不再裸子串命中；
- bash/zsh 两份实现保持对称，加"shell 门命中形态"边界矩阵测试锚定（误报组 + 漏报组双向断言）；
- 验收应含 bash/zsh PTY 端到端与三入口（隐式 NL / `??` 显式 / command）同一密钥的一致性用例。

## 验证口径说明

上表数据来自判定器函数级 focused 对照（非 PTY 端到端），复现方法：从 bash.rs 原样抽出 `_cosh_command_has_secret` 在 bash 5.x 下逐句执行。

kongche-jbw:
## Why

The shell secret gate uses raw substring matches that redact normal prose while missing short credentials next to CJK text. Bash, Zsh, and Rust also disagree on JWT, assignment, Bearer, and Cookie shapes, leaving history and journal privacy dependent on the entry path.

## What changed

Replace substring checks with boundary-aware credential shapes shared by the Bash and Zsh contracts. Align Rust redaction for short `sk-` tokens and structured, quoted, escaped, shell-word-concatenated, or curl-option-attached Cookie forms. Attached curl headers now decode static option prefixes before accepting `--header=` or any single short-option cluster ending in `H`, including `-H`, `-sH`, and `-fsSLH`. Static Cookie shell words retain independent trailing arguments; dynamic prefixes or values, backticks, and ambiguous compact multi-pair headers redact the whole field.

## Related issue

Closes #2155

## User / Agent impact

Normal prose such as `sk-hynix`, `npm_package_version`, and Cookie/Bearer questions remains visible in local evidence. Short CJK-adjacent credentials and structured secret assignments are consistently excluded from native history and redacted before journal persistence.

## Risk and compatibility

The change intentionally adjusts the security-sensitive classification boundary but does not change public CLI, configuration, or cross-component contracts. Static Cookie values preserve unrelated command arguments; commands with dynamic Cookie construction may lose provider-facing context because privacy takes precedence when the shell-word boundary cannot be proven.

## Validation

Local Linux aarch64 on commit `b26e7b71`:

- `cargo test --package cosh-shell --lib evidence::redaction::tests -- --nocapture` — 8 passed.
- `cargo test --package cosh-shell --lib redaction_boundary_tests -- --nocapture` — 3 passed.
- `cargo test --package cosh-shell --test shell_host shell_secret_detectors_match_canonical_shapes_and_boundaries -- --nocapture` — 1 passed.
- `cargo test --locked -p cosh-shell --test shell_host raw_relay_bash_excludes_secrets_from_history_and_journal -- --nocapture --test-threads=1` — 1 passed.
- `cargo test --locked -p cosh-shell --test shell_host raw_relay_zsh_excludes_secrets_from_history_and_journal -- --nocapture --test-threads=1` — 1 passed.
- `cargo clippy --package cosh-shell --lib -- -D warnings` — passed.
- `cargo clippy --package cosh-shell --test shell_host -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- `bash -n crates/cosh-shell/src/shell_host/marker/bash.sh` — passed.
- `zsh -n crates/cosh-shell/src/shell_host/marker/zsh_secret_detector.zsh` — passed.
- `crates/cosh-shell/scripts/check-layout.sh` — passed.
- `git diff --check` — passed.

The regression matrix covers escaped-space, escaped-colon, adjacent-quote, attached `-H`, combined `-sH` / `-fsSLH`, statically quoted or escaped option prefixes, direct `-HCookie:`, and `--header=` spellings in Rust, Bash, and Zsh. It also covers dynamic option-prefix fail-closed behavior, nested substitution operators (`;`, `&&`, and pipelines), dynamic `then case`, comment and here-doc forms, compact multi-pair headers, safe internal-substring controls, and preservation of trailing arguments. Full workspace gates and provider validation were not run because this is a focused redaction change; broader coverage remains with CI.

## Documentation and rollback

No documentation changes are needed because this corrects internal privacy classification behavior. Roll back by reverting commit `b26e7b71`.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
@@ -5,6 +5,10 @@ use regex::Regex;
 use crate::tools::{classify_command_interaction, OutputStability};
 use crate::types::{CommandBlock, CommandStatus};
 
+mod shell_word;
+
+use shell_word::redact_shell_cookie_headers;
+
 const PROVIDER_COMMAND_MAX_BYTES: usize = 4 * 1024;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -93,12 +97,20 @@ fn redact_home_path(value: &str) -> String {
 
 pub(crate) fn redact_sensitive_text(text: &str) -> (String, bool) {
     let (mut redacted, mut changed) = redact_private_key_blocks(text);
+    let next = conservative_cookie_header_pattern()
+        .replace_all(&redacted, "$prefix<redacted>")
+        .into_owned();
+    changed |= next != redacted;
+    redacted = next;
+    let (next, cookie_changed) = redact_shell_cookie_headers(&redacted);
+    redacted = next;
+    changed |= cookie_changed;
     for (pattern, replacement) in [
-        (cookie_header_pattern(), "$prefix<redacted>"),
         (authorization_pattern(), "$prefix$scheme <redacted>"),
         (bearer_pattern(), "$prefix<redacted>"),
         (url_password_pattern(), "$prefix<redacted>@"),
         (sensitive_flag_pattern(), "$prefix<redacted>"),
+        (sensitive_cookie_assignment_pattern(), "$prefix<redacted>"),
         (sensitive_assignment_pattern(), "$prefix<redacted>"),
         (github_token_pattern(), "<redacted>"),
         (opaque_token_pattern(), "<redacted>"),
@@ -152,12 +164,13 @@ fn private_key_marker_range(line: &str, marker: &str) -> Option<(usize, usize)> 
     Some((start, start + marker_end + "PRIVATE KEY-----".len()))
 }
 
-fn cookie_header_pattern() -> &'static Regex {
+fn conservative_cookie_header_pattern() -> &'static Regex {
     static PATTERN: OnceLock<Regex> = OnceLock::new();
     PATTERN.get_or_init(|| {
-        // The pattern is a compile-time constant covered by the tests below.
-        Regex::new(r"(?im)(?P<prefix>^(?:set-cookie|cookie)\s*:\s*).*$")
-            .unwrap_or_else(|_| unreachable!("static cookie pattern must compile"))
+        Regex::new(
+            r#"(?im)(?P<prefix>(?:^(?:set-cookie|cookie)\s*:\s*|\s(?:set-cookie|cookie)\s*:\s+))[^=\s;,]+\s*=\s*[^\r\n]*"#,
+        )
+        .unwrap_or_else(|_| unreachable!("static cookie pattern must compile"))
     })
 }
 
@@ -185,7 +198,7 @@ fn url_password_pattern() -> &'static Regex {
     static PATTERN: OnceLock<Regex> = OnceLock::new();
     PATTERN.get_or_init(|| {
         // The pattern is a compile-time constant covered by the tests below.
-        Regex::new(r"(?i)(?P<prefix>(?-u:\b)[a-z][a-z0-9+.-]*://[^/\s:@]+:)[^@/\s]+@")
+        Regex::new(r"(?i)(?P<prefix>(?-u:\b)[a-z][a-z0-9+.-]*://[^/\s:@]*:)[^@/\s]+@")
             .unwrap_or_else(|_| unreachable!("static URL password pattern must compile"))
     })
 }
@@ -230,8 +243,7 @@ fn sensitive_assignment_pattern() -> &'static Regex {
                    dashscope[_-]?api[_-]?key|openai[_-]?api[_-]?key|
                    client[_-]?secret|security[_-]?token|refresh[_-]?token|
                    access[_-]?token|github[_-]?token|id[_-]?token|
-                   password|passphrase|passwd|api[_-]?key|apikey|token|secret|
-                   cookie|set[_-]?cookie)
+                   password|passphrase|passwd|api[_-]?key|apikey|token|secret)
                 ["']?
                 \s*(?:=|:)\s*
             )
@@ -247,6 +259,28 @@ fn sensitive_assignment_pattern() -> &'static Regex {
     })
 }
 
+fn sensitive_cookie_assignment_pattern() -> &'static Regex {
+    static PATTERN: OnceLock<Regex> = OnceLock::new();
+    PATTERN.get_or_init(|| {
+        // Cookie colons are reserved for structured JSON or name=value headers.
+        Regex::new(
+            r#"(?ix)
+            (?P<prefix>
+                (?:^|[^A-Za-z0-9_-])["']?(?:cookie|set[_-]?cookie)["']?\s*=\s*|
+                ["'](?:cookie|set[_-]?cookie)["']\s*:\s*
+            )
+            (?:
+                "(?:\\(?:\r?\n|[^\r\n])|[^"\\])*(?:"|$)|
+                '[^']*(?:'|$)|
+                \\(?:\r?\n|[^\r\n])|
+                [^\s;&|()<>"'\\]
+            )+
+            "#,
+        )
+        .unwrap_or_else(|_| unreachable!("static cookie assignment pattern must compile"))
+    })
+}
+
 fn github_token_pattern() -> &'static Regex {
     static PATTERN: OnceLock<Regex> = OnceLock::new();
     PATTERN.get_or_init(|| {
@@ -259,9 +293,9 @@ fn github_token_pattern() -> &'static Regex {
 fn opaque_token_pattern() -> &'static Regex {
     static PATTERN: OnceLock<Regex> = OnceLock::new();
     PATTERN.get_or_init(|| {
-        // The pattern is a compile-time constant covered by the tests below.
+        // A five-character sk- suffix needs a digit to exclude sk-hynix prose.
         Regex::new(
-            r"(?-u:\b)(?:sk-[A-Za-z0-9_-]{10,}|sk_(?:live|test)_[A-Za-z0-9]{10,}|glpat-[A-Za-z0-9_-]{10,}|npm_[A-Za-z0-9]{20,}|hf_[A-Za-z0-9]{20,}|AIza[A-Za-z0-9_-]{20,}|(?i:xox.)-[A-Za-z0-9-]{10,})(?-u:\b)",
+            r"(?-u:\b)(?:sk-(?:[A-Za-z0-9_-]{6,}|[0-9][A-Za-z0-9_-]{4}|[A-Za-z0-9_-][0-9][A-Za-z0-9_-]{3}|[A-Za-z0-9_-]{2}[0-9][A-Za-z0-9_-]{2}|[A-Za-z0-9_-]{3}[0-9][A-Za-z0-9_-]|[A-Za-z0-9_-]{4}[0-9])|sk_(?:live|test)_[A-Za-z0-9]{10,}|glpat-[A-Za-z0-9_-]{10,}|npm_[A-Za-z0-9]{20,}|hf_[A-Za-z0-9]{20,}|AIza[A-Za-z0-9_-]{20,}|(?i:xox.)-[A-Za-z0-9-]{10,})(?-u:\b)",
         )
         .unwrap_or_else(|_| unreachable!("static opaque token pattern must compile"))
     })
@@ -348,9 +382,10 @@ mod tests {
             "cookie=assignment-cookie set-cookie=assignment-set-cookie ",
             "'https://example.test/?client_secret=query-value&next=ok'\n",
             "https://user:url-password@example.test/path\n",
-            r#"{"api_key":"json-value","password":"json-password"}"#,
+            r#"{"api_key":"json-value","password":"json-password","cookie":"json-cookie-value"}"#,
             "\nAuthorization: Basic dXNlcjpwYXNz\n",
             "Cookie: session=private; other=value\n",
+            "Set-Cookie: csrf=set-cookie-header-value; Path=/\n",
         );
 
         let (redacted, changed) = redact_sensitive_text(input);
@@ -371,9 +406,11 @@ mod tests {
             "url-password",
             "json-value",
             "json-password",
+            "json-cookie-value",
             "dXNlcjpwYXNz",
             "session=private",
             "other=value",
+            "set-cookie-header-value",
         ] {
             assert!(!redacted.contains(secret), "{redacted}");
         }
--- a/src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
@@ -1,0 +1,556 @@
+//! Finds static Cookie header shell-word boundaries without executing shell syntax.
+
+#[derive(Clone, Copy)]
+enum InitialContext {
+    Unquoted,
+    Quoted(u8),
+}
+
+#[derive(Clone, Copy, Eq, PartialEq)]
+enum ShellContext {
+    Unquoted,
+    SingleQuoted,
+    DoubleQuoted,
+}
+
+enum ScanResult {
+    Precise {
+        value_len: usize,
+        quote_closed: bool,
+    },
+    Dynamic,
+}
+
+struct CookieHeaderMatch {
+    start: usize,
+    prefix_end: usize,
+    value_start: usize,
+    prefix_quote: Option<u8>,
+    initial_context: InitialContext,
+}
+
+enum HeaderSearchResult {
+    Match(CookieHeaderMatch),
+    Dynamic,
+    None,
+}
+
+enum HeaderParseResult {
+    Match(CookieHeaderMatch),
+    Dynamic,
+    NoMatch,
+}
+
+struct StaticChar {
+    byte: u8,
+    raw_start: usize,
+    raw_end: usize,
+    context_before: ShellContext,
+    context_after: ShellContext,
+}
+
+enum StaticCharResult {
+    Char(StaticChar),
+    Dynamic,
+    End(ShellContext),
+}
+
+enum CandidateStartResult {
+    Match(ShellContext),
+    Dynamic,
+    NoMatch,
+}
+
+enum StaticOptionPrefix {
+    Decoded {
+        bytes: Vec<u8>,
+        context: ShellContext,
+    },
+    Dynamic,
+}
+
+/// Redacts static Cookie header values through their complete shell word.
+///
+/// Dynamic shell syntax is redacted as a whole because finding its boundary
+/// requires the complete Bash and Zsh grammars.
+pub(super) fn redact_shell_cookie_headers(text: &str) -> (String, bool) {
+    let mut output = String::with_capacity(text.len());
+    let mut cursor = 0;
+    let mut changed = false;
+
+    loop {
+        let header = match find_cookie_header(text, cursor) {
+            HeaderSearchResult::Match(header) => header,
+            HeaderSearchResult::Dynamic => return ("<redacted>".to_string(), true),
+            HeaderSearchResult::None => break,
+        };
+        let ScanResult::Precise {
+            value_len,
+            quote_closed,
+        } = shell_word_value_len(&text[header.value_start..], header.initial_context)
+        else {
+            return ("<redacted>".to_string(), true);
+        };
+
+        output.push_str(&text[cursor..header.start]);
+        output.push_str(&text[header.start..header.prefix_end]);
+        output.push_str("<redacted>");
+        if let Some(prefix_quote) = header.prefix_quote {
+            let value_keeps_prefix_quote = matches!(header.initial_context, InitialContext::Quoted(quote) if quote == prefix_quote);
+            if !value_keeps_prefix_quote || quote_closed {
+                output.push(prefix_quote as char);
+            }
+        }
+        cursor = header.value_start + value_len;
+        changed = true;
+    }
+    output.push_str(&text[cursor..]);
+
+    (output, changed)
+}
+
+fn find_cookie_header(text: &str, search_start: usize) -> HeaderSearchResult {
+    let bytes = text.as_bytes();
+    for start in search_start..bytes.len() {
+        if !could_start_cookie_header(bytes[start]) {
+            continue;
+        }
+        match is_cookie_header_candidate_start(bytes, start) {
+            CandidateStartResult::Match(context) => {
+                match parse_cookie_header(bytes, start, context) {
+                    HeaderParseResult::Match(header) => {
+                        return HeaderSearchResult::Match(header);
+                    }
+                    HeaderParseResult::Dynamic => return HeaderSearchResult::Dynamic,
+                    HeaderParseResult::NoMatch => {}
+                }
+            }
+            CandidateStartResult::Dynamic => {
+                for context in [
+                    ShellContext::Unquoted,
+                    ShellContext::SingleQuoted,
+                    ShellContext::DoubleQuoted,
+                ] {
+                    if !matches!(
+                        parse_cookie_header(bytes, start, context),
+                        HeaderParseResult::NoMatch
+                    ) {
+                        return HeaderSearchResult::Dynamic;
+                    }
+                }
+            }
+            CandidateStartResult::NoMatch => {}
+        }
+    }
+
+    HeaderSearchResult::None
+}
+
+fn could_start_cookie_header(byte: u8) -> bool {
+    matches!(byte, b'c' | b'C' | b's' | b'S' | b'\'' | b'"' | b'\\')
+}
+
+fn is_cookie_header_candidate_start(bytes: &[u8], start: usize) -> CandidateStartResult {
+    if start == 0 || is_shell_word_boundary(bytes[start - 1]) {
+        return CandidateStartResult::Match(ShellContext::Unquoted);
+    }
+
+    let word_start = bytes[..start]
+        .iter()
+        .rposition(|byte| is_shell_word_boundary(*byte))
+        .map_or(0, |boundary| boundary + 1);
+    match decode_static_option_prefix(&bytes[word_start..start]) {
+        StaticOptionPrefix::Decoded {
+            bytes: option_prefix,
+            context,
+        } if option_prefix == b"--header=" || is_attached_short_header_option(&option_prefix) => {
+            CandidateStartResult::Match(context)
+        }
+        StaticOptionPrefix::Dynamic => CandidateStartResult::Dynamic,
+        StaticOptionPrefix::Decoded { .. } => CandidateStartResult::NoMatch,
+    }
+}
+
+fn decode_static_option_prefix(raw_prefix: &[u8]) -> StaticOptionPrefix {
+    let mut decoded = Vec::with_capacity(raw_prefix.len());
+    let mut cursor = 0;
+    let mut context = ShellContext::Unquoted;
+
+    loop {
+        match next_static_char(raw_prefix, cursor, context) {
+            StaticCharResult::Char(character) => {
+                decoded.push(character.byte);
+                cursor = character.raw_end;
+                context = character.context_after;
+            }
+            StaticCharResult::Dynamic => return StaticOptionPrefix::Dynamic,
+            StaticCharResult::End(final_context) => {
+                return StaticOptionPrefix::Decoded {
+                    bytes: decoded,
+                    context: final_context,
+                };
+            }
+        }
+    }
+}
+
+fn is_attached_short_header_option(option_prefix: &[u8]) -> bool {
+    option_prefix.starts_with(b"-")
+        && !option_prefix.starts_with(b"--")
+        && option_prefix.ends_with(b"H")
+        && option_prefix[1..]
+            .iter()
+            .all(|byte| byte.is_ascii_alphanumeric() || *byte == b'#')
+}
+
+fn parse_cookie_header(
+    bytes: &[u8],
+    start: usize,
+    initial_context: ShellContext,
+) -> HeaderParseResult {
+    for expected_prefix in [b"set-cookie:".as_slice(), b"cookie:".as_slice()] {
+        let mut cursor = start;
+        let mut context = initial_context;
+        if !consume_static_literal(bytes, &mut cursor, &mut context, expected_prefix) {
+            continue;
+        }
+
+        match consume_static_whitespace(bytes, &mut cursor, &mut context) {
+            StaticCharResult::Dynamic => return HeaderParseResult::Dynamic,
+            StaticCharResult::Char(_) | StaticCharResult::End(_) => {}
+        }
+        let prefix_end = cursor;
+        let prefix_quote = context_quote(context);
+        let mut name_seen = false;
+
+        loop {
+            match next_static_char(bytes, cursor, context) {
+                StaticCharResult::Char(character) if character.byte == b'=' && name_seen => {
+                    return cookie_header_match(start, prefix_end, prefix_quote, character);
+                }
+                StaticCharResult::Char(character)
+                    if character.byte.is_ascii_whitespace() && name_seen =>
+                {
+                    match consume_static_whitespace(bytes, &mut cursor, &mut context) {
+                        StaticCharResult::Dynamic => return HeaderParseResult::Dynamic,
+                        StaticCharResult::Char(_) | StaticCharResult::End(_) => {}
+                    }
+                    return match next_static_char(bytes, cursor, context) {
+                        StaticCharResult::Char(character) if character.byte == b'=' => {
+                            cookie_header_match(start, prefix_end, prefix_quote, character)
+                        }
+                        StaticCharResult::Dynamic => HeaderParseResult::Dynamic,
+                        StaticCharResult::Char(_) | StaticCharResult::End(_) => {
+                            HeaderParseResult::NoMatch
+                        }
+                    };
+                }
+                StaticCharResult::Char(character)
+                    if matches!(character.byte, b'=' | b';' | b',') =>
+                {
+                    return HeaderParseResult::NoMatch;
+                }
+                StaticCharResult::Char(character) => {
+                    name_seen = true;
+                    cursor = character.raw_end;
+                    context = character.context_after;
+                }
+                StaticCharResult::Dynamic => return HeaderParseResult::Dynamic,
+                StaticCharResult::End(_) => return HeaderParseResult::NoMatch,
+            }
+        }
+    }
+
+    HeaderParseResult::NoMatch
+}
+
+fn cookie_header_match(
+    start: usize,
+    prefix_end: usize,
+    prefix_quote: Option<u8>,
+    equals: StaticChar,
+) -> HeaderParseResult {
+    HeaderParseResult::Match(CookieHeaderMatch {
+        start,
+        prefix_end,
+        value_start: equals.raw_end,
+        prefix_quote,
+        initial_context: initial_context(equals.context_after),
+    })
+}
+
+fn consume_static_literal(
+    bytes: &[u8],
+    cursor: &mut usize,
+    context: &mut ShellContext,
+    expected: &[u8],
+) -> bool {
+    for expected_byte in expected {
+        let StaticCharResult::Char(character) = next_static_char(bytes, *cursor, *context) else {
+            return false;
+        };
+        if !character.byte.eq_ignore_ascii_case(expected_byte) {
+            return false;
+        }
+        *cursor = character.raw_end;
+        *context = character.context_after;
+    }
+
+    true
+}
+
+fn consume_static_whitespace(
+    bytes: &[u8],
+    cursor: &mut usize,
+    context: &mut ShellContext,
+) -> StaticCharResult {
+    loop {
+        match next_static_char(bytes, *cursor, *context) {
+            StaticCharResult::Char(character) if character.byte.is_ascii_whitespace() => {
+                *cursor = character.raw_end;
+                *context = character.context_after;
+            }
+            StaticCharResult::Char(character) => {
+                *cursor = character.raw_start;
+                *context = character.context_before;
+                return StaticCharResult::Char(character);
+            }
+            other => return other,
+        }
+    }
+}
+
+fn next_static_char(
+    bytes: &[u8],
+    mut cursor: usize,
+    mut context: ShellContext,
+) -> StaticCharResult {
+    loop {
+        let Some(&byte) = bytes.get(cursor) else {
+            return StaticCharResult::End(context);
+        };
+        match context {
+            ShellContext::Unquoted => match byte {
+                b'\'' => {
+                    context = ShellContext::SingleQuoted;
+                    cursor += 1;
+                }
+                b'"' => {
+                    context = ShellContext::DoubleQuoted;
+                    cursor += 1;
+                }
+                b'\\' => match bytes.get(cursor + 1) {
+                    Some(b'\r') if bytes.get(cursor + 2) == Some(&b'\n') => cursor += 3,
+                    Some(b'\n') => cursor += 2,
+                    Some(&escaped) => {
+                        return StaticCharResult::Char(StaticChar {
+                            byte: escaped,
+                            raw_start: cursor,
+                            raw_end: cursor + 2,
+                            context_before: context,
+                            context_after: context,
+                        });
+                    }
+                    None => return StaticCharResult::End(context),
+                },
+                b'$' | b'`' => return StaticCharResult::Dynamic,
+                boundary if is_shell_word_boundary(boundary) => {
+                    return StaticCharResult::End(context);
+                }
+                _ => {
+                    return StaticCharResult::Char(StaticChar {
+                        byte,
+                        raw_start: cursor,
+                        raw_end: cursor + 1,
+                        context_before: context,
+                        context_after: context,
+                    });
+                }
+            },
+            ShellContext::SingleQuoted => {
+                if byte == b'\'' {
+                    context = ShellContext::Unquoted;
+                    cursor += 1;
+                } else {
+                    return StaticCharResult::Char(StaticChar {
+                        byte,
+                        raw_start: cursor,
+                        raw_end: cursor + 1,
+                        context_before: context,
+                        context_after: context,
+                    });
+                }
+            }
+            ShellContext::DoubleQuoted => match byte {
+                b'"' => {
+                    context = ShellContext::Unquoted;
+                    cursor += 1;
+                }
+                b'\\' => match bytes.get(cursor + 1) {
+                    Some(b'\r') if bytes.get(cursor + 2) == Some(&b'\n') => cursor += 3,
+                    Some(b'\n') => cursor += 2,
+                    Some(&escaped) if matches!(escaped, b'$' | b'`' | b'"' | b'\\') => {
+                        return StaticCharResult::Char(StaticChar {
+                            byte: escaped,
+                            raw_start: cursor,
+                            raw_end: cursor + 2,
+                            context_before: context,
+                            context_after: context,
+                        });
+                    }
+                    Some(_) => {
+                        return StaticCharResult::Char(StaticChar {
+                            byte,
+                            raw_start: cursor,
+                            raw_end: cursor + 1,
+                            context_before: context,
+                            context_after: context,
+                        });
+                    }
+                    None => return StaticCharResult::End(context),
+                },
+                b'$' | b'`' => return StaticCharResult::Dynamic,
+                _ => {
+                    return StaticCharResult::Char(StaticChar {
+                        byte,
+                        raw_start: cursor,
+                        raw_end: cursor + 1,
+                        context_before: context,
+                        context_after: context,
+                    });
+                }
+            },
+        }
+    }
+}
+
+fn initial_context(context: ShellContext) -> InitialContext {
+    match context {
+        ShellContext::Unquoted => InitialContext::Unquoted,
+        ShellContext::SingleQuoted => InitialContext::Quoted(b'\''),
+        ShellContext::DoubleQuoted => InitialContext::Quoted(b'"'),
+    }
+}
+
+fn context_quote(context: ShellContext) -> Option<u8> {
+    match context {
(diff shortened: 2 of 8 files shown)


src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
```diff
@@ -265,23 +273,21 @@ _cosh_command_has_secret() {
   canonical_assignment_key+='client[_-]?secret|security[_-]?token|refresh[_-]?token|'
   canonical_assignment_key+='access[_-]?token|github[_-]?token|id[_-]?token|'
   canonical_assignment_key+='password|passphrase|passwd|api[_-]?key|apikey|token|secret|'
-  canonical_assignment_key+='cookie|set[_-]?cookie)'
-  local canonical_assignment_pattern="(^|[^a-z0-9_-])['\"]?${canonical_assignment_key}['\"]?[[:space:]]*(=|:)[[:space:]]*[^[:space:]]"
-  if [[ "$lower" =~ $canonical_assignment_pattern ]]; then
+  canonical_assignment_key+='authorization)'
+  local assignment_value="['\"]?[^[:space:]]"
+  local canonical_assignment_pattern="(^|[^a-z0-9_-])['\"]?${canonical_assignment_key}['\"]?[[:space:]]*(=|:)[[:space:]]*${assignment_value}"
+  local canonical_flag_pattern="(^|[^a-z0-9_-])--${canonical_assignment_key}([[:space:]]+|=)[[:space:]]*${assignment_value}"
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 29bc63b5. Bash and Zsh now classify both `--cookie value` / `--cookie=value` and `--set-cookie value` as sensitive, with symmetric matrix coverage. `Test cosh-ng fast checks` and `Test cosh-ng` both pass on the updated head.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
```diff
@@ -265,23 +276,23 @@ _cosh_command_has_secret() {
   canonical_assignment_key+='client[_-]?secret|security[_-]?token|refresh[_-]?token|'
   canonical_assignment_key+='access[_-]?token|github[_-]?token|id[_-]?token|'
   canonical_assignment_key+='password|passphrase|passwd|api[_-]?key|apikey|token|secret|'
-  canonical_assignment_key+='cookie|set[_-]?cookie)'
-  local canonical_assignment_pattern="(^|[^a-z0-9_-])['\"]?${canonical_assignment_key}['\"]?[[:space:]]*(=|:)[[:space:]]*[^[:space:]]"
-  if [[ "$lower" =~ $canonical_assignment_pattern ]]; then
+  canonical_assignment_key+='authorization)'
+  local assignment_value="['\"]?[^[:space:]]"
+  local canonical_assignment_pattern="(^|[^a-z0-9_-])['\"]?${canonical_assignment_key}['\"]?[[:space:]]*(=|:)[[:space:]]*${assignment_value}"
```
ikunkun-sys: [P1] Keep prefixed credential assignments out of native history

This boundary treats `_` as part of the surrounding identifier, while this commit removes the old per-key substring fallback. On this head, both Bash and Zsh return safe for `MEMORY_OPENAI_API_KEY=supersecret`, `MY_TOKEN=supersecret`, `FOO_PASSWORD=supersecret`, and `AUTH_REFRESH_TOKEN=supersecret`; the parent Bash detector returned sensitive. `MEMORY_OPENAI_API_KEY` is a documented real credential variable in this repository. Since `_cosh_scrub_sensitive_native_history` deletes an entry only when this detector succeeds, the literal value can now remain in native history. The new `MY_OPENAI_API_KEY = public-name` safe control cannot distinguish that dummy value from an actual credential, so it codifies the leak.

ikunkun-sys (changes requested):
Requesting changes for one current-head privacy regression. The narrowed shell-side assignment boundary lets prefixed credential variables bypass native-history protection; see the inline finding. Focused Rust redaction tests and the Bash/Zsh detector matrix passed, but direct parent/current probes reproduce this regression on both shells.

Reviewed commit: `e3081236bde39c1020ac202d4d06ffb7f8516f17`

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
kongche-jbw: Fixed in a10a3573. The Bash/Zsh assignment boundary now accepts underscore-separated canonical credential suffixes, so `MEMORY_OPENAI_API_KEY`, `MY_TOKEN`, `FOO_PASSWORD`, and `AUTH_REFRESH_TOKEN` are sensitive without restoring the old substring fallback. The previous `MY_OPENAI_API_KEY = public-name` control is now positive; `MY_OPENAI_API_KEY_LABEL` and `MY_TOKEN_COUNT` pin the non-suffix negative boundary. The focused symmetric detector matrix passes, and both `Test cosh-ng fast checks` and `Test cosh-ng` pass on the updated head.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
```diff
@@ -265,23 +276,24 @@ _cosh_command_has_secret() {
   canonical_assignment_key+='client[_-]?secret|security[_-]?token|refresh[_-]?token|'
   canonical_assignment_key+='access[_-]?token|github[_-]?token|id[_-]?token|'
   canonical_assignment_key+='password|passphrase|passwd|api[_-]?key|apikey|token|secret|'
-  canonical_assignment_key+='cookie|set[_-]?cookie)'
-  local canonical_assignment_pattern="(^|[^a-z0-9_-])['\"]?${canonical_assignment_key}['\"]?[[:space:]]*(=|:)[[:space:]]*[^[:space:]]"
-  if [[ "$lower" =~ $canonical_assignment_pattern ]]; then
+  canonical_assignment_key+='authorization)'
+  local assignment_value="['\"]?[^[:space:]]"
+  # Real environment variables may prefix canonical keys with an underscore.
+  local canonical_assignment_pattern="(^|[^a-z0-9-])['\"]?${canonical_assignment_key}['\"]?[[:space:]]*(=|:)[[:space:]]*${assignment_value}"
```
ikunkun-sys: [P1] Keep unseparated credential suffix assignments out of native history

This boundary only restores underscore-separated prefixes. On this head, both Bash and Zsh still return safe for valid shell identifiers such as `AUTHTOKEN=supersecret`, `myAccessToken=supersecret`, `dbPassword=supersecret`, `serviceSecret=supersecret`, and `customApiKey=supersecret`; the parent detector returned sensitive for all five. An isolated Bash history probe confirms these entries are retained, whereas `MEMORY_OPENAI_API_KEY=supersecret` is now correctly removed. The Rust assignment redactor has no equivalent left-boundary restriction, so the native-history and journal paths still disagree and the previous privacy regression is only partially fixed.

ikunkun-sys (changes requested):
Requesting changes for one current-head privacy regression. The underscore-prefix cases from the previous review are fixed, but valid unseparated/camelCase credential suffix assignments still bypass the Bash/Zsh native-history gate; see the inline finding. Focused Rust redaction tests and the Bash/Zsh detector matrix pass, while direct parent/current and isolated-history probes reproduce the remaining gap.

Reviewed commit: `a10a3573d17dd595face20c0500ae7de700b9f7b`

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
kongche-jbw: Fixed in `88c23dc8`. Bash and Zsh now match canonical credential suffixes across complete valid shell assignment identifiers, covering `AUTHTOKEN`, `myAccessToken`, `dbPassword`, `serviceSecret`, and `customApiKey` without restoring arbitrary substring matching. `MY_OPENAI_API_KEY_LABEL` and `MY_TOKEN_COUNT` remain negative suffix controls. The focused symmetric detector test passed locally, and the updated head passed `Test cosh-ng fast checks`, `Test cosh-ng`, `Build cosh-ng release`, and `Test cosh-ng prebuilt action`.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
```diff
@@ -156,8 +157,10 @@ fn cookie_header_pattern() -> &'static Regex {
     static PATTERN: OnceLock<Regex> = OnceLock::new();
     PATTERN.get_or_init(|| {
         // The pattern is a compile-time constant covered by the tests below.
-        Regex::new(r"(?im)(?P<prefix>^(?:set-cookie|cookie)\s*:\s*).*$")
-            .unwrap_or_else(|_| unreachable!("static cookie pattern must compile"))
+        Regex::new(
+            r#"(?im)(?P<prefix>(?:^|[\s'"])(?:set-cookie|cookie)\s*:\s*)[^=\s;,]+\s*=\s*[^\r\n'"]*"#,
```
ikunkun-sys: [P1] Consume quoted Cookie values before replacing the header

This value class stops before either quote, so a valid quoted Cookie value makes the match end immediately after `=`. Running the current production `redact_sensitive_text` on `curl -H 'Cookie: session="quoted-cookie-value-42"'` returns `curl -H 'Cookie: <redacted>"quoted-cookie-value-42"'`: `changed` is true while the literal value remains, and `Set-Cookie` behaves the same way. Bash and Zsh classify these inputs as sensitive, but Rust-only redaction consumers can still persist or render the credential suffix.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
```diff
@@ -243,20 +243,31 @@ _cosh_command_has_secret() {
...
-    sk-*|sk_live_*|sk_test_*|*" sk-"*|*"=sk-"*|*":sk-"*|*"\"sk-"*|*"'sk-"*|*" sk_live_"*|*" sk_test_"*|*"=sk_live_"*|*"=sk_test_"*)
+    *"-----begin "*"private key-----"*)
       return 0
       ;;
   esac
-  # Keep recovered-history privacy aligned with Rust's canonical assignment detector.
+  # ASCII boundaries allow credentials next to CJK text without matching inside
+  # identifiers such as npm_package_version or shf_test.
+  local bearer_pattern='(^|[^a-z0-9_])bearer[[:space:]]+[a-z0-9._~+/=-]+([^a-z0-9_]|$)'
+  local url_password_pattern='[a-z][a-z0-9+.-]*://[^/[:space:]:@]+:[^@/[:space:]]+@'
```
ikunkun-sys: [P1] Keep empty-username URL passwords out of history

The required username characters before `:` exclude valid userinfo such as `http://:s3cr3t42@example.test/`. The parent detector classified this input as sensitive, while both current Bash and Zsh classify it as safe; curl accepts the URL and proceeds to the connection attempt. The current production Rust redactor also returns `changed=false` because it has the same restriction. Since native-history cleanup only deletes an entry when this detector succeeds, the password can remain in native history and persisted evidence.

ikunkun-sys (changes requested):
Requesting changes for two current-head privacy gaps. The assignment-suffix cases from the previous review are fixed, but quoted Cookie header values are only partially redacted, and valid URL userinfo with an empty username bypasses both shell detectors and the Rust redactor. See the inline findings.

Focused detector/redaction tests, Bash/Zsh history E2E, formatting, all-target Clippy, layout/inventory checks, and shell syntax pass.

Reviewed commit: `88c23dc85cd8f1397138dbf6a0163d1fc10e1125`

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
kongche-jbw: Fixed in `e1f1b5d5`. Rust now handles single-quoted, double-quoted, and unquoted Cookie header boundaries separately, so quoted Cookie/Set-Cookie values—including escaped inner quotes—are consumed completely while the outer shell quote is preserved. The reviewer input, the corresponding Set-Cookie form, a raw quoted header, and an escaped-quote form are covered by the focused redaction matrix. Local redaction tests passed (8 existing + 3 boundary tests), and all four cosh-ng CI jobs passed on the updated head.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
kongche-jbw: Fixed in `e1f1b5d5`. The Bash, Zsh, and Rust URL userinfo patterns now allow an empty username while still requiring a non-empty password. The exact `http://:s3cr3t42@example.test/` case is covered together with an empty-password negative control, the symmetric Bash/Zsh detector matrix, and both Bash/Zsh native-history/journal E2E tests. All focused local tests and the four cosh-ng CI jobs passed on the updated head.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
```diff
@@ -152,12 +155,34 @@ fn private_key_marker_range(line: &str, marker: &str) -> Option<(usize, usize)>
...
-fn cookie_header_pattern() -> &'static Regex {
+fn single_quoted_cookie_header_pattern() -> &'static Regex {
     static PATTERN: OnceLock<Regex> = OnceLock::new();
     PATTERN.get_or_init(|| {
-        // The pattern is a compile-time constant covered by the tests below.
-        Regex::new(r"(?im)(?P<prefix>^(?:set-cookie|cookie)\s*:\s*).*$")
-            .unwrap_or_else(|_| unreachable!("static cookie pattern must compile"))
+        // Keep the outer shell quote while consuming quotes inside Cookie values.
+        Regex::new(
+            r#"(?im)(?P<prefix>'(?:set-cookie|cookie)\s*:\s*)[^=\s;,]+\s*=\s*(?:\\(?:\r?\n|[^\r\n])|[^'\\\r\n])*"#,
```
ikunkun-sys: [P1] Redact adjacent shell-word segments in Cookie values

This pattern stops at the closing single quote and never consumes an adjacent unquoted or differently quoted segment, even though Bash and Zsh concatenate those segments into one header word. For `curl -H 'Cookie: session='quoted-cookie-value-42`, both shells produce `Cookie: session=quoted-cookie-value-42`, but the current production `redact_sensitive_text` returns `curl -H 'Cookie: <redacted>'quoted-cookie-value-42` with `changed=true`. The shell detector classifies the input as sensitive, but Rust-only redaction consumers can still persist or render the literal suffix. The same residual occurs for double-to-single concatenation and `Set-Cookie`; the new tests cover quotes within one segment but not shell word concatenation.

ikunkun-sys (changes requested):
Requesting changes for one current-head privacy gap. The prior quoted Cookie and empty-username URL findings are fixed, but valid Bash/Zsh adjacent shell-word segments still let a literal Cookie value survive the Rust redactor; see the inline finding.

Focused detector/redaction tests, Bash/Zsh history E2E, formatting, all-target Clippy, layout/inventory checks, and shell syntax pass.

Reviewed commit: `e1f1b5d5f294fe9963a4893a7a436a435bc99d83`

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
kongche-jbw: Fixed in `31c463e0`. The quoted Cookie/Set-Cookie patterns now preserve the outer header quote while consuming every immediately adjacent unquoted, single-quoted, double-quoted, or escaped shell-word segment, stopping at whitespace or a shell separator. Coverage includes the exact reported case, double-to-single concatenation, Set-Cookie with a spaced quoted segment, and a control proving a separate `--region cn` argument is preserved. Local redaction tests passed (8 existing + 3 boundary tests), and all four cosh-ng CI jobs passed on the updated head.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
```diff
@@ -152,12 +161,53 @@ fn private_key_marker_range(line: &str, marker: &str) -> Option<(usize, usize)>
...
+        // Preserve the outer quote after consuming every adjacent shell-word segment.
+        Regex::new(
+            r#"(?imx)
+            (?P<prefix>'(?:set-cookie|cookie)\s*:\s*)
+            [^=\s;,]+\s*=\s*[^'\r\n]*(?P<suffix>'|$)
+            (?:
+                "(?:\\(?:\r?\n|[^\r\n])|[^"\\])*(?:"|$)|
+                '[^']*(?:'|$)|
+                \\(?:\r?\n|[^\r\n])|
+                [^\s;&|()<>"'\\]
```
ikunkun-sys: [P1] Redact command-substitution segments in Cookie values

This unquoted arm consumes the `$` before a command substitution but stops at the excluded `(`, so it never consumes the rest of `$()`; the backtick form similarly stops at the first space. On this exact head, production `redact_sensitive_text` turns `curl -H 'Cookie: session='$(printf quoted-cookie-value-42)` into `curl -H 'Cookie: <redacted>'(printf quoted-cookie-value-42)` with `changed=true`, leaving the literal value. Bash and Zsh both construct `Cookie: session=quoted-cookie-value-42` from the input. Shell gating catches the command, but Rust-only redaction consumers can still persist or render the secret; the new tests cover adjacent literal and quoted segments, not command substitutions.

ikunkun-sys (changes requested):
Requesting changes for one current-head privacy gap. The previous adjacent literal/quoted shell-word finding is fixed, but valid command-substitution segments (`$()` and backticks) still let a Cookie value survive the Rust redactor; see the inline finding.

Direct current-production probes and Bash/Zsh argument construction reproduce the leak. Focused redaction/detector tests, Bash/Zsh history/journal E2E, formatting, all-target Clippy, layout/inventory checks, shell syntax, and relevant CI pass, but do not cover command substitution.

Reviewed commit: `31c463e08e3744502bad43460eea4002273759ee`

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
kongche-jbw: Fixed in `d6231090`. The adjacent Cookie/Set-Cookie shell-word grammar now recognizes both `$()` and backtick command-substitution segments before the generic unquoted arm, so substitution bodies containing spaces are consumed instead of leaving a literal suffix. Coverage includes the exact reported `$()` case, nested `$()` with a following `--region cn` stop-boundary control, and a Set-Cookie backtick substitution containing a spaced value. Local redaction tests passed (8 existing + 3 boundary tests). Workspace, release, and prebuilt CI passed on the first attempt; fast checks passed on a bounded rerun after an unrelated `cosh-platform` `ETXTBSY` test-runner failure.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
```diff
@@ -152,12 +161,57 @@ fn private_key_marker_range(line: &str, marker: &str) -> Option<(usize, usize)>
...
-        // The pattern is a compile-time constant covered by the tests below.
-        Regex::new(r"(?im)(?P<prefix>^(?:set-cookie|cookie)\s*:\s*).*$")
-            .unwrap_or_else(|_| unreachable!("static cookie pattern must compile"))
+        // Preserve the outer quote after consuming every adjacent shell-word segment.
+        Regex::new(
+            r#"(?imx)
+            (?P<prefix>'(?:set-cookie|cookie)\s*:\s*)
+            [^=\s;,]+\s*=\s*[^'\r\n]*(?P<suffix>'|$)
+            (?:
+                \$\([^\r\n]*\)|\$\([^\r\n]*$|
```
ikunkun-sys: [P1] Redact line-continued command substitutions

Both `$()` alternatives exclude `\r` and `\n`; with multiline mode enabled, the unterminated fallback stops at the first physical line. On this exact head, production `redact_sensitive_text` turns `curl -H 'Cookie: session='$(printf '%s' \\\n  quoted-cookie-value-42)` into `curl -H 'Cookie: <redacted>'\n  quoted-cookie-value-42)` with `changed=true`, leaving the literal value. Bash and Zsh both construct `Cookie: session=quoted-cookie-value-42` from this valid continuation. Shell gating catches the command, but Rust-only redaction consumers can still persist or render the secret, and the new tests cover only single-line substitutions.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
```diff
@@ -152,12 +161,57 @@ fn private_key_marker_range(line: &str, marker: &str) -> Option<(usize, usize)>
...
+fn double_quoted_cookie_header_pattern() -> &'static Regex {
+    static PATTERN: OnceLock<Regex> = OnceLock::new();
+    PATTERN.get_or_init(|| {
+        Regex::new(
+            r#"(?imx)
+            (?P<prefix>"(?:set-cookie|cookie)\s*:\s*)
+            [^=\s;,]+\s*=\s*(?:\\(?:\r?\n|[^\r\n])|[^"\\\r\n])*
+            (?P<suffix>"|$)
+            (?:
+                \$\([^\r\n]*\)|\$\([^\r\n]*$|
```
ikunkun-sys: [P2] Preserve arguments after the Cookie substitution

This duplicated `[^\r\n]*` body is greedy, so it closes at the last `)` on the line rather than the end of the adjacent substitution. A current-production probe with `curl -H 'Cookie: session='$(printf quoted-cookie-value-42) --region "$(uname)"` returns only `curl -H 'Cookie: <redacted>'`, deleting the independent `--region "$(uname)"` argument. The boundary tests explicitly preserve a following `--region cn`, but do not cover a later parenthesized argument, so provider-facing command facts and persisted evidence can lose non-secret execution context.

ikunkun-sys (changes requested):
Requesting changes for one current-head P1 privacy gap and one P2 evidence-integrity regression. Same-line `$()` and backtick Cookie values from the previous review are now redacted, but a valid line-continued `$()` still leaves the literal value in Rust-redacted output, while the greedy same-line arm can erase independent trailing arguments; see the inline findings.

Direct current-production probes and Bash/Zsh argument construction reproduce both behaviors. Focused redaction/detector tests, Bash/Zsh history/journal E2E, formatting, all-target Clippy, layout/inventory checks, shell syntax, relevant CI, and CLA pass, but do not cover these boundaries.

Reviewed commit: `d6231090b3464dd037772240315399ab873da61f`

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
kongche-jbw: Fixed in `adc7a364`. Quoted Cookie/Set-Cookie values now use a bounded shell-word scanner that tracks quotes, escapes, backticks, balanced command substitutions, and backslash-newline continuations, so the exact multiline `$()` input is fully redacted. The focused redaction suites passed locally (8 existing + 3 boundary tests), cosh-shell lib Clippy and layout checks passed, and all four cosh-ng CI jobs passed on the updated head.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction.rs
kongche-jbw: Fixed in `adc7a364`. The same shell-word scanner stops at the matching close of the adjacent substitution instead of the last `)` on the line, preserving the independent `--region "$(uname)"` argument exactly. The reviewer input is covered by a regression assertion; focused local tests, Clippy/layout checks, and all four cosh-ng CI jobs pass on the updated head.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
```diff
@@ -0,0 +1,204 @@
...
+                    contexts.push(ShellScanContext::Backtick);
+                }
+                b'(' => {
+                    cursor += 1;
+                    if let Some(ShellScanContext::CommandSubstitution(depth)) = contexts.last_mut()
+                    {
+                        *depth += 1;
+                    }
+                }
+                b')' => {
```
ikunkun-sys: [P1] Keep `case` pattern terminators inside `$()`

This arm decrements command-substitution depth for every bare `)`, but `case ... pattern)` uses the same token without closing `$()`. On this exact head, the production scanner turns `curl -H 'Cookie: session='$(case x in x) printf '%s' quoted-cookie-value-42;; esac)` into `curl -H 'Cookie: <redacted>' printf '%s' quoted-cookie-value-42;; esac)` with `changed=true`, leaving the literal value. Bash and Zsh both construct `Cookie: session=quoted-cookie-value-42` from the input. Shell gating catches the command, but Rust-only redaction consumers can still persist or render the secret; the new scanner tests cover balanced parentheses, not `case` pattern terminators.

ikunkun-sys (changes requested):
Requesting changes for one current-head privacy gap. The new bounded shell-word scanner fixes the previously reported multiline `$()` leak and greedy trailing-argument deletion, but it treats every bare `)` inside command substitution as a nesting close. Valid `case ... pattern)` syntax therefore ends the scan early and leaves the Cookie value literal; see the inline finding.

Direct current-production scanner probes and Bash/Zsh argument construction reproduce the leak. Focused redaction/detector tests, Bash/Zsh history/journal E2E, formatting, all-target Clippy, layout/inventory checks, shell syntax, relevant CI, and CLA pass, but do not cover `case` pattern terminators.

Reviewed commit: `adc7a3647653d4a1f0382acb6fd74d20cbbd44ff`

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
kongche-jbw: Fixed in `02d4f5e4`. The bounded scanner now tracks command-position `case`, its switch word and `in`, pattern/body transitions, `;;`/`;&`/`;;&`, and matching `esac`, so a pattern terminator no longer closes the surrounding `$()`. The exact reviewer input is fully redacted, while a `case` word used as an ordinary `printf` argument remains a negative syntax control and the following `--region cn` is preserved. Local redaction tests passed (8 existing + 3 boundary tests), cosh-shell lib Clippy and layout checks passed, and all four cosh-ng CI jobs passed on the updated head.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
```diff
@@ -0,0 +1,359 @@
...
+                }
+                b'(' => {
+                    cursor += 1;
+                    if let Some(ShellScanContext::CommandSubstitution { paren_depth, .. }) =
+                        contexts.last_mut()
+                    {
+                        *paren_depth += 1;
+                    }
+                }
+                b')' => {
```
ikunkun-sys: [P1] Keep legal `)` tokens inside `$()`

This arm only protects `)` when `active_case_phase` is already `Pattern`. The word path clears `command_start` after `then`, so valid `curl -H 'Cookie: session='$(if true; then case x in x) printf '%s' quoted-cookie-value-42;; esac; fi)` never records the nested case and closes here at `x)`. On this exact head the production scanner returns `curl -H 'Cookie: <redacted>' printf '%s' quoted-cookie-value-42;; esac; fi)`, leaving the literal value. A valid command substitution with `)` inside a shell comment fails at the same arm because comments are not tracked. Bash and Zsh both construct `Cookie: session=quoted-cookie-value-42` from these inputs. Subsequent redactors do not match this ordinary literal, so Rust-only journal, audit, activity, and provider paths can still persist or render it.

ikunkun-sys (changes requested):
Requesting changes for one current-head P1 privacy gap. The new `case` tracker fixes the prior direct `$(case ...)` report, but the scanner still treats legal `)` tokens as the outer command-substitution close when a `case` follows a control keyword such as `then`, or when the token appears in a shell comment. See the inline finding.

Direct current-production scanner probes and Bash/Zsh argument construction reproduce both paths. Focused redaction tests, the detector matrix, Bash/Zsh history/journal E2E, formatting, all-target Clippy, layout/inventory checks, shell syntax, relevant CI, and CLA pass, but do not cover these grammar contexts.

Reviewed commit: `02d4f5e4da424be4c1c61d2c9166e036b5ebe7ed`

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
kongche-jbw: Fixed in `1498eb40`. I removed the incomplete case/parenthesis parser instead of adding more grammar states. Static Cookie shell words are still redacted precisely and preserve independent trailing arguments; dynamic expansion, backticks, adjacent/process substitutions, and ambiguous compact multi-pair headers now fail closed to whole-field `<redacted>`. Regression coverage includes the exact `then case` and comment paths, a here-doc variant, process substitution, prior partial-redaction ordering, compact multi-cookie input, and a static `--region cn` preservation control. Focused redaction tests passed (8 existing + 3 boundary), lib Clippy/rustfmt/layout/diff checks passed, and all required CI checks on `1498eb40` are green, including `Test cosh-ng`, fast checks, release build, and prebuilt action.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
```diff
@@ -0,0 +1,239 @@
...
+            .unwrap_or_else(|_| unreachable!("static cookie pattern must compile"))
+    })
+}
+
+fn compact_unquoted_cookie_header_prefix_pattern() -> &'static Regex {
+    static PATTERN: OnceLock<Regex> = OnceLock::new();
+    PATTERN.get_or_init(|| {
+        // No whitespace after the colon distinguishes a shell argument from a
+        // free-text header, whose remaining line may contain more cookies.
+        Regex::new(r#"(?im)(?P<prefix>(?:^|\s)(?:set-cookie|cookie)\s*:)[^=\s;,]+="#)
```
ikunkun-sys: [P1] Match static shell spellings of the Cookie prefix

This matcher only recognizes a literal `Cookie:` in one raw source segment, so the new fail-closed path is never reached when Bash or Zsh constructs the same static header with escapes or adjacent quoting. On this exact head, `curl -H Cookie:\ session=quoted-cookie-value-42`, `curl -H Cookie\:session=quoted-cookie-value-42`, and `curl -H Cookie':'session=quoted-cookie-value-42` all return `changed=false` from the current Rust redaction sequence, and both shell detectors classify them as safe. Bash and Zsh construct `Cookie: session=quoted-cookie-value-42` or `Cookie:session=quoted-cookie-value-42`; the first form uses the standard shell spelling for keeping a space in one argument. The literal value can therefore remain in native history and Rust-only persisted or rendered evidence. Please recognize static shell syntax in the prefix, or fail closed before requiring a literal prefix, and cover these forms in both detector and redaction tests.

ikunkun-sys (changes requested):
Requesting changes for one current-head credential-redaction bypass. The previous `then case` and shell-comment finding is fixed, but valid static Bash/Zsh spellings that escape or quote characters inside the `Cookie:` prefix bypass both the native-history detectors and the Rust redactor; see the inline finding.

Direct current-production probes return `changed=false`, while Bash and Zsh construct the real Cookie header and both detectors classify the commands as safe. Focused redaction/detector and Bash/Zsh history/journal tests, formatting, all-target Clippy, shell syntax, synthetic merge, relevant CI, and CLA pass, but do not cover prefix-level shell syntax.

Reviewed commit: `1498eb40f0e8c73f0e423494f001e4c499141aa6`

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
kongche-jbw: Fixed in 0a1cd0777bf596db6c069b58dea1cecfe09cb94e. Rust now decodes only the static Cookie/Set-Cookie prefix across unquoted escapes and adjacent quote segments, then delegates value handling to the existing shell-word scanner; dynamic values still fail closed. The Bash and Zsh history gates apply the same conservative static spelling check. Added the three exact reproductions to Rust redaction and both shell detector tests, plus Set-Cookie, quoted-name, safe-control, and trailing-argument coverage. Local validation passed (8 redaction tests, 3 boundary tests, the dual-shell detector test, scoped Clippy, fmt, and layout audit). Required CI is green: Test cosh-ng 9m26s after one unrelated PTY timing retry, fast checks 6m14s, release build 4m25s, and prebuilt action 1m10s.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
```diff
@@ -0,0 +1,460 @@
...
+    }
+    output.push_str(&text[cursor..]);
+
+    (output, changed)
+}
+
+fn find_cookie_header(text: &str, search_start: usize) -> HeaderSearchResult {
+    let bytes = text.as_bytes();
+    for start in search_start..bytes.len() {
+        if start != 0 && !is_shell_word_boundary(bytes[start - 1]) {
```
ikunkun-sys: [P1] Preserve Cookie detection for attached `-H` arguments

This boundary requires the Cookie prefix to start a raw shell word, but curl permits its `-H` argument to be attached directly to the option. `curl -H"Cookie: session=quoted-cookie-value-42"` and the single-quoted equivalent are valid standard spellings: Bash and Zsh produce one `-HCookie: session=quoted-cookie-value-42` argument, and curl accepts it. On the previous head, the Rust redactor removed the value and both detectors classified these forms as sensitive; on this exact head, the production redaction sequence returns `changed=false` and both detectors classify them as safe. The new Rust search skips the opening quote because it follows `H`, while the shell normalization removes that quote and leaves `-hcookie:`, which also fails its prefix boundary. The literal value can therefore remain in native history and Rust-only persisted or rendered evidence; the new matrix only covers headers separated from `-H` by whitespace.

ikunkun-sys (changes requested):
Requesting changes for one current-head privacy regression. The previous escaped/quoted Cookie-prefix finding is fixed, but the new raw-boundary search and normalized Bash/Zsh detector regress standard curl spellings whose header argument is attached directly to `-H`; see the inline finding.

Direct previous/current production probes show that `1498eb40` redacted and classified these inputs as sensitive, while `0a1cd077` returns `changed=false` and both native-history detectors classify them as safe. Bash and Zsh produce the expected single `-HCookie: ...` argument, and curl accepts it. Focused redaction/detector and Bash/Zsh history/journal tests, formatting, all-target Clippy, layout, shell syntax, synthetic merge, relevant CI, and CLA pass, but do not cover the attached-option boundary.

Reviewed commit: `0a1cd0777bf596db6c069b58dea1cecfe09cb94e`

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
kongche-jbw: Fixed in b9110b69. Rust now statically decodes the attached option prefix across quotes and escapes, then accepts --header= or any single short-option cluster ending in uppercase H, including -H, -sH, and -fsSLH; dynamic prefixes remain fail-closed. Bash and Zsh apply the symmetric normalized short-cluster rule. Regression coverage includes the exact attached forms, quoted and escaped option prefixes, dynamic fail-closed behavior, internal-substring and non-H controls, and trailing-argument preservation. Required CI is green: Test cosh-ng 10m17s, fast checks 5m56s, release build 4m22s, and prebuilt action 1m17s.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
```diff
@@ -0,0 +1,538 @@
...
+    loop {
+        match next_static_char(raw_prefix, cursor, context) {
+            StaticCharResult::Char(character) => {
+                decoded.push(character.byte);
+                cursor = character.raw_end;
+                context = character.context_after;
+            }
+            StaticCharResult::Dynamic => return StaticOptionPrefix::Dynamic,
+            StaticCharResult::End(final_context) => {
+                return if final_context == ShellContext::Unquoted {
```
ikunkun-sys: [P1] Redact attached headers when the complete option word is quoted

This requires the decoded prefix to finish unquoted, but `curl "-HCookie: session=whole-option-secret-42" https://example.test` reaches `Cookie` while still inside the outer double quote. Bash and Zsh produce the valid single argument `-HCookie: session=whole-option-secret-42`, and a loopback curl probe confirms that the Cookie is sent. On this exact head the production Rust redaction sequence returns `changed=false`, so consumers that rely on Rust redaction can persist or render the literal even though the Bash/Zsh history detector catches it. Preserve the static prefix context through header parsing instead of rejecting this quoted whole-word form, and add single- and double-quoted regression cases.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
```diff
@@ -265,23 +276,31 @@ _cosh_command_has_secret() {
...
+  local assignment_value="['\"]?[^[:space:]]"
+  # Match canonical suffixes across complete shell assignment identifiers.
+  local canonical_assignment_pattern="(^|[^a-z0-9_-])['\"]?([a-z_][a-z0-9_]*)?${canonical_assignment_key}['\"]?[[:space:]]*(=|:)[[:space:]]*${assignment_value}"
+  local canonical_flag_pattern="(^|[^a-z0-9_-])--${canonical_assignment_key}([[:space:]]+|=)[[:space:]]*${assignment_value}"
+  local cookie_key='(cookie|set[_-]?cookie)'
+  local cookie_flag_pattern="(^|[^a-z0-9_-])--${cookie_key}([[:space:]]+|=)[[:space:]]*[^[:space:]]"
+  local cookie_assignment_pattern="(^|[^a-z0-9_-])['\"]?${cookie_key}['\"]?[[:space:]]*=[[:space:]]*[^[:space:]]"
+  local cookie_json_pattern="(^|[^a-z0-9_-])['\"]${cookie_key}['\"][[:space:]]*:[[:space:]]*['\"][^'\"]+['\"]"
+  local cookie_header_pattern="(^|[[:space:]'\"])${cookie_key}[[:space:]]*:[[:space:]]*[^=[:space:];,]+[[:space:]]*=[^[:space:]]*"
+  local attached_cookie_header_pattern="(^|[[:space:]])(-[[:alnum:]#]*h|--header=)${cookie_key}[[:space:]]*:[[:space:]]*[^=[:space:];,]+[[:space:]]*=[^[:space:]]*"
```
ikunkun-sys, line 289: [P1] Fail closed on dynamic attached-header prefixes in native history

After `HEADER_OPTION=-H`, `curl ${HEADER_OPTION}"Cookie: session=dynamic-option-secret-42" https://example.test` is a valid command: both Bash and Zsh construct `-HCookie: session=dynamic-option-secret-42`, and loopback probes confirm that curl sends the Cookie. The current Bash and Zsh detectors both return safe because quote removal leaves the `${HEADER_OPTION}` expansion in front of `Cookie:`, which cannot match this static short-option regex. Native-history cleanup therefore retains the literal. Rust already treats the equivalent dynamic prefix as whole-field `<redacted>`; apply the same fail-closed policy to both shell gates and cover the form in the symmetric detector and history E2E matrices.

ikunkun-sys (changes requested):
Two P1 secret-handling gaps remain on this exact head. One bypasses Rust evidence redaction when the complete attached header option is quoted; the other bypasses the Bash/Zsh native-history gate when the option prefix is dynamically expanded. Both forms execute as valid curl header arguments and were reproduced against the production paths, but are absent from the current test matrix.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
kongche-jbw: Fixed in `5b56d09f`. The Rust scanner now carries the final shell context from the statically decoded option prefix into Cookie header parsing, so complete `-HCookie: ...` words wrapped in either single or double quotes are precisely redacted while trailing arguments are preserved. Both quoted regressions pass; the complete required CI set is green, including `Test cosh-ng`.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
kongche-jbw: Fixed in `5b56d09f`. Bash and Zsh now fail closed when dynamic shell syntax precedes a header-shaped Cookie in the same command segment, matching the Rust privacy contract. The symmetric detector matrix covers variable expansion, command substitution, and backticks; Bash/Zsh native-history E2E tests also verify that the literal is absent from history, journal, and ledger. The complete required CI set is green, including `Test cosh-ng`.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
```diff
@@ -265,23 +276,34 @@ _cosh_command_has_secret() {
...
+  local dynamic_cookie_header_pattern='(\$|`)[^;&|]*'
```
ikunkun-sys, line 290: [P1] Keep nested substitution separators inside the dynamic prefix

This character class stops at every `;`, `&`, or `|`, including operators nested inside the command substitution that is constructing the attached header option. On this exact head, both Bash and Zsh classify `curl $(true; printf -- -H)"Cookie: session=semicolon-dynamic-secret-42" https://example.test` as safe, although both shells produce the single option argument `-HCookie: session=semicolon-dynamic-secret-42`; loopback curl probes confirm that both send the Cookie. The `$(true && printf -- -H)` and `$(printf -- -H | cat)` forms fail the same way. Because the history hooks delete or reject an entry only when this detector succeeds, the literal remains in native history. The Rust path does redact it, so the two privacy gates still disagree; the current matrix covers only a command substitution without internal operators.

ikunkun-sys (changes requested):
Requesting changes for one current-head P1. The exact variable-prefix case from the previous review is fixed, but the new dynamic Cookie detector treats shell control operators inside a command substitution as outer command boundaries. Valid dynamic header options containing `;`, `&&`, or a pipeline therefore still bypass Bash and Zsh native-history protection. Reviewed commit: `5b56d09f0df6d9a894ffd8758327e172855dd7be`.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/bash.sh
kongche-jbw: Fixed in `b26e7b71`. Bash and Zsh now recognize a closed `$()` or backtick substitution as part of the dynamic option prefix, so `;`, `&&`, and pipelines inside that substitution no longer terminate the match; outer shell separators still stop the same-word suffix. The symmetric detector matrix covers all three reported forms plus backticks, nested substitution, and an outer-separator safe control. Both native-history E2E tests execute the three reported forms and verify the literal is absent from history, journal, and ledger. Required CI is green; the first `Test cosh-ng` attempt had an unrelated raw-cli PTY prompt-count failure, its exact test passed locally, and the failed job rerun passed.

ikunkun-sys (approved):

PR #3001, opened 2026-08-31
fix(sight): match modern oom-kill dmesg format

jfeng18:
## Problem

`was_pid_oom_killed` only matches the `"Killed process <pid>`" summary line in dmesg. On kernel 5.0+ (commit ef8444ea01d7), memcg OOM kills emit only the structured `oom-kill:...,pid=<pid>,...` line — the summary line is absent. Since containers use cgroup memory limits, memcg OOM is the only OOM form in containerized deployments, so OOM crash attribution is silently missed there.

Verified on kernel 6.6.102: memcg OOM shows only `oom-kill:constraint=CONSTRAINT_MEMCG,...,pid=N,...`; global OOM shows both formats.

## Fix

Recognise both formats via `line_matches_oom_kill` with whole-token pid comparison (no prefix matching).

## Validation

- ECS (kernel 6.6.102): 6/6 OOM tests pass, including 4 new discriminative tests for the structured format
- fmt / clippy clean

Extracted from #661 (f450c8d2).

### DIFF
--- a/src/agentsight/src/interruption/oom_recovery.rs
+++ b/src/agentsight/src/interruption/oom_recovery.rs
@@ -285,7 +285,8 @@ fn match_agent_name(comm: &str) -> Option<&'static str> {
 /// when an agent process disappears, we check dmesg to determine if it
 /// was killed by the OOM killer (vs normal exit, SIGKILL, segfault, etc.).
 ///
-/// Returns `true` if the PID appears in a "Killed process" OOM line in dmesg.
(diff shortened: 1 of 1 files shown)


chengshuyi (approved):
LGTM. 结构化 oom-kill 行匹配逻辑正确：全 token pid 比较防前缀误判，测试覆盖真实 6.6.x memcg OOM 场景，签名未变调用方无影响。

PR #3004, opened 2026-09-01
feat(cosh-ng): [shell] route zsh prompts

### LINKED ISSUE #1943: [cosh-ng] zsh: slash-bearing natural-language prompts bypass interception (bash fixed in #1942)

kongche-jbw:
## Why

#2918 moved slash-bearing natural-language classification into Rust, but Zsh still had ownership gaps around non-leading path prompts and coalesced Tab input. The secret hardening merged in #2996 also treated operators inside `${parameter#pattern}` as outer word boundaries, allowing dynamic Cookie headers to bypass native-history gates and Rust redaction.

## What changed

Enhanced-marker Zsh sessions keep slash, space-prefixed, and potential Han-leading path prompts Rust-owned until submission. Ordinary ASCII prefixes return byte-identically to ZLE as soon as classification resolves.

Tab is handed to ZLE before retained typeahead is replayed. A bounded 50 ms handoff lets the widget finish, then Rust cancels any widget-rewritten buffer and replays all retained bytes as the next logical line. Later bytes are appended, and EOF waits out the same deadline before draining them. If an earlier submit routes to Agent in one coalesced read, the remainder follows the new Cosh owner instead of stale ZLE ownership.

Bash and Zsh native-history detectors now recognize Cookie headers following a closed braced parameter expansion. Rust word scanning treats a precisely closed `${...}` expansion as one shell word and fails closed when nested syntax makes the closing brace ambiguous.

The implementation installs no ZLE wrappers, private key sequences, inherited control descriptors, signals, or shell prompt deadlines. The local 50 ms deadline is limited to the ZLE Tab handoff.

## Related issue

closes #1943

Follow-up hardening for #2996.

## User / Agent impact

Eligible path prompts are intercepted before Zsh executes them. Custom Tab widgets receive Tab normally; retained typeahead drains without another key, while bytes after an Agent ownership cutover cannot reach a stale ZLE widget. Dynamically assembled Cookie headers no longer remain in native history or Rust-backed persistence and display paths.

## Risk and compatibility

The change affects Enhanced Assisted Zsh sessions and secret-redaction boundaries. Candidate buffering remains limited to eligible prefixes, and the Tab handoff retains input for at most 50 ms before cancel-and-replay. Plain braced expansions become one shell word only when a precise matching close is established; ambiguous nested shell syntax fails closed to whole-command redaction.

## Validation

- `cargo test --package cosh-shell --lib --quiet` (1464 passed)
- `cargo test --package cosh-shell --lib zsh_` (12 passed)
- `cargo test --package cosh-shell --test raw_cli raw_cli_zsh_hands_buffered_tab_to_zle_before_follow_up_input -- --nocapture` (1 passed)
- `cargo test --package cosh-shell --test raw_cli raw_cli_zsh_routes_space_prefixed_path_prompt_before_zle -- --nocapture` (1 passed)
- `cargo test --package cosh-shell --lib path_prompt -- --test-threads=1` (6 passed)
- `cargo test --package cosh-shell --test raw_cli path_prompt -- --test-threads=1` (7 passed)
- `cargo test --package cosh-shell --lib evidence::redaction::tests -- --test-threads=1` (8 passed)
- `cargo test --package cosh-shell --lib redaction_boundary_tests -- --test-threads=1` (3 passed)
- `cargo test --package cosh-shell --test shell_host shell_secret_detectors_match_canonical_shapes_and_boundaries -- --test-threads=1` (1 passed)
- `bash -n crates/cosh-shell/src/shell_host/marker/bash.sh`
- `zsh -n crates/cosh-shell/src/shell_host/marker/zsh_secret_detector.zsh`

No ECS, real-provider, or full-workspace gate was run; broader coverage is delegated to CI.

## Documentation and rollback

No documentation changes are required. Revert `68ad7198` to remove parameter-pattern redaction hardening, and revert `b651b6d1` to remove Zsh candidate buffering. Shell-only mode remains the immediate runtime fallback.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
@@ -155,10 +155,7 @@ fn is_cookie_header_candidate_start(bytes: &[u8], start: usize) -> CandidateStar
         return CandidateStartResult::Match(ShellContext::Unquoted);
     }
 
-    let word_start = bytes[..start]
-        .iter()
-        .rposition(|byte| is_shell_word_boundary(*byte))
-        .map_or(0, |boundary| boundary + 1);
+    let word_start = shell_word_start(bytes, start);
     match decode_static_option_prefix(&bytes[word_start..start]) {
         StaticOptionPrefix::Decoded {
             bytes: option_prefix,
@@ -169,6 +166,50 @@ fn is_cookie_header_candidate_start(bytes: &[u8], start: usize) -> CandidateStar
         StaticOptionPrefix::Dynamic => CandidateStartResult::Dynamic,
         StaticOptionPrefix::Decoded { .. } => CandidateStartResult::NoMatch,
     }
+}
+
+fn shell_word_start(bytes: &[u8], end: usize) -> usize {
+    let mut word_start = 0;
+    let mut cursor = 0;
+    while cursor < end {
+        if bytes[cursor] == b'$' && bytes.get(cursor + 1) == Some(&b'{') {
+            cursor = parameter_expansion_end(bytes, cursor + 2, end);
+            continue;
+        }
+        if is_shell_word_boundary(bytes[cursor]) {
+            word_start = cursor + 1;
+        }
+        cursor += 1;
+    }
+    word_start
+}
+
+fn parameter_expansion_end(bytes: &[u8], mut cursor: usize, end: usize) -> usize {
+    let mut depth = 1usize;
+    while cursor < end {
+        match bytes[cursor] {
+            b'$' if bytes.get(cursor + 1) == Some(&b'{') => {
+                depth += 1;
+                cursor += 2;
+            }
+            // Quotes and nested substitutions require the complete shell
+            // grammar to prove which brace closes this expansion. Consume
+            // the remaining prefix so the caller treats it as dynamic.
+            b'\'' | b'"' | b'`' => return end,
+            b'$' if bytes.get(cursor + 1) == Some(&b'(') => return end,
+            b'<' | b'>' if bytes.get(cursor + 1) == Some(&b'(') => return end,
+            b'}' => {
+                depth -= 1;
+                cursor += 1;
+                if depth == 0 {
+                    return cursor;
+                }
+            }
+            b'\\' => cursor = skip_shell_escape(bytes, cursor),
+            _ => cursor += 1,
+        }
+    }
+    end
 }
 
 fn decode_static_option_prefix(raw_prefix: &[u8]) -> StaticOptionPrefix {
--- a/src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
@@ -13,9 +13,11 @@ use super::event_parser::{
 use super::event_sender::RawInputEventSink;
 use super::generation::{LineSubmitCounter, UserPtyInputGeneration};
 use super::mode::new_delay_input_mode;
+use super::path_prompt_candidate::zsh_path_candidate_should_hold;
 use super::soft_newline::{
     contains_soft_newline_sequence, draft_text_from_bytes, render_soft_newline_markers,
 };
+use super::spawn::ZshPathPromptBuffering;
 use super::{write_all_pty, MainPromptGate, PromptGhostRoute, RawInputEvent, RawInputMode, CTRL_C};
 
 pub(super) struct InputRelayContext<'a> {
@@ -35,6 +37,9 @@ pub(super) struct InputRelayContext<'a> {
     /// until the next prompt_ready marker and later bytes fall back to the
     /// Rust intercept path instead of leaking into a foreground process.
     pub(super) slash_route_enabled: bool,
+    /// Zsh-only capability that keeps slash candidates out of ZLE until Rust
+    /// can route or submit the complete line.
+    pub(super) zsh_path_prompt_buffering: Option<&'a mut ZshPathPromptBuffering>,
 }
 
 /// Writes real user bytes to the PTY, bumping the shared input generation
@@ -114,6 +119,16 @@ fn relay_passthrough_input_with_policy(
     emit_activity: bool,
     pending_shell_submits: usize,
 ) -> io::Result<bool> {
+    if relay
+        .zsh_path_prompt_buffering
+        .as_deref()
+        .is_some_and(ZshPathPromptBuffering::has_deferred_tab_typeahead)
+    {
+        if let Some(buffering) = relay.zsh_path_prompt_buffering.as_deref_mut() {
+            buffering.defer_tab_typeahead(bytes);
+        }
+        return Ok(true);
+    }
     if relay.line_buffer.force_agent_intercept && relay.line_buffer.is_active() {
         relay.line_buffer.soft_newline_enabled = true;
         relay.line_buffer.push(bytes);
@@ -138,6 +153,18 @@ fn relay_passthrough_input_with_policy(
         return candidate::relay_candidate_line(relay, emit_activity, pending_shell_submits);
     }
     relay_native_passthrough(bytes, relay, emit_activity, pending_shell_submits)
+}
+
+pub(super) fn replay_deferred_zsh_tab_typeahead(
+    bytes: &[u8],
+    relay: &mut InputRelayContext<'_>,
+) -> io::Result<()> {
+    // A separate relay turn gives ZLE time to finish the Tab widget. Cancel
+    // its potentially rewritten buffer before replaying preserved input as
+    // the next logical line.
+    relay_native_passthrough(&[CTRL_C], relay, true, 0)?;
+    relay_passthrough_input_with_policy(bytes, relay, true, 0)?;
+    Ok(())
 }
 
 pub(super) fn relay_prompt_ghost_input(
@@ -338,8 +365,21 @@ fn relay_native_passthrough(
     let complete_paste = bytes.starts_with(BRACKETED_PASTE_START);
     if relay.line_buffer.is_active()
         || complete_paste
-        || starts_native_intercept_candidate(bytes, relay.native_line_state)
+        || starts_native_intercept_candidate(
+            bytes,
+            relay.native_line_state,
+            relay.zsh_path_prompt_buffering.is_some(),
+        )
     {
+        if let Some(handled) = tab_handoff::relay_coalesced_zsh_tab(
+            bytes,
+            relay,
+            emit_activity,
+            pending_shell_submits,
+            complete_paste,
+        )? {
+            return Ok(handled);
+        }
         // Route flags must consider the whole draft so far: a bracketed
         // paste opener may arrive as its own chunk (or split mid-delimiter,
         // #1721) before the payload decides CJK vs slash (#1721 D13).
@@ -351,8 +391,21 @@ fn relay_native_passthrough(
         .concat();
         relay.line_buffer.soft_newline_enabled = native_candidate_allows_soft_newline(&combined);
         relay.line_buffer.push(bytes);
+        let zsh_path_candidate = relay.zsh_path_prompt_buffering.is_some();
+        let hold_zsh_path_candidate = zsh_path_candidate
+            && !relay.line_buffer.saw_paste()
+            && zsh_path_candidate_should_hold(&relay.line_buffer.bytes)
+            && !relay.line_buffer.bytes.contains(&b'\t');
+        let return_resolved_zsh_space_prefix = zsh_path_candidate
+            && relay.line_buffer.bytes.first() == Some(&b' ')
+            && !hold_zsh_path_candidate;
         if !relay.line_buffer.in_paste()
-            && native_candidate_should_return_to_shell(relay.input_classifier, relay.line_buffer)
+            && !hold_zsh_path_candidate
+            && (return_resolved_zsh_space_prefix
+                || native_candidate_should_return_to_shell(
+                    relay.input_classifier,
+                    relay.line_buffer,
+                ))
         {
             return flush_candidate_line_to_shell(relay, emit_activity, pending_shell_submits);
         }
@@ -633,6 +686,7 @@ mod candidate;
 mod exit_tracker;
 mod path_prompt_submit;
 mod soft_newline_upgrade;
+mod tab_handoff;
 use bash_submission_guard::{
     bash_submission_has_leading_whitespace, bash_submission_needs_guard, guarded_bash_submission,
     history_private_submission,
--- a/src/cosh-ng/crates/cosh-shell/src/raw_input/relay/path_prompt_submit.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/raw_input/relay/path_prompt_submit.rs
@@ -15,6 +15,11 @@ pub(super) fn route_missing_path_submission(
     relay: &mut InputRelayContext<'_>,
     pending_shell_submits: usize,
 ) -> io::Result<bool> {
+    // Zsh path candidates remain Rust-owned until Enter. If editing or a
+    // control key already returned the line to ZLE, keep it Shell-owned.
+    if relay.zsh_path_prompt_buffering.is_some() {
+        return Ok(false);
+    }
     if !relay.main_prompt_gate.is_path_prompt_ready() || pending_shell_submits > 0 {
         return Ok(false);
     }
@@ -53,15 +58,15 @@ pub(super) fn route_missing_path_submission(
     if let Ok(mut mode) = relay.input_mode.lock() {
         *mode = new_delay_input_mode();
     }
-    // Bytes typed in this read have not reached Readline yet. Write them
-    // before Ctrl-U so Bash echoes the exact user submission, then accept the
-    // cleared line to repaint PS1. The command itself is never executed.
-    let mut repaint = Vec::with_capacity(submit + 2);
-    repaint.extend_from_slice(&bytes[..submit]);
-    repaint.extend_from_slice(b"\x15\r");
     let _ = relay
         .input_events
         .send(RawInputEvent::SyntheticPromptRepaint);
+    // Bytes typed in this read have not reached Readline yet. Write them
+    // before Ctrl-U so Bash echoes the exact user submission, then accept
+    // the cleared line to repaint PS1. The command itself never executes.
+    let mut repaint = Vec::with_capacity(submit + 2);
+    repaint.extend_from_slice(&bytes[..submit]);
+    repaint.extend_from_slice(b"\x15\r");
     write_user_bytes_to_pty(
         relay.master,
         relay.input_generation,
@@ -111,14 +116,21 @@ pub(super) fn route_candidate_missing_path_submission(
         *mode = new_delay_input_mode();
     }
     let sensitive = redact_sensitive_text(&input).1;
-    let _ = relay
-        .input_events
-        .send(RawInputEvent::UserInterceptWithRouting {
+    let event = if input.starts_with('/') {
+        RawInputEvent::UserInterceptWithRouting {
             input,
             reason: InterceptReason::NaturalLanguage,
             cwd,
             sensitive,
-        });
+        }
+    } else {
+        RawInputEvent::NativePathPromptIntercept {
+            input,
+            cwd,
+            sensitive,
+        }
+    };
+    let _ = relay.input_events.send(event);
     send_shell_input_state(true, relay.input_events);
     true
 }
--- a/src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
@@ -1,0 +1,96 @@
+//! Zsh candidate handoff when Tab and later input share one terminal read.
+
+use std::io;
+
+use super::{
+    relay_native_passthrough, relay_passthrough_input_with_policy, send_held_input_events,
+    InputRelayContext,
+};
+
+pub(super) fn relay_coalesced_zsh_tab(
+    bytes: &[u8],
+    relay: &mut InputRelayContext<'_>,
+    emit_activity: bool,
+    pending_shell_submits: usize,
+    complete_paste: bool,
+) -> io::Result<Option<bool>> {
+    if relay.zsh_path_prompt_buffering.is_none()
+        || complete_paste
+        || relay.line_buffer.saw_paste()
+        || !relay.line_buffer.pending_partial_bytes().is_empty()
+    {
+        return Ok(None);
+    }
+    let Some(tab) = bytes.iter().position(|byte| *byte == b'\t') else {
+        return Ok(None);
+    };
+    if let Some(submit) = bytes
+        .iter()
+        .position(|byte| matches!(byte, b'\n' | b'\r'))
+        .filter(|submit| *submit < tab)
+    {
+        let submit_end = submit
+            + usize::from(
+                bytes.get(submit) == Some(&b'\r') && bytes.get(submit + 1) == Some(&b'\n'),
+            )
+            + 1;
+        let routed = relay_passthrough_input_with_policy(
+            &bytes[..submit_end],
+            relay,
+            emit_activity,
+            pending_shell_submits,
+        )?;
+        if submit_end < bytes.len() {
+            if !routed {
+                relay_passthrough_input_with_policy(
+                    &bytes[submit_end..],
+                    relay,
+                    emit_activity,
+                    pending_shell_submits,
+                )?;
+            } else {
+                // A handled submit cut over to a Cosh owner. Match the relay's
+                // ownership-crossing rule instead of reinterpreting suffix
+                // bytes through the stale ZLE owner.
+                send_held_input_events(&bytes[submit_end..], relay.input_events);
+            }
+        }
+        return Ok(Some(true));
+    }
+    if tab + 1 == bytes.len() {
+        return Ok(None);
+    }
+
+    let (through_tab, remainder) = bytes.split_at(tab + 1);
+    relay_native_passthrough(through_tab, relay, emit_activity, pending_shell_submits)?;
+    if let Some(submit) = remainder
+        .iter()
+        .position(|byte| matches!(byte, b'\n' | b'\r'))
+    {
+        if submit > 0 {
+            relay_passthrough_input_with_policy(
+                &remainder[..submit],
+                relay,
+                emit_activity,
+                pending_shell_submits,
+            )?;
+        }
+        let submit_end = submit
+            + usize::from(
+                remainder.get(submit) == Some(&b'\r') && remainder.get(submit + 1) == Some(&b'\n'),
+            )
+            + 1;
+        if submit_end < remainder.len() {
+            // PTYs do not preserve write boundaries, so the current read
+            // cannot safely cancel a buffer that the Tab widget may rewrite.
+            // Hold later type-ahead until the next read, then cancel and
+            // replay it as the next logical line.
+            if let Some(buffering) = relay.zsh_path_prompt_buffering.as_deref_mut() {
+                buffering.defer_tab_typeahead(&remainder[submit_end..]);
+            }
+        }
+        return Ok(Some(true));
+    }
+    relay_passthrough_input_with_policy(remainder, relay, emit_activity, pending_shell_submits)
+        .map(Some)
+}
--- a/src/cosh-ng/crates/cosh-shell/src/shell_host/bootstrap.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/shell_host/bootstrap.rs
@@ -10,6 +10,8 @@ use std::time::{Duration, SystemTime};
 use nix::libc;
 use nix::pty::openpty;
 
+use crate::raw_input::ZshPathPromptBuffering;
+
 use super::adapter::{BashAdapter, ShellAdapter, ZshAdapter};
 use super::auth::{generate_marker_token, marker_script_with_token};
 use super::lifecycle::push_shell_started_event;
@@ -31,6 +33,7 @@ pub(super) struct PtySession {
     pub(super) parser: OscParser,
     pub(super) recovery_request_file: PathBuf,
     pub(super) handoff_request_file: PathBuf,
+    pub(super) zsh_path_prompt_buffering: Option<ZshPathPromptBuffering>,
 }
 
 pub(super) fn start_bash_session(config: &ShellHostConfig) -> io::Result<PtySession> {
@@ -90,6 +93,9 @@ fn start_shell_session(
         fs::set_permissions(&path, fs::Permissions::from_mode(0o600))?;
         Some(path)
     };
+    let zsh_path_prompt_buffering = (marker.is_some()
+        && adapter.supports_zsh_path_prompt_buffering())
+    .then(ZshPathPromptBuffering::new);
 
     let pty = openpty(Some(&config.winsize), None).map_err(nix_to_io)?;
     let master = unsafe { File::from_raw_fd(pty.master.into_raw_fd()) };
@@ -188,6 +194,7 @@ fn start_shell_session(
         parser,
         recovery_request_file,
         handoff_request_file,
+        zsh_path_prompt_buffering,
     })
 }
 
--- a/src/cosh-ng/crates/cosh-shell/src/evidence/redaction_boundary_tests.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/evidence/redaction_boundary_tests.rs
@@ -281,6 +281,9 @@ fn redacts_embedded_cookie_headers_and_keeps_bare_token_prose() {
         "curl -H Cookie:session=$COOKIE_VALUE --region cn",
         "curl -s${HEADER_OPTION}Cookie:session=quoted-cookie-value-42 --region cn",
         r#"curl ${HEADER_OPTION}"Cookie: session=dynamic-option-secret-42" --region cn"#,
+        r#"curl ${HEADER_OPTION#*;}"Cookie: session=parameter-pattern-secret-42" https://example.test"#,
+        r#"curl ${HEADER_OPTION#*|}"Cookie: session=parameter-pipe-secret-42" https://example.test"#,
+        r#"curl ${HEADER_OPTION:-$(printf '}' >/dev/null; printf -- -H)}"Cookie: session=parameter-nested-secret-42""#,
         "curl -H 'Cookie: session='<(printf quoted-cookie-value-42) --region cn",
         "request Cookie:session=quoted-cookie-value-42; other=second-cookie-value",
         "Cookie: first=first-cookie-value\ncurl -H 'Cookie: session='$(printf quoted-cookie-value-42)",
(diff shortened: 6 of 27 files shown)


src/cosh-ng/crates/cosh-shell/src/shell_host/bootstrap.rs
```diff
@@ -90,6 +92,16 @@ fn start_shell_session(
         fs::set_permissions(&path, fs::Permissions::from_mode(0o600))?;
         Some(path)
     };
+    let (zsh_path_prompt_submit, zsh_path_prompt_child) =
+        if marker.is_some() && adapter.supports_zsh_path_prompt_submit() {
+            let (parent, child) = UnixStream::pair()?;
+            parent.set_read_timeout(Some(Duration::from_millis(500)))?;
+            parent.set_write_timeout(Some(Duration::from_millis(500)))?;
+            clear_close_on_exec(child.as_raw_fd())?;
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in `b23e4afe`. Replaced the inheritable bidirectional socket with one-way request/ack pipes. Zsh reopens both with `sysopen -o cloexec` and closes the bootstrap descriptors before registering the ZLE handler. The regression launches an external `sh` and verifies that neither internal descriptor is inherited. Local targeted tests and clippy passed; GitHub `Test cosh-ng`, fast checks, release build, and prebuilt-action checks all pass.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_submit.sh
```diff
@@ -0,0 +1,16 @@
+# Rust owns the verdict. This FD handler invokes only protected ZLE builtins;
+# it never wraps user accept-line, changes keymaps, or replays history hooks.
+_cosh_path_prompt_submit() {
+  local request
+  if ! IFS= read -r -u "$1" request || [[ "$request" != submit ]]; then
+    return 0
+  fi
+  zle .kill-whole-line || return 1
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in `b23e4afe`. The FD handler is now a `zle -F -w` widget that clears `BUFFER` and `CURSOR` directly, then finalizes through protected `.accept-line`; it no longer invokes a kill widget. The regression pre-seeds and verifies that both `CUTBUFFER` and `killring` remain unchanged. Local targeted tests and clippy passed; GitHub `Test cosh-ng`, fast checks, release build, and prebuilt-action checks all pass.

src/cosh-ng/crates/cosh-shell/src/shell_host/bootstrap.rs
```diff
@@ -90,6 +94,35 @@ fn start_shell_session(
...
+            for fd in [
+                request_child.as_raw_fd(),
+                request_parent.as_raw_fd(),
+                acknowledgment_parent.as_raw_fd(),
+                acknowledgment_child.as_raw_fd(),
+            ] {
+                set_close_on_exec(fd)?;
+            }
+            clear_close_on_exec(request_child.as_raw_fd())?;
+            clear_close_on_exec(acknowledgment_child.as_raw_fd())?;
```
ikunkun-sys: **[P1] Keep bootstrap descriptors out of user startup children**

These two descriptors are inheritable as soon as zsh executes, but `zsh_marker_script` sources the user `.zshrc` at `zsh.rs:18-39` and only runs `_cosh_setup_path_prompt_submit` at `zsh.rs:694`. Therefore any external or background process launched by `.zshrc` inherits both the request reader and acknowledgment writer while `COSH_ZSH_PATH_PROMPT_*_FD` still names them. Such a process can consume `submit\n` and prewrite `accepted\n`, causing Rust to report interception while ZLE has not cleared `BUFFER`; the intercepted text can remain executable, or the relay can time out and reap the shell. The regression launches `sh` only after setup, so it does not cover this startup window.

ikunkun-sys (changes requested):
Requesting changes for the P1 inline finding on b23e4afe. CI and the targeted local checks are green, but the Zsh bootstrap still exposes the control pipes to processes launched from the user startup file before the CLOEXEC reopen runs.

src/cosh-ng/crates/cosh-shell/src/shell_host/bootstrap.rs
kongche-jbw: Fixed in `e2e4c484`. I split channel initialization into an early preparation phase and the later ZLE registration phase. The generated marker now reopens both endpoints with `cloexec`, unsets their exported bootstrap names, and closes the inheritable descriptors before sourcing the user `.zshenv`, `.zprofile`, `.zlogin`, or `.zshrc`. The regression launches an external `sh` from the user `.zshrc` with both descriptor numbers and verifies neither is present, while the later routing and post-setup inheritance checks still pass. Local path-prompt tests (5 unit, 6 raw-cli), clippy, fmt, and layout passed; GitHub `Test cosh-ng`, fast checks, release build, and prebuilt-action checks all pass.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_prepare.sh
```diff
@@ -0,0 +1,19 @@
+# Bootstrap descriptors must be replaced before user startup files can launch
+# children. The replacements stay in zsh but close automatically on exec.
+_cosh_bootstrap_request_fd="${COSH_ZSH_PATH_PROMPT_REQUEST_FD:-}"
+_cosh_bootstrap_acknowledgment_fd="${COSH_ZSH_PATH_PROMPT_ACK_FD:-}"
+unset COSH_ZSH_PATH_PROMPT_REQUEST_FD COSH_ZSH_PATH_PROMPT_ACK_FD
+unset _COSH_PATH_PROMPT_REQUEST_FD _COSH_PATH_PROMPT_ACK_FD
+if [[ -n "$_cosh_bootstrap_request_fd" && -n "$_cosh_bootstrap_acknowledgment_fd" ]]; then
+  typeset -g _COSH_PATH_PROMPT_REQUEST_FD _COSH_PATH_PROMPT_ACK_FD
+  if ! zmodload zsh/system \
+      || ! sysopen -r -o cloexec -u _COSH_PATH_PROMPT_REQUEST_FD "/dev/fd/$_cosh_bootstrap_request_fd" \
```
ikunkun-sys: **[P1] Keep control pipes out of fork-only Zsh children**

`sysopen -o cloexec` protects only an `exec`; a Zsh subshell or background function forked from `.zshrc` retains both descriptors and the `_COSH_PATH_PROMPT_*_FD` parameters. I reproduced this on e2e4c484 by starting a `.zshrc` subshell that read `submit` and wrote `accepted`: Rust rendered “The input was intercepted before the shell ran it”, but the next Enter executed the original `/nonexistent-cosh-fork/SKILL.md ...` buffer and zsh reported “no such file or directory”. The new regression launches an external `sh`, so it cannot exercise this fork-only path.

ikunkun-sys (changes requested):
Requesting changes for the P1 inline finding on e2e4c484. CI, fmt, layout, clippy, 5 path-prompt unit tests, and 6 raw-cli tests pass, but an end-to-end .zshrc fork-only worker can still consume the control request and acknowledge it without clearing ZLE BUFFER, allowing the supposedly intercepted command to execute on the next Enter.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_submit.sh
```diff
@@ -0,0 +1,19 @@
+# Rust owns the verdict. This private-sequence widget clears BUFFER directly
+# and invokes the protected accept-line builtin without wrapping user widgets.
+typeset -gr _COSH_PATH_PROMPT_SUBMIT_SEQUENCE=$'\e[99;99u'
```
ikunkun-sys: **[P1] Keep user-generated CSI-u away from the internal submit widget**

This sequence is not private to Rust. `CSI 99;99u` is a valid [Kitty keyboard-protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/) encoding: key code 99 is `c`, and modifier 99 is a valid modifier bitset; user input reaches the same PTY. On 1f92eafe I typed `echo private-sequence-command-ran` at an ordinary prompt and then sent these bytes from the user side. This widget cleared the whole `BUFFER` and accepted an empty line, so the command silently disappeared without entering path-prompt routing.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_submit.sh
```diff
@@ -0,0 +1,19 @@
...
+  local keymap
+  bindkey -M main "$_COSH_PATH_PROMPT_SUBMIT_SEQUENCE" _cosh_path_prompt_submit 2>/dev/null || return 1
+  for keymap in ${(f)"$(bindkey -l)"}; do
+    [[ "$keymap" == main || "$keymap" == .safe ]] && continue
+    bindkey -M "$keymap" "$_COSH_PATH_PROMPT_SUBMIT_SEQUENCE" _cosh_path_prompt_submit 2>/dev/null || true
+  done
+}
+zle -N _cosh_path_prompt_submit
+autoload -Uz add-zle-hook-widget
+add-zle-hook-widget zle-line-init _cosh_bind_path_prompt_submit
```
ikunkun-sys: **[P1] Do not assume this line-init hook remains last**

A user `precmd` hook can append another `zle-line-init` hook after this registration. That later hook runs after `_cosh_bind_path_prompt_submit` and can rebind the sequence. In an end-to-end `.zshrc` probe on 1f92eafe, the later hook rebound it to a user widget: Rust rendered “The input was intercepted before the shell ran it”, but the next Enter executed the original `/nonexistent-cosh-late/SKILL.md ...` buffer and zsh reported “no such file or directory”. The regression defines its user line-init widget before the marker registers this hook, so it only covers the ordering where ANOLISA remains last.

ikunkun-sys (changes requested):
Requesting changes for two P1 findings on 1f92eafe. The previous FD-inheritance defect is removed and the completed checks pass, but the replacement in-band ZLE sequence is both user-generatable and overwriteable by a later line-init hook. End-to-end probes showed an ordinary command draft being silently discarded, and a path prompt reported as intercepted but then executed on the next Enter. GitHub Test cosh-ng is still pending at submission time.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_prepare.sh
kongche-jbw: Fixed in `fbe2e13c`. Removed the request/ack pipes and their environment variables entirely. Rust now targets only the interactive Zsh process with `SIGURG`; the ZLE signal widget clears `BUFFER` and submits through protected `.accept-line`, so fork-only children have no control endpoint to consume or spoof. The regression retains the reviewer fork-only `.zshrc` worker and verifies no channel is exposed. Local path-prompt tests, clippy, fmt, and layout passed; all GitHub checks, including `Test cosh-ng`, now pass.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_submit.sh
kongche-jbw: Fixed in `fbe2e13c`. Removed the CSI-u binding and all other reserved input bytes. Submission now uses a process-directed signal, while user CSI `99;99u` remains entirely user/ZLE-owned. The regression types an ordinary command, sends the exact reviewer sequence, and verifies the command still executes before exercising path-prompt routing. Local targeted gates and all GitHub checks pass.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_submit.sh
kongche-jbw: Fixed in `fbe2e13c`. Removed `_cosh_bind_path_prompt_submit` and the `zle-line-init` hook, so routing no longer depends on keymap or hook ordering. Rust signals the Zsh process directly and waits for the subsequent trusted `prompt_ready` boundary before reporting interception; if the handler does not complete, it cannot emit a false success. The regression keeps a line-init rebind of the old CSI sequence and verifies the path prompt is still cleared and routed. Local targeted gates and all GitHub checks pass.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_submit.sh
```diff
@@ -0,0 +1,12 @@
...
+_cosh_path_prompt_submit() {
+  BUFFER=''
+  CURSOR=0
+  zle .accept-line
+}
+_cosh_path_prompt_signal() {
+  [[ -o zle ]] && zle _cosh_path_prompt_submit
+}
+zle -N _cosh_path_prompt_submit
+trap _cosh_path_prompt_signal URG
```
ikunkun-sys: **[P1] Do not take ownership of the user's SIGURG trap**

This installs a process-global Zsh trap, so sourcing the marker discards any existing user SIGURG handler, while a user hook can later replace this handler and disable path-prompt submission. On fbe2e13c I reproduced the latter with a `precmd` that restores its own `URG` trap: the user handler ran, the ZLE buffer was not submitted, and Rust timed out and terminated the raw shell. The submission mechanism cannot rely on exclusive, persistent ownership of a user-visible signal trap.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/path_prompt_submit.rs
```diff
@@ -83,6 +105,32 @@ pub(super) fn route_missing_path_submission(
     Ok(true)
 }
 
+fn wait_for_zsh_prompt(main_prompt_gate: &super::super::MainPromptGate) -> io::Result<()> {
+    let deadline = Instant::now() + Duration::from_secs(2);
```
ikunkun-sys: **[P1] Do not fail the shell on a fixed two-second prompt wait**

Prompt completion includes user-defined `precmd` work, whose duration has no two-second contract. On fbe2e13c, a normal `precmd() { sleep 3; }` let the signal widget submit the line but made this deadline expire before the next prompt; the error then terminated the raw shell and no Agent route or follow-up was delivered. A slow but valid prompt must not be classified as submission failure that tears down the session.

ikunkun-sys (changes requested):
Requesting changes for two P1 findings on fbe2e13c. The previous descriptor-inheritance and in-band CSI-u defects are removed, and all executed GitHub checks pass, but the SIGURG replacement introduces two raw-shell termination paths: a user hook can replace or be replaced by the process-global trap, and a normal precmd longer than two seconds exceeds the fixed prompt deadline. Both behaviors were reproduced end to end on this exact head, and neither duplicates an existing thread.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh_path_prompt_submit.sh
kongche-jbw: Fixed in `c8a0bcc8`. I removed the SIGURG trap and the entire signal/ZLE submission path. Enhanced-marker Zsh sessions now keep leading-slash candidates Rust-owned until Enter, so no user signal handler is installed, replaced, or required. The raw-cli regression installs and invokes a user `TRAPURG` while verifying the path prompt still routes. Local targeted gates and all GitHub checks, including `Test cosh-ng`, pass.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/path_prompt_submit.rs
kongche-jbw: Fixed in `c8a0bcc8`. The Zsh route no longer submits through ZLE or waits for a subsequent prompt, so the fixed two-second deadline and its shell-terminating error path are gone. The regression includes a three-second `precmd` and verifies routing plus subsequent shell input remain healthy. Local targeted gates and all GitHub checks, including `Test cosh-ng`, pass.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/path_prompt_submit.rs
```diff
@@ -15,6 +15,11 @@ pub(super) fn route_missing_path_submission(
     relay: &mut InputRelayContext<'_>,
     pending_shell_submits: usize,
 ) -> io::Result<bool> {
+    // Zsh slash candidates remain Rust-owned until Enter. If editing or a
+    // control key already returned the line to ZLE, keep it Shell-owned.
+    if relay.zsh_path_prompt_buffering.is_some() {
```
ikunkun-sys, line 20: **[P1] Keep Zsh missing-path routing for non-leading-slash prompts**

This early return disables submit-time missing-path classification for every Enhanced Zsh session, while the new Rust candidate buffer only starts when the first visible byte is `/`. The existing cross-shell regression `raw_cli_routes_slash_bearing_han_prompt_before_shell_execution` now fails consistently on this head: Zsh executes `你读一下，并安装这个skill：/nonexistent-cosh-2913/SKILL.md` and prints `no such file or directory`, although Bash routes the same prompt to the Agent. A path-bearing natural-language prompt must not become executable solely because the path is not the first token.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
```diff
@@ -351,7 +355,11 @@ fn relay_native_passthrough(
         .concat();
         relay.line_buffer.soft_newline_enabled = native_candidate_allows_soft_newline(&combined);
         relay.line_buffer.push(bytes);
+        let hold_zsh_path_candidate = relay.zsh_path_prompt_buffering.is_some()
+            && !relay.line_buffer.saw_paste()
+            && relay.line_buffer.bytes.first() == Some(&b'/');
         if !relay.line_buffer.in_paste()
+            && !hold_zsh_path_candidate
```
ikunkun-sys, line 376: **[P1] Hand Tab to ZLE before a buffered Enter**

This hold also suppresses the Tab handoff in `native_candidate_should_return_to_shell`. On c8a0bcc8, with a custom Tab widget that rewrites `BUFFER` to `echo TAB-WIDGET-COMMAND-RAN`, typing `/definitely-not-command`, then Tab, then Enter makes Rust replay the buffered text, Tab, and Enter together; the widget rewrites the line and the queued Enter executes it immediately. A control key that returns ownership to ZLE must be delivered before any later submission byte, otherwise a transformed command can run without the user seeing and confirming it.

ikunkun-sys (changes requested):
Requesting changes for two P1 findings on c8a0bcc8. The previous signal-trap and fixed prompt-deadline defects are removed, and all executed GitHub checks plus the targeted positive checks pass. However, the replacement Rust-side buffering regresses non-leading-slash path prompts and can replay Tab together with Enter, allowing a ZLE rewrite to execute without user confirmation. Both behaviors were reproduced end to end on this exact head and do not duplicate existing threads.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/path_prompt_submit.rs
kongche-jbw: Fixed in `73ed143c`. Enhanced Zsh now starts a Rust-owned candidate for a non-ASCII line prefix as well as a leading slash, so the Han-leading missing-path prompt reaches the shared submit-time classifier before ZLE can execute it. Ordinary ASCII commands remain ZLE-owned, and non-leading prompts keep the stable Agent-input confirmation. The existing cross-shell regression now passes for both Bash and Zsh. Local targeted gates and all GitHub checks pass.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
kongche-jbw: Fixed in `73ed143c`. Tab is excluded from the Zsh candidate hold and is written to ZLE immediately, before any later bytes can be processed. Added both a direct relay regression and an end-to-end custom Tab widget test that cancels the rewritten draft and verifies the transformed command never executes. Local targeted gates and all GitHub checks pass.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
```diff
@@ -171,6 +168,44 @@ fn is_cookie_header_candidate_start(bytes: &[u8], start: usize) -> CandidateStar
...
+
+fn parameter_expansion_end(bytes: &[u8], mut cursor: usize, end: usize) -> usize {
+    let mut depth = 1usize;
+    while cursor < end {
+        match bytes[cursor] {
+            b'$' if bytes.get(cursor + 1) == Some(&b'{') => {
+                depth += 1;
+                cursor += 2;
+            }
+            b'}' => {
```
ikunkun-sys, line 201: **[P1] Do not close braced expansions on quoted inner braces**

`parameter_expansion_end` decrements `depth` for every `}`, even when it belongs to a quoted word inside `$(...)`. For example, `curl ${HEADER_OPTION:-$(printf '}' >/dev/null; printf -- -H)}"Cookie: session=parameter-secret-42"` is valid in both Bash and Zsh and expands to one `-HCookie: session=...` argument. On this head the shell history gate classifies it as sensitive, but `redact_shell_cookie_headers` returns the original string with `changed=false`: the quoted `}` ends this scan early and the following `;` moves `word_start` past the dynamic option. This leaves the Cookie value unredacted on durable evidence/provider surfaces. Honor the nested command/quote context here, or fail closed whenever the matching outer brace cannot be proven.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
```diff
@@ -351,7 +359,12 @@ fn relay_native_passthrough(
         .concat();
         relay.line_buffer.soft_newline_enabled = native_candidate_allows_soft_newline(&combined);
         relay.line_buffer.push(bytes);
+        let hold_zsh_path_candidate = relay.zsh_path_prompt_buffering.is_some()
+            && !relay.line_buffer.saw_paste()
+            && relay.line_buffer.bytes.first() == Some(&b'/')
+            && !relay.line_buffer.bytes.contains(&b'\t');
```
ikunkun-sys, line 374: **[P1] Split Tab from later bytes before returning ownership to ZLE**

This clears the hold based on the presence of Tab, but the input reader passes arbitrary chunks of up to 8192 bytes and no layer splits a chunk at Tab. If an active `/...` candidate receives `b"\t\r"` in one read, `line_buffer.push` stores both bytes and `flush_candidate_line_to_shell` writes the buffered path, Tab, and Enter to the PTY as one batch. A custom Tab widget can rewrite `BUFFER`, then the queued Enter immediately accepts the replacement—the same unintended execution boundary as the prior thread. The new direct and E2E tests send a standalone Tab followed later by Ctrl-C, so they do not cover coalescing. Hand only the bytes through Tab to ZLE before processing the remainder under shell ownership, and cover a one-batch Tab-plus-Enter input.

ikunkun-sys (changes requested):
Requesting changes for two P1 findings on b04047e. The previous non-leading-slash routing regression is fixed, and all executed GitHub checks plus fmt, clippy, the fragmented-path E2E, the standalone-Tab E2E, and the existing redaction/marker tests pass. However, Tab and Enter can still be replayed together when they are coalesced in one input read, and the new braced-expansion scanner can miss a real dynamically constructed Cookie header when an inner command substitution contains a quoted `}`. Both have concrete execution or redaction impact on this exact head.

src/cosh-ng/crates/cosh-shell/src/evidence/redaction/shell_word.rs
kongche-jbw: Fixed in `397f27e0`. The braced-expansion scanner now stops claiming a precise outer close when quotes, command substitution, process substitution, or backticks introduce nested shell syntax; that ambiguity fails closed to whole-command redaction. Added the exact quoted-inner-`}` Cookie PoC plus a matching `Crookie` safe control. Targeted fmt, clippy, redaction tests, and all GitHub checks pass.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay.rs
kongche-jbw: Fixed in `675ef28e`. For a Zsh candidate read containing Tab plus later bytes, Rust now hands bytes through Tab to ZLE first and preserves intervening edits, but suppresses a submit coalesced into the same PTY read; PTYs do not preserve producer write boundaries, so a fresh Enter is required after the widget-rendered line is visible. Added a direct one-batch `b"\t\r"` relay test and an end-to-end custom-widget regression proving the rewritten command is not executed. Targeted fmt, clippy, raw-cli tests, and all GitHub checks pass.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/path_prompt_submit.rs
```diff
@@ -15,6 +15,11 @@ pub(super) fn route_missing_path_submission(
...
+        return Ok(false);
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in `8c50ef9d`. Instead of re-enabling the Bash-specific `Ctrl-U` fallback in Zsh, the relay now owns an unresolved leading-space prefix from its first (possibly fragmented) read and keeps it Rust-owned once it resolves to a slash/Han path prompt. A normal ASCII command is returned to ZLE byte-identically as soon as it resolves; raw Tab remains deliberately editor-owned and keeps the existing coalesced-input protection. The E2E splits the first space into its own read, installs a custom `accept-line`, and proves the prompt is routed without reaching ZLE; the inverse ASCII and Tab controls also pass. Targeted fmt, clippy, unit/raw-cli tests, layout audit, and all GitHub checks pass.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
```diff
@@ -0,0 +1,50 @@
...
+    complete_paste: bool,
+) -> io::Result<Option<bool>> {
+    if relay.zsh_path_prompt_buffering.is_none()
+        || complete_paste
+        || relay.line_buffer.saw_paste()
+        || !relay.line_buffer.pending_partial_bytes().is_empty()
+    {
+        return Ok(None);
+    }
+    let Some(tab) = bytes
```
ikunkun-sys: **[P1] Process an earlier submit before a later Tab**

This searches the whole read for Tab before considering whether an Enter precedes it. With an active `/definitely-not-command` candidate, a single `b"\r\t"` read falls through because Tab is the final byte; the outer relay then sees Tab and flushes the buffered path plus Enter to ZLE. On acfbe65c, an end-to-end custom-ZLE probe produced `zsh: no such file or directory: /definitely-not-command`, proving that Zsh executed the path instead of Agent routing handling the earlier Enter. The input reader accepts arbitrary chunks of up to 8192 bytes, so routing semantics cannot depend on these two keys landing in separate reads.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
```diff
@@ -0,0 +1,50 @@
...
+        // later read-ahead depends on the suppressed line boundary.
+        if submit > 0 {
+            relay_passthrough_input_with_policy(
+                &remainder[..submit],
+                relay,
+                emit_activity,
+                pending_shell_submits,
+            )?;
+        }
+        return Ok(Some(true));
```
ikunkun-sys, line 92: **[P1] Preserve the tail after suppressing a coalesced submit**

When `remainder` contains Enter, this forwards only the bytes before it and returns, discarding both the submit and every later byte in the same read. On acfbe65c, sending a path candidate followed by one `b"\t\r echo PR3004-TRAILING-INPUT-RAN\r"` chunk ran the custom Tab widget, but the marker command never appeared; a command sent in a later read still executed. Thus ordinary type-ahead is silently lost whenever it is coalesced after the protected Enter.

ikunkun-sys (changes requested):
Requesting changes for two P1 findings on acfbe65c. The leading-space routing change, the prior redaction regression, the targeted Zsh E2E tests, raw-input unit suite, fmt, layout, and workspace all-targets clippy pass, and all executed GitHub checks are green. However, a coalesced Enter-before-Tab read can still execute a path candidate in Zsh instead of routing it to the Agent, while a Tab-before-Enter read silently discards every byte after the suppressed submit. Both behaviors were reproduced end to end on this exact head and do not duplicate an existing thread.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
kongche-jbw: Addressed in 9514a380. The relay now splits at an earlier submit boundary before it evaluates a later Tab, with unit and Zsh PTY coverage for a qualifying slash-bearing natural-language prompt followed by `b"\r\t"`. The exact English-only `/definitely-not-command` sample remains intentionally Shell-owned under #1943’s acceptance contract (English typo paths stay native); the corrected invariant is that the later Tab can no longer change the earlier submit decision. All required checks on 9e557cd7 passed, including Test cosh-ng, fast checks, release build, and the prebuilt action.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
kongche-jbw: Fixed in 9514a380. Bytes after the protected submit are now retained in the Zsh route state instead of being dropped. On the next terminal read, cosh first cancels the potentially widget-rewritten ZLE buffer and then replays the retained bytes as the next logical line. The regression verifies the Tab widget still runs, its rewritten command does not execute, and coalesced trailing type-ahead is preserved. All required checks on 9e557cd7 passed.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
```diff
@@ -0,0 +1,86 @@
...
+            )
+            + 1;
+        relay_passthrough_input_with_policy(
+            &bytes[..submit_end],
+            relay,
+            emit_activity,
+            pending_shell_submits,
+        )?;
+        if submit_end < bytes.len() {
+            relay_passthrough_input_with_policy(
```
ikunkun-sys: **[P1] Re-dispatch the remainder under the new input owner**

Processing `bytes[..submit_end]` can classify the natural-language path prompt, emit the Agent intercept, and change `input_mode` to `Delay`, but this recursive call sends the remaining Tab back through passthrough without consulting that new owner. On 9e557cd7, I added a custom `^I` widget to the qualifying leading-space Zsh PTY case and sent the existing single `b"\n\t"` chunk: the Agent response was rendered and `tab-after-agent-submit-reached-zle` was also printed, proving the later key escaped into ZLE. Route or defer the suffix through the current mode rather than continuing in the stale passthrough context.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
```diff
@@ -0,0 +1,86 @@
...
+                remainder.get(submit) == Some(&b'\r') && remainder.get(submit + 1) == Some(&b'\n'),
+            )
+            + 1;
+        if submit_end < remainder.len() {
+            // PTYs do not preserve write boundaries, so the current read
+            // cannot safely cancel a buffer that the Tab widget may rewrite.
+            // Hold later type-ahead until the next read, then cancel and
+            // replay it as the next logical line.
+            if let Some(buffering) = relay.zsh_path_prompt_buffering.as_deref_mut() {
+                buffering.defer_tab_typeahead(&remainder[submit_end..]);
```
ikunkun-sys, line 89: **[P1] Drain a submitted tail without another input read**

These bytes may already contain a complete next line and Enter, but `take_tab_typeahead` is called only at the start of a later `relay_passthrough_input_with_policy` invocation. With `/definitely-not-command` followed by one `b"\t\r: > \"$HOME/deferred-tail-ran\"\r"` chunk, the file was still absent after one second and appeared only after sending unrelated `echo wake\r`; `finish_input_relay` also never drains this field at EOF. The safety hold must schedule or drain the retained submitted tail independently of a future user keystroke, otherwise ordinary type-ahead stalls indefinitely or is dropped on shutdown.

ikunkun-sys (changes requested):
Requesting changes for two P1 findings on 9e557cd7. The earlier-submit routing decision and tail retention are improved, and all executed GitHub checks plus fmt, layout, workspace all-targets clippy, raw-input 234/234, targeted redaction checks, and the added Zsh PTY tests pass. However, bytes after an Agent route cutover are still sent through stale passthrough ownership, allowing a coalesced Tab to run a user ZLE widget, while a complete retained tail cannot run until unrelated new input arrives and is discarded at EOF. Both behaviors were reproduced end to end on this exact head; all 13 existing threads are resolved and neither current-head failure mode is already reported.

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
kongche-jbw: Fixed in b651b6d1. The prefix relay now returns whether submission routed to Cosh; after that ownership cutover, the same-read suffix follows the held-input path instead of being recursively dispatched through stale ZLE passthrough. The Zsh PTY regression installs a custom Tab widget and verifies the marker is absent. All required checks passed on 68ad7198 (10 successful, 0 failing).

src/cosh-ng/crates/cosh-shell/src/raw_input/relay/tab_handoff.rs
kongche-jbw: Fixed in b651b6d1. Retained Tab typeahead now has a bounded 50 ms deadline in the input relay, appends bytes that arrive during the handoff, drains from the timeout path without another read, and is also drained at EOF after the same deadline. The PTY test holds stdin open with no follow-up read, and the unit test covers EOF drainage. All required checks passed on 68ad7198 (10 successful, 0 failing).

ikunkun-sys (approved):
Approved at `68ad7198`. The two prior P1 findings are fixed: coalesced Enter→Tab no longer crosses the Agent ownership cutover into ZLE, and retained Tab→Enter→tail input now drains autonomously (including at EOF) without executing the widget rewrite. No new blocking findings. I re-ran fmt/layout, workspace all-targets clippy, raw-input 234/234, the two targeted Zsh PTY regressions, targeted redaction checks, and an independent no-follow-up-input Zsh probe. Current GitHub checks are green.

PR #3007, opened 2026-09-01
feat(sec-core): add v2 foundation types

RemindD:
## Risk and compatibility


kongche-jbw (commented):
Review baseline: `572e93b0dc07...5818fcda003f`

**[P1] Reject inconsistent mapping aggregates**

`src/agent-sec-core/v2/crates/policy/asc-policy-types/src/target.rs:191` makes
`overallRelation` an independent public wire field. An Adapter can return
`policy.relation=unsupported` with `overallRelation=exact`, and serde accepts the report.
Because a translated plan is explicitly subject to the PCP mapping gate, a gate that reads the
advertised aggregate can dispatch a plan whose detailed evidence says it is unsupported.
Possible direction: derive the aggregate, or validate it against the policy, scope, and guarantee
relations, with a regression test that rejects an inconsistent report.

**[P2] Reuse the shared foundation revision and resource ID**

`src/agent-sec-core/v2/crates/policy/asc-policy-types/src/identifiers.rs:103` redefines
`ResourceId`, and line 108 redefines `Revision`, instead of using `asc-foundation-types`.
The two resource IDs even disagree on the maximum length (256 versus 128 bytes), while the two
revisions are distinct Rust types. A value can therefore cross one contract and fail the next,
and every consumer must add conversions between nominally shared primitives.
Possible direction: depend on and re-export the foundation types, keep only policy-local IDs here,
and add a cross-crate wire-compatibility test.

**[P2] Canonicalize port lists before duplicate detection**

`src/agent-sec-core/v2/crates/policy/asc-policy-types/src/resource.rs:455` rejects zero and
duplicate ports but accepts both `[80, 443]` and `[443, 80]`. Two host matchers with the same
pattern and those lists both validate because matcher equality is order-sensitive, even though
they select the same endpoints. This bypasses the duplicate-matcher gate and gives the same
Canonical IR semantics multiple wire forms.
Possible direction: require strictly ascending ports or normalize before matcher comparison, and
cover permuted lists in the duplicate-matcher regression test.

**[P2] Return paths that exist in the serialized document**

`src/agent-sec-core/v2/crates/policy/asc-policy-types/src/ir.rs:425` forwards Atom errors
without the `when` prefix, and `resource.rs:236` forwards path-matcher errors without `path`.
An unknown target reports `rules[0].atom.target` instead of
`rules[0].when.atom.target`; an invalid glob reports
`resources[0].matchers[0].pattern` instead of
`resources[0].matchers[0].path.pattern`. Clients cannot locate the failing field even though
`ValidationError` promises a stable JSON-style path.
Possible direction: add prefixes at each containing field and test that every reported path exists
in the serialized payload.

kongche-jbw (commented):
Review baseline: `d47952085771...9ed335cc4f36`

**[P1] Make translated plans independently gateable**

`src/agent-sec-core/v2/crates/policy/asc-policy-types/src/target.rs:62` represents a successful
translation without source-coverage evidence or validation against the Adapter's advertised
artifact contracts. If an Adapter omits one deny rule, or returns a contract ID/media type it did
not advertise, serde still accepts `Translated`; the caller sees only opaque bytes, diagnostics,
digests, and capabilities, so it cannot detect the weaker or mismatched plan before dispatch.
`Rejected` only helps when the same Adapter notices its own omission.
Possible direction: validate a translated result against the source Binding and negotiated
descriptor, carrying complete per-rule/guarantee evidence where needed, and add regressions for a
missing source rule and an unadvertised artifact contract.

**[P2] Reject trailing separators in normalized paths**

`src/agent-sec-core/v2/crates/policy/asc-policy-types/src/resource.rs:412` accepts both `/vault`
and `/vault/` for `Exact` or `Prefix`, although those variants promise normalized paths. For a
directory these spellings resolve to the same entry, but `FileMatcher` equality treats them as
different, so both pass the duplicate-matcher gate and give one resource set multiple canonical
wire forms. Possible direction: reject a trailing slash except for `/`, and cover the two
spellings in the duplicate-matcher regression test.

edonyzpc (approved):
LGTM

PR #3016, opened 2026-09-01
fix(sight): use append_policy_delta to prevent fail-open enforcement

viron-xiao:
## Problem

PR #2979 (commit 066fb14) switched the runtime delta injection path from `append_policy_delta` to `reload_policy_delta`. The RELOAD path performs fire-and-forget writes to the BPF ring **without verifying** that the kernel actually installed the policy rules.

This causes `install_prepared_locked` to report `state=Enforced` while **no rules are active** in the BPF capability engine — a **silent fail-open** security regression.

### Evidence

On Alibaba Cloud Linux 4 / kernel 7.1.10 with current main:
- `enforce_unlink_demo`: reports `state=Enforced`, but `rm` succeeds and file is deleted
- HTTP `POST /bindings` with `block unlink` DSL: reports `state=enforced`, file deleted, no violation

### Root Cause

`reload_policy_delta` uses raw `self.submit(...)` which writes to the BPF ring and returns Ok immediately. The prebuilt BPF object does not process RELOAD tags (-1/-2/-3), so submissions are silently discarded.

In contrast, `append_policy_delta` uses `submit_expect_count` which polls the BPF map counter after each submission, failing loudly when the kernel rejects or silently drops a rule.

## Fix

One-line change: switch back to `append_policy_delta(control_pid, id, &compiled.bytes)`.

The control process is already bound with `AUTH_BIND_RULE` authority via `bind_state` (line 241), so the APPEND admission check passes. This was the original working path before the RELOAD switch.

## Verification

Tested on 47.111.114.232 (Alibaba Cloud Linux 4 / kernel 7.1.10):

| Test | Before fix | After fix |
|------|-----------|-----------|
| `enforce_unlink_demo` | `rm` succeeds, FAIL | `rm: Operation not permitted`, PASS |
| HTTP `/bindings` + `block unlink` | File deleted, no violation | File survives, violation `effect=block, blocked=true` |
| Node.js multi-threaded gateway | File deleted | `child_process.spawn` tool blocked |
| `DELETE /bindings` | 204 ✓ | 204 ✓ |
| Path ≥64 bytes | `compile_failure` ✓ | `compile_failure` ✓ |

### DIFF
--- a/src/agentsight/crates/agentsight-enforcer/src/actplane.rs
+++ b/src/agentsight/crates/agentsight-enforcer/src/actplane.rs
@@ -246,10 +246,13 @@ impl ActPlaneBackend {
                 cleanup,
             ));
         }
-        if let Err(error) = self.reload.reload_policy_delta(id, &compiled.bytes) {
+        if let Err(error) = self
+            .reload
+            .append_policy_delta(control_pid, id, &compiled.bytes)
+        {
             let cleanup = self.cleanup_binding(&request, id);
             return Err(kernel_error_with_cleanup(
-                "reload policy delta",
+                "append policy delta",
                 error,
                 cleanup,
             ));


jfeng18 (approved):
## Deep review — verified the fail-open diagnosis and the fix's safety at every layer

### The fail-open on main (confirmed by code, not just the report)

Traced the full RELOAD path on main:

1. `reload_policy_delta` (ebpf-ifc-engine/src/lib.rs:3072) submits `CAP_REQ_RELOAD_COUNTS/UPDATE/RULE` (tags **-1/-2/-3**) via raw `submit()` and returns `Ok(())` unconditionally.
2. The BPF side (`capability.bpf.h` `cap_request_cb`) handles **only** `CAP_REQ_APPEND_UPDATE(-4)` and `CAP_REQ_APPEND_RULE(-5)`. Tags -1/-2/-3 do not exist anywhere in BPF C code — I also checked the actplane patch series (0005 adds the Rust-side RELOAD helpers only, zero BPF changes).
3. An unmatched tag falls into the "normal capability delta request" arm, where the first field is read as `caller_pid` — i.e. the tag value (-1/-2/-3) — and `r->caller_pid != ctx->current_pid` → `cap_count(CAP_STAT_DROP)`, silently discarded.
4. Userspace still returns `Ok(())` → `install_prepared_locked` reports `state=Enforced` with zero active rules. **Silent fail-open confirmed.**

### Why the one-line revert is sound (the part I probed hardest)

`ReloadHandle.append_policy_delta` already exists on main with the full fail-loud machinery: `precheck_append_authority` (userspace authority check per update/rule), `submit_expect_count` (polls the `ts_counts` slot after every submission), and `restore_append_state` rollback on partial failure. The call-site prerequisites all hold:

- `bind_state(control_pid, id, control_state)` (actplane.rs:225-243) writes `cap_task[control_pid]=id` and `cap_state[id]=control_state` with `AUTH_BIND_RULE | … | TARGET_SELF | gate_mask=u64::MAX | label_mask=u64::MAX` *before* the delta injection — both the userspace precheck and the BPF-side `cap_check_fields` read exactly these maps.
- `submit_raw` self-triggers the drain with `libc::syscall(SYS_getpid)` after every submit, so the drain callback runs in the enforcer's own context and `cap_can_submit_for` passes trivially.

### On #2979's original motivation — the one genuine question

#2979 switched to RELOAD citing "map visibility timing between userspace writes and BPF drain callbacks". That explanation does not hold architecturally (userspace `bpf_map_update_elem` is synchronously visible to BPF), but two real scenarios can still make APPEND fail:

- **Drain-context race**: the drain hook is `sys_enter_getpid` on *any* process, so an agent's getpid can drain the ringbuf before the enforcer's self-trigger fires; `cap_can_submit_for(agent_pid, enforcer_pid)` then requires `AUTH_DELEGATE` the agent doesn't hold → REJECT. Crucially, with `submit_expect_count` this surfaces as a **loud error** (count never advances → `Err` → rollback → binding fails cleanly), never as a false `Enforced`.
- **pid-namespace mismatch**: if the enforcer runs namespaced (the SecCore/AP-02 context #2979 was built for), `std::process::id()` differs from BPF's global pid → same loud-failure outcome.

So the revert trades #2979's integration blocker for an *honest* failure mode. Given main currently reports Enforced while enforcing nothing, this is the right emergency call — an unimplemented scenario should fail loudly, not lie.

### Independent gates (ECS, kernel 6.6.102)

fmt ✓ / `clippy -p agentsight-enforcer --all-targets -D warnings` ✓ / enforcer tests: 40 passed, 0 failed (install path exercised via mock backend). CI on the PR is green. The author's real-machine evidence table (rm → `Operation not permitted`, HTTP binding → violation with `effect=block, blocked=true`, DELETE and ≥64-byte paths unchanged) covers the regression surface.

### Follow-ups I'd suggest (non-blocking)

1. The RELOAD protocol is now dead code on the userspace side with no BPF counterpart — either implement the -1/-2/-3 handlers in BPF or delete `reload_policy_delta`, otherwise the next caller hits the same trap. Its doc comment actively advertises a protocol the kernel does not implement.
2. Nothing in CI loads the real BPF engine and asserts enforcement *effectiveness* — that is how a fail-open shipped green through 42f5e3b8. A root-level e2e assertion (`rm` must fail after install) would close the class, not just this instance.
3. The drain-context race deserves a deterministic trigger eventually (dedicated syscall or per-CPU drain), since today correctness relies on the submitter's getpid winning the race.

Approving — this is the correct minimal fix for a security regression on main, with verified fail-loud semantics on the restored path.

PR #3022, opened 2026-09-01
fix(sight): resolve PID namespace for BPF cap_task seed

viron-xiao:
## Problem

When the protected agent process runs inside a PID namespace (containers, WSL2, systemd-nspawn), the user-supplied `root_pid` is the namespace-local PID. However, the BPF `handle_fork` hook uses `bpf_get_current_pid_tgid() >> 32`, which returns the **global kernel PID**.

If `seed_label_in_domain` writes the namespace PID into `cap_task`, `handle_fork` queries with the kernel PID and gets a miss — child processes never inherit the domain, and file deletion protection silently fails.

### Evidence (from SecCore testing on WSL2 6.18)

- User-space PID: gateway=121550, child=500469→500470
- Kernel tracepoint PID: 2364/24869/... (different set)
- `cap_task[121550]` exists (written by user-space seed)
- `handle_fork` queries `cap_task[2364]` → Not found → no propagation
- File deleted, no violation

### Root Cause

`bpf_get_current_pid_tgid()` returns `task->tgid` (global kernel PID), not the `/proc` namespace PID. In root PID namespace they are identical; in nested namespaces they differ.

## Fix

Read `/proc/<pid>/status` `NSpid:` line to translate the namespace PID to the global kernel PID before writing `cap_task`. The first value in `NSpid` is always the global PID.

```rust
fn resolve_kernel_pid(ns_pid: i32) -> i32 {
    // /proc/<pid>/status contains:
    //   NSpid: <global_pid> <ns_pid>
    // When only one level, we're in root namespace — no translation needed.
}

### DIFF
--- a/src/agentsight/crates/agentsight-enforcer/src/actplane.rs
+++ b/src/agentsight/crates/agentsight-enforcer/src/actplane.rs
@@ -44,6 +44,10 @@ pub const ACTPLANE_REVISION: &str = "a62e5d9d96f91101cda019519053e950d532380a";
 struct ActiveBinding {
     binding: Binding,
     credential_policy: Option<CredentialExfiltrationPolicy>,
+    /// When the target process runs inside a PID namespace, this holds the
+    /// global kernel PID resolved by [`resolve_kernel_pid`].  `None` when
+    /// the namespace PID equals the kernel PID (root namespace).
+    kernel_pid: Option<i32>,
     reasons: Vec<String>,
     rule_names: Vec<String>,
     label_names: HashMap<u64, String>,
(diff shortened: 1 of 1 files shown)


jfeng18 (approved):
## Deep review — PID namespace resolution for cap_task seed

### Problem statement (verified against the BPF side)

The claim is that `handle_fork` (`process.bpf.c:2511`) reads `bpf_get_current_pid_tgid() >> 32` — the global kernel PID — and looks it up in `cap_task`. If `seed_label_in_domain` writes the namespace-local PID, `cap_fork` (capability.bpf.h:546) queries with the global PID and misses. Verified: `handle_fork` → `te_fork` → `cap_fork(ppid, cpid)` where `ppid` is the global PID from the tracepoint context. Correct.

### The fix (resolve_kernel_pid)

The `/proc/<pid>/status` `NSpid:` approach is the standard way to resolve namespace PIDs from host context. Key semantics verified:

- `NSpid` lists PIDs from innermost (the process's own namespace) to outermost (the reader's namespace). Since the enforcer reads host `/proc`, the first value is the global (host-namespace) PID. Correct.
- Root namespace: single value, no translation needed — the code only translates when `parts.next().is_some()` (multiple levels). Correct.
- Read failure (process already exited): returns the original PID, no panic. Correct.

### What I probed and found sound

- **No other cap_task write points need the same fix**: `seed_label_in_domain` and `unbind_pid_from_domain` are the only paths that write to `cap_task`. The fix covers both via `kernel_pid` propagation through `ActiveBinding` and `cleanup_binding`.
- **`protect_pid` uses `control_pid`** (the enforcer's own PID, which runs on the host) — no translation needed. Correct.
- **cleanup_binding symmetry**: `seed` uses `kernel_pid`, `cleanup` uses `kernel_pid.unwrap_or(request.root_pid)`. Symmetric — same PID is used for seed and cleanup. Correct.
- **BPF-side `cap_fork` is called from `te_fork`** which is called from `handle_fork` (sched_process_fork tracepoint). The tracepoint context gives the global PID of the parent, not the namespace PID. The fix aligns the userspace seed with what the BPF side expects. Correct.

### What's missing

**No tests for `resolve_kernel_pid`**. It's a pure function reading a well-defined file format, but zero unit tests cover the three branches: (a) root namespace (single NSpid value → return unchanged), (b) nested namespace (multiple values → return first), (c) read failure (process exited → return unchanged). A test writing a synthetic `/proc/self/status` string to a temp path and asserting the translation would be trivial and would prevent a future regression where the parsing logic changes.

The author's real-machine evidence (WSL2 6.18, file deleted vs blocked) covers the happy path, but a unit test would make the fix self-documenting and CI-verifiable.

### Independent gates

fmt ✓ / clippy ✓ / enforcer tests pass on ECS. CI green on the PR.

### Verdict

The fix is correct and minimal. The one gap is the missing unit test for `resolve_kernel_pid` — the three branches (root ns, nested ns, read failure) each need one discriminating test. I'd ask for that before merge, but it's a small ask.

**Review state: APPROVE with the test request as a follow-up comment.**

PR #3025, opened 2026-09-02
fix(sight): pin Dockerfile RPM glob to TARGETARCH

jfeng18:
Follow-up to #2976 review comment: the RPM glob `/tmp/rpms/*/agentsight-*.rpm` matched both `rpms/x86_64/` and `rpms/aarch64/` in a multi-arch build context, so dnf could see a conflict or install the wrong arch.

Maps Docker `TARGETARCH` (amd64/arm64) onto the RPM directory naming (x86_64/aarch64) via a case statement, so only the target arch dir is globbed.

### DIFF
--- a/src/agentsight/packaging/docker/Dockerfile
+++ b/src/agentsight/packaging/docker/Dockerfile
@@ -19,11 +19,17 @@ LABEL org.opencontainers.image.title="AgentSight" \
 
 RUN sed -i -e "s/cloud.aliyuncs/aliyun/g" /etc/yum.repos.d/*.repo
 
-# rpm-build.sh emits arch dirs in RPM naming (x86_64/aarch64); copy the whole
-# tree and glob one level down so Docker's TARGETARCH naming (amd64/arm64)
-# mismatch cannot break the build.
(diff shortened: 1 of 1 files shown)


chengshuyi (approved):
LGTM. Verified the TARGETARCH→RPM arch mapping against rpm-build.sh output layout and confirmed docker-nightly (Dockerfile.alinux) is unaffected. Two nits noted verbally: the 'rpm-build.sh emits arch dirs' comment is slightly inaccurate, and the header could spell out rpms/x86_64|aarch64 — non-blocking.

chengshuyi (approved):
LGTM. Verified: the multi-arch context glob conflict is real (rpmbuild emits x86_64/aarch64 dirs; old */agentsight-*.rpm glob would hand both arches to dnf). TARGETARCH mapping is correct and fail-fast on unsupported arch is good. Does not affect docker-nightly (uses docker/Dockerfile.alinux). Two non-blocking nits: (1) comment 'rpm-build.sh emits arch dirs' is slightly inaccurate — rpm-build.sh only produces the source tarball; the arch dirs come from rpmbuild's standard output assembled by the builder; (2) header comment could spell out rpms/x86_64|aarch64/ explicitly.

PR #3041, opened 2026-09-02
fix(sight): resolve namespace PID to host PID with init-ns self-check

viron-xiao:
## Problem

PR #3022's `resolve_kernel_pid` was a no-op. It never changed the input PID because `/proc/<pid>/status NSpid[0]` always equals the input when read from the same namespace.

Additionally, the serve layer's `validate_target_identity` rejected namespace-local PIDs (e.g. container PID 1) before they reached the enforcer, because PID 1 on the host is systemd with a different start_time.

## Fix

Replace `resolve_kernel_pid` with `resolve_to_host_pid` that uses **pid + start_time joint matching**:

1. **Fast path**: `/proc/<pid>/stat` exists and `start_time` matches → return `NSpid[0]` (host PID). O(1).
2. **Slow path**: scan `/proc/*/status` for innermost `NSpid == input_pid` with matching `start_time` → return host PID. O(n), runs once per binding.
3. **Fallback**: no match → return input unchanged.

Resolution runs in **both** serve (`validate_target_identity`) and enforcer (`prepare_binding`), **before** start-time validation.

Also adds **init-ns self-check**: compares `/proc/self/ns/pid` with `/proc/1/ns/pid`; if different, `file_delete_guard` reports `false` (fail-closed).

## Verification (real nested PID namespace on 47.111)

| Test | Input | Result |
|------|-------|--------|
| T1: host PID, basic block | pid=host | **PASS** (fast path, rm→EPERM) |
| T2: host PID, multi-level fork | gateway→bash→sh→rm | **PASS** |
| T3: ns-local PID=1, container target | unshare --pid, pst matching | **PASS** (slow path, resolved 1→42971, rm→EPERM) |
| health | init-ns check | `file_delete_guard=true` |

T3 was previously FAIL with PR #3022 (`invalid_target` / no-op resolve).

## Files changed

- `crates/agentsight-enforcer/src/actplane.rs` — enforcer-side resolve + init-ns field + health gate
- `src/enforcement/target.rs` — shared `resolve_to_host_pid`, `is_init_pid_namespace`, helpers
- `src/enforcement.rs` — re-exports
- `src/server/enforcement.rs` — serve-side resolve before validate

### DIFF
--- a/src/agentsight/crates/agentsight-enforcer/src/actplane.rs
+++ b/src/agentsight/crates/agentsight-enforcer/src/actplane.rs
@@ -45,7 +45,7 @@ struct ActiveBinding {
     binding: Binding,
     credential_policy: Option<CredentialExfiltrationPolicy>,
     /// When the target process runs inside a PID namespace, this holds the
-    /// global kernel PID resolved by [`resolve_kernel_pid`].  `None` when
+    /// global kernel PID resolved by [`resolve_to_host_pid`].  `None` when
     /// the namespace PID equals the kernel PID (root namespace).
     kernel_pid: Option<i32>,
     reasons: Vec<String>,
@@ -98,6 +98,10 @@ pub struct ActPlaneBackend {
     stop: Arc<AtomicBool>,
     poller: Mutex<Option<JoinHandle<()>>>,
     lifecycle: Mutex<()>,
+    /// True when the enforcer is running in the init PID namespace.
+    /// File-delete-guard requires this for correct cap_task seeding and
+    /// domain propagation.  When false, health reports file_delete_guard=false.
+    in_init_pidns: bool,
 }
 
 impl ActPlaneBackend {
@@ -151,6 +155,7 @@ impl ActPlaneBackend {
             stop,
             poller: Mutex::new(Some(poller)),
             lifecycle: Mutex::new(()),
+            in_init_pidns: is_init_pid_namespace(),
         })
     }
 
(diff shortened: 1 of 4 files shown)


jfeng18 (commented):
## Deep review — namespace PID resolution with init-ns self-check

### The core claim: #3022's `resolve_kernel_pid` was a no-op

Verified. The old code read `/proc/<ns_pid>/status` and returned `NSpid[0]` when multiple values existed. But:

- If `ns_pid` is a valid host PID (a process exists with that host PID), NSpid[0] = ns_pid — no translation, and if it's a PID collision, we seed the WRONG process into `cap_task`.
- If `ns_pid` is NOT a valid host PID (namespace-local from a container), `/proc/<ns_pid>/status` doesn't exist on the host — read fails, return ns_pid unchanged.

So #3022 was a no-op in both cases. The new `resolve_to_host_pid` fixes this with start_time disambiguation: fast path (host PID exists + start_time matches) → NSpid[0]; slow path (scan /proc for innermost NSpid + start_time match) → host PID; fallback → return unchanged.

### What I verified

- **NSpid ordering**: kernel lists outermost-to-innermost; the code checks `nspid.last()` for the innermost (namespace-local) match and returns `nspid[0]` for the host PID. Correct.
- **start_time disambiguation**: prevents PID collision from seeding the wrong process. The `validate_target_identity` in the serve layer previously rejected container PID 1 because PID 1 on the host is systemd with a different start_time — this was the actual failure mode the PR fixes.
- **Init-ns self-check**: `is_init_pid_namespace()` compares `/proc/self/ns/pid` with `/proc/1/ns/pid`. Correct for the intended deployment (enforcer on host). Gates `file_delete_guard` — fail-closed when not in init ns.
- **Idempotent resolution**: serve layer resolves in `apply_binding`, enforcer resolves again in `prepare_binding`. Second resolution is a no-op (host PID already resolved). Redundant but harmless.
- **Tests**: 8 new tests cover fast path, fallback, wrong start_time, NSpid chain reading, init-ns check, and end-to-end. The slow path (nested namespace) requires root + unshare — the author's real-machine verification (T3 on 47.111) covers it.

### Issues found

1. **Clippy failure**: `zombie_processes` at `target.rs:335` — the test spawns `sleep 10` and kills it via `nix_kill` but never calls `.wait()` on the child handle. Fix: bind the child and call `.wait()` after the kill, or use `Command::new("sleep").arg("10").spawn()` → `child.wait()` after the assertions.

2. **Code duplication**: `is_init_pid_namespace()` exists in both `target.rs` (marked `#[allow(dead_code)]`) and `actplane.rs` (the one actually used by the enforcer). The target.rs comment says "Used by agentsight-enforcer crate (actplane.rs has its own copy)" — this is misleading. Either remove the target.rs version or have the enforcer use the shared one.

3. **`nix_kill` naming**: the helper is named `nix_kill` but uses `Command::new("kill")` instead of `nix::sys::signal::kill`. If the intent was to avoid a nix dependency, the name should reflect that.

### Independent gates (ECS, kernel 6.6.102)

fmt ✓ / 28 target tests pass (all new resolve tests) / clippy fails on the zombie-process issue above.

### Verdict

The fix is correct and addresses a real bug. The clippy issue needs fixing before merge. The code duplication and naming issues are minor.

**Review state: COMMENT — fix the clippy zombie_processes error before merge.**

Daydreamer-Li (approved):

PR #3046, opened 2026-09-03
fix(sight): retry startup db purge on lock

Daydreamer-Li:
## Description

The startup size-based purge for `agentsight.db` gave up on the first `SQLITE_BUSY`/`SQLITE_LOCKED` error. When a large (500MB+) database is reopened right after the previous process exits, it collides with that process's exit-time WAL checkpoint, which holds the write lock past the 500ms `busy_timeout` — so the oversized DB was left uncleaned at startup until the insert-count purge eventually compensated (an unbounded delay on idle machines). This retries the startup purge with `1s/3s/9s` backoff (~13s cap), retrying only lock errors while keeping other errors (e.g. corruption) terminal.

## Related Issue

refs #2870

## Risk and compatibility

- Behavior-preserving on the happy path and on non-lock errors. Only adds bounded retries for transient locks; if all attempts are locked out the purge is skipped exactly as before (now with an explanatory WARN).
- Worst-case added startup latency is ~13s, and only on a large-DB restart that is actively locked; no change on the common path.
- No config, schema, or public API changes.

## Validation

- `cargo clippy --lib --all-targets -- -D warnings`
- `cargo test --lib storage::unified` — 15 passed (incl. new `retryable_lock_matches_busy_and_locked` / `retryable_lock_rejects_terminal_errors`)

Lock-classification is covered by unit tests; the timing-dependent lock collision itself is not reproduced in CI by design.

## Documentation and rollback

- No docs affected. Rollback is a straight revert of the single commit; the pre-existing insert-count purge continues to enforce the size cap.

### DIFF
--- a/src/agentsight/src/storage/sqlite/token.rs
+++ b/src/agentsight/src/storage/sqlite/token.rs
@@ -490,10 +490,11 @@ impl TokenStore {
             )",
             self.table_name, self.table_name
         );
-        let deleted = self
-            .conn
-            .execute(&sql, params![limit as i64])
-            .map_err(|e| anyhow::anyhow!("Failed to delete oldest token records: {e}"))?;
+        // Propagate the rusqlite error unwrapped (like the sibling stores) so a
+        // transient SQLITE_BUSY/SQLITE_LOCKED stays downcastable and the startup
+        // purge's lock-retry classifier can recognize it; wrapping it in a
+        // formatted `anyhow!` here would erase the cause and make the retry a no-op.
+        let deleted = self.conn.execute(&sql, params![limit as i64])?;
         Ok(deleted)
     }
 
--- a/src/agentsight/src/storage/unified.rs
+++ b/src/agentsight/src/storage/unified.rs
@@ -38,7 +38,7 @@
 use anyhow::Result;
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU64, Ordering};
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use super::sqlite::connection::default_base_path;
 use super::sqlite::{AuditStore, HttpStore, TokenConsumptionStore, TokenStore};
@@ -183,12 +183,29 @@ impl Storage {
         // crash or a config change), prune immediately rather than waiting for
         // the next purge interval.
         if storage.max_db_size_bytes > 0 {
-            if let Err(e) = storage.purge_oversized() {
-                log::warn!("Startup size-based cleanup failed: {e}");
-            }
+            storage.startup_purge_with_retry();
         }
 
         Ok(storage)
+    }
+
+    /// Run the startup size-based purge, retrying on transient lock contention.
+    ///
+    /// Reopening a large database right after the previous process exits often
+    /// collides with that process's exit-time WAL checkpoint (see
+    /// [`Drop for Storage`]), which holds the write lock long enough to exceed
+    /// the connection `busy_timeout`; the purge then fails with `SQLITE_BUSY` /
+    /// `SQLITE_LOCKED`. This scenario is most likely on exactly the large
+    /// databases that need the startup purge, so retry with backoff instead of
+    /// giving up on the first collision.
+    ///
+    /// Only lock errors are retried — any other error (e.g. a corrupt
+    /// database) is terminal and must not stall startup for the full backoff.
+    /// When every attempt is locked out, the purge is skipped; the
+    /// insert-count-triggered purge in [`Self::store`] compensates once traffic
+    /// resumes.
+    fn startup_purge_with_retry(&self) {
+        purge_with_backoff(|| self.purge_oversized(), &STARTUP_PURGE_BACKOFF);
     }
 
     /// Create a new Storage with default SQLite config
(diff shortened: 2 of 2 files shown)


src/agentsight/src/storage/unified.rs
```diff
@@ -595,6 +638,34 @@ fn purge_share(rows: u64, pct: f64) -> usize {
...
+
+/// Whether a [`Storage::purge_oversized`] error is a transient SQLite lock
+/// (`SQLITE_BUSY` / `SQLITE_LOCKED`) worth retrying, as opposed to a terminal
+/// error (corrupt database, disk failure, …) that a retry cannot fix.
+///
+/// Store methods propagate `rusqlite::Error` unwrapped through `anyhow`, so the
+/// lock code is recoverable via `downcast_ref`.
+fn is_retryable_lock(err: &anyhow::Error) -> bool {
+    matches!(
+        err.downcast_ref::<rusqlite::Error>(),
```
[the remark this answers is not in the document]
Daydreamer-Li: Fixed in the latest revision (force-pushed).

Confirmed root cause: `TokenStore::delete_oldest_batch` wrapped the rusqlite error in a formatted `anyhow!("...: {e}")`, erasing the cause. `is_retryable_lock` classifies via `downcast_ref::<rusqlite::Error>()`, so a `SQLITE_BUSY`/`SQLITE_LOCKED` raised on the token table was seen as `None` → treated as terminal → the backoff never ran for that path.

Fix: propagate the error unwrapped via `?`, matching the sibling `audit`/`http`/`token_consumption` stores, so the SQLite code stays downcastable and the retry classifier recognizes it. Added a comment at the call site documenting why plain `?` (not a re-wrapped `anyhow!`) is required here.

Also extracted the retry loop into a testable `purge_with_backoff(purge, backoff)` (backoff injected so tests use zero delays) with unit tests covering the retry-then-succeed, locked-out, and terminal-error branches, which also resolves the incremental coverage gate failure.

src/agentsight/src/storage/unified.rs
```diff
@@ -183,14 +183,31 @@ impl Storage {
         // crash or a config change), prune immediately rather than waiting for
         // the next purge interval.
         if storage.max_db_size_bytes > 0 {
-            if let Err(e) = storage.purge_oversized() {
-                log::warn!("Startup size-based cleanup failed: {e}");
-            }
+            storage.startup_purge_with_retry();
```
jfeng18, line 186: Non-blocking: this call chain (`AgentSight::new` → `create_storage` → here, src/unified.rs:356) deepens a pre-existing tension with `src/UNIFIED_AGENTS.md` rule 3 — "`AgentSight::new()` does no I/O or blocking operations". The startup purge already blocked before this PR; the retry now adds up to ~13s of sleep on that path. Worth either carving out an explicit, budgeted exception for the startup purge in `UNIFIED_AGENTS.md`, or noting the deviation. No code change requested.

src/agentsight/src/storage/sqlite/token.rs
```diff
@@ -490,10 +490,11 @@ impl TokenStore {
...
         );
-        let deleted = self
-            .conn
-            .execute(&sql, params![limit as i64])
-            .map_err(|e| anyhow::anyhow!("Failed to delete oldest token records: {e}"))?;
+        // Propagate the rusqlite error unwrapped (like the sibling stores) so a
+        // transient SQLITE_BUSY/SQLITE_LOCKED stays downcastable and the startup
+        // purge's lock-retry classifier can recognize it; wrapping it in a
+        // formatted `anyhow!` here would erase the cause and make the retry a no-op.
+        let deleted = self.conn.execute(&sql, params![limit as i64])?;
```
jfeng18, line 497: Non-blocking precision note: the comment is right that a formatted `anyhow!("...: {e}")` erases the cause — but `.with_context("Failed to delete oldest token records")` would have kept both the context message and downcastability (anyhow's `context_downcast` forwards to the inner error; verified against anyhow 1.0.102). Bare `?` is fine; flagging the option in case the log context is missed.

jfeng18 (approved):
Approve — reviewed against the code, not the description:

- Mechanism confirmed: `busy_timeout` is 500ms (`sqlite/connection.rs:50`) and `Drop for Storage` runs a TRUNCATE checkpoint (`storage/unified.rs:601`), so a large DB reopened right after the previous exit can genuinely outlast the first purge attempt.
- Error-chain audit: all four stores' `delete_oldest_batch` (plus audit/http `count()`) propagate `rusqlite::Error` unwrapped, so every lock error that can escape `purge_oversized` stays downcastable for `is_retryable_lock`; `checkpoint_busy` failures are handled in-loop and never reach the classifier — consistent.
- Extended SQLite codes are masked to their primary code by libsqlite3-sys (`result_code & 0xff`), so e.g. `BUSY_SNAPSHOT` still classifies as retryable.
- Tests discriminate by construction (zero-delay injected backoff, retry counts asserted); CI green on `e7180cde`; the `Fixes: 33ed3846` trailer attributes the origin correctly.

Two non-blocking notes as inline comments; no changes requested for this PR.

Minor wording: the "~13s cap" counts only the sleeps — each failed attempt also burns its `busy_timeout` and any partial delete work before a mid-purge lock loss (each DELETE is its own autocommit transaction), so true worst-case added startup latency is somewhat higher. Consider rewording the PR body when convenient.

The lock-collision timing itself is unit-level only (as declared); acceptable given the extracted pure retry loop.

PR #3050, opened 2026-09-03
feat(cosh-ng): [shell] load login PATH

### LINKED ISSUE #3049: [cosh-ng] feat: discover PATH from Bash login profiles
## Component

cosh-ng

A CLI installer may persist a PATH entry only in `~/.bash_profile` and then tell the user to restart the terminal or run `exec $SHELL`. A normal interactive `cosh` session is non-login by default, so both its PATH bootstrap (`bash -ic`) and its managed Bash session load `~/.bashrc`, not `~/.bash_profile`.

This makes the newly installed command remain unavailable after restarting a normal terminal when the installer wrote only the Bash login profile. The observed Qoder CLI installation writes a dispatcher PATH block that prepends `$HOME/.qoder/entry`; when that block exists only in `~/.bash_profile`, `qoder` is not found in a default Cosh terminal.

Current behavior on `up/main` at `6712b3d6593e3e82afc5acacd529d93f7fbd3a4c`:

- Default Cosh launch sets `login_shell = false`.
- Enhanced native-mode PATH bootstrap invokes `bash -ic`.
- The managed Bash marker sources `~/.bashrc` for non-login sessions.
- `~/.bash_profile`, `~/.bash_login`, and `~/.profile` are supported only for `cosh -l`, `cosh --login`, or login-style `argv[0]`.
- `exec $SHELL` without `-l` does not load `~/.bash_profile`.

Minimal reproduction:

1. Put a unique PATH entry in `~/.bash_profile` only.
2. Ensure the entry is absent from `~/.bashrc` and the parent process PATH.
3. Start a normal interactive `cosh` session.
4. Observe that the PATH entry and command are unavailable.
5. Start `cosh -l` and observe that they become available.

After a supported CLI installer updates a standard Bash startup file, a fresh Cosh terminal should make the installed command available, or Cosh should provide a clear diagnostic explaining which startup file was skipped and how to enter a login session.

The resulting contract should preserve standard Bash behavior and avoid blindly sourcing both login and interactive startup files, because startup files may have side effects.

Acceptance coverage should include:

- Non-login Bash with PATH changes in `~/.bashrc`.
- Login Bash with the standard `~/.bash_profile` / `~/.bash_login` / `~/.profile` precedence.
- The installer compatibility case where PATH is written only to a login profile.
- `--isolated` continuing to skip user startup files.
- Guidance for `exec $SHELL` versus `exec "$SHELL" -l`.

## Possible approach

Evaluate whether the pre-launch PATH bootstrap can safely inspect a login-shell PATH separately and merge only PATH entries, without importing arbitrary shell state into a non-login session. If that would violate Bash semantics, add startup diagnostics and documentation that identify the active login/non-login mode and the startup files Cosh loads.

kongche-jbw:
## Why

CLI installers can persist PATH entries only in Bash login profiles while a normal Cosh terminal starts a non-login Bash session. A fresh terminal then misses the installed command even though its dispatcher was added to `~/.bash_profile`.

## What changed

Probe Bash login startup separately for default non-login sessions and import only login-specific PATH additions. Run that supplemental login probe on a bounded isolated PTY initialized from the caller terminal size so profile terminal guards match the managed shell. Launch the PTY probe through a dedicated supervisor process; on Linux only that supervisor becomes a child subreaper, so it terminates and reaps profile-started descendants without claiming unrelated Cosh children. Fail closed when equivalent containment is unavailable. Preserve the managed shell's non-login behavior, standard profile precedence, isolated-mode bypass, and existing Zsh behavior.

## Related issue

closes #3049

## User / Agent impact

Commands installed through Bash login profiles become available in a fresh default Cosh terminal. Login-profile PATH additions retain their placement relative to inherited command directories.

## Risk and compatibility

The additional interactive login Bash child runs on an isolated PTY and executes the user's login profile, which may have external side effects; only PATH is imported into Cosh. The dedicated Linux supervisor terminates only descendants of this disposable probe, while other profile side effects remain possible. Set `COSH_SHELL_BOOTSTRAP_PATH=0` to disable startup PATH probing.

## Validation

Local Linux validation on the current revision:

- `cargo test --locked --package cosh-shell --bin cosh-shell runtime::startup::tests:: -- --test-threads=4` (24 passed)
- `cargo test --locked --package cosh-shell --test shell_host termios::parent_lifecycle::startup_path_probe_ -- --test-threads=1` (4 passed)
- `cargo clippy --locked --package cosh-shell --all-targets -- -D warnings`

The fallback-anchor regression fails before the fix in both the merge tests and a real enhanced Cosh PTY session. Coverage checks prepend/append command collisions, synthetic versus inherited `/usr/local/bin` anchors, terminal guards and dimensions, detached-daemon cleanup, and timeout cleanup. Fallback directories are appended only after merging the actual inherited/non-login PATH with login additions. Broader regression coverage runs in CI.

## Documentation and rollback

No documentation files changed. Disable the behavior with `COSH_SHELL_BOOTSTRAP_PATH=0` or revert commit `0ad122d1`.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
@@ -1,10 +1,15 @@
 use std::collections::HashSet;
 use std::ffi::OsString;
-use std::io::{IsTerminal, Write};
-use std::os::unix::process::ExitStatusExt;
+use std::fmt;
+use std::fs::File;
+use std::io::{IsTerminal, Read, Write};
+use std::os::unix::process::{CommandExt, ExitStatusExt};
 use std::path::Path;
-use std::process::{Command, ExitStatus};
-use std::time::Duration;
+use std::process::{Command, ExitStatus, Stdio};
+use std::time::{Duration, Instant};
+
+use nix::pty::Winsize;
+use wait_timeout::ChildExt;
 
 use crate::diagnostics::health::{
     record_startup_health_recommendations, HealthFindingCategory, HealthScanReport, HealthSeverity,
@@ -49,7 +54,10 @@ const RESET: &str = "\x1b[0m";
 const LOGO_MIN_WIDTH: u16 = 42;
 const STARTUP_HEALTH_ROW_WAIT: Duration = Duration::from_millis(150);
 const STARTUP_AUTH_HINT_WAIT: Duration = Duration::from_millis(150);
+const BOOTSTRAP_PATH_PROBE_TIMEOUT: Duration = Duration::from_secs(5);
+const BOOTSTRAP_PATH_MAX_OUTPUT_BYTES: usize = 1024 * 1024;
 
+mod descendants;
 mod recommendations;
 #[cfg(test)]
 use recommendations::{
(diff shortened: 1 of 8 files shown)


src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
```diff
@@ -163,39 +164,158 @@ fn startup_hooks_enabled() -> bool {
...
+}
+
+const BASH_NON_LOGIN_PATH_PROBES: &[BootstrapPathProbe] = &[
+    BootstrapPathProbe {
+        flags: "-ic",
+        source: "Bash interactive startup",
+    },
+    BootstrapPathProbe {
+        flags: "-lc",
+        source: "Bash login startup",
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 07836f12. The startup probe now runs in its own process group with a 5-second deadline, bounded output capture, and process-group termination on timeout; a timed-out login probe leaves the non-login PATH result intact. Added bash_login_probe_timeout_preserves_non_login_path; local scoped tests passed (21 startup, 7 adapter, and 1 PTY), and the refreshed cosh-ng CI checks are green.

src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
```diff
@@ -163,39 +168,248 @@ fn startup_hooks_enabled() -> bool {
...
+    source: &'static str,
+}
+
+const BASH_NON_LOGIN_PATH_PROBES: &[BootstrapPathProbe] = &[
+    BootstrapPathProbe {
+        flags: "-ic",
+        source: "Bash interactive startup",
+    },
+    BootstrapPathProbe {
+        flags: "-lc",
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in a895f593. The separate Bash login probe now uses interactive-login semantics (-lic), while merge_bootstrap_paths still subtracts the non-login and inherited PATH entries before promoting login-only additions. The regression profile gates its PATH update on the i flag in $-, so it fails with -lc and passes through the production probe plan. All 21 startup tests, scoped Clippy, and the refreshed cosh-ng CI checks passed.

src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
```diff
@@ -163,39 +168,248 @@ fn startup_hooks_enabled() -> bool {
...
+fn bootstrap_path_command(shell: &str, flags: &str) -> Command {
+    let mut command = Command::new(shell);
+    command
         .arg(flags)
         .arg("printf '\\n__COSH_PATH_BEGIN__%s__COSH_PATH_END__\\n' \"$PATH\"")
         .env("COSH_SHELL_BOOTSTRAP_PATH", "0")
-        .output()
-    else {
-        return;
+        .stdin(Stdio::null())
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 673dd5ef. Only the supplemental Bash login probe now runs in its own controlling PTY, preserving the existing pipe behavior for the non-login Bash and Zsh probes; the existing 5-second deadline, bounded capture, and process-group cleanup remain in place. Added the shell-host PTY regression startup_path_probe_preserves_login_profile_tty_semantics with both stdin and stdout TTY guards. The 21 startup tests, exact PTY regression, layout audit, package all-targets Clippy, and refreshed cosh-ng CI all passed.

src/cosh-ng/crates/cosh-shell/src/shell_host/bootstrap.rs
```diff
@@ -191,6 +189,42 @@ fn start_shell_session(
     })
 }
 
+pub(crate) fn spawn_profile_probe_on_pty(mut command: Command) -> io::Result<(Child, File)> {
+    let (master, slave) = open_pty_pair(None)?;
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 76163d64. The runtime now passes the already captured ShellHostConfig terminal size through the PATH probe into openpty, so the supplemental login shell receives the same nonzero rows and columns as the managed shell. Extended startup_path_probe_preserves_login_profile_terminal_semantics to require both TTY file descriptors and an exact `stty size` value of `37 113`. The 21 startup tests, exact PTY regression, package all-targets Clippy, layout audit, and all required CI checks passed.

src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
```diff
@@ -163,39 +170,311 @@ fn startup_hooks_enabled() -> bool {
...
+}
+
+const BASH_NON_LOGIN_PATH_PROBES: &[BootstrapPathProbe] = &[
+    BootstrapPathProbe {
+        flags: "-ic",
+        source: "Bash interactive startup",
+        io: BootstrapPathProbeIo::Pipes,
+    },
+    BootstrapPathProbe {
+        flags: "-lic",
```
ikunkun-sys, line 194: **[P1] Avoid orphaning profile-started daemons**

Adding a full `-lic` probe to every non-login Bash launch executes `.bash_profile`, but the caller imports only PATH. A common `eval "$(ssh-agent -s)"` stanza therefore starts an agent and sets `SSH_AGENT_PID`/`SSH_AUTH_SOCK` only inside the throwaway probe; the managed shell receives neither variable, while the daemon survives with PPID 1, so each Cosh launch leaks another unusable agent. I reproduced this at `76163d64` with a temporary HOME and cleared SSH variables: the managed shell reported both as unset while `/usr/bin/ssh-agent -s` remained alive after Cosh exited. The timeout/process-group cleanup does not help because ssh-agent daemonizes successfully before the probe returns.

ikunkun-sys (changes requested):
Requesting changes for one P1 correctness and resource-leak issue found on the current head. The supplemental login-shell probe executes the full Bash profile in a disposable environment and imports only PATH, so profile-started daemons can survive without the variables needed to use or manage them.

src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
kongche-jbw: Fixed in 2e4fa0f9. The supplemental PTY login probe now enters a Linux child-subreaper scope before spawning Bash, records pre-existing children, and recursively terminates and reaps newly adopted descendants after the probe exits, including processes that daemonize out of the original process group. On platforms where equivalent containment is unavailable, the supplemental probe fails closed. Added startup_path_probe_reaps_detached_profile_daemon, which starts a setsid --fork daemon, verifies PATH is still imported, and asserts the daemon is gone. The 21 startup tests (including default concurrency), both exact PTY regressions, all-targets Clippy, layout audit, and all required CI checks passed.

src/cosh-ng/crates/cosh-shell/src/runtime/startup/descendants.rs
```diff
@@ -0,0 +1,199 @@
...
+            self.cleanup()?;
+            self.restore()?;
+            self.armed = false;
+            Ok(())
+        }
+
+        fn cleanup(&self) -> io::Result<()> {
+            let deadline = Instant::now() + CLEANUP_TIMEOUT;
+            loop {
+                let owned = direct_children()?
```
ikunkun-sys: **[P1] Scope cleanup to probe-owned descendants**

`direct_children().difference(&self.baseline)` treats every process-wide child created after `enter()` as belonging to the login probe, although `SUBREAPER_LOCK` only serializes probes and does not block other threads from spawning children. The following `kill_child`/`reap_child` loop can therefore SIGKILL or consume an unrelated child before its owner waits. On `2e4fa0f9`, `cargo test --quiet --package cosh-shell --bin cosh-shell -- --test-threads=16` failed 23 tests with exit 137 or `ECHILD`; the identical command on `76163d64` passed 2622 tests with 1 ignored, and the new head passes when forced to one test thread. The green CI uses `run_canonical_units cosh-shell cosh-shell 1`, so it does not exercise this concurrency. Please make descendant ownership structural rather than inferring it from a process-wide before/after set.

ikunkun-sys (changes requested):
The previous ssh-agent leak is fixed for both successful and timed-out profile probes. Requesting changes for one new P1 child-ownership bug in the containment implementation: the process-wide baseline can classify, kill, and reap unrelated children created concurrently by other threads. The required CI is green, but its canonical cosh-shell binary suite is explicitly pinned to one test thread and does not cover this interaction.

src/cosh-ng/crates/cosh-shell/src/runtime/startup/descendants.rs
kongche-jbw: Fixed in bfc5ee02. The main cosh-shell process no longer enables PR_SET_CHILD_SUBREAPER or infers ownership from a process-wide before/after child set. It now launches the supplemental PTY probe through a dedicated internal supervisor process; only that supervisor becomes a subreaper, and it refuses to start containment if it already owns a child, so every adopted process is structurally a descendant of the profile probe. The daemon regression now asserts that the profile Bash parent is distinct from the Cosh wrapper, and a new timeout regression verifies that both the supervisor and a detached daemon are gone. The reported 16-thread command passed on the final tree (2622 passed, 1 ignored), along with 21 startup tests, both daemon paths, PTY semantics, all-targets Clippy, layout audit, and all required CI checks.

src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
```diff
@@ -163,39 +171,371 @@ fn startup_hooks_enabled() -> bool {
...
+                .filter(|entry| !entry.is_empty() && !existing.contains(entry))
+                .collect::<Vec<_>>()
+                .join(":")
+        })
+    } else {
+        None
+    };
+
+    let mut paths = Vec::with_capacity(discovered.len() + 3);
+    if let Some(path) = login_only.as_deref().filter(|path| !path.is_empty()) {
```
ikunkun-sys: **[P1] Preserve appended login-profile PATH precedence**

`login_only` is collected without its position anchors and then pushed before both the non-login and inherited PATHs, so an entry intentionally appended by `.bash_profile` becomes high priority in Cosh. With inherited `PATH=/usr/bin:/bin` and `export PATH="$PATH:$HOME/late-bin"`, standard login Bash resolves `ls` as `/usr/bin/ls`, while the current Cosh launch resolves it as `$HOME/late-bin/ls`. This can execute a different binary than the user's login shell. The merge needs to preserve whether each login-only addition appeared before or after the inherited entries; regression coverage should include both prepend and append collisions.

ikunkun-sys (changes requested):
One blocking correctness issue remains on the current head. The dedicated supervisor fixes the earlier process-wide child-ownership problem, and the current CI plus targeted concurrency/PTY validation are green, but the login-profile PATH merge can still invert the user's intended command precedence.

src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
kongche-jbw: Fixed in c4d33148. The Bash non-login merge now builds the non-login/base PATH first and inserts only login-specific additions at their neighboring base anchors, so prepended directories stay ahead of inherited commands while appended directories remain behind them; login-profile removal or reordering of existing directories is still not imported. Added executable-collision coverage for both prepend and append cases plus an internal anchor-order regression. Local startup tests (23), the 16-thread cosh-shell suite (2624 passed, 1 ignored), Clippy, format, and layout checks passed. Required CI is green after rerunning two unrelated relay timing flakes; both exact tests also passed 3/3 local reruns.

src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
```diff
@@ -163,39 +171,398 @@ fn startup_hooks_enabled() -> bool {
...
+    }
+}
+
+fn merge_path_additions_at_anchors(overlay: &str, base: &str) -> String {
+    let mut merged = base
+        .split(':')
+        .filter(|entry| !entry.is_empty())
+        .map(str::to_string)
+        .collect::<Vec<_>>();
+    let anchors = merged.iter().cloned().collect::<HashSet<_>>();
```
ikunkun-sys, line 506: **[P1] Exclude synthetic fallback directories from PATH anchors**

`base` already contains the fallback directories appended at lines 483–484, so this set treats `/usr/local/bin` as an inherited anchor even when the parent and non-login PATH contain only `/usr/bin:/bin`. On a host where `/etc/profile` prepends `/usr/local/bin`, a user profile containing `export PATH="$HOME/login-only:$PATH"` produces an overlay with `login-only` before that directory. The merge inserts `login-only` at the fallback `/usr/local/bin` position, which is after `/usr/bin`, reversing the user's explicit prepend. In a temporary HOME on this commit, login Bash resolved `ls` to `$HOME/login-only/ls`, while a real enhanced Cosh PTY session resolved `/usr/bin/ls`; the resulting Cosh PATH placed `login-only` after `/usr/bin:/bin:/opt/homebrew/bin`. The existing `runtime::startup::tests::bash_non_login_probe_reads_bashrc_then_interactive_login_path` also fails at startup_tests.rs:646, including in an isolated single-thread run. This can select the wrong installed command despite the simple prepend/append unit tests passing.

ikunkun-sys (changes requested):
Requesting changes for one P1 PATH-ordering issue on c4d33148: synthetic fallback directories are treated as inherited PATH anchors, so a login-profile prepend can still end up behind system commands.

Validation from the completed local review of this same commit:
- Startup suite: 22 passed, 1 failed. The failing bash_non_login_probe_reads_bashrc_then_interactive_login_path test also failed when run alone with --exact --test-threads=1.
- Binary suite with --test-threads=16: 2623 passed, 1 failed, 1 ignored; the same PATH test failed.
- Real Cosh PTY reproduction confirmed different command resolution from login Bash.
- All three supervisor PTY tests passed.

The PR head and threads were refreshed before submission; the head is unchanged and current GitHub CI checks are green. The local failure depends on the host's login-profile PATH layout.

src/cosh-ng/crates/cosh-shell/src/runtime/startup.rs
kongche-jbw: Fixed in 0ad122d1818d1955c728a65fc8dbfb6ad39b3441. Synthetic fallback directories are now appended only after login additions are merged against the actual inherited/non-login PATH, so a login-profile prepend stays ahead of system commands. Added deterministic prepend/append collision and fallback-versus-inherited anchor coverage, plus a real enhanced Cosh PTY regression that executes the login directory's `ls`. Both merge regressions and the PTY regression failed before the fix. On the final revision, all 24 startup tests, all 4 PATH-probe PTY tests, scoped all-targets Clippy, format, and layout checks passed; refreshed CI (main cosh-ng tests, fast checks, release build, and all other applicable checks) is green without retries.

ikunkun-sys (approved):
Re-reviewed 0ad122d1818d1955c728a65fc8dbfb6ad39b3441 against the previous revision. No remaining actionable findings.

The previous PATH-precedence blocker is fixed: login-only additions are merged against actual inherited/non-login PATH entries before synthetic fallback directories are appended. The previously failing real Bash profile test and the real enhanced Cosh PTY command-collision regression now pass.

Independent local validation on this exact head (Linux):
- 24 runtime startup tests passed.
- All 4 PATH-probe PTY tests passed, covering terminal semantics, command precedence, detached-daemon cleanup, and timeout cleanup.
- cargo fmt --all -- --check passed.
- cargo clippy --locked --package cosh-shell --all-targets -- -D warnings passed.
- cosh-shell layout audit passed.

Validation was scoped; the full workspace suite and non-Linux runtime behavior were not tested locally. Immediately before submission, head/base were unchanged, all 12 applicable remote checks were successful, and all 8 historical review threads were resolved.

PR #3058, opened 2026-09-04
fix(cosh-ng): [shell] preserve zsh slash commands

kongche-jbw:
## Why

The Rust path-prompt classifier only checks filesystem availability. A Zsh function or alias may resolve the same slash-bearing first token, so it must remain Shell-owned.

## What changed

At each trusted Zsh prompt, Cosh receives a bounded snapshot of slash-bearing functions and aliases. A path prompt routes to Agent only when that complete snapshot excludes its first token.

## Related issue

#1943 · #3004

## User / Agent impact

Zsh function and alias commands with slash-bearing names execute natively even when their arguments contain natural language. Missing path prompts retain the existing Agent route.

## Risk and compatibility

The Zsh marker carries a bounded namespace snapshot only at prompt boundaries. An unavailable or oversized snapshot keeps the input Shell-owned. Any registered `zle-line-init` widget, including same-name registration and distribution defaults, makes the snapshot unavailable because it may change the namespace after the prompt marker.

## Validation

- `cargo test --package cosh-shell --lib path_prompt -- --test-threads=1` (7 passed)
- `cargo test --package cosh-shell --test raw_cli path_prompt -- --test-threads=1` (7 passed)
- `cargo test --package cosh-shell --test raw_cli agent_input::raw_cli_zsh_keeps_slash_named_function_and_alias_shell_owned -- --exact` (1 passed)
- `git diff --check`

No ECS, real-provider, manual-terminal UX, release build, or full-workspace gate was run; broader coverage is delegated to CI.

Current-head review validation (Linux, Zsh 5.9):

- The new self-bound line-init regression fails on the previous implementation and passes after the fix; both same-name and differently named hooks execute the slash function natively.
- `cargo test --locked -p cosh-shell --test raw_cli agent_input::raw_cli_zsh_ -- --test-threads=1`: 10 passed; the unrelated `raw_cli_zsh_agent_prompt_restore_suppresses_partial_line_marker` fails locally and also fails with the original `6dd2356b` marker implementation.
- `cargo test --locked -p cosh-shell --test raw_cli agent_input::raw_cli_zsh_keeps_slash -- --test-threads=1`: all 3 final regression tests passed; the alias fixture clears distribution hooks so it exercises the complete namespace snapshot.
- `cargo clippy --locked -p cosh-shell --all-targets -- -D warnings`, formatting, layout audit, and diff hygiene passed.

## Documentation and rollback

No documentation changes are required. Revert `bde6c92c7e3c57e63d84d5ec437e4806b46e5780` to restore the prior Zsh routing behavior.

### DIFF
--- a/src/cosh-ng/crates/cosh-shell/src/input/path_prompt.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/input/path_prompt.rs
@@ -1,6 +1,7 @@
 //! Conservative classification for natural-language submissions whose first
 //! shell token contains a path separator.
 
+use std::collections::HashSet;
 use std::ffi::CString;
 use std::os::unix::ffi::OsStrExt;
 use std::path::{Component, Path, PathBuf};
@@ -40,6 +41,63 @@ impl PartialEq for ShellPromptCwd {
 }
 
 impl Eq for ShellPromptCwd {}
+
+/// Slash-bearing command names and suffix aliases Zsh resolves without a
+/// filesystem path.
+///
+/// Prompt markers refresh this snapshot after the shell has completed its
+/// own parsing and startup hooks. An unavailable snapshot is intentionally
+/// not treated as proof that a name is missing.
+#[derive(Debug, Default)]
+struct ShellPathNamespace {
+    names: HashSet<String>,
+    suffixes: HashSet<String>,
+}
+
+#[derive(Debug, Clone, Default)]
+pub(crate) struct ShellPathCommandNames(Arc<Mutex<Option<ShellPathNamespace>>>);
+
+impl ShellPathCommandNames {
+    pub(crate) fn set(&self, names: Option<Vec<String>>, suffixes: Option<Vec<String>>) {
+        if let Ok(mut current) = self.0.lock() {
+            *current = names
+                .zip(suffixes)
+                .map(|(names, suffixes)| ShellPathNamespace {
+                    names: HashSet::from_iter(names),
+                    suffixes: HashSet::from_iter(suffixes),
+                });
+        }
+    }
+
+    pub(crate) fn excludes_first_token(&self, input: &str) -> bool {
+        let first_token = input
+            .trim_matches([' ', '\t'])
+            .split_ascii_whitespace()
+            .next()
+            .unwrap_or_default();
+        self.0
+            .lock()
+            .ok()
+            .and_then(|namespace| {
+                namespace.as_ref().map(|namespace| {
+                    !namespace.names.contains(first_token)
+                        && Path::new(first_token)
+                            .extension()
+                            .and_then(|suffix| suffix.to_str())
+                            .is_none_or(|suffix| !namespace.suffixes.contains(suffix))
+                })
+            })
+            .unwrap_or(false)
+    }
+}
+
+impl PartialEq for ShellPathCommandNames {
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
+    }
+}
+
+impl Eq for ShellPathCommandNames {}
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) struct PathPromptIntercept {
--- a/src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
+++ b/src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
@@ -132,6 +132,27 @@ _cosh_json_escape() {
 _cosh_now_ms() {
   date +%s000
 }
+_cosh_shell_path_names_fragment() { # Line-init can mutate the namespace after this snapshot.
+  local -a names suffixes quoted_names quoted_suffixes; local name total=0
+  if [[ -n "${widgets[zle-line-init]:-}" ]]; then REPLY=',"shell_path_names":null,"shell_path_suffixes":null'; return; fi
+  for name in ${(k)functions} ${(k)aliases} ${(k)galiases}; do
+    [[ "$name" == */* ]] && names+=("$name")
+  done
+  suffixes=("${(k)saliases}"); if (( ${#names} > 128 || ${#suffixes} > 128 )); then REPLY=',"shell_path_names":null,"shell_path_suffixes":null'; return; fi
+  for name in "${names[@]}"; do
+    if [[ "$name" == *[[:cntrl:]]* || "$name" == *'"'* || "$name" == *\\* ]] || (( (total += ${#name} + 3) > 8192 )); then
+      REPLY=',"shell_path_names":null,"shell_path_suffixes":null'; return
+    fi
+    quoted_names+=("\"$name\"")
+  done
+  for name in "${suffixes[@]}"; do
+    if [[ "$name" == *[[:cntrl:]]* || "$name" == *'"'* || "$name" == *\\* ]] || (( (total += ${#name} + 3) > 8192 )); then
+      REPLY=',"shell_path_names":null,"shell_path_suffixes":null'; return
+    fi
+    quoted_suffixes+=("\"$name\"")
+  done
+  REPLY=",\"shell_path_names\":[${(j:,:)quoted_names}],\"shell_path_suffixes\":[${(j:,:)quoted_suffixes}]"
+}
 _cosh_emit_marker() {
   local event="$1"
   local command="$2"
@@ -143,6 +164,8 @@ _cosh_emit_marker() {
   # lines carry a token, every other marker stays byte-identical.
   local handoff_fragment=""
   local physical_cwd_fragment=""
+  local shell_path_names_fragment=""
+  local REPLY=""
   if [[ -n "${_COSH_HANDOFF_TOKEN:-}" ]]; then
     handoff_fragment=",\"handoff\":\"$(_cosh_json_escape "$_COSH_HANDOFF_TOKEN")\""
   fi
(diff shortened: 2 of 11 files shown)


src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
```diff
@@ -158,8 +181,9 @@ _cosh_emit_marker() {
     if [[ "$physical_cwd" == /* ]]; then
       physical_cwd_fragment=",\"physical_cwd\":\"$(_cosh_json_escape "$physical_cwd")\""
     fi
+    shell_path_names_fragment="$(_cosh_shell_path_names_fragment)"
```
ikunkun-sys: **[P2] Refresh the proof after ZLE line initialization**

This snapshot is emitted from `_cosh_precmd_marker`, before Zsh invokes `zle-line-init`, so it can already be stale when the user starts typing. On this head, register `function zle-line-init { function 路径/late { print -r -- late-path-command:$*; }; }; zle -N zle-line-init`, then submit `路径/late 帮我运行一下`: Cosh routes it to the Agent, while `whence -w 路径/late` reports `function`. The snapshot therefore does not prove at submission time that Zsh cannot resolve the token.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
```diff
@@ -132,6 +132,28 @@ _cosh_json_escape() {
 _cosh_now_ms() {
   date +%s000
 }
+_cosh_shell_path_names_fragment() {
+  local -a names escaped_names; local name total=0
+  for name in ${(k)functions}; do
+    [[ "$name" == */* ]] && names+=("$name")
+  done
+  for name in ${(k)aliases}; do
+    [[ "$name" == */* ]] && names+=("$name")
```
ikunkun-sys: **[P2] Account for suffix aliases before proving absence**

Zsh stores `alias -s txt=...` in `saliases`, and that alias applies to slash-bearing words even though the alias key itself contains no slash. With `alias -s txt='print -r -- suffix-path-command:'`, native Zsh executes `/missing/path.txt 帮我运行一下` through the suffix alias; the same submission through this head's real raw Zsh path is intercepted by the Agent. Enumerating slash-bearing keys from `functions` and `aliases` therefore cannot establish that the token is unresolved.

src/cosh-ng/crates/cosh-shell/src/input/path_prompt.rs
```diff
@@ -41,6 +42,49 @@ impl PartialEq for ShellPromptCwd {
...
+/// not treated as proof that a name is missing.
+#[derive(Debug, Clone, Default)]
+pub(crate) struct ShellPathCommandNames(Arc<Mutex<Option<HashSet<String>>>>);
+
+impl ShellPathCommandNames {
+    pub(crate) fn set(&self, names: Option<Vec<String>>) {
+        let names = names.map(|names| {
+            names
+                .into_iter()
+                .filter(|name| name.len() <= 1024 && !name.chars().any(char::is_control))
```
ikunkun-sys: **[P2] Do not silently drop names from a complete snapshot**

The producer bounds `${#name}` by Zsh characters, while `String::len()` here counts UTF-8 bytes; filtering a rejected entry still leaves the outer snapshot as `Some`, so the consumer treats an incomplete set as proof of absence. A function name made from five 80-Han-character path components is 404 characters but 1204 bytes: Zsh sends it, this filter drops it, and the real raw Zsh path routes its invocation to the Agent even though `whence` reports the name as a function. This violates the snapshot's fail-closed completeness invariant.

ikunkun-sys (changes requested):
Requesting changes for three additional current-head correctness gaps in the Zsh namespace proof. Each was reproduced through the real `cosh-shell raw fake --shell zsh` path at commit `2cf1693b14cebb8b8ca8a720f10ec0455edbf1a7`. The two existing unresolved findings about global aliases and prompt-time command substitutions also remain applicable. `Test cosh-ng fast checks` is currently red because `zsh_candidate_submit_precedes_later_tab_in_same_read` fails.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
```diff
@@ -132,6 +132,28 @@ _cosh_json_escape() {
...
+  if (( ${#names} > 128 )); then
+    printf '%s' ',"shell_path_names":null'
+    return
+  fi
+  for name in "${names[@]}"; do
+    if (( ${#name} > 1024 || (total += ${#name} + 3) > 8192 )); then
+      printf '%s' ',"shell_path_names":null'
+      return
+    fi
+    escaped_names+=("\"$(_cosh_json_escape "$name")\"")
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 6dd2356b. Prompt serialization now builds the namespace JSON in-process, avoiding a command substitution per name. Targeted raw-CLI coverage, Test cosh-ng fast checks, and Test cosh-ng passed.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
```diff
@@ -132,6 +132,28 @@ _cosh_json_escape() {
...
   date +%s000
 }
+_cosh_shell_path_names_fragment() {
+  local -a names escaped_names; local name total=0
+  for name in ${(k)functions}; do
+    [[ "$name" == */* ]] && names+=("$name")
+  done
+  for name in ${(k)aliases}; do
+    [[ "$name" == */* ]] && names+=("$name")
+  done
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 6dd2356b. The prompt snapshot now includes galiases, with a raw-CLI regression confirming a slash-named global alias stays Shell-owned. Test cosh-ng fast checks and Test cosh-ng passed.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
kongche-jbw: Fixed in 6dd2356b. The marker carries saliases separately and the router treats a matching path suffix as Shell-owned. A raw-CLI suffix-alias regression and CI passed.

src/cosh-ng/crates/cosh-shell/src/input/path_prompt.rs
kongche-jbw: Fixed in 6dd2356b. Rust now retains names accepted by the Zsh producer, and raw-CLI coverage invokes a slash name exceeding 1024 UTF-8 bytes. Test cosh-ng fast checks and Test cosh-ng passed.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
kongche-jbw: Fixed in 6dd2356b. A re-bound zle-line-init widget now makes the namespace proof unavailable, so the candidate stays Shell-owned; no ZLE wrapper is installed. A raw-CLI regression covers the mutation. Test cosh-ng fast checks and Test cosh-ng passed.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
```diff
@@ -132,6 +132,27 @@ _cosh_json_escape() {
 _cosh_now_ms() {
   date +%s000
 }
+_cosh_shell_path_names_fragment() {
+  local -a names suffixes quoted_names quoted_suffixes; local name total=0
+  if [[ -n "${widgets[zle-line-init]:-}" && "${widgets[zle-line-init]}" != "user:zle-line-init" ]]; then REPLY=',"shell_path_names":null,"shell_path_suffixes":null'; return; fi
```
ikunkun-sys: **[P2] Treat a self-bound zle-line-init as mutable**

The guard exempts `user:zle-line-init`, but that is exactly the descriptor produced by the standard same-name registration `zle-line-init() { ... }; zle -N zle-line-init`. On this head, defining `路径/late` from that hook and submitting `路径/late 帮我运行一下` through `cosh-shell raw fake --shell zsh` still produces an AI request instead of invoking the function. The added regression binds `zle-line-init` to a differently named `_cosh_test_line_init`, whose descriptor is `user:_cosh_test_line_init`, so it never exercises this exemption. Treat the self-bound descriptor as making the snapshot unavailable as well.

ikunkun-sys (changes requested):
Requesting changes for one remaining current-head correctness gap at 6dd2356bcebea296438245ec6970a8cb0906c828. The global and suffix alias handling, multibyte-name retention, prompt-hot-path serialization, and Enter-to-Tab regression are fixed and independently revalidated; GitHub CI is green. The original same-name zle-line-init reproduction still fails because the new guard explicitly exempts user:zle-line-init.

src/cosh-ng/crates/cosh-shell/src/shell_host/marker/zsh.rs
kongche-jbw: Fixed in bde6c92c. Removed the user:zle-line-init exemption: any registered line-init widget now makes the namespace snapshot unavailable. Added a same-name regression alongside the existing differently named hook case; it fails on the previous implementation and passes with the fix. The no-hook and alias fixtures explicitly clear distribution hooks to keep complete-snapshot routing covered. Final-head fast checks, full cosh-ng tests, release build, and prebuilt action all passed.

ikunkun-sys (approved):
Re-reviewed the current head bde6c92c7e3c57e63d84d5ec437e4806b46e5780. The same-name zle-line-init gap is fixed: both the targeted raw-CLI regression and the original real-Zsh PTY reproduction now keep the slash function Shell-owned. The complete namespace snapshot, missing-path routing, Enter-to-Tab regression, formatting, layout, diff hygiene, and GitHub CI also pass. No actionable findings remain.

PR #3091, opened 2026-09-07
feat(chore): honor explicit rustup servers

### LINKED ISSUE #3090: build-all.sh should respect explicit rustup server configuration
### Component

project-wide (`scripts/build-all.sh`)

### What happened?

While checking the build script, I noticed that an explicitly configured `RUSTUP_DIST_SERVER` is not consistently respected. For example, a user might deliberately select the official Rust servers for a machine outside China:

```bash
export RUSTUP_DIST_SERVER=https://static.rust-lang.org
export RUSTUP_UPDATE_ROOT=https://static.rust-lang.org/rustup
```

The script can still replace these settings:

- `_configure_cargo_mirror()` selects another mirror if the configured server fails its pinned-toolchain probe. A temporary timeout can therefore change the user's chosen source.
- When `install_rust()` reaches `rustup update stable`, it independently selects a mirror and overrides both variables for that command, even if the explicitly configured server passed the earlier probe.

Both automatic selection lists put China-based mirrors before the official server. This makes it difficult to reliably choose a source suited to the user's network, including networks outside China or environments with an approved internal mirror.

### How can we reproduce it?

Code inspected at commit `8bf150c24482de42cf5f4be580b56d8c6bf3a376`. Isolated checks with mocked successful/failed probes show:

1. With an explicit official source and successful probes, the pinned-toolchain configuration keeps the official source, but the stable selector returns the Tsinghua mirror. The stable update call then uses that selection.
2. If the explicitly configured source fails its probe and another candidate passes, the configuration function replaces both variables with the selected mirror.

This report is based on code inspection and isolated selection checks, rather than a full build failure log from a US machine.

Relevant code:

- [Configured-source fallback](https://github.com/alibaba/anolisa/blob/8bf150c24482de42cf5f4be580b56d8c6bf3a376/scripts/build-all.sh#L1035-L1046)
- [Stable update overrides](https://github.com/alibaba/anolisa/blob/8bf150c24482de42cf5f4be580b56d8c6bf3a376/scripts/build-all.sh#L798-L814)

### Expected behavior

Please treat explicitly set `RUSTUP_DIST_SERVER` and `RUSTUP_UPDATE_ROOT` as user configuration and preserve them throughout installation, updates, and builds. If that source is unavailable, report the failure clearly rather than automatically replacing it.

Automatic mirror selection would still be useful when no source has been configured. Explicit configuration should provide a reliable way to opt out of that selection.

kongche-jbw:
## Why

`build-all.sh` can replace an explicitly configured Rustup server after a failed probe or when updating stable. Users need a reliable way to select endpoints appropriate for their network.

## What changed

Preserve `RUSTUP_DIST_SERVER` and `RUSTUP_UPDATE_ROOT` whenever either has a nonempty user-provided value, letting Rustup report download failures from the configured endpoints. Capture that choice before automatic configuration so stable updates inherit it; users without overrides retain the existing pinned/stable mirror selection.

## Related issue

Closes #3090

## User / Agent impact

Explicit endpoint settings now survive dependency setup, stable updates, and repeated build-time configuration. Setting only one variable preserves that value and leaves the other unset for Rustup's default behavior.

## Risk and compatibility

An unavailable explicitly configured endpoint is no longer replaced automatically. Unset the overrides to restore automatic selection. Automatic mirror order and Cargo registry configuration are unchanged.

## Validation

- macOS arm64, Bash 5.3.15: `bash tests/test-build-all-runtime-deps.sh` — all 36 tests passed.
- New regression coverage checks both variables together, either variable alone, repeated configuration with failed mocked probes, and stable-update inheritance. It also verifies separate pinned/stable mirror selection with no overrides.
- The explicit-configuration regression failed against the original implementation before the fix.
- `bash -n scripts/build-all.sh tests/test-build-all-runtime-deps.sh` and `git diff --check` passed.
- The local copilot-shell pre-commit hook could not start because `lint-staged` was missing. Its configured JS/TS/JSON/Markdown patterns do not include these Shell files, so it was skipped for this commit; the Shell checks above passed.
- Network and package operations were stubbed; no full Rust build or US-network download was performed.

## Documentation and rollback

No documentation files changed. Inline comments explain why the override is captured before automatic configuration. Revert this commit to restore the previous behavior.

### DIFF
--- a/scripts/build-all.sh
+++ b/scripts/build-all.sh
@@ -788,6 +788,11 @@ install_rust() {
     }
 
     _source_cargo
+    # Capture overrides before automatic mirror selection exports its defaults.
+    local rustup_server_override=false
+    if [[ -n "${RUSTUP_DIST_SERVER:-}" || -n "${RUSTUP_UPDATE_ROOT:-}" ]]; then
+        rustup_server_override=true
+    fi
     _configure_cargo_mirror
 
     if _rust_ver_ok; then
@@ -800,8 +805,10 @@ install_rust() {
     # RUSTUP_DIST_SERVER remains selected for sec-core's pinned Rust toolchain.
     if cmd_exists rustup; then
         info "Updating via rustup ..."
-        local stable_picked stable_dist stable_update_root
-        stable_picked=$(_pick_rustup_stable_mirror 2>/dev/null || echo "")
+        local stable_picked="" stable_dist stable_update_root
+        if ! $rustup_server_override; then
+            stable_picked=$(_pick_rustup_stable_mirror 2>/dev/null || echo "")
+        fi
         if [[ -n "$stable_picked" ]]; then
             stable_dist="${stable_picked%%|*}"
             stable_update_root="${stable_picked##*|}"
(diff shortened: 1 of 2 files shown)


ikunkun-sys (approved):
Reviewed commit 88528acfb47e643d389de046cf1918cc2464a2c0. No actionable findings. Explicit Rustup endpoint settings are preserved through dependency setup, stable updates, and repeated build configuration; automatic pinned/stable mirror selection remains intact when neither endpoint is configured.

Validation on Linux x86_64 with Bash 4.4: all 36 runtime dependency tests passed, along with 10 additional cases exercising external child-process environment inheritance and repeated do_build calls across both/individual/unset/empty endpoint settings and old/current toolchains. Shell syntax and diff checks passed. The separate endpoint defaults were checked against the Rustup documentation. Network and installation operations were stubbed; no real download or full build was performed locally.

Before submission, head/base were unchanged, CI reported 11 successful and 15 skipped checks with no failures or pending checks, and there were no review threads.

PR #3142, opened 2026-09-08
feat(anolisa): honor editable config files

### LINKED ISSUE #3137: [TestAgent][anolisa] bug: type="config" 文件编辑仍触发 sha256_mismatch 组件 failed（#2269 未修复仍复现）
## Component

anolisa

## What happened?

layout 声明 `type = "config"` 的文件仍被按不可变文件校验：用户编辑配置文件后 `anolisa doctor` 报 `sha256_mismatch`，组件进入 `failed`：

```
$ anolisa install cfg-comp --backend raw
installed cfg-comp 1.0.0

$ echo "user edit" >> .../components/cfg-comp/settings.toml   # type="config" 文件

$ anolisa doctor cfg-comp
Doctor: 1 checked, 0 ok, 0 degraded, 1 failed

cfg-comp failed (1.0.0) scope=system
  error [health_sha256_mismatch] health check 'integrity:.../components/cfg-comp/settings.toml' reported 'sha256_mismatch'
```

对照组（同组件的普通 data 文件被编辑）同样报 mismatch——**config 类型声明对完整性校验行为没有任何影响**，`FileKind::Config`（manifest.rs，"Configuration file (purge-scoped on uninstall)"）未参与 integrity 语义。

原始 issue #2269 于 2026-08-10 关闭，未关联任何修复 PR（merged），anolisa 0.3.10 上仍 100% 复现。

## How can we reproduce it?

环境：Alinux 4 x86_64，anolisa 0.3.10（file:// raw fixture 仓库）。

1. 组件契约声明 config 文件 + 普通 data 文件：

```toml
[[component.layout.files]]
source = "share/settings.toml"
target = "{datadir}/components/cfg-comp/settings.toml"
mode = "0644"
type = "config"

[[component.layout.files]]
source = "data/plain.dat"
target = "{datadir}/components/cfg-comp/plain.dat"
mode = "0644"
```

2. `anolisa install cfg-comp --backend raw` → exit 0。
3. 对 settings.toml 追加一行（合法用户配置编辑）。
4. `anolisa doctor cfg-comp` → settings.toml 报 `sha256_mismatch` → 组件 `failed`。

复现稳定性：3/3。

## Expected behavior

- `type = "config"` 文件的编辑不应触发组件 `failed`（config 语义上就是给用户改的）。
- 普通不可变文件的篡改仍应被检出（对照组验证未过度放行——当前对照组正常报错，说明只是 config 分支缺失，不是整体失效）。

## Diagnostic output

anolisa 0.3.10-1.alnx4.x86_64（ECS `<ECS-IP>`，Alinux 4）。

kongche-jbw:
## Why

Editing a raw-installed `type = "config"` file makes `doctor` report
`sha256_mismatch` because installation drops the manifest file kind.

## What changed

Carry each file's kind and mode from its actual source mapping through archive
expansion, placement, and state recording. Remove destination-based inference,
which cannot distinguish separate source directories sharing or overlapping a
destination. Config files skip content hashing only after the shared integrity
probe's path, type, permission, and capability checks.

Legacy hydration recovers exact config declarations and directory declarations
only when every matching directory agrees on config semantics. Ambiguous legacy
files retain digest checking; conflicting directory modes are not guessed.
The rebase preserves main's post-install-hook digest refresh.

## Related issue

Closes #3137

## User / Agent impact

New raw installs retain accurate file kinds regardless of declaration order.
Config edits remain healthy; ordinary data and executable tampering is detected.
Legacy exact and unambiguous config declarations work without reinstalling.
For ambiguous legacy directory mappings, back up edited configs and reinstall to
record accurate kinds. Diagnostics remain read-only and do not interrupt services.

## Risk and compatibility

Owned state gains `kind = "config"`; older binaries cannot deserialize it.
File digests, capability rollback, and managed adapter inventory are retained.
No CLI flags, dependencies, or update/uninstall policy change.

## Validation

Linux ARM64, Rust 1.93.1, rebased onto `89c45da03`:

- `cargo check --locked`
- `cargo test --workspace --locked` — 2,483 passed, 1 existing ignored test

The expanded local-repository install/doctor regression reproduced the review
finding before the fix (`Config` instead of `File`). It covers both declaration
orders, distinct sources sharing a destination, overlapping destinations,
different source modes, exact mappings, config edits, immutable tampering,
unambiguous legacy recovery, conservative ambiguous legacy checks, and unchanged
state bytes after diagnostics. Core tests preserve missing-file, mode, Linux
capability, symlink, non-regular-file, and out-of-bounds failures.

macOS ARM64: the expanded regression, check, and docs pass. Workspace Clippy
reports only the existing unused capability error variants. Full-suite testing
encounters existing platform failures: the capability rollback missing-log test
and the managed-source symlink-root test. The latter was also reproduced on
unmodified `89c45da03`; Linux workspace gates pass without exclusions.

## Documentation and rollback

Updated both README languages, both user-guide languages, and the component
contract. Before downgrading, back up state and convert owned-file
`kind = "config"` entries to `kind = "file"`, restoring the old digest behavior.
Changelog and version changes remain release work.

### Ship Note

- Changed: editable raw config content no longer fails integrity checks.
- Known limitations: legacy recovery requires the installed manifest; mixed
  directory declarations without source provenance retain hashing until reinstall.
- Not covered: native RPM config policy and preserving edited configs on update.
- Verification: Linux workspace gates and install/doctor regression above.

### DIFF
--- a/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/owned_ops.rs
+++ b/src/anolisa/crates/anolisa-cli/src/commands/tier1/install/owned_ops.rs
@@ -31,8 +31,8 @@ use anolisa_core::state::{FileOwner, ObjectKind, OwnedFile, OwnedFileKind, Servi
 use anolisa_core::state_store::StateStore;
 use anolisa_core::transaction::restore_backup_file;
 use anolisa_core::{
-    CapabilityRequest, FileKind, ResolvedInstallFile, ResolvedLifecycleHooks, ServiceActivation,
-    ServiceRequest, ServiceRunOutcome, ServiceScope, apply_capabilities, apply_services,
+    CapabilityRequest, FileKind, ResolvedLifecycleHooks, ServiceActivation, ServiceRequest,
+    ServiceRunOutcome, ServiceScope, apply_capabilities, apply_services,
     capability_for_install_mode, deactivate_services, run_hooks, service_for_install_mode,
     user_service_for_install_mode,
 };
@@ -67,7 +67,7 @@ fn rollback_capability_requests(
         .iter()
         .filter(|file| {
             file.owner == FileOwner::Anolisa
-                && file.kind == OwnedFileKind::File
+                && file.kind != OwnedFileKind::Symlink
                 && !file.capabilities.is_empty()
                 && backups.iter().any(|backup| backup.dest == file.path)
         })
@@ -440,7 +440,6 @@ impl OwnedOps for RawReplayOps<'_> {
                 &self.placed,
                 &manifest_path,
                 &prepared.manifest_toml,
-                &prepared.files,
                 &self.applied_capabilities,
             ),
             services: self.service_refs(&prepared.services),
@@ -1373,7 +1372,6 @@ impl OwnedOps for RawInstallOps<'_> {
                 &self.placed,
                 &manifest_path,
                 &prepared.manifest_toml,
-                &prepared.files,
                 &self.applied_capabilities,
             ),
             services: prepared
@@ -1470,7 +1468,6 @@ fn owned_file_rows(
     placed: &[InstalledFile],
     manifest_path: &Path,
     manifest_toml: &str,
-    contract_files: &[ResolvedInstallFile],
     applied_capabilities: &[CapabilityRequest],
 ) -> Vec<OwnedFile> {
     let mut files: Vec<OwnedFile> = placed
@@ -1485,15 +1482,13 @@ fn owned_file_rows(
             },
             kind: if f.referent.is_some() {
                 OwnedFileKind::Symlink
+            } else if f.kind == FileKind::Config {
+                OwnedFileKind::Config
             } else {
                 OwnedFileKind::File
             },
             referent: f.referent.clone(),
-            mode: expected_mode_for_path(&f.path, contract_files).or_else(|| {
-                (f.referent.is_none())
-                    .then(|| recorded_mode(&f.path))
-                    .flatten()
-            }),
+            mode: expected_mode(f),
             capabilities: capabilities_for_path(&f.path, applied_capabilities),
         })
         .collect();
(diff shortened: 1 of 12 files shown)


src/anolisa/crates/anolisa-cli/src/commands/tier1/install/owned_ops.rs
```diff
@@ -1493,36 +1497,43 @@ fn owned_file_rows(
...
-                }
-                None => "0755",
-            };
-            let octal = raw.trim().strip_prefix("0o").unwrap_or(raw.trim());
-            u32::from_str_radix(octal, 8)
-                .ok()
-                .filter(|mode| *mode <= 0o7777)
-                .map(|mode| format!("{mode:04o}"))
-        })
+        .max_by_key(|file| (file.dest == path, file.dest.components().count()))
```
ikunkun-sys: **[P1] Shared directory destinations bypass immutable-file integrity checks**

A valid manifest can map `source = "data/", type = "data"` followed by `source = "config/", type = "config"` to the same target, e.g. `{datadir}/components/agentsight`, with `mode = "0644"` on both. An archive containing `data/plain.dat` and `config/settings.toml` installs successfully because the expanded file destinations are distinct. Here both directory candidates have the same key, so `max_by_key` selects the last mapping and `owned_file_rows` persists **both** files as `Config`. I reproduced overwriting `plain.dat` after installation: `doctor` exits 0 with no findings and `status` reports healthy despite the stale recorded digest; the base revision detects the same tampering. Reversing the manifest order instead misclassifies the editable config as immutable. The same destination-only selection in `hydrate_owned_file_contracts` (`common.rs:1006-1021`) reproduces both outcomes after removing persisted config kinds, so existing installations are also affected. Directory-prefix specificity alone cannot recover which source mapping actually supplied each installed file.

ikunkun-sys (changes requested):
Requesting changes for one reproducible integrity-check regression on `abe032195c9bf14c0a7dedeea37c56142b2a93e3`.

Distinct directory sources can legally install different files into the same destination directory. The new contract lookup assigns both sources the last matching directory's kind, allowing ordinary data tampering to pass `doctor` and `status`. Legacy contract hydration has the same issue. The inline finding includes the trigger and observed results.

Validation: Linux x86_64, Rust 1.93.1. All 38 targeted tests passed: the new install/doctor regression (1), core integrity tests (21), owned install operation tests (6), and managed adapter inventory tests (10). Separate real-CLI probes used isolated XDG roots and local `file://` raw repositories, exercised both manifest orders and new/legacy state, and confirmed diagnostics did not rewrite state. With identical ordinary-file tampering, base `80301fb9e4a9fa047f95e8c943b1db1884c0b9af` reports `health_sha256_mismatch` and doctor exits 2; the affected PR-head cases report healthy and exit 0. Config-only editing in the reverse mapping order still incorrectly fails.

The current CI checks have passed, but the added regression does not cover distinct directory sources sharing a destination.

src/anolisa/crates/anolisa-cli/src/commands/tier1/install/owned_ops.rs
kongche-jbw: Fixed in ae8b84633 and rebased onto main (89c45da03), preserving the post-install-hook digest fix.

Removed destination-based kind/mode lookup entirely. The install runner now carries each file's actual mapping kind and mode through directory expansion, preview, placement, and state recording. Shared and overlapping destinations therefore cannot exchange config exemptions or permissions.

Legacy records cannot recover missing source provenance from destination prefixes. Hydration now exempts exact config declarations or unanimous config directory candidates only; mixed candidates retain SHA-256 checks, and disagreeing directory modes are not inferred. This deliberately means an edited legacy config with ambiguous directory mappings may still report a mismatch until the user backs it up and reinstalls. That limitation is documented instead of silently accepting immutable-file tampering.

The regression reproduced Config-vs-File misclassification before the fix. Both mapping orders now pass for shared destinations, overlapping directories, distinct modes, new config edits, immutable-file tampering, unambiguous legacy recovery, and conservative ambiguous legacy checks, without diagnostics rewriting state. Linux ARM64 / Rust 1.93.1: fmt, check, Clippy with warnings denied, all 2,483 workspace tests, and docs pass.

ikunkun-sys (approved):
Re-reviewed `ae8b84633d45abfed49135ac8b7fd0efabbc3e79` against base `89c45da03c92fe89a5a003f62c877362add16c4b`, including the range-diff from the previously reviewed revision. No remaining blocking findings.

The previous P1 is fixed: each placed file now retains the kind and mode of its actual source mapping, and destination-based inference has been removed. Independent real-CLI probes confirm that shared and nested destinations, in both declaration orders, detect ordinary-file tampering while allowing new-install config content edits. Config permission drift still fails. Legacy files with ambiguous directory mappings retain digest checking, consistent with the documented backup/reinstall limitation, and diagnostics leave state bytes unchanged.

Validation on Linux x86_64 with Rust 1.93.1: 94 targeted tests passed (53 install-runner, 21 integrity, 10 managed-inventory, 6 owned-operation, 1 expanded config regression, and 3 post-install-hook/rollback regressions), plus the isolated raw `file://` CLI probes above. The rebase retains main's post-hook digest refresh. Current CI checks passed. macOS was not independently retested in this review.

PR #3149, opened 2026-09-08
fix(memory): accept openclaw capability consent

ikunkun-sys:
## Why

OpenClaw gates capability-declaring plugins behind install-time consent: `openclaw plugins install` rejects them (rc=1) unless `--accept-capabilities` is passed. The gate shipped in OpenClaw **2026.8.1** — verified in the published 2026.8.1 npm tarball (the installer registers the flag and `docs/plugins/manage-plugins.md` documents the consent section; the CLI reference mentions the flag only since 2026.9.1, which is why the boundary is easy to miscount — the nightly host in #3099 happened to run 2026.9.2). `agent-memory`'s install.sh only passes `--force --dangerously-force-unsafe-install`, so `memory-anolisa` can never be installed on 2026.8.1+ hosts, violating install.sh's declared OpenClaw >= 5.0.0 compatibility contract and breaking the 9 OpenClaw E2E cases at their `ensure_memory_plugin_installed()` fixture.

## What changed

`src/agent-memory/adapters/agent-memory/openclaw/scripts/install.sh` probes `openclaw plugins install --help` and appends `--accept-capabilities` (both the default and `AGENT_MEMORY_SAFE_INSTALL=1` paths) only when the host advertises it:

- Help output is captured into a variable and matched with a whole-token regex — the same form as tokenless's installer and anolisa-core's `help_lists_flag()`. A `cmd | grep -q` pipeline can silently drop the match under `set -o pipefail` (grep exits on match → writer gets SIGPIPE), and prefix-matching would confuse near-miss options (`--no-accept-capabilities`, `--accept-capabilities-only`).
- Probe outcomes are logged distinctly: consent granted, flag not advertised (pre-2026.8.1 host, base flags), or probe failure (stderr WARNING with rc and captured output; degrades to base flags rather than failing closed so every host the manifest claims to support keeps installing).
- User guides (en/zh) document that both install entry points grant the declared-capability consent, with a troubleshooting row for the consent error, per specs/documentation-standard.md §5.

This deliberately differs from the unconditional append suggested in the issue: pre-2026.8.1 releases abort on the unknown option (`OpenClaw does not recognize option`, verified on 2026.5.27 with rc=1), so an unconditional flag would trade the new failure for an old one.

## Related issue

fixes #3099

## User / Agent impact

`memory-anolisa` installs successfully on OpenClaw 2026.8.1+ and continues to install on older releases. Consent grants and probe degradation are now visible in install output. No configuration or plugin behavior changes.

## Risk and compatibility

[...]

Issue #3134, opened 2026-09-08
test(agent-memory): wire install-script-test.ts into CI — the Test agent-memory job has no Node step


Forrest-ly:
### Component

- CI: `.github/workflows/ci.yaml` → job `Test agent-memory` (line 1028)
- agent-memory OpenClaw adapter: `src/agent-memory/Makefile:151-153`, `src/agent-memory/adapters/agent-memory/openclaw/tests/unit/install-script-test.ts`

### Background

PR #3100 added a unit test that drives the real `scripts/install.sh` with a fake `openclaw` CLI, pinning the `--accept-capabilities` gating (append the flag **iff** `plugins install --help` advertises it), the manual-consent hint echoing the argv actually used, and the failure diagnostics for a legacy installer / an unreadable help probe. It also added the entry point:

```make
test-openclaw-plugin: ## Run the OpenClaw adapter install-script test (needs node/npm)
	cd $(OPENCLAW_PLUGIN_SRC_DIR) && npm install --legacy-peer-deps --no-audit --no-fund
	cd $(OPENCLAW_PLUGIN_SRC_DIR) && npx tsx --test tests/unit/install-script-test.ts
```

**Nothing calls that target.** `Test agent-memory` only runs `cargo fmt --check` / `clippy` / `cargo test`, so the "revert the gating and the tests go red" protection currently exists only when someone remembers to run it locally. The analogous test in the sibling component *is* wired: `Test agent-sec-core` runs `make test-openclaw-plugin-coverage` right after `Setup Node.js` (`.github/workflows/ci.yaml:527-537`), covering `deploy-script-test.ts`.

### Why this needs a maintainer

The agent-side token has scopes `gist, read:org, repo` — no `workflow` — so GitHub rejects any push that touches `.github/workflows/ci.yaml`:

```text
refusing to allow an OAuth App to create or update workflow `.github/workflows/ci.yaml` without `workflow` scope
```

So the change has to be made by a human, or by a token that carries `workflow`.

### Option A — copy the agent-sec-core precedent (patch from the PR #3100 description)

```diff
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ test-agent-memory
       - name: Run tests
         run: |
           cd src/agent-memory
           cargo test --locked
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
[...]

PR #3169, opened 2026-09-08
fix(memory): gate unsafe-install flag on help

### LINKED ISSUE #3161: fix(memory): stale unsafe-install comment and deprecated --dangerously-force-unsafe-install in OpenClaw install.sh
Follow-up to #3149 (merged), split out per review guidance to keep that P1 hotfix's blast radius small. Component: agent-memory (OpenClaw plugin installer); the same pattern exists in tokenless's install.sh.

`src/agent-memory/adapters/agent-memory/openclaw/scripts/install.sh:37-49` still defaults to `--dangerously-force-unsafe-install`, with a comment explaining that OpenClaw's install-time scanner flags `child_process.spawn` as a dangerous pattern and that `AGENT_MEMORY_SAFE_INSTALL=1` opts into the regular (blocking) safe-install scan.

On OpenClaw >= 2026.8.1 that rationale is stale:

- Install-time dangerous-code scanning was removed entirely; the flag is a deprecated no-op. Its own 2026.8.1 option description reads "Deprecated no-op; security.installPolicy may still block", and #3099's install log confirmed "is deprecated and no longer affects plugin installs".
- So the script passes an ineffective flag by default, and documents a scanner that no longer exists on current hosts. `AGENT_MEMORY_SAFE_INSTALL`'s documented distinction (bypass vs. blocking scan) collapses — both paths are equivalent on >= 2026.8.1.

The driver side already models this correctly: `anolisa-core/src/adapter/openclaw.rs` classifies the option from `plugins install --help` into `UnsafeInstallSupport::{Effective, DeprecatedNoOp, Unsupported}`, refuses `--allow-unsafe-plugin-install` when it is absent or a no-op, and points operators at `security.installPolicy` instead. The script and driver now diverge in semantics on the same host.

- install.sh probes the unsafe flag's advertised effectiveness from `plugins install --help` and passes it only when effective; on deprecated-no-op hosts it omits the flag and logs why, matching anolisa-core's classification.
- The stale scanner comments rewritten to describe current host behavior (pre-2026.8.1 bypass rationale kept as historical context for old hosts).
- `AGENT_MEMORY_SAFE_INSTALL` semantics re-examined: either removed or redefined so it means something on current hosts.
- The equivalent note block in `src/tokenless/adapters/tokenless/openclaw/scripts/install.sh` ("--dangerously-force-unsafe-install is required because…") converges on the same handling.

## Possible approach

Mirror anolisa-core's `unsafe_install_support()`: locate the option's description line in the installer help, treat "no-op"/"no op" as `DeprecatedNoOp`, and branch flag passing accordingly. The probe machinery (help captured into a variable) already landed in #3149, so this extends that block rather than adding a second CLI call.

Forrest-ly:
## Why

`src/agent-memory/adapters/agent-memory/openclaw/scripts/install.sh` passed `--dangerously-force-unsafe-install` unconditionally and documented an install-time scanner that current OpenClaw hosts no longer run, while the driver already models the option correctly (`anolisa-core/src/adapter/openclaw.rs` classifies it into `UnsafeInstallSupport::{Effective, DeprecatedNoOp, Unsupported}`, refuses `--allow-unsafe-plugin-install` when it is absent or a no-op, and points operators at `security.installPolicy`). Same host, two semantics — and `AGENT_MEMORY_SAFE_INSTALL`'s documented distinction (bypass vs. blocking scan) collapsed to nothing on a no-op host.

The issue dates this to OpenClaw >= 2026.8.1. I re-derived the boundary from the published npm packages instead of reusing it, and it is **earlier** — the option description each release registers in its `dist/plugins-cli-*.js`:

| OpenClaw | published `--dangerously-force-unsafe-install` description | classification |
|---|---|---|
| 2026.5.26 | `Bypass built-in dangerous-code install blocking (plugin hooks may still block)` | effective |
| 2026.6.1 | same | effective (last stable that bypasses) |
| 2026.6.2-beta.1 | `Deprecated no-op; install policy and plugin hooks may still block` | no-op |
| 2026.6.5 | `Deprecated no-op; security.installPolicy may still block` | no-op (first stable) |
| 2026.6.8 / 2026.6.11 / 2026.6.34 / 2026.8.1 / 2026.9.2 | same | no-op |

2026.6.5's own CLI reference: "`--dangerously-force-unsafe-install` is deprecated and is now a no-op. OpenClaw no longer runs built-in install-time dangerous-code blocking for plugin installs." (2026.8.1's: "The deprecated `--dangerously-force-unsafe-install` flag remains a no-op.") So this boundary is the 2026.6.2 line — **not** the 2026.8.1 capability-consent boundary from #3149, and not the "2026.9.2+" that 2bcabb24 wrote into the tokenless guides. Both are corrected here.

## What changed

**Commit 1 — `fix(memory): gate unsafe-install flag on help`**

- install.sh classifies the option from the help the consent probe (#3149) already captures — no second CLI call — with the same line-scoped, whole-token read `unsafe_install_support()` performs, and appends the flag only while it is effective.
- Distinct log lines per outcome: bypass passed (host still scans), omitted as a deprecated no-op (naming `security.installPolicy`), omitted as not advertised, or kept because the host could not be classified.
- A failed probe is now its own state (`unknown`) instead of collapsing into "absent": the legacy bypass is kept so pre-2026.6.2 hosts still install, with a WARNING naming the opt-out — degrade rather than fail closed, matching the shape #3149 review accepted.
- A rejected install on a no-op host points at the operator-owned `security.installPolicy`, mirroring the driver's `DeprecatedNoOp` advice and tokenless's installer.
- `AGENT_MEMORY_SAFE_INSTALL` is **kept but redefined**: it declines a bypass that would actually be sent, which is a real choice on effective/unclassified hosts; on a no-op host the script says it changes nothing instead of implying two paths the host treats identically. Not removed — dropping it would silently start sending the bypass to operators who explicitly declined it on exactly the hosts where it still does something.
- Stale scanner comments rewritten, keeping the pre-2026.6.2 rationale as historical context.
- New `src/agent-memory/tests/test-openclaw-adapter-install.sh` (bash only, no node/npm), wired into `make test` through a new `test-adapters` target and into the memory branch of `tests/run-all-tests.sh`. This is the committed test #3149 deferred, and it gives #3134 a target file: CI wiring is a single `bash tests/test-openclaw-adapter-install.sh` step in `Test agent-memory` with **no** Node setup — out of reach for this token (no `workflow` scope).
- Two fixtures are verbatim `plugins install --help` captures from real releases (2026.5.22, 2026.8.1), so the classifier is pinned against commander's real rendering and line wrapping rather than hand-written help text.
- en/zh user guides document the gating and the probe-failure fallback, plus two troubleshooting rows (documentation-standard §5).

**Commit 2 — `docs(tokenless): correct unsafe no-op boundary`**

The behavioral convergence the issue asks for already landed in 2bcabb24 (#3113). What remained was the note comment still describing the scanner as unconditional (comment-only; tokenless's own adapter test passes unmodified) and the four guide spots dating the no-op to "OpenClaw 2026.9.2+", which told operators that 2026.6.5–2026.9.1 hosts still enforce the scan and still receive the bypass.

## Related issue

fixes #3161

Also delivers the test follow-up #3149 deferred and gives #3134 a file to wire.

## User / Agent impact

- Current hosts (2026.6.2-beta.1 onward, i.e. every release since June 2026): argv becomes `plugins install <dir> --force [--accept-capabilities]`; the ineffective token is dropped and the log says why. No functional change — those hosts ignored the flag.
- 2026.6.1 and earlier: argv byte-identical to before this change.
- `AGENT_MEMORY_SAFE_INSTALL=1` still declines the bypass where one would be sent, and now reports that it changes nothing where none would be.
- No plugin behavior, config, or manifest change; `compatibleVersions: ">=5.0.0"` still holds.

## Risk and compatibility

The security posture narrows: an install-time bypass is no longer sent by default where it does nothing, and a no-op host's rejection now names `security.installPolicy` instead of implying a version problem. The one path that still sends the bypass without classifying the host is a failed help probe — logged as a WARNING, keeping #3149's degrade-rather-than-fail-closed choice. That trades a possible unknown-option error on a hypothetical future host (which would fail the consent probe too) for keeping every legacy host installable; `AGENT_MEMORY_SAFE_INSTALL=1` is the escape hatch.

Known limitation, deliberately identical to `unsafe_install_support()` and tokenless's installer: the no-op check reads only the first help line that lists the flag, so a `no-op` marker wrapped onto a continuation line would go undetected. Real 2026.8.1/2026.9.2 help renders `--dangerously-force-unsafe-install    Deprecated no-op; security.installPolicy` / `may still block (default: false)`, so the marker stays on the option line today. Changing that idiom must happen on all three sides at once.

## Validation

- `bash src/agent-memory/tests/test-openclaw-adapter-install.sh` → 16 PASS: effective, no-op, uppercase `NO-OP`, near-match (`--dangerously-force-unsafe-install-only`), absent, real 2026.5.22 help, real 2026.8.1 help, probe failure, `AGENT_MEMORY_SAFE_INSTALL=1` on no-op / effective / unclassified hosts, policy-rejected install, hook opt-in after success, exactly one help probe per run, missing-CLI skip.
- Mutation-checked that test: always pass the flag → caught by the no-op hosts; never pass it → caught by `legacy_effective`; substring match instead of whole-token → caught by `unsafe_near`.
- `shellcheck` clean on both install.sh files, the new test, and `tests/run-all-tests.sh`.
- `bash src/tokenless/tests/test-openclaw-adapter-install.sh` → passes unmodified.
- `make -C src/agent-memory test-adapters` → green.
- `bash scripts/docs-lint.sh` and `python3 scripts/docs-link-check.py` → pass.
- Real-host end-to-end, forwarding only the read-only help probe to the real CLI and stubbing every mutating call (nothing was installed or configured):
  - **OpenClaw 2026.5.22** (the CLI on this machine) → `plugins install <dir> --force --dangerously-force-unsafe-install`, byte-identical to the pre-change argv, and `--accept-capabilities` correctly withheld.
  - **OpenClaw 2026.8.1** and **2026.9.2 (`3928bad`** — the build #3099/#3113 report; npm packages run under Node 22.23.2 in a temp prefix) → `plugins install <dir> --force --accept-capabilities`, no-op token dropped, reason logged.
- No Rust code changed, so `cargo fmt/clippy/test` are untouched; the `Test agent-memory` job covers them. The new bash test is not in CI yet (see #3134).

## Documentation and rollback

Docs: `docs/user-guide/{en,zh}/token-saving/agent-memory.md` (gating paragraph including the probe-failure fallback, two troubleshooting rows) and `docs/user-guide/{en,zh}/token-saving/tokenless/{framework-integration,troubleshooting}.md` (boundary correction). CHANGELOG and the RPM spec changelog are intentionally untouched — per documentation-standard §5 they are written only in release version-bump PRs, and the historical `0.1.0-3` entry that introduced `AGENT_MEMORY_SAFE_INSTALL` stays as a record of what that release did.

Rollback: revert `7bcbb1ac` (script + test + Makefile + memory docs) to restore the unconditional flag. `f8dbcd3d` is comment/doc-only.


docs/user-guide/en/token-saving/agent-memory.md
```diff
@@ -117,6 +117,8 @@ anolisa adapter status agent-memory
...
 
+The standalone `install.sh` negotiates the unsafe-install bypass the same way: it passes `--dangerously-force-unsafe-install` only while the installer advertises that option as effective. OpenClaw 2026.6.1 and earlier run an install-time safety scan that blocks `child_process` plugins non-interactively — this plugin spawns the agent-memory MCP server over stdio — so those hosts receive the bypass. OpenClaw 2026.6.5 and later dropped install-time dangerous-code blocking and list the option as a deprecated no-op, so they never receive it; there install-time safety is decided by the operator-owned `security.installPolicy`, which no script flag can override. `AGENT_MEMORY_SAFE_INSTALL=1` declines the bypass on hosts where it still has effect and changes nothing on current ones — the install log states which case applied. If the `plugins install --help` probe itself fails, the host cannot be classified: the script keeps the bypass so pre-2026.6.2 hosts still install, logs a WARNING, and `AGENT_MEMORY_SAFE_INSTALL=1` declines it there too.
```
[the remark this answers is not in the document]
Forrest-ly: Fixed in 97fde8e8 — added an `### OpenClaw` subsection to `src/agent-memory/README.md` and `src/agent-memory/README_zh.md` (semantically equivalent per documentation-standard §4.6) summarizing: the install command, that both optional flags are negotiated from `plugins install --help`, the verified unsafe-install boundary (OpenClaw 2026.6.1 and earlier still get the bypass; current hosts advertise it as a deprecated no-op and get nothing, with `security.installPolicy` in charge), and what the redefined `AGENT_MEMORY_SAFE_INSTALL=1` does on each. Kept to a summary with a link to the user guide as the full reference, per §4.5. `scripts/docs-lint.sh` and `scripts/docs-link-check.py` both pass.

ikunkun-sys (commented):
Reviewed 97fde8e8bda47d6c53bef9c772aa3836723bd4b9. One P2 finding is attached inline: the new failure diagnostic attributes unrelated install errors to security.installPolicy and advises relaxing it.

Validation: make -C src/agent-memory test-adapters passed (16 PASS lines); the Tokenless adapter installer tests passed (12 PASS lines); Bash syntax and diff whitespace checks passed. Reproduced the diagnostic issue by running the actual agent-memory installer against a stub CLI whose help advertises the deprecated no-op and whose install command returns an EACCES error. No real plugin was installed.

The existing README finding is addressed by the current head. Executed CI checks are successful, but this head's agent-memory CI does not run the new Bash test. GitHub currently reports CONFLICTING / DIRTY against main; this review covers the PR head before conflict resolution.

src/agent-memory/adapters/agent-memory/openclaw/scripts/install.sh
```diff
@@ -76,9 +68,91 @@ elif [ -n "$INSTALL_HELP" ]; then
...
+        ;;
+esac
+
 env -u OPENCLAW_HOME OPENCLAW_STATE_DIR="$OPENCLAW_STATE_DIR" "$OPENCLAW_BIN" plugins install "$PLUGIN_DIR" \
     "${INSTALL_ARGS[@]}" || {
     echo "[${COMPONENT}] openclaw CLI install failed — check OpenClaw version >= 5.0.0" >&2
+    if [ "$UNSAFE_SUPPORT" = "noop" ]; then
+        echo "[${COMPONENT}]       this OpenClaw treats --dangerously-force-unsafe-install as a deprecated" >&2
+        echo "[${COMPONENT}]       no-op, so the rejection comes from the operator-owned" >&2
+        echo "[${COMPONENT}]       security.installPolicy — relax that policy, not this script." >&2
```
ikunkun-sys: [P2] Do not attribute every install failure to security.installPolicy

This branch checks only whether the host advertises the unsafe-install option as a deprecated no-op; it does not establish why the install failed. On such a host, an ordinary filesystem permission error also produces the assertion that “the rejection comes from” security.installPolicy and the instruction to “relax that policy”. I reproduced this with the actual installer and a stub CLI: the help advertises the deprecated no-op, then installation prints `EACCES: permission denied, mkdir extensions/memory-anolisa` and exits 1; the script still emits that policy diagnosis. This sends operators toward an unrelated remedy and encourages weakening a security policy that did not cause the failure.

Forrest-ly: 确认是合理意见，已修复（`66b05248`）。

分支刚 rebase 到 `2388b370`（见上方评论），这段诊断现在位于 main 新增的 `if [ "$INSTALL_RC" -ne 0 ]` 处理器内、通用失败行之后；行为与你复现的一致：no-op 宿主上任何失败都会断言"rejection comes from security.installPolicy"并让运维去放宽策略。

修复思路对齐 `2388b370` 给 consent opt-out 定的纪律（只有 OpenClaw 输出带上文档化的拒绝短语才归因，否则保持通用 rc=1），但没有借用它的 transcript：那套 tee 机制只存在于 withheld 路径，正是为了让默认安装把 CLI 的 stderr 留在运维自己的终端上；而且 OpenClaw 对 `security.installPolicy` 拒绝没有文档化的稳定短语可 grep。所以两条 note 都改成只陈述脚本自己做了什么（省略了该参数 / 拒绝了该参数），把补救手段改成条件式，并指向运维在这条路径上本来就能读到的 CLI 输出：

```
note: this OpenClaw advertises --dangerously-force-unsafe-install as a
deprecated no-op, so the script sent no bypass and cannot shape install-time
safety on this host. Read the CLI output above for the actual cause; only
if it names security.installPolicy is that operator-owned policy what to
relax — not this script.
```

`AGENT_MEMORY_SAFE_INSTALL=1` 那条同样弱化为"若上方 CLI 输出归因于安装期扫描，再取消该变量"。

测试补了你的复现场景 `no-op host, unrelated install error`（`TEST_HELP=noop` + `TEST_GATE=unrelated`，stub 抛 `EACCES`）：仍要说明 deprecated no-op，但断言不再出现 `the rejection comes from` / `relax that policy, not this script`，且必须出现"Read the CLI output above for the actual cause"。原有 `no-op host blocked by install policy` 场景改为通过条件式措辞 `if it names security.installPolicy` 钉住该提示。本地 37 个场景全 PASS，`shellcheck -S warning` 干净。

用户指南 troubleshooting 那一行以"install.sh 报安装被 `security.installPolicy` 拦截"为前提，仍然成立，未改动。

Forrest-ly: Fixed — in two steps, both on this branch: `66b05248` (the notes no longer assert a cause) and `77721af0` (the one cause the script *can* verify is now diagnosed, and it ranks first).

Your point was that the branch classified the **host**, not the **failure**. Both notes now state only what the script itself did — sent no bypass, or declined one — and defer the cause to the CLI output the operator can already read, naming `security.installPolicy` only as a conditional ("only if it names security.installPolicy is that operator-owned policy what to relax").

On top of that, `77721af0` adds a real verdict for your repro. OpenClaw materialises the plugin under `${OPENCLAW_STATE_DIR}/extensions/` — the same path `uninstall.sh` removes — so `install_target_writable()` walks to the nearest existing ancestor and tests it with `-w`. With `extensions/` unwritable, the script now prints this instead of any policy advice:

```text
[agent-memory] openclaw CLI install failed — check OpenClaw version >= 5.0.0
[agent-memory]       the install target /home/u/.openclaw/extensions/memory-anolisa is not
[agent-memory]       writable by the user running this script, which on its own fails the install with a
[agent-memory]       filesystem-permission error (EACCES). Fix that directory's ownership/permissions — or
[agent-memory]       point OPENCLAW_STATE_DIR at a writable state directory — and re-run before drawing any
[agent-memory]       conclusion about security.installPolicy: an unwritable target is not a policy refusal,
[agent-memory]       so do not relax that policy for it.
```

Two things I deliberately did **not** do:

- **No grep for a policy-rejection phrase.** The consent attribution can do that because `Plugin "…" requires capability consent` is documented; nothing publishes a `security.installPolicy` rejection string — the 2026.8.1 help only documents `--acknowledge-install-policy-warning` as "Acknowledge security.installPolicy warnings without prompting; blocks and failures remain terminal". An invented pattern would silently never fire and the operator would lose the note entirely.
- **No transcript on this path.** The `tee` exists only on the consent-withheld path so a default install keeps the CLI's stderr on stderr and its TTY-gated prompts intact; extending it to every no-op host would also swallow the interactive policy prompt that same help advertises. The `-w` probe needs no capture, so the stream contract is unchanged — pinned by a new `channel_scenario 'no-op host failure keeps CLI stderr on stderr'`.

Tests (`src/agent-memory/tests/test-openclaw-adapter-install.sh`, 41 scenarios, ~4.5 s):

- your repro with a writable target → conditional note, no policy blame, and no invented permission verdict;
- your repro with `extensions/` at mode 500 → the verdict above, conditional note suppressed (skipped under root, where `W_OK` is granted regardless of mode);
- the verified verdict outranks both the policy-block note (`TEST_POLICY_REJECT=1`) and the declined-bypass note (`AGENT_MEMORY_SAFE_INSTALL=1` on a scan-blocking host), so nobody is sent to unset a switch or relax a policy for a permission failure;
- the stub's EACCES line is your verbatim text, `EACCES: permission denied, mkdir extensions/memory-anolisa`.

Both user guides also lost the troubleshooting row that prescribed "relax `security.installPolicy`" for any failure on a no-op host — the same misattribution, in docs — and gained an unwritable-target row plus a no-op-host row pointing at the `openclaw` output.

Unrelated to your comment but in `77721af0`: the classifier's line-scoped read used to scan the whole help capture one bash iteration at a time, which made the suite's 200k-line `big` probe cost ~18 s per run (~42 s for the suite, on a CI step documented as "~1 s"). A `grep -F` pre-filter narrows the capture first — whole input read, no `-q`/`-m` early exit, so nothing SIGPIPEs the producer under `pipefail` — and the whole-token match plus the no-op read still happen in bash on the option's own line. Suite is back to ~4.5 s.

Forrest-ly: 补充说明（同一分支追加 `77721af0`，在 `66b05248` 之上）：其中一处取代了上一条评论的决定——用户指南的 troubleshooting 行还是改了。理由是 `66b05248` 之后脚本已不再"报安装被 `security.installPolicy` 拦截"，那一行的前提和它给出的"放宽 `security.installPolicy`"处置在两个语言版本里都成了过期指引，正是本次评审反对的那类归因。现在拆成两行：安装目标不可写（`EACCES`，修目录权限/换 `OPENCLAW_STATE_DIR`，不要为此放宽策略），以及 no-op 宿主安装失败（以 `openclaw` 输出为准，只有它点名 `security.installPolicy` 时才涉及该策略）。

`77721af0` 还补上了脚本自己能核实的那一类原因：`install_target_writable()` 检查 `${OPENCLAW_STATE_DIR}/extensions/`（即 `uninstall.sh` 删除的同一路径）最近的存在祖先是否可写，命中时给出文件系统权限判定并明确 "an unwritable target is not a policy refusal, so do not relax that policy for it"，排序在两条条件式 note 之前；它不需要 transcript，所以 `2388b370` 定的流约定不变（新增 channel 场景钉住）。你的复现场景因此有两种钉法：目标可写 → 只出条件式 note、不得凭空断言权限问题；目标 mode 500 → 出权限判定、条件式 note 被抑制（root 下跳过该场景）。套件 41 个场景全 PASS，约 4.5 s（顺带把 help 分类的整段 bash 逐行扫描换成 `grep -F` 预筛，此前 200k 行的 `big` 探测让整套跑到约 42 s）。

以上细节见上一条英文回复。thread 已 resolve，若还有不同意见请 reopen。

src/agent-memory/README.md
```diff
@@ -33,8 +33,22 @@ The bundled plugin (`memory-anolisa`) is deployed by
...
+openclaw gateway restart
+```
+
+Both optional installer flags are negotiated from `openclaw plugins install
+--help`: `--accept-capabilities` is passed only when the host advertises that
+exact option, and `--dangerously-force-unsafe-install` only while the host still
+advertises the bypass as effective (OpenClaw 2026.6.1 and earlier). Current hosts
+list it as a deprecated no-op and never receive it — there install-time safety
+follows the operator-owned `security.installPolicy`, and neither switch changes
+the argv; the install log states which case applied.
```
ikunkun-sys: [P2] Keep capability consent distinct from the deprecated unsafe bypass

The statement that neither switch changes argv on current hosts is incorrect for AGENT_MEMORY_ACCEPT_CAPABILITIES; README_zh.md repeats the same claim. Only the unsafe-install bypass is a deprecated no-op. With the actual installer and the committed OpenClaw 2026.8.1 help fixture, AGENT_MEMORY_ACCEPT_CAPABILITIES=1 produces `plugins install <dir> --force --accept-capabilities`, whereas setting it to 0 produces `plugins install <dir> --force`. A stub enforcing the advertised consent gate succeeds in the first case and makes the script exit 3 in the second. This wording contradicts the opt-out documented immediately above and can lead operators to believe that withholding capability consent has no effect on current hosts.

ikunkun-sys (commented):
Re-reviewed 77721af0552e9cc1db6b62a10298a2c9f658b1d3 after the rebase and follow-up changes. The previous unconditional security.installPolicy attribution is fixed. One P2 documentation inconsistency remains, attached inline; I found no new blocking installer-logic issue.

Validation: the 41 agent-memory installer scenarios passed as a non-root user, including the permission and stream-channel scenarios; the Tokenless installer tests passed. Independently exercised the actual script with the committed OpenClaw 2026.8.1 help fixture and a stub consent gate: enabling consent passes --accept-capabilities and succeeds, while withholding consent omits it and exits 3. No real plugin was installed.

Executed CI checks are successful, including the now-wired Bash installer test. GitHub currently reports MERGEABLE / BLOCKED with no approval recorded.

src/agent-memory/README.md
Forrest-ly: Confirmed — the claim was wrong for `AGENT_MEMORY_ACCEPT_CAPABILITIES`, and `README_zh.md` repeated it. Fixed in `a9aa4b02` (docs only, appended on top of `77721af0`).

Reproduced your case first against the committed 2026.8.1 help fixture with a stub that enforces the advertised consent gate — it matches exactly:

| `AGENT_MEMORY_ACCEPT_CAPABILITIES` | argv | rc |
| --- | --- | --- |
| `1` | `plugins install <dir> --force --accept-capabilities` | 0 |
| `0` | `plugins install <dir> --force` | 3 |

So only the unsafe-install bypass is inert on current hosts; the consent switch still shapes the argv there, and withholding it is what makes a gated install fail.

Both READMEs now describe each switch on its own terms instead of collapsing them into one sentence:

- `--accept-capabilities` is passed only when the host advertises that exact option — current hosts do, so `AGENT_MEMORY_ACCEPT_CAPABILITIES` still shapes the argv there: `1` appends `--accept-capabilities` to `openclaw plugins install <dir> --force`, while `0` omits it and a host that gates consent then rejects the install.
- `--dangerously-force-unsafe-install` is passed only while the host still advertises the bypass as effective (OpenClaw 2026.6.1 and earlier). Current hosts list it as a deprecated no-op and never receive it under either setting, so `AGENT_MEMORY_SAFE_INSTALL` changes nothing there and install-time safety follows the operator-owned `security.installPolicy`.

That also clears the contradiction with the opt-out paragraph immediately above, and `README_zh.md` carries the same split (两个开关并不对称 → consent 仍改变 argv，只有 deprecated no-op 覆盖参数不受 `AGENT_MEMORY_SAFE_INSTALL` 影响).

Checked the user guides for the same sentence: `docs/user-guide/en/token-saving/agent-memory.md` and its `zh` counterpart already scope the "changes nothing on current ones" claim to `AGENT_MEMORY_SAFE_INSTALL` alone, so nothing to correct there.

Tests after the change: 41/41 installer scenarios (`make -C src/agent-memory test-openclaw-install`), `scripts/docs-lint.sh`, `scripts/docs-link-check.py` — all green.

ikunkun-sys (approved):
Reviewed a9aa4b0218d2ee639f7808a73d2199fd02405c80. No remaining findings.

Both previous P2 findings are addressed: installation failures no longer unconditionally blame security.installPolicy, and the English/Chinese READMEs correctly distinguish the active capability-consent switch from the deprecated unsafe-install bypass. All three review threads are resolved.

Range-diff confirms the latest update only changes the two READMEs; installer code and tests are unchanged from the reviewed 77721af0. The prior 41 agent-memory installer scenarios (run as non-root) and Tokenless installer tests remain applicable. For this head, docs-lint, relative-link checks and diff whitespace checks passed. Current executed CI checks are successful. No real plugin installation was performed.

PR #3024, opened 2026-09-02
test(cosh-ng): [shell] pin PS2 ownership

kongche-jbw:
## Why

Current main already preserves Bash PS2 ownership in interactive sessions, but the open issues lack stable current-main nested-PTY evidence. This test-only change pins the existing behavior without modifying production code.

## What changed

- Cover Enhanced environment PS2 and startup override in login and non-login shells.
- Cover Native login and non-login PS2 ownership.
- Reject out-of-order or duplicate prompt transcripts.
- Bound malformed continuation probes with a child-reaping watchdog.

## Related issue

relates #2537

relates #2540

## User / Agent impact

None. Production behavior is unchanged.

## Risk and compatibility

Low risk. The diff is test-only and changes no public CLI, API, configuration, privileged path, migration, or runtime contract.

## Validation

Local Linux nested PTY validation:

- cargo test --package cosh-shell --test shell_host parity::secondary_prompt_oracle_rejects_wrong_order_and_duplicates -- --exact --nocapture
- cargo test --package cosh-shell --test shell_host parity::bash_secondary_prompt_stays_shell_owned -- --exact --nocapture
- cargo test --package cosh-shell --test shell_host parity::bash_secondary_prompt_watchdog_reaps_unclosed_continuation -- --exact --nocapture
- cargo fmt --all -- --check
- crates/cosh-shell/scripts/check-layout.sh
- git diff --check

The PTY matrix covers six Bash scenarios, and the malformed-continuation case reaches its internal timeout and reaps the child.

## Documentation and rollback

No documentation changes are needed. Roll back by reverting commit 0df928289a3531d9a2ef4c4b05afc0d8d4985d81.

PR #3176, opened 2026-09-09
feat(cosh-ng): [shell] define prompt contract

kongche-jbw:
## Why

Enhanced uses Bash `PROMPT_COMMAND` for bounded prompt integration, so an exported user scalar no longer reaches child processes after integration. Document this compatibility boundary explicitly and pin the behavior instead of restoring `export` on Cosh's internal hook text.

## What changed

Extend the existing prompt-environment regression to 16 real PTY sessions: exported nonempty scalar, array, empty scalar, and unassigned export, each in login/non-login Native and Enhanced shells. The user startup files establish the fixtures after system profiles run, so distro-added prompt hooks do not alter the test inputs. Check user hook execution order, variable attributes, child environment values, reassigned-array attributes, and the absence of a marker rcfile in the Native control. Add matching English/Chinese guidance and README links; production behavior is unchanged.

## Related issue

Refs #2540 and #2598; supplements #2915. This PR deliberately does not auto-close either issue: it records the agreed PROMPT_COMMAND boundary, while the original Core-107/platform acceptance still needs an explicit disposition. PS2 is independent and has existing coverage from #3024.

## User / Agent impact

| Integration | Session representation | Child environment |
| --- | --- | --- |
| Enhanced, Bash 5.1+ | Combined array; existing export attribute retained | No PROMPT_COMMAND value, including when the original exported scalar was empty |
| Enhanced, Bash 4.3–5.0 | Combined scalar without export attribute | Neither the original user scalar nor Cosh's internal hook text is exported |
| Native control | Bash's own representation and attributes | Exported scalar retains its original value; arrays are not exported by Bash |

User hooks configured before integration keep the execution order supported by the selected Bash version. Both Enhanced Assisted and Enhanced Shell-only have these limits: Shift+Tab does not remove integration. Configure hooks independently in child shells' startup files, or choose Native when Bash's own prompt-variable behavior is required; Native provides no Cosh hooks, observation, or insights. `--resume` always selects Enhanced, even when Native is configured; omit it when Native behavior is required.

This boundary concerns Bash PROMPT_COMMAND only. It does not relax PS2 behavior, ordinary environment-variable semantics, user hook execution, job control, exit status, or terminal restoration.

## Risk and compatibility

Tests and documentation only; no runtime, dependency, CLI, or configuration changes. The Native PTY control directly launches Bash without a Cosh marker rcfile; it is not an installed-package, console-login, or PAM oracle.

## Validation

Ubuntu 24.04.3, Linux aarch64, Bash 5.2.21:

```bash
cd src/cosh-ng
cargo test --locked -p cosh-shell --test shell_host parity::bash_prompt_command_attributes_and_child_environment_follow_integration_contract -- --exact --nocapture
rustfmt --edition 2021 --check crates/cosh-shell/tests/shell_host/parity.rs
```

- Exact regression: 1 test passed, all 16 PTY sessions executed, 0 failed/ignored.
- Temporarily appended `;:` to PROMPT_COMMAND at the beginning of the user startup fixture to model a system startup hook: the previous revision failed the Native scalar child-environment assertion; the corrected fixture passed all 16 PTY sessions. Temporary source changes and logs were removed.
- `bash scripts/docs-lint.sh`, changed-document relative links, bilingual parity, source verification of the resume override, and `git diff --check` passed.
- Bash 4.3–5.0 expectations are encoded but were not executed locally. The original Alibaba Cloud Linux 3 host was not rerun. No ALinux4/Core-107 rerun, dual-architecture acceptance, installed-package/PAM check, manual screenshot, or real-provider validation is claimed. Broader regression coverage remains with CI.

## Documentation and rollback

Updated both component READMEs and the English/Chinese interactive-terminal overview. Reverting this commit removes the documentation and expanded regression only; existing runtime behavior is unchanged. Users requiring the Native boundary can select `COSH_SHELL_INTEGRATION=native cosh` at startup.

### DIFF
--- a/docs/user-guide/en/user-entrypoint/cosh-ng/shell/overview.md
+++ b/docs/user-guide/en/user-entrypoint/cosh-ng/shell/overview.md
@@ -44,6 +44,34 @@ Approved Shell commands in enhanced integration stay in the foreground Shell,
 so prompts, output, job control, and `Ctrl+C` remain usable. See
 [Tool approval](approval.md) for the safety rules.
 
+## Bash prompt compatibility
+
+Enhanced combines user `PROMPT_COMMAND` hooks with Cosh's prompt hooks. User
+hooks configured before integration keep their execution order under the
+selected Bash version. Both Assisted and Shell-only use this integration.
+
+The variable's representation and child environment have these limits:
+
+- Bash 5.1 and newer use an array. An existing export attribute is retained,
+  but Bash does not export array values. A previously exported scalar
+  `PROMPT_COMMAND` therefore no longer reaches child processes, including
+  when its original value was an empty string.
+- Bash 4.3–5.0 use a combined scalar without the export attribute. This keeps
+  Cosh's internal hook text out of child processes; the original user scalar
+  is not exported either.
+
+Configure prompt hooks in each interactive shell's startup files when child
+shells should initialize them independently. Choose Native at startup when
+the session must preserve Bash's own prompt-variable representation and
+environment behavior. Native provides no Cosh hooks, observation, or insights;
+switching Enhanced to Shell-only does not remove the limits above.
+
+`--resume` always selects Enhanced, even with `COSH_SHELL_INTEGRATION=native`
+or `shell.integration = "native"`. Omit `--resume` when Native behavior is required.
+
+These limits apply to Bash `PROMPT_COMMAND`, not ordinary environment variables
+or the interactive continuation prompt `PS2`.
+
 ## Sessions and proactive help
 
 - Enhanced sessions are persisted by cosh-core and scoped to the workspace
--- a/src/cosh-ng/crates/cosh-shell/tests/shell_host/parity.rs
+++ b/src/cosh-ng/crates/cosh-shell/tests/shell_host/parity.rs
@@ -275,56 +275,150 @@ fn enhanced_bash_preserves_last_argument_across_prompt_boundary() {
 }
 
 #[test]
-fn enhanced_bash_keeps_internal_prompt_hooks_out_of_child_env() {
+fn bash_prompt_command_attributes_and_child_environment_follow_integration_contract() {
     if !bash_available() {
         return;
     }
 
-    let work_dir = std::env::temp_dir().join(format!(
-        "cosh-shell-prompt-command-env-{}-{}",
-        std::process::id(),
-        unique_suffix()
-    ));
-    let config = ShellHostConfig::new("prompt-command-env", &work_dir)
-        .with_integration(ShellIntegration::Enhanced)
-        .with_env("PROMPT_COMMAND", "printf ''")
-        .with_env("LANG", "C.UTF-8")
-        .with_env("LC_ALL", "C.UTF-8");
-    let output = run_scripted_bash(
-        &config,
-        &[ScriptedInput::command(
-            "printf '__PROMPT_COMMAND__=%s\\n' \"$(declare -p PROMPT_COMMAND)\"; \
-             if env | grep -q '^PROMPT_COMMAND='; then \
-               printf '__CHILD_PROMPT_COMMAND__=present\\n'; \
-             else \
-               printf '__CHILD_PROMPT_COMMAND__=absent\\n'; \
-             fi; \
-             (PROMPT_COMMAND=('printf p1' 'printf p2'); \
-             printf '__REASSIGNED_PROMPT_COMMAND__=%s\\n' \
-             \"$(declare -p PROMPT_COMMAND)\")",
-        )],
-    )
-    .expect("scripted bash");
+    let supports_array = bash_supports_prompt_command_array();
+    for (fixture, inherited, startup, expected_trace) in [
+        ("scalar", "USER_PROMPT_TRACE+=scalar", "", "scalar"),
+        (
+            "array",
+            "",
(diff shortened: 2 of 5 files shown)


src/cosh-ng/crates/cosh-shell/tests/shell_host/parity.rs
```diff
@@ -275,56 +275,145 @@ fn enhanced_bash_preserves_last_argument_across_prompt_boundary() {
...
+                let child_env = read("child-env");
+                let child_value = child_env
+                    .lines()
+                    .find_map(|line| line.strip_prefix("PROMPT_COMMAND="));
+                let expected_child = if !enhanced && matches!(fixture, "scalar" | "empty") {
+                    Some(inherited)
+                } else {
+                    None
+                };
+                assert_eq!(child_value, expected_child, "{context}: {child_env}");
```
ikunkun-sys, line 401: [P2] Account for system startup hooks in the Native login expectation

For login_shell=true, Native launches bash --login -i, so the temporary HOME does not isolate /etc/profile or /etc/bashrc. On this Alibaba Cloud Linux 3 host, /etc/profile sources /etc/bashrc, which appends ;/etc/sysconfig/bash-prompt-history to the exported PROMPT_COMMAND. The scalar fixture leaves that value intact, but expected_child is still Some(inherited), so this assertion deterministically fails with USER_PROMPT_TRACE+=scalar;/etc/sysconfig/bash-prompt-history versus USER_PROMPT_TRACE+=scalar. I reproduced this with Bash 4.4.20 and the CI-pinned Rust 1.88.0, including outside the sandbox. This flags normal Native Bash startup behavior as a regression and aborts the remaining matrix cases on affected hosts.

ikunkun-sys (changes requested):
One reproducible P2 test portability regression needs correction: the Native login fixture assumes system startup files leave the inherited PROMPT_COMMAND unchanged.

Validation on Alibaba Cloud Linux 3, Linux x86_64, Bash 4.4.20, using the CI integration toolchain Rust 1.88.0:

```bash
cargo +1.88.0 test --locked --offline -p cosh-shell --test shell_host parity::bash_prompt_command_attributes_and_child_environment_follow_integration_contract -- --exact --nocapture
```

The test fails at scalar/login=true/Native with the child value USER_PROMPT_TRACE+=scalar;/etc/sysconfig/bash-prompt-history instead of USER_PROMPT_TRACE+=scalar. Running the compiled test outside the sandbox reproduces the same assertion failure. The parent revision's original regression passes outside the sandbox with Rust 1.98.1. The 16-session matrix does not complete because this assertion aborts the loop.

The current-head cosh-ng checks and Docs Lint are successful; that CI result does not cover this host startup configuration. No production-code changes are introduced by this PR.

docs/user-guide/en/user-entrypoint/cosh-ng/shell/overview.md
```diff
@@ -44,6 +44,31 @@ Approved Shell commands in enhanced integration stay in the foreground Shell,
...
+  `PROMPT_COMMAND` therefore no longer reaches child processes, including
+  when its original value was an empty string.
+- Bash 4.3–5.0 use a combined scalar without the export attribute. This keeps
+  Cosh's internal hook text out of child processes; the original user scalar
+  is not exported either.
+
+Configure prompt hooks in each interactive shell's startup files when child
+shells should initialize them independently. Choose Native at startup when
+the session must preserve Bash's own prompt-variable representation and
+environment behavior. Native provides no Cosh hooks, observation, or insights;
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in 4aa0dfe57048e78ee2282ac42764b1417c3845cb. Both English and Chinese compatibility guides now state that --resume selects Enhanced even when Native is configured through COSH_SHELL_INTEGRATION or shell.integration, and instruct users requiring Native behavior to omit --resume.

src/cosh-ng/crates/cosh-shell/tests/shell_host/parity.rs
kongche-jbw: Fixed in 4aa0dfe57048e78ee2282ac42764b1417c3845cb. The user startup fixture now resets and exports PROMPT_COMMAND after system startup files run, then applies the scalar/array/empty/unassigned case. This keeps all 16 Native/Enhanced login/non-login combinations deterministic while retaining the exact attribute, hook-order, child-environment, and marker-rcfile assertions. Production startup behavior is unchanged.

ikunkun-sys (approved):
Re-reviewed the complete current diff at 4aa0dfe57048e78ee2282ac42764b1417c3845cb and compared it with the previous revision. No remaining actionable findings.

The previously reported P2 is fixed: the user startup files establish PROMPT_COMMAND after system profiles have run, so system-added hooks no longer alter the fixture. On the original Alibaba Cloud Linux 3 x86_64 host with Bash 4.4.20 and the CI integration toolchain Rust 1.88.0, the exact regression passes and completes all 16 PTY sessions:

```bash
cargo +1.88.0 test --locked --offline -p cosh-shell --test shell_host parity::bash_prompt_command_attributes_and_child_environment_follow_integration_contract -- --exact --nocapture
```

The English and Chinese guides now disclose that --resume selects Enhanced even when Native is configured; this matches the runtime implementation. Targeted rustfmt using Rust 1.97.1, changed-document relative links and prompt anchors, and git diff --check pass. Current-head relevant CI checks are successful and both review threads are resolved.

Bash 5.1+ was not executed locally in this re-review. No installed-package/PAM or ALinux4/Core-107 acceptance is claimed.

PR #3177, opened 2026-09-09
fix(memory): align corpus supplement with host contract

Forrest-ly:
## Why

Three defects in the agent-memory OpenClaw adapter's corpus supplement and its test fixtures, all invisible to CI: `Test agent-memory` runs `cargo fmt` / `clippy` / `test` only, so nothing typechecks or runs the adapter's Node side. `npm run build:check` (`tsc --project tsconfig.json --noEmit`) fails on a clean `main` checkout with **4 errors** and has been red since 2026-05-26; two of the three defects below are contract bugs rather than stale annotations.

**1. The corpus supplement never reported the window it returned** (`src/index.ts:515`)

`api.registerMemoryCorpusSupplement({... get()})` answered with `{corpus, path, title, content}`. The host contract — the plugin SDK's `MemoryCorpusGetResult` (`plugins/memory-state.d.ts:25`) — also requires `fromLine` and `lineCount`, and the host spreads the supplement's result straight into the JSON it hands the model:

```js
const { content, ...rest } = supplement;
return { ...rest, text: content };
```

So a windowed `memory_read` served by this corpus came back without saying *which* window it contained, while the identical request served by the host's builtin reader does report `from` / `lines` (`buildMemoryReadResult`). Two corpora, two response shapes, and the model has no way to paginate this one.

Reading the host's builtin reader also exposed two live bugs in our own slicing:

| input | before | host builtin | after |
|---|---|---|---|
| `fromLine: 0` | `start = -1` → `Array.slice(-1, n)` → the document's **last line**, reported as nothing | `Math.max(1, from ?? 1)` → line 1 | line 1, reported |
| `fromLine: -3` | third line **from the end** | line 1 | line 1, reported |
| `lineCount: 0` | falsy → **the rest of the document** | `Math.max(1, lines ?? …)` → 1 line | 1 line, reported |

The last one matters for a token-saving component: the request that should have returned nothing returned everything.

**2. The supplement claimed the host's own corpus id, and then advertised a read route that did not exist** (`src/index.ts:497,529`)

Every result was stamped `corpus: "memory"` — the id the host stamps on its *builtin* hits (`createMemorySearchTool`: `surfacedMemoryResults = memoryResults.map((r) => ({ ...r, corpus: "memory" }))`), while the host's compiled-wiki supplement labels its own hits `corpus: "wiki"` (`toWikiSearchResult`). The field is how the model tells corpora apart inside one merged `memory_search` response, and how it picks the `corpus` argument to echo back to `memory_get`. Claiming the host's id broke both directions:

- agent-memory hits were indistinguishable from `MEMORY.md` hits in a `corpus=all` merge;
- a model echoing the label back was routed to the builtin reader (`readAgentMemoryFile`, over the workspace `MEMORY.md` + `memory/*.md`), which cannot see this store. `memory_get` consults supplements only for `corpus=wiki` and, after a failed builtin read, for `corpus=all` — and its `corpus` enum is `memory | wiki | all`. So **every path this supplement surfaced was a dead end** in the shared tool flow: search returned it, read could not.

Fixing the id alone was not enough, as @ikunkun-sys found in review: `memory_get` consults supplements only *after* the builtin read has **thrown** (`executeMemoryReadResult` → `resolveMemoryReadFailureResult`), and the host's `readMemoryFile` does not throw inside its own namespace. For every shape its `isMemoryPath` admits — `MEMORY.md`, `dreams.md`, anything under `memory/` — it *resolves*: with the workspace file's content when one exists, and with `{text: "", path}` when none does. So a hit this store held at `memory/sample.md` came back correctly labelled, advertised `corpus=all`, and still never reached `get()`; the model got empty text, or another store's file. The route the label promised was a property of the **path**, not of the corpus.

**3. Both test fixtures predate `AgentMemoryConfig.sessionId` / `.sessionDir`** (`tests/smoke-test.ts:51`, `tests/unit/mcp-client-test.ts:98,103`)

`McpStdioClient` takes `AgentMemoryConfig`, which has carried both fields since `bc00bbe5`, but each fixture was written with only the five fields that existed on paper — TS2345 ×3.

## What changed

- **`src/corpus.ts` (new)** — the supplement's host-contract helpers as pure functions, testable without a live plugin API or MCP client (the same shape `safety.ts` and `keyword-extract.ts` already use):
  - `sliceCorpusWindow(text, fromLine?, lineCount?)` → `{content, fromLine, lineCount}`: the window *actually returned*, with inputs clamped the way the host's builtin reader clamps them.
  - `AGENT_MEMORY_CORPUS` / `toCorpusSearchResult()` / `AGENT_MEMORY_CORPUS_READ_HINT`: the corpus identity. The id is this plugin's own (`agent-memory`) rather than the wiki's, because `memory_get` walks the registered supplements and returns the first non-null answer — borrowing `corpus: "wiki"` would make the two corpora interchangeable to the model and let whichever registered first shadow the other. Since the enum has no value naming our corpus, each search hit also carries the route that *does* reach a supplement as its `provenanceLabel` (`"agent-memory (read via memory_get corpus=all)"`); `get()` omits the hint because the route is chosen by then.
  - `isHostBuiltinMemoryPath()` / `AGENT_MEMORY_READ_NAMESPACE` / `toCorpusReadHandle()` / `fromCorpusReadHandle()`: the read handle that makes that label true. A hit whose store path lands in the host's builtin namespace is advertised as `agent-memory/<path>` — `isMemoryPath` tests the whole relative path, so the handle is foreign to the builtin reader, the read throws, and `corpus=all` falls through to this supplement. Paths outside that namespace are advertised unchanged: they already reach the supplement, and namespacing them too would spend tokens on a route that needs no repair. A store path that already begins with the namespace is prefixed as well, so the mapping stays injective and the store's own `agent-memory/notes.md` cannot be confused with a namespaced `notes.md`. The host's predicate is mirrored rather than imported (no plugin-sdk entry point exports `isMemoryPath`) and deliberately over-eager — an unnecessary prefix still routes here, a missed one silently swallows the read.
- **`src/index.ts`** — `get()` delegates the window and reports the two required fields, resolves the read handle to a store path before addressing MCP, and echoes the handle back so a follow-up windowed read keeps routing here; both `search()` and `get()` label results with the plugin's own corpus id. No other behaviour change.
- **`tests/unit/corpus-test.ts` (new)** — 29 cases: the window (whole document, explicit window, open-ended window, window past EOF, `fromLine` 0 / negative, `lineCount` 0 / negative, fractional bounds, non-finite bounds, trailing-newline accounting, empty document); the corpus identity (own id, never one the host resolves without a supplement, read hint names `corpus=all`, MCP score passed through unrescaled); and the handle as a pure mapping (builtin-namespace shapes incl. the host's normalisation and case-insensitive `dreams.md`, near-misses that must *not* be namespaced, handle round-trip, injectivity on a path that already starts with the namespace).
- **`tests/unit/corpus-routing-test.ts` (new)** — 9 cases that pin the reviewer's repro against the pinned host's **own** exported `readMemoryFile` (`openclaw/plugin-sdk/memory-core-host-engine-storage`), not a restatement of its logic, over two workspaces — one bare, one holding a `MEMORY.md` and a `memory/sample.md` whose content differs from the store's: every advertised hit routes to the supplement and returns store content; every advertised handle makes the host's reader throw `path required`; the *bare* store path is answered by the host itself (`"builtin one\nbuiltin two"` when the file exists, `{text: ""}` when it does not) with zero supplement calls; and `notes/observed/sample.md` still routes un-namespaced. With `toCorpusReadHandle` reduced to the identity this suite fails 6 of 9.
- **`tests/smoke-test.ts` / `tests/unit/mcp-client-test.ts`** — fixtures name `sessionId` / `sessionDir`. The smoke test points its child at a throwaway dir created only when the suite will actually run, rather than at `/run/anolisa/sessions` (root-owned 0700 via the tmpfiles.d snippet, and not guaranteed to exist on a dev host); the unit-test one stays inert because that binary path never spawns.

An absent `lineCount` still means "to the end of the document", exactly as before. The host applies its char budget only on the builtin path, so capping here would silently change what operators get back from a plugin whose reads are already bounded by `maxReadBytes` — a separate decision, deliberately not folded into a contract fix.

Commits are independent: the window fix, the corpus-id-and-routing fix and the fixture repair can each be reviewed (or dropped) on their own.

## Verification

Node v22.21.1, dependencies installed per `src/agent-memory/Makefile` `test-openclaw-plugin`; branch rebased onto current `main`, which now carries `48d1002` (#3155):

- `npm run build:check` — main: **4 errors**; this branch: **0**
- `npm test` — **126 tests / 0 fail** (`sliceCorpusWindow` 13/13, `corpus identity` 4/4, `host builtin namespace` 4/4, `corpus read handle` 8/8, `memory_get routing through the pinned host` 9/9). The 5 `resolveConfig` failures this branch reported earlier were the evaluation-order defect #3155 fixes; the rebase picks that fix up, so the suite is green end to end.
- `npm run build` — esbuild bundles the change (`dist/index.js` 239.0kb)
- `npm run smoke` — skips cleanly with no binary present, and creates no temp dir when skipped

Host-side claims were read off the pinned `openclaw@2026.5.7` artifact (`dist/tools-B3jFTrn6.js`, `dist/read-file-D9O26XT9.js`, `dist/internal-DcS7JcH6.js`, `dist/cli-D-ubyXOM.js`, `dist/plugin-sdk/src/plugins/memory-state.d.ts`) and, for the routing suite, executed against it — not assumed from the docs.

## Notes

- No docs update: `specs/documentation-standard.md` §5 triggers on new/modified CLI flags, config options, install method, architecture/protocol changes and new components. This is none of those — it complies with protocols the SDK already defines — and no agent-memory doc mentions the corpus supplement. The read handle is internal to the supplement's own results: not a path an operator configures, not a file the store grows. CHANGELOG is release-PR-only per the same section.
- **Residual limitation, host-side**: the model still has to choose `corpus=all` or `corpus=wiki` to read one of our hits, because `memory_get`'s enum has no value naming a third-party corpus — a read that omits `corpus` still lands on the builtin reader and fails. Closing that fully needs a host change (accept a supplement corpus id in the enum, or fall back to supplements on *any* read the builtin namespace did not claim). Worth raising with whoever owns the memory-core corpus routing.
- **Residual limitation, operator-side**: an operator who points `extraPaths` — or a qmd collection root — at a path this store also uses reintroduces the shadowing for that one path. Neither setting is visible to a plugin, so the handle cannot account for them; the same host-side fallback would close this too.
- **Follow-up, deliberately not changed here**: cross-corpus score comparability. Our BM25 path returns negative ranks (`src/index/store.rs:470` — "more negative = worse"), the LIKE fallback returns a non-negative term frequency, and hybrid returns RRF values around 1/(60+rank); the host normalises its own BM25 with `bm25RankToScore` (`relevance / (1 + relevance)` → (0,1)) and the wiki supplement emits 0..100 page scores. Merged `corpus=all` ranking therefore mixes four scales, and our hits sort last whenever the builtin side has any hit at all. `balanceCorpora` keeps them in the response, so this is a ranking defect rather than a visibility one, and one supplement cannot fix a merge policy — no rescaling is applied in this PR.

Tracked in Multica as AGE-6861, AGE-6875 and AGE-6926 (Developer autopilot scans).

### DIFF
--- a/src/agent-memory/adapters/agent-memory/openclaw/src/corpus.ts
+++ b/src/agent-memory/adapters/agent-memory/openclaw/src/corpus.ts
@@ -1,0 +1,67 @@
+/**
+ * Line-window slicing for the memory corpus supplement.
+ *
+ * OpenClaw's `MemoryCorpusGetResult` (plugin-sdk `plugins/memory-state`)
+ * requires a supplement to report the window it actually returned: `fromLine`
+ * is the 1-based first line and `lineCount` the number of lines in `content`.
+ * The host spreads that result straight into the `memory_read` response, so
+ * these two fields are what the model sees — and the host's builtin reader
+ * (`buildMemoryReadResult`) clamps the same two inputs, which means a
+ * supplement that echoes the *requested* window instead of the *returned* one
+ * makes the two corpora disagree about what "from line 0, 10 lines" produced.
+ *
+ * Pure and standalone on purpose: the supplement itself needs a live plugin
+ * API and MCP client, this does not, so the window arithmetic is testable.
+ */
+
+export type CorpusWindow = {
+  /** The selected lines, joined with "\n". */
+  content: string;
+  /** 1-based first line of `content`; always >= 1. */
+  fromLine: number;
+  /** Number of lines in `content`; 0 when the window starts past the end. */
+  lineCount: number;
+};
+
+/** 1-based start line, clamped exactly like the host's builtin reader.
+ *  Without the clamp a 0 or negative request reaches `Array.slice` as a
+ *  negative index and silently returns the document's *tail*, and a non-finite
+ *  value propagates NaN into the reported window. */
+function normalizeFromLine(fromLine?: number): number {
+  if (fromLine === undefined || !Number.isFinite(fromLine)) return 1;
+  return Math.max(1, Math.trunc(fromLine));
+}
+
+/** Requested line count, or undefined for "to the end of the document".
+ *  Clamped to >= 1 for the same reason: to the host's builtin reader a
+ *  non-positive count means one line, not the whole file — and this plugin's
+ *  entire job is keeping payloads out of the context window. */
+function normalizeLineCount(lineCount?: number): number | undefined {
+  if (lineCount === undefined || !Number.isFinite(lineCount)) return undefined;
+  return Math.max(1, Math.trunc(lineCount));
+}
+
+/**
+ * Select a line window from a document body.
+ *
+ * `text` is split on "\n" and the selected lines are re-joined, so a document
+ * ending in a newline carries one trailing empty line — the same accounting the
+ * host's builtin reader uses, which keeps `fromLine + lineCount` pagination
+ * consistent across corpora.
+ */
+export function sliceCorpusWindow(
+  text: string,
+  fromLine?: number,
+  lineCount?: number,
+): CorpusWindow {
+  const lines = text.split("\n");
+  const start = normalizeFromLine(fromLine);
+  const requested = normalizeLineCount(lineCount);
+  const end = requested === undefined ? lines.length : start - 1 + requested;
+  const selected = lines.slice(start - 1, end);
+  return {
+    content: selected.join("\n"),
+    fromLine: start,
+    lineCount: selected.length,
+  };
+}
--- a/src/agent-memory/adapters/agent-memory/openclaw/src/corpus.ts
+++ b/src/agent-memory/adapters/agent-memory/openclaw/src/corpus.ts
@@ -1,5 +1,6 @@
 /**
- * Line-window slicing for the memory corpus supplement.
+ * Corpus identity, read handles and line-window slicing for the memory corpus
+ * supplement.
  *
  * OpenClaw's `MemoryCorpusGetResult` (plugin-sdk `plugins/memory-state`)
  * requires a supplement to report the window it actually returned: `fromLine`
@@ -10,8 +11,14 @@
  * supplement that echoes the *requested* window instead of the *returned* one
  * makes the two corpora disagree about what "from line 0, 10 lines" produced.
  *
+ * The path a search hit advertises is the other half of the same contract: the
+ * model echoes it back to `memory_get`, and whether that read ever reaches this
+ * supplement is a property of the path's *shape*, not of its corpus — see
+ * `toCorpusReadHandle`.
+ *
  * Pure and standalone on purpose: the supplement itself needs a live plugin
- * API and MCP client, this does not, so the window arithmetic is testable.
+ * API and MCP client, this does not, so the window arithmetic and the path
+ * mapping are testable.
  */
 
 export type CorpusWindow = {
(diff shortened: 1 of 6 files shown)


src/agent-memory/adapters/agent-memory/openclaw/src/corpus.ts
```diff
@@ -0,0 +1,132 @@
...
+ * `memory_get`'s `corpus` enum is `memory | wiki | all`, so the model cannot
+ * echo `agent-memory` back; with no route it retries the default builtin read
+ * and lands in the dead end described above. `corpus=all` is the value that
+ * reaches a supplement once the builtin read has failed, which is exactly this
+ * case. The hint costs a few tokens per hit and replaces a failed read plus its
+ * retry — a net saving for a component whose job is token economy. Only search
+ * hits carry it: by the time `get()` answers, the route is already chosen.
+ */
+export const AGENT_MEMORY_CORPUS_READ_HINT =
+  "agent-memory (read via memory_get corpus=all)";
```
ikunkun-sys, line 120: [P2] The advertised read route still misses valid corpus paths

This hint is attached to every search hit, but OpenClaw 2026.5.7 only falls back to supplements when the builtin read throws. Its readMemoryFile returns a successful {text: "", path} for a missing MEMORY.md or memory/*.md, and returns the builtin file when one exists. Consequently, an agent-memory hit at memory/sample.md or MEMORY.md still cannot be reliably read by following this new hint: it yields empty text or content from the other store, without calling this supplement.

Reproduced using the pinned host's createMemorySearchTool/createMemoryGetTool, this PR's plugin registration, and mocked MCP store responses: search(corpus=all) surfaced all three agent-memory hits with this hint; get(path="notes/observed/sample.md", corpus="all", from=2, lines=1) called the supplement and returned "store two", while missing memory/sample.md and MEMORY.md returned empty text with zero supplement calls. After creating a different workspace MEMORY.md, the same read returned "builtin two", also bypassing the supplement. Thus the route repair remains incomplete even when the model explicitly follows corpus=all; this is separate from the documented limitation when corpus is omitted.

ikunkun-sys (changes requested):
Reviewed commit 953d3dfff81cb78611d65e5a228d17f27e6c2a22. One P2 correctness issue remains in the new read-route hint: corpus=all does not reliably reach the supplement for valid store paths that also belong to the host's builtin memory namespace. See the inline reproduction.

Validation: build:check and build passed; all 17 new tests passed. Running the unit files with node --import tsx produced 100/105 passing, versus 83/88 on the base, with the same five pre-existing config failures. Smoke skipped because no agent-memory binary was available. The routing probe used the pinned openclaw@2026.5.7 tool implementation and this PR's plugin registration, with mocked MCP responses; it was not a real MCP server end-to-end run.

At submission, head/base were unchanged, CI had no failed checks, mergeable was MERGEABLE, mergeStateStatus was BLOCKED, reviewDecision was REVIEW_REQUIRED, and there were no existing review threads.

src/agent-memory/adapters/agent-memory/openclaw/src/corpus.ts
Forrest-ly: Fixed in [`b04e869d`](https://github.com/alibaba/anolisa/commit/b04e869d) — the branch is also rebased onto current `main`, so it now carries `48d1002` (#3155) and the 5 `resolveConfig` failures from your verification notes are gone.

You are right that the label promised a route the path could block: `memory_get` consults supplements only after the builtin read *throws*, and `readMemoryFile` resolves rather than throws for every shape its `isMemoryPath` admits — `MEMORY.md`, `dreams.md`, `memory/**` — returning the workspace file, or `{text: "", path}` when the workspace has none.

Rather than narrow the claim, the supplement now makes it true. `search()` advertises each hit through `toCorpusReadHandle`, which prefixes `agent-memory/` onto exactly the paths the host answers itself. `isMemoryPath` tests the whole relative path, so the handle is foreign to the builtin reader: the read throws, `corpus=all` falls through to `get()`, which strips the prefix before addressing the store and echoes the handle back, so a follow-up windowed read keeps routing here instead of being handed a bare path the host answers itself.

Paths outside the host's namespace are advertised unchanged — they already reached the supplement, and namespacing them too would spend tokens on a route that needs no repair. A store path that already begins with `agent-memory/` is prefixed as well, so the mapping stays injective and the store's own `agent-memory/notes.md` cannot be confused with a namespaced `notes.md`.

New `tests/unit/corpus-routing-test.ts` (9 cases) pins your repro against the pinned host's own exported `readMemoryFile` — not a restatement of its logic — over two workspaces, one bare and one holding a `MEMORY.md` and a `memory/sample.md` with different content:

- every advertised hit routes to the supplement and returns *store* content, in both workspaces;
- every advertised handle makes the host's reader throw `path required`;
- the bare store path is what the host answers itself — `"builtin one\nbuiltin two"` when the workspace file exists, `{text: ""}` when it does not, in both cases with zero supplement calls;
- `notes/observed/sample.md` still reaches the supplement un-namespaced.

With `toCorpusReadHandle` reduced to the identity, that suite fails 6 of its 9 cases.

`isMemoryPath` is mirrored in `corpus.ts` rather than imported, because no plugin-sdk entry point exports it, and mirrored over-eagerly: an unnecessary prefix still routes here, a missed one silently swallows the read. The residual is recorded in the commit body — an operator who points `extraPaths`, or a qmd collection root, at a path this store also uses reintroduces the shadowing for that one path, and neither is visible to a plugin, so closing that needs a host-side change.

Verification after the rebase, Node v22.21.1: `npm run build:check` 0 errors; `npm test` 126/126 (was 100/105); `npm run build` bundles at 239.0kb.

ikunkun-sys (approved):
Re-reviewed b04e869d5fce0c3e77741ea3deca93de48c169d2 against base d279176d8389faed706cfe23763009903d173829. No remaining blocking findings. The previous P2 read-routing finding is fixed: search results namespace paths that the builtin reader would otherwise claim, the supplement decodes them before addressing the store, and returned paths retain the handle for subsequent windowed reads.

Compared the revisions with git range-diff and reviewed the complete current diff. Independently exercised the pinned openclaw@2026.5.7 createMemorySearchTool/createMemoryGetTool through this PR's plugin registration with mocked MCP responses: 36 scenarios passed across bare and shadowing workspaces, corpus=all/wiki, builtin-shaped paths, paths already containing the agent-memory/ prefix, and Unicode paths. Initial and continuation reads returned the correct store content; missing store files remained unavailable.

Local validation on Node v24.15.0: build:check and build passed; running all unit test files with node --import tsx passed 126/126, including the prior five config failures now fixed by the rebase. Smoke skipped because no agent-memory binary was available. Real MCP server end-to-end execution, qmd, and custom extraPaths configurations were not validated; the documented host/configuration limitations remain outside this approval's verified scope.

Immediately before submission, head/base were unchanged and CI had no failed checks.

PR #3187, opened 2026-09-09
fix(deps): install Node.js for memory builds

Forrest-ly:
## Why

`src/agent-memory/Makefile` has needed Node.js since the OpenClaw adapter landed: `install:` → `build:` → `build-openclaw-plugin` runs `npm install --legacy-peer-deps` + `npm run build` (esbuild) to produce `adapters/agent-memory/openclaw/dist/index.js` (`Makefile:121`).

`scripts/build-all.sh` never declared that requirement for the `memory` component. It was missing from **both** places that matter:

- the `install_node` gate — `want_component cosh || want_component sec-core || want_component sight` (twice: dry-run plan and the real call)
- `source_build_runtime_dependencies()` — which injects build-only language runtimes and had a single `sight)` case

`memory` is in `DEFAULT_COMPONENTS`, so a plain `./scripts/build-all.sh` on a Node-less host reported the dependency set as satisfied and then died inside `make`:

```text
### memory dry-run dependency plan (before)
DRY-RUN: detect Linux distribution and package manager
DRY-RUN: preflight platform capabilities before user dependency setup
DRY-RUN: check/install Rust toolchain if needed          <-- no Node.js line
DRY-RUN: verify all runtime dependencies after user dependency setup

### selected_runtime_dependencies, COMPONENTS=(memory) (before)
(empty)                     # COMPONENTS=(sight) -> sight|node|language-runtime|...|>=20|

### system-mode preflight, COMPONENTS=(memory), node absent from the system PATH (before)
==> Runtime dependency preflight
[ok]    Selected component runtime dependencies are available
preflight exit=0

### what the build then does (make with node/npm absent from PATH)
cd adapters/agent-memory/openclaw && npm install --legacy-peer-deps --no-audit --no-fund
/bin/sh: npm: command not found
make: *** [Makefile:121: build-openclaw-plugin] Error 127
```

That is exactly the late, unactionable failure the preflight exists to prevent — no missing-dependency report, no package hint, no retry command.

## What changed

**`scripts/build-all.sh`** (commit 1)

- `memory` added to the `install_node` gate in both the dry-run plan and the real path, matching the `d9d0f567 ("fix(scripts): install Node.js for agentsight builds")` precedent.
- `memory)` case added to `source_build_runtime_dependencies()` emitting `memory|node|language-runtime|node --version|nodejs|nodejs||>=20|` — the same record shape `sight` uses, so Node ≥ 20 (the repository-wide minimum documented in `docs/BUILDING.md` §3) is preflighted, reported and re-verified after install.
- `install_node`'s step header now names every component that triggers it. `"Node.js (for copilot-shell)"` was already stale for `sec-core` and `sight` before this change, and would have been actively misleading once `memory` started calling it.

**`tests/test-build-all-runtime-deps.sh`** (commit 1) — the file the `check-build-runtime-deps` CI job runs:

- `test_manifest_parser_covers_component_dependencies` now also asserts the `memory` source-build record.
- `test_memory_source_build_sets_up_node` — user-mode `do_install_deps` with `COMPONENTS=(memory)` calls `install_node` and the full preflight then passes.
- `test_memory_dry_run_plan_lists_node_setup` — keeps the dry-run plan aligned with the real flow (the same invariant `d9d0f567` called out).

**Docs** (commit 2) — `docs/BUILDING.md` + `docs/BUILDING_zh.md` (Node.js row and `agent-memory` row) and `src/agent-memory/README.md` (Requirements) now name Node.js ≥ 20 + npm for source builds, so the documented prerequisites and the enforced preflight agree.

**Why this shape and not a manifest entry.** The alternative was a `[[component.dependencies]]` block in `src/anolisa/manifests/components/agent-memory/component.toml`. That would make Node a *runtime* contract for the component, which is wrong: the installed MCP server is a Rust binary that never shells out to `node`, and the RPM ships a prebuilt `dist/`, so packaged installs (`anolisa install agent-memory`, `yum install agent-memory`) need no Node at all. `source_build_runtime_dependencies()` is the existing mechanism for exactly this build-only case, which is why `sight` uses it.

## Related issue

`no-issue:` found while re-verifying agent-memory build/test health on `main`. The closest open issue, #3134, asks for the adapter's TypeScript tests to be wired into the `Test agent-memory` CI job (that needs a token with `workflow` scope, since it edits `.github/workflows/ci.yaml`); this PR is about `build-all.sh`'s dependency preflight and touches no workflow file.

Tracked in Multica as AGE-6890 (Developer autopilot scan; no claimable development task was open, so this run re-verified agent-memory build health on `main` and fixed what it found).

## User / Agent impact

- `./scripts/build-all.sh` (default six components), `--component memory`, and `--deps-only` now set up Node for `memory`.
- **User mode**: missing Node is installed through the existing nvm/npm path before the build, so the memory build no longer dies at `npm: command not found`.
- **System mode**: missing or too-old Node is reported up front as a manual blocker, before any package mutation, with the retry command:

```text
==> Runtime dependency preflight
[error] Missing runtime dependencies; no component files were installed:
  memory: node [language-runtime] - requires node >=20

[info]  Install these language runtimes manually:
  memory: node >=20 in /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

[info]  Then retry:
  <repo>/scripts/build-all.sh --component memory --system
```

- No change to `anolisa install agent-memory`, to the RPM, or to the component's runtime dependency contract.

## Risk and compatibility

The checked box is the documented build prerequisite (`docs/BUILDING.md`, `src/agent-memory/README.md`) and the dependency-plan output gaining a `DRY-RUN: check/install Node.js if needed` line for `memory`.

One deliberate narrowing: in **system** mode a host carrying Node 18 could previously build the adapter (esbuild accepts ≥ 18) and is now stopped by the preflight at ≥ 20. That aligns `memory` with the Node ≥ 20 policy already enforced for `cosh` (manifest), `sec-core` and `sight`, and with the statement already in `docs/BUILDING.md`: "A system install that needs Node.js requires Node.js 20 or newer in the standard system PATH." Node 18 is upstream EOL, so no supported host loses a working build path.

**Pre-existing sibling gap, intentionally not touched here.** `docs/BUILDING.md` also lists `tokenless` and `ws-ckpt` among the components whose plugin builds use npm, and neither is in the `install_node` gate either (`src/tokenless/Makefile:43` even runs `node -p` at parse time). They have the same late-failure shape, but choosing their minimum Node version is a decision for those components' owners — `tokenless` runs `node --test` on TypeScript sources, which wants a newer floor than the ≥ 20 used here. Flagging it rather than guessing.

## Validation

Alibaba Cloud Linux 3 (rpm), Node v22.21.1, npm 10.9.4, cargo 1.96.0, clean worktree at `main` (`43d6ec87`):

| Check | Command | Result |
|---|---|---|
| Dependency plan (after) | `./scripts/build-all.sh --component memory --deps-only --dry-run` | now prints `DRY-RUN: check/install Node.js if needed` |
| Dependency record (after) | `source scripts/build-all.sh; COMPONENTS=(memory); selected_runtime_dependencies` | `memory\|node\|language-runtime\|node --version\|nodejs\|nodejs\|\|>=20\|` |
| System-mode preflight, node absent (after) | harness with `RUNTIME_SYSTEM_PATH` pointing at an empty dir | stops with the `memory: node [language-runtime]` manual-blocker report shown above (before: `[ok] ... available`, `exit=0`) |
| Late-failure repro (before) | `make -C src/agent-memory build-openclaw-plugin` with node/npm removed from `PATH` | `/bin/sh: npm: command not found`, `Makefile:121`, `Error 127` |
| Happy path, real run | `NPM_REGISTRY=https://registry.npmjs.org/ CARGO_HOME=$(mktemp -d) ./scripts/build-all.sh --component memory --deps-only` | `==> Node.js (for copilot-shell, agent-sec-core, agentsight, agent-memory)` → `Node.js v22.21.1 already installed, skipping` → preflight ok → `Dependency setup complete` (registry and `CARGO_HOME` pinned so the run mutates no host config) |
| Regression suite | `bash tests/test-build-all-runtime-deps.sh` | **38 ok / 0 not ok** (36 pre-existing + 2 new); this is the script the `check-build-runtime-deps` job runs |
| Negative check | revert only `scripts/build-all.sh`, keep the tests | all three new assertions fail (`agent-memory source Node dependency was not collected`, `agent-memory source build did not set up its Node dependency`, `expected 'DRY-RUN: check/install Node.js if needed'`) — the tests are not vacuous |
| Docs gates | `bash scripts/docs-lint.sh`, `python3 scripts/docs-link-check.py` | naming convention ok, en/zh tree parity ok, all relative links resolve |
| Shell lint | `shellcheck -S warning scripts/build-all.sh` | 7 warnings before and after, identical SC codes — no new findings |
| Syntax | `bash -n` on both changed shell files | ok |

Not run: `cargo test` for the crate (untouched by this change) and a full `--install` (would mutate the host).

## Documentation and rollback

- Docs updated in the same PR, per `specs/documentation-standard.md` §5: `docs/BUILDING.md` and `docs/BUILDING_zh.md` (Node.js row + `agent-memory` row, kept semantically equivalent as a bilingual pair) and `src/agent-memory/README.md` Requirements. `README_zh.md` has no Requirements section at all — a pre-existing asymmetry against `README.md` that this PR neither widens nor repairs.
- Rollback: revert both commits. No state, schema, packaging or workflow change; nothing to migrate. The two commits are independent — reverting only the docs commit leaves the fix in place with stale docs, reverting only the fix leaves the docs claiming a requirement the script does not enforce.

### DIFF
--- a/scripts/build-all.sh
+++ b/scripts/build-all.sh
@@ -613,7 +613,7 @@ node_version_satisfies_on_path() {
 }
 
 install_node() {
-    step "Node.js (for copilot-shell)"
+    step "Node.js (for copilot-shell, agent-sec-core, agentsight, agent-memory)"
     local REQUIRED="20.0.0"
     local NVM_INSTALL_MAJOR="24"
 
@@ -631,12 +631,157 @@ install_node() {
         if [[ -s "$NVM_DIR/nvm.sh" ]]; then source "$NVM_DIR/nvm.sh"; fi
     }
 
+    _node_command_path() {
+        if [[ "$INSTALL_MODE" == "system" ]]; then
+            echo "$RUNTIME_SYSTEM_PATH"
+        else
+            echo "$PATH"
+        fi
+    }
+
+    _npm_ok() {
+        PATH="$(_node_command_path)" npm -v &>/dev/null
+    }
+
+    # Several distributions ship nodejs without npm, while the plugin builds
+    # (`build-openclaw-plugin`, copilot-shell, agentsight) call npm directly.
+    # A node-only toolchain therefore is not a ready Node.js setup.
+    #
+    # `nvm install-latest-npm` is *not* a way to recover a missing npm: it
+    # starts by running `npm --version` and aborts with "Unable to obtain npm
+    # version" when that fails (see nvm_install_latest_npm in nvm.sh), so it
+    # needs the very npm that is gone. `nvm install <version>` does not help
+    # either: for a version that is already installed it exits 1 with
+    # "<version> is already installed." before unpacking anything. An
+    # nvm-managed Node keeps npm in <prefix>/lib/node_modules/npm behind the
+    # <prefix>/bin/npm and <prefix>/bin/npx symlinks, so restore exactly that
+    # from the registry tarball instead. It needs only curl and tar (both
+    # already required by this script) plus a writable prefix, never sudo, and
+    # leaves the existing Node install untouched.
+    #
+    # The npm release each Node.js release shipped with is recorded in the
+    # dist index, so prefer that: it is by construction inside the engine range
+    # of the node binary next to it, whereas the registry `latest` tag lands
+    # npm 12 on Node 20, which npm itself reports as unsupported.
+    _bundled_npm_version() {
+        local mirror="${1%/}" node_version="$2"
+        curl -fsSL --connect-timeout 10 --max-time 60 "$mirror/index.json" \
+            | tr '{' '\n' | grep -F "\"version\":\"$node_version\"" \
+            | head -1 | grep -o '"npm":"[0-9][0-9.]*"' | head -1 \
+            | sed 's/.*"\([0-9][0-9.]*\)"/\1/' || true
+    }
+
+    _latest_npm_version() {
+        local registry="$1" version=""
+        version="$(curl -fsSL --connect-timeout 10 --max-time 30 \
+                "$registry/-/package/npm/dist-tags" \
+            | grep -o '"latest"[[:space:]]*:[[:space:]]*"[0-9][^"]*"' \
+            | head -1 | sed 's/.*"\([0-9][^"]*\)"$/\1/' || true)"
+        if [[ -z "$version" ]]; then
+            version="$(curl -fsSL --connect-timeout 10 --max-time 30 \
+                    "$registry/npm/latest" \
+                | grep -o '"version"[[:space:]]*:[[:space:]]*"[0-9][^"]*"' \
+                | head -1 | sed 's/.*"\([0-9][^"]*\)"$/\1/' || true)"
+        fi
+        echo "${version:-}"
+    }
+
+    _restore_nvm_npm() {
+        local node_bin="$1"
+        local prefix modules npm_dir registry node_version version archive
+        prefix="$(cd "$(dirname "$node_bin")/.." && pwd)" || return 1
+        modules="$prefix/lib/node_modules"
+        npm_dir="$modules/npm"
+        registry="${npm_config_registry:-$NPM_REGISTRY}"
+        registry="${registry%/}"
+        node_version="${prefix##*/}"
+
+        info "Restoring npm for the nvm-managed Node.js in $prefix ..."
+        version=""
+        if [[ "$node_version" == v[0-9]* ]]; then
+            version="$(_bundled_npm_version \
+                "${NVM_NODEJS_ORG_MIRROR:-https://nodejs.org/dist}" "$node_version")"
+            if [[ -n "$version" ]]; then
+                info "Node.js $node_version shipped with npm $version"
+            fi
+        fi
+        if [[ -z "$version" ]]; then
+            version="$(_latest_npm_version "$registry")"
+            if [[ -n "$version" ]]; then
+                warn "Could not read the Node.js dist index; using the registry's latest npm $version, which may not support Node.js $node_version"
+            fi
+        fi
+        if [[ -z "$version" ]]; then
+            warn "Could not resolve an npm version for Node.js $node_version"
+            return 1
+        fi
+
+        archive="$(mktemp "${TMPDIR:-/tmp}/anolisa-npm-XXXXXX.tgz")" || return 1
+        if ! curl -fsSL --connect-timeout 10 --max-time 180 \
+                -o "$archive" "$registry/npm/-/npm-$version.tgz"; then
+            warn "Could not download npm $version from $registry"
+            rm -f "$archive"
+            return 1
+        fi
+
+        # Strip the archive's package/ prefix without touching sibling global packages.
+        rm -rf "$npm_dir"
+        if ! mkdir -p "$npm_dir" || ! tar -xzf "$archive" --strip-components=1 -C "$npm_dir"; then
+            warn "Could not unpack npm $version into $npm_dir"
+            rm -f "$archive"
+            return 1
+        fi
+        rm -f "$archive"
+        chmod +x "$npm_dir/bin/npm-cli.js" "$npm_dir/bin/npx-cli.js" 2>/dev/null || true
+        ln -sfn "../lib/node_modules/npm/bin/npm-cli.js" "$prefix/bin/npm"
+        ln -sfn "../lib/node_modules/npm/bin/npx-cli.js" "$prefix/bin/npx"
+
+        if _npm_ok; then
+            ok "npm $(PATH="$(_node_command_path)" npm -v) restored for the nvm-managed Node.js"
+            return 0
+        fi
+        warn "Unpacked npm $version into $npm_dir but it still does not run"
+        return 1
+    }
+
+    _ensure_npm() {
+        if _npm_ok; then
+            return 0
+        fi
+        if [[ "$INSTALL_MODE" == "system" ]]; then
+            # System installs never mutate package state: the operator owns
+            # language runtimes. Components whose build actually calls npm
+            # (agent-memory) declare it as a source-build dependency record,
+            # where the preflight already turns a missing npm into a manual
+            # blocker, so this stays a warning here.
+            warn "npm was not found in $RUNTIME_SYSTEM_PATH; the plugin builds invoke npm"
+            return 0
+        fi
+        warn "Node.js is available but npm is missing; the plugin builds invoke npm"
+        local node_bin
+        node_bin="$(PATH="$(_node_command_path)" command -v node 2>/dev/null || true)"
+        if [[ -n "$node_bin" && "$node_bin" == "${NVM_DIR:-$HOME/.nvm}"/* ]]; then
+            _restore_nvm_npm "$node_bin" && return 0
+            warn "Could not restore npm inside ${NVM_DIR:-$HOME/.nvm}; falling back to the $npm_pkg package"
+        fi
+        info "Installing $npm_pkg via $PKG_BASE ..."
+        if [[ "$PKG_BASE" == "deb" ]]; then sudo apt-get update -y 2>/dev/null || true; fi
+        # shellcheck disable=SC2086
+        sudo $PKG_INSTALL $npm_pkg || true
+        if _npm_ok; then
+            ok "npm $(PATH="$(_node_command_path)" npm -v) installed"
+            return 0
+        fi
+        die "Failed to install npm; the plugin builds run npm install and npm run build. Install the '$npm_pkg' package, or reinstall the Node.js version your nvm manages, then retry"
+    }
+
     _configure_npm_mirror
 
     if _node_ver_ok; then
         local command_path="$PATH"
         [[ "$INSTALL_MODE" == "system" ]] && command_path="$RUNTIME_SYSTEM_PATH"
         ok "Node.js $(PATH="$command_path" node -v) already installed, skipping"
+        _ensure_npm
         return 0
     fi
 
@@ -657,6 +802,7 @@ install_node() {
             local command_path="$PATH"
             [[ "$INSTALL_MODE" == "system" ]] && command_path="$RUNTIME_SYSTEM_PATH"
             ok "Node.js $(PATH="$command_path" node -v) installed via package manager"
+            _ensure_npm
             return 0
         fi
         warn "Package manager install did not satisfy version requirement"
@@ -737,6 +883,7 @@ install_node() {
     _configure_npm_mirror
 
     if _node_ver_ok; then
+        _ensure_npm
         ok "Node.js $(node -v), npm $(npm -v)"
         info "nvm was sourced for this session; open a new terminal (or run: source ~/.bashrc) to persist"
     else
@@ -1411,8 +1558,8 @@ do_install_deps() {
                 echo "DRY-RUN: preflight platform capabilities before user dependency setup"
             fi
         fi
-        if want_component cosh || want_component sec-core || want_component sight; then
-            echo "DRY-RUN: check/install Node.js if needed"
+        if want_component cosh || want_component sec-core || want_component sight || want_component memory; then
+            echo "DRY-RUN: check/install Node.js and npm if needed"
         fi
         if want_component cosh || want_component sec-core || want_component cosh-ng || want_component sight; then
             echo "DRY-RUN: check/install build tools if needed"
@@ -1447,7 +1594,7 @@ do_install_deps() {
         fi
     fi
 
-    if want_component cosh || want_component sec-core || want_component sight; then
+    if want_component cosh || want_component sec-core || want_component sight || want_component memory; then
         install_node
     fi
 
@@ -1980,6 +2127,17 @@ source_build_runtime_dependencies() {
         case "$component" in
             sight)
                 echo 'sight|node|language-runtime|node --version|nodejs|nodejs||>=20|'
+                ;;
+            memory)
+                # `make install` bundles the OpenClaw adapter from source
+                # (npm install + esbuild), so Node.js is a source-build
+                # requirement even though the installed MCP server itself is
+                # a Rust binary that never shells out to node.
+                echo 'memory|node|language-runtime|node --version|nodejs|nodejs||>=20|'
+                # npm is a separate package on several distributions and
+                # `build-openclaw-plugin` calls it directly, so a node-only
+                # toolchain must not pass the source-build preflight.
+                echo 'memory|npm|language-runtime|npm --version|npm|npm|||'
                 ;;
         esac
     done < <(runtime_install_components)
--- a/src/agent-memory/README.md
+++ b/src/agent-memory/README.md
@@ -81,6 +81,7 @@ Profile gating (basic/advanced/expert) controls tool visibility per deployment.
 
 - Linux (x86_64 / aarch64)
 - Rust ≥ 1.85 (for source build)
+- Node.js ≥ 20 and npm (for source build — bundles the OpenClaw adapter)
 - Optional: embedding provider (OpenAI or Ollama) for vector search
 
 ## License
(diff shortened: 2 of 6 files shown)


scripts/build-all.sh
```diff
@@ -1981,6 +1981,13 @@ source_build_runtime_dependencies() {
             sight)
                 echo 'sight|node|language-runtime|node --version|nodejs|nodejs||>=20|'
                 ;;
+            memory)
+                # `make install` bundles the OpenClaw adapter from source
+                # (npm install + esbuild), so Node.js is a source-build
+                # requirement even though the installed MCP server itself is
+                # a Rust binary that never shells out to node.
+                echo 'memory|node|language-runtime|node --version|nodejs|nodejs||>=20|'
```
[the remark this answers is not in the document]
Forrest-ly: Fixed in b857581f — npm is now validated and provisioned instead of being implied by node.

- `source_build_runtime_dependencies` emits a second memory record, `memory|npm|language-runtime|npm --version|npm|npm|||`, so both the system-mode preflight and the post-setup user-mode preflight fail when npm is absent (previously the node-only record let both pass).
- `install_node` no longer returns on a node-only toolchain: a new `_ensure_npm` runs on all three success paths (already-installed, package-manager, nvm). It probes npm on the same command path as node, provisions it via `nvm install-latest-npm` when node lives under `$NVM_DIR`, otherwise installs the distro `npm` package, and dies with an actionable message if neither works.
- System installs keep the existing "language runtime is a user decision" policy: `_ensure_npm` dies with `System npm is required in <path>; install it manually and retry` rather than mutating package state.
- The dry-run plan line now reads `check/install Node.js and npm if needed`.

Tests: added `test_memory_preflight_requires_npm_alongside_node`, `test_install_node_provisions_missing_npm`, `test_install_node_system_requires_npm`, and extended the memory source-dependency and deps-phase assertions to the two-record shape. `bash tests/test-build-all-runtime-deps.sh` → 41 ok / 0 not ok. `shellcheck --severity=warning` reports the same pre-existing findings as main (no new ones).

src/agent-memory/README.md
```diff
@@ -81,6 +81,7 @@ Profile gating (basic/advanced/expert) controls tool visibility per deployment.
 
 - Linux (x86_64 / aarch64)
 - Rust ≥ 1.85 (for source build)
+- Node.js ≥ 20 and npm (for source build — bundles the OpenClaw adapter)
```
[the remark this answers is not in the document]
Forrest-ly: Fixed in b857581f — `src/agent-memory/README_zh.md` now has a `## 环境要求` section mirroring the English `## Requirements`, placed before `## 许可证` (same position as `src/anolisa/README_zh.md` and `src/blaze/README_zh.md`):

- Linux（x86_64 / aarch64）
- Rust ≥ 1.85（源码构建）
- Node.js ≥ 20 与 npm（源码构建 —— 打包内置的 OpenClaw adapter）
- 可选：用于向量检索的 embedding 提供方（OpenAI 或 Ollama）

`bash scripts/docs-lint.sh` passes (naming convention + en/zh tree parity) and `python3 scripts/docs-link-check.py` reports all relative links resolve.

scripts/build-all.sh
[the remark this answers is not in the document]
Forrest-ly: Addendum in 5aca49f6 — the system-mode half of this needed a second pass.

The first cut died whenever npm was absent from `RUNTIME_SYSTEM_PATH`, which broke `Source Build System Install (Ubuntu 22.04)` (`sec-core-source-code-build.yaml:153-180`): that job symlinks only `node` into `/usr/local/bin` and then runs `build-all.sh --component sec-core --system`, so npm legitimately stays outside the system path there while the build resolves it from the runner PATH.

System mode now warns and returns instead — no package mutation, operator-owned language runtime, matching the existing node policy in this function. The hard requirement stays where it is load-bearing: the `memory|npm` source-build record makes the system-mode preflight report npm as a manual language-runtime blocker before any agent-memory install proceeds, and user-mode setup still provisions npm.

Tests: `test_install_node_system_warns_when_npm_is_not_in_system_path` (asserts the warning and zero `sudo` calls) and `test_memory_system_preflight_blocks_on_missing_npm`; suite is 42 ok / 0 not ok.

scripts/build-all.sh
```diff
@@ -631,12 +631,63 @@ install_node() {
...
+            return 0
+        fi
+        warn "Node.js is available but npm is missing; the plugin builds invoke npm"
+        local node_bin
+        node_bin="$(PATH="$(_node_command_path)" command -v node 2>/dev/null || true)"
+        if [[ -n "$node_bin" && "$node_bin" == "${NVM_DIR:-$HOME/.nvm}"/* ]]; then
+            info "Installing npm for the nvm-managed Node.js ..."
+            _source_nvm
+            if cmd_exists nvm; then
+                nvm install-latest-npm &>/dev/null || true
```
ikunkun-sys: [P2] The nvm recovery command requires the npm that is missing

When Node.js >=20 is nvm-managed but npm is absent or cannot run, this branch invokes `nvm install-latest-npm`. That command first runs `npm --version` and returns 2 with `Unable to obtain npm version` if it cannot obtain a version, so it cannot bootstrap the missing npm (see [nvm v0.40.3 implementation](https://github.com/nvm-sh/nvm/blob/v0.40.3/nvm.sh#L186-L205)). The error is suppressed here and execution falls through to `sudo apt-get install -y npm` / the RPM equivalent. In an isolated reproduction with real Node.js and nvm, no npm, and unavailable sudo, `install_node` ends with `Failed to install npm` and exit 1. Thus the newly introduced nvm provisioning path cannot repair the condition it is entered for, and a user-local Node installation still requires a successful system package installation to proceed. The added provisioning test only exercises the distro-package stub and misses this branch.

ikunkun-sys (changes requested):
Reviewed 5aca49f689a686eae7a3ce1b69ffa15d9bfacdc6. One P2 correctness issue remains in the newly added nvm npm-provisioning branch; see the inline finding.

Validation: all 42 runtime-dependency regression tests passed, along with six isolated PATH preflight cases (user/system mode, missing npm, complete toolchain, old Node), shell syntax checks, and the memory dependency dry-run. The npm recovery failure was reproduced with real Node.js v24.15.0 and nvm v0.40.4, npm absent from the isolated PATH, and a sudo stub that records calls and denies installation. The pinned nvm v0.40.3 has the same npm-presence prerequisite. No real package installation or full component build was performed.

Current CI has no failing checks; remaining checks are skipped. The two earlier review threads are resolved.

scripts/build-all.sh
Forrest-ly: Fixed in d5b6d3a5 — you are right, and I reproduced it on a real fixture.

**The bug.** `nvm install-latest-npm` cannot bootstrap a missing npm: `nvm_install_latest_npm` runs `npm --version` first and returns 2 with `Unable to obtain npm version` when it cannot (nvm v0.40.3 `nvm.sh:197-217`), and `&>/dev/null || true` hid that. So an nvm-managed Node >= 20 with no npm fell straight through to the distro package and, with no sudo, died. Official Node v20.19.5 unpacked into `$NVM_DIR/versions/node`, npm removed, isolated PATH with no npm, sudo stub refusing — against 5aca49f6:

```
[ok]    Node.js v20.19.5 already installed, skipping
[warn]  Node.js is available but npm is missing; the plugin builds invoke npm
[info]  Installing npm for the nvm-managed Node.js ...
[info]  Installing npm via deb ...
[error] Failed to install npm; the plugin builds run npm install and npm run build
exit 1
```

**The fix.** `nvm install-latest-npm` is gone. `_restore_nvm_npm` restores npm the way nvm itself lays it out — an nvm-managed Node keeps npm in `<prefix>/lib/node_modules/npm` behind the `<prefix>/bin/npm` and `<prefix>/bin/npx` symlinks — by fetching the npm tarball from the configured registry (`npm_config_registry`, else `$NPM_REGISTRY`), unpacking it into that prefix, and recreating both symlinks. Only curl and tar, both already required by this script; never sudo; the installed Node is left untouched.

Why not the other candidates:

- `nvm install <version>` — for the version being repaired it prints `<version> is already installed.`, runs `nvm use` and returns without extracting anything (`nvm.sh:3540-3544`), so it only helps when the remote latest differs from what is installed. `nvm uninstall` + `nvm install` would work but destroys a working Node if the network fails, and `install_node` has no nvm-level rollback.
- corepack — ships yarn/pnpm, not npm, while `build-openclaw-plugin` calls `npm install` / `npm run build` directly.
- distro `npm` package — kept, but now only as a fallback once the restore has failed; a user-local Node should not need root to get its own npm back.

Version choice: the npm release is read from the Node.js dist index (`NVM_NODEJS_ORG_MIRROR`, which `_configure_npm_mirror` already sets), since it records the npm each Node release shipped with — the restored npm is then by construction inside the engine range of the node binary next to it. The registry `latest` tag alone is not: it lands npm 12 on Node 20, which npm itself reports as unsupported (`npm v12.0.2 does not support Node.js v20.19.5`). Registry `latest` stays as the fallback when the dist index is unreachable, with a warning saying so.

**Diagnostics.** Nothing in this path is suppressed any more. Each failure reports: `Could not resolve an npm version for Node.js vX`, `Could not download npm V from <registry>`, `Could not unpack npm V into <modules>`, `The npm V archive did not contain a package/ directory`, `Could not move the unpacked npm into <dir>`, `Unpacked npm V into <dir> but it still does not run`, and `Could not restore npm inside $NVM_DIR; falling back to the npm package` before the distro attempt. `sudo $PKG_INSTALL $npm_pkg` also lost its `2>/dev/null`, and the final `die` now names both ways out.

**Verification** — same real fixture, after the fix:

```
[info]  Restoring npm for the nvm-managed Node.js in .../versions/node/v20.19.5 ...
[info]  Node.js v20.19.5 shipped with npm 10.8.2
[ok]    npm 10.8.2 restored for the nvm-managed Node.js
exit 0
```

`bin/npm` and `bin/npx` come out byte-identical to the official tarball's (`npm -> ../lib/node_modules/npm/bin/npm-cli.js`), `npm -v` / `npx -v` report 10.8.2 with no engine warning, and a real `npm install ms` in a scratch dir succeeds (1 package added, `require('ms')` works). Zero sudo calls.

**Tests** — the nvm branch is covered now, on an isolated PATH that carries no npm so a host npm cannot satisfy it:

- `test_install_node_restores_npm_for_nvm_managed_node_without_sudo` — hermetic `file://` registry plus `file://` dist index; asserts the dist-index version (10.8.2) wins over registry `latest` (99.0.1), npm/npx are linked into the prefix, the restored npm actually runs, and the sudo log stays empty.
- `test_install_node_npm_restore_falls_back_to_registry_latest` — no dist index, so registry `latest` is used and the warning is emitted.
- `test_install_node_reports_nvm_npm_restore_failure_and_tries_the_package` — your exact scenario: unreachable index and registry plus a refusing sudo, so the exit is non-zero with every step visible and the package fallback attempted.

All three fail against 5aca49f6 and pass against d5b6d3a5. `bash tests/test-build-all-runtime-deps.sh`: 45 ok / 0 not ok. `shellcheck scripts/build-all.sh` gains no findings (one pre-existing SC2086 goes away).

While in there: `test_install_node_system_warns_when_npm_is_not_in_system_path` asserted `SUDO_CALLS -eq 0` after running `install_node` in a subshell, so the counter it read was always the parent's 0 — it now counts through a file.

**Commit shape.** AGENTS.md §13 asks for `--fixup=b857581f` + autosquash rather than a separate commit for an in-PR bug. I pushed a standalone commit instead: the autosquash would have to be replayed under 5aca49f6, which rewrites the same `_ensure_npm`, so it means rewriting two already-reviewed commits and invalidating the `b857581f` / `5aca49f6` SHAs cited in the resolved threads above, and it would mark this thread outdated while it is still open. Say the word and I will squash the npm work into one commit before merge; a squash merge makes it moot either way.

scripts/build-all.sh
```diff
@@ -631,12 +631,164 @@ install_node() {
...
+
+        archive="$(mktemp "${TMPDIR:-/tmp}/anolisa-npm-XXXXXX.tgz")" || return 1
+        if ! curl -fsSL --connect-timeout 10 --max-time 180 \
+                -o "$archive" "$registry/npm/-/npm-$version.tgz"; then
+            warn "Could not download npm $version from $registry"
+            rm -f "$archive"
+            return 1
+        fi
+
+        rm -rf "$npm_dir" "$modules/package"
```
ikunkun-sys: [P2] npm restoration deletes an unrelated global package

`modules` is the existing Node prefix's `lib/node_modules`, so `$modules/package` is also the installation location of a legitimate global package named `package`; it is not a temporary directory owned by this recovery attempt. Entering npm recovery unconditionally removes that package before extracting the npm archive, then moves the extracted `package/` into `npm/`, leaving the unrelated global installation gone. I reproduced this with real Node.js 24.15.0 and npm 11.12.1: globally install a local package whose package.json declares `"name": "package"` into the disposable nvm prefix, make npm absent from its isolated PATH, and run `install_node` with a local registry/dist-index fixture and refusing sudo stub. Restoration returns 0 and npm works, but the previously installed `lib/node_modules/package` no longer exists. This changes user-owned global packages outside the npm repair target even on a successful recovery.

ikunkun-sys (changes requested):
Re-reviewed d5b6d3a5f1e14205a0588a2c1d8c7b495dc987d9. The previous nvm install-latest-npm finding is fixed: the new restoration path successfully restored npm 10.8.2 for official Node.js 20.19.5 without sudo and passed the memory dependency preflight. One new P2 issue remains: restoration deletes an unrelated globally installed package named `package`; see the inline finding.

Validation: all 45 runtime-dependency regression tests, shell syntax checks, and diff checks passed. A real Node.js 24.15.0/npm 11.12.1 fixture reproduced the global-package removal while restoration returned success; the restored npm also installed a local dependency that Node successfully loaded. No full component build or real system package installation was performed. Current CI has no failing checks; other checks are skipped.

scripts/build-all.sh
ikunkun-sys: Fixed in 81d93795a007454629bae330707a71f568e13562 (autosquashed into the npm-restoration commit, preserving its original author). The archive is now extracted directly into the npm directory with `--strip-components=1`, so restoration never removes or stages files in the sibling global `package` directory.

Extended the recovery regression test to preserve the existing global package metadata and module contents; it fails against the previous implementation and passes after the fix. All 45 regression tests and shell syntax/diff checks pass, with no new ShellCheck warnings. A real Node.js 24.15.0/npm 11.12.1 fixture also preserved the existing global installation, restored npm without sudo, and successfully installed and loaded a local dependency.

ikunkun-sys (approved):
Re-reviewed the full change at 1be9f2bb8fd573381591a766613937569caabbf0. No remaining actionable findings. The missing-npm recovery and unrelated global-package deletion findings are fixed, and all four review threads are resolved.

The squashed commit has exactly the same file tree as the previously validated 81d93795a revision. The existing validation therefore still applies: all 45 runtime-dependency tests passed, the preservation assertion failed against the buggy implementation, and a real Node.js 24.15.0/npm 11.12.1 fixture restored npm without sudo while preserving the global package and successfully installing/loading a local dependency. Shell syntax and diff checks passed, with no new ShellCheck warnings. I also checked the newer main-branch changes at the affected integration points; the build-all script and its regression tests are unchanged there, and the agent-memory Makefile updates retain the npm build dependency.

I contributed the targeted global-package preservation fix and the history squash. This approval reflects the final code review and the validation above; it is not a claim of an independent second reviewer. Full component builds were not run locally. At the latest check, CI had no failures, with some builds still running; normal CI gates must complete before merge.

PR #3202, opened 2026-09-10
feat(skillfs): cache transformed SKILL.md reads

### LINKED ISSUE #1488: [skillfs] perf: cache transformed SKILL.md reads
## Problem

A transformed virtual `SKILL.md` is currently read and processed in full for
every FUSE `read()` callback. A multi-block read therefore reopens the selected
source file, reruns the directive and OS-adapter pipeline, allocates a complete
result, and then discards everything outside the requested offset range.

`getattr` independently runs the same full pipeline to report the transformed
size. This repeats CPU and allocation work on a hot Agent read path.

## Current behavior

`open()` pins the selected `ActiveTarget` in `HandleEntry`, preventing an
already-open handle from switching between Current, Snapshot, and Hidden.
However, the handle does not retain raw or transformed bytes.

For Current, each `read()` resolves the pinned target back to the live source
path and reads its current contents. The target remains Current, but a live
edit between two block reads can make one handle observe slices derived from
different source versions. Snapshot reads are stable only because trusted
snapshots are required to be immutable.

## Expected behavior

A read-only open of a transformed virtual `SKILL.md` should capture one
complete Agent-visible result. All subsequent reads on that handle should
return slices of those same bytes. Activation or live-source changes should
affect only newly opened handles.

A bounded cross-open LRU should then allow `getattr`, `open`, and `read` to
reuse an exact transformed result when both the selected source identity and
the complete transform-pipeline identity still match.

### Per-open stable view

- Add optional transformed bytes to `HandleEntry`.
- Populate them during `open()` for `O_RDONLY` virtual `SKILL.md` files that
  run through the transform pipeline.
- Serve subsequent `read()` requests by offset/size slicing only.
- Release the bytes with the handle.
- Keep a handle opened on Current stable across live edits; a fresh open sees
  the new live content.
- Keep a handle opened on a Snapshot pinned to that snapshot with no live
  fallback.

### Bounded cross-open LRU

- Share exact transformed results among `getattr`, `open`, and later handles.
- Key entries by the selected physical source identity, pinned activation
  target, and a content-free identity for the complete pipeline.
- Include directive-stage environment identity, enabled stage order,
  OS-adapter target, and OS rule digest in the pipeline identity.
- Detect source changes with an identity strong enough for atomic replacement
  and in-place updates; do not rely on pathname alone.
- Bound the cache by both entries and bytes, and release it on unmount.
- Keep diagnostics content-free and expose hit, miss, invalidation, and
  eviction behavior for tests or metrics.

## Non-goals

- Caching `O_WRONLY` or `O_RDWR` handles.
- Changing write visibility or invalidation semantics for mutable handles.
- Caching staging, pending-install, raw, or ordinary passthrough files.
- Transforming files other than `SKILL.md`.
- Persisting transformed Skill content to disk.
- Adding model, network, or subprocess work to the read path.
- Changing activation selection or allowing Snapshot-to-live fallback.

## Security and compatibility

Per-open stability is an intentional semantic contract: an old read-only
handle may continue returning an older Current view until close, while a new
open observes the latest source. Cache reuse must never cross Skill identity,
Current/Snapshot target, snapshot version, source-file identity, target OS, or
pipeline configuration.

The LRU must contain only in-memory transformed bytes, remain strictly bounded,
and never log content. If source identity changes during cache population, the
implementation must retry safely or avoid inserting the result rather than
publishing bytes under a stale key.

## Acceptance criteria

- [ ] Multiple partial reads on one eligible handle run the full pipeline once.
- [ ] Offsets, EOF behavior, and reported transformed size remain correct.
- [ ] A Current handle keeps its captured bytes after an in-place edit and
      atomic replacement; a fresh handle sees the new content.
- [ ] A Snapshot handle remains on its selected snapshot across activation
      changes and never falls back to live source.
- [ ] Writable, staging, pending-install, raw, and passthrough handles retain
      their existing behavior and do not use the transformed-content cache.
- [ ] `getattr` and a matching read-only open can reuse one valid LRU entry.
- [ ] Changes to source identity, directive environment, enabled stages,
      adapter target, rule digest, or pinned target cause a cache miss.
- [ ] The LRU enforces deterministic entry and byte limits under concurrent
      access without cross-Skill or cross-target reuse.
- [ ] Focused tests demonstrate pipeline invocation counts and cache
      invalidation without production hot-path test counters.
- [ ] `cargo +1.86.0 fmt --all -- --check`,
      `cargo +1.86.0 clippy --workspace --all-targets -- -D warnings`,
      `cargo +1.86.0 test --workspace`,
      `cargo +1.86.0 doc --workspace --no-deps`, and `scripts/test.sh` pass on
      Linux with real FUSE.

## Related work

- #1450 defines the read-time transform pipeline.
- #1484 implements activation pinning and the initial transform stages.
- The repeated full-scan behavior and per-handle optimization were identified
  during review of #1484.

kongche-jbw:
## Why

Transformed `SKILL.md` reads rerun the full pipeline for every FUSE block and
attribute query. A live edit between blocks can also mix source versions within
one read-only handle.

## What changed

Read-only transformed opens now capture immutable bytes and serve subsequent
reads by slicing, using direct I/O to keep different handles' versions separate.
A mount-local LRU shares results across lookup/getattr and later opens, bounded
by 128 entries and 8 MiB of transformed payload. Reuse requires matching Skill,
physical source identity and content digest, activation target and complete
pipeline fingerprint;
source changes detected during population prevent insertion. The two commits
separate per-open semantics from cross-open reuse.

## Related issue

closes #1488

## User / Agent impact

An already-open Current handle keeps its captured view after live edits or
replacement, while a new open observes the new content. Snapshot handles retain
their selected bytes and never fall back to live source. Mutable, raw, staging,
pending-install, passthrough and generated skill-discover paths retain their
existing behavior.

## Risk and compatibility

Captured handles bypass kernel page caching/readahead and do not enable shared
mmap. Inode attributes describe the current target; an old handle's reads and
EOF use its captured length. Cache eviction does not revoke open handles. A separate mount-wide budget
limits captured handles to 64 MiB and 1,024 handles, conservatively charging
each handle in full even for shared payloads. New opens return ENOMEM when
either limit is reached; closing handles restores capacity. Temporary
read/transform allocations remain outside these retained-memory limits.
Every new open and attribute lookup reads and hashes the selected source bytes
to detect changes even when stat timestamps collide. Cache hits still avoid
transformation and transformed-result allocation. Reading a concurrently
written live file does not provide a transactional snapshot.
No cache content is persisted or logged, and no new dependency is introduced.

## Validation

Linux aarch64, Rust 1.86.0, real FUSE:

- `cargo +1.86.0 fmt --all -- --check`
- `cargo +1.86.0 clippy --workspace --all-targets -- -D warnings`
- `cargo +1.86.0 test --workspace`: 1,612 passed, 7 existing ignored
- `cargo +1.86.0 doc --workspace --no-deps`: passed with existing rustdoc warnings
- `RUSTUP_TOOLCHAIN=1.86.0 scripts/test.sh`: passed

Focused checks cover flat/Hermes old and new handles, partial reads and EOF,
snapshot removal, attribute/open reuse, complete pipeline identity, same-size
edits with restored mtime, identical stat identities, population races, deterministic LRU eviction,
concurrent capacity accounting and reference lifetime. A real FUSE stress
test retains eight near-8-MiB versions while making 24 edits/replacements,
checks mount-wide backpressure across Skills, and recovers admission on close. Independent agent review
of the full change found no actionable issues.

## Documentation and rollback

Updated `src/skillfs/docs/design/read-time-transform-pipeline.md` with captured
handle semantics, direct I/O compatibility, cache identity, limits and lifetime.
Renamed the filesystem parent to `src/fs.rs` and updated the crate layout guide.
No configuration or persistent-data migration is required. To roll back, revert
the cross-open cache commit followed by the per-open capture commit and remount.

### DIFF
--- a/src/skillfs/crates/skillfs-fuse/src/fs.rs
+++ b/src/skillfs/crates/skillfs-fuse/src/fs.rs
@@ -35,6 +35,7 @@ mod events;
 mod paths;
 mod policy;
 mod read_resolution;
+mod transform_cache;
 
 // ---------------------------------------------------------------------------
 // Filesystem Implementation
@@ -59,6 +60,7 @@ pub struct SkillFs {
     /// runs in-memory string work. Environment detection happens only when the
     /// directive stage is enabled.
     transform_pipeline: TransformPipeline,
+    transform_cache: transform_cache::TransformCache,
     /// View configuration loaded from skillfs-views.toml (if present).
     views_config: Option<ViewsConfig>,
     /// Optional reader-visible root for paths emitted by `skill-discover`.
@@ -213,6 +215,7 @@ impl SkillFs {
             handles: HandleManager::new(),
             inodes: InodeManager::new(),
             transform_pipeline,
+            transform_cache: transform_cache::TransformCache::default(),
             views_config,
             skill_discover_root: None,
             source_dirfd,
--- a/src/skillfs/crates/skillfs-fuse/src/fs/callbacks/read.rs
+++ b/src/skillfs/crates/skillfs-fuse/src/fs/callbacks/read.rs
@@ -3,6 +3,7 @@
 use std::os::unix::fs::FileExt;
 use std::path::Path;
 
+use fuser::consts::FOPEN_DIRECT_IO;
 use fuser::{FUSE_ROOT_ID, ReplyData, ReplyEmpty, ReplyOpen, Request};
 use tracing::{debug, warn};
 
@@ -27,6 +28,22 @@ impl SkillFs {
         reply: ReplyData,
     ) {
         debug!(ino, offset, size, "read");
+
+        if offset < 0 {
+            reply.error(libc::EINVAL);
+            return;
+        }
+        // Captured reads must survive source and namespace changes after open.
+        if let Some(content) = self
+            .handles
+            .with_handle(fh, |entry| entry.transformed.clone())
+            .flatten()
+        {
+            let start = (offset as usize).min(content.len());
+            let end = start.saturating_add(size as usize).min(content.len());
+            reply.data(&content.as_bytes()[start..end]);
+            return;
+        }
 
         let path = match self.inodes.get_path(ino) {
             Some(p) => p,
@@ -511,6 +528,9 @@ impl SkillFs {
                             if policy_decided {
                                 self.metric_policy_fallback();
                             }
+                            if matches!(&path_type, PathType::SkillMd { .. }) {
+                                physical = dir.join("SKILL.md");
+                            }
                             if let PathType::Passthrough { relative_path, .. } = &path_type {
                                 // I4: grace bypass — if the path matches the
                                 // post-publish whitelist, open from physical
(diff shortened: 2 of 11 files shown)


src/skillfs/crates/skillfs-fuse/src/fs/callbacks/read.rs
```diff
@@ -654,6 +674,54 @@ impl SkillFs {
...
+                            Some(err),
+                            None,
+                        );
+                        reply.error(err);
+                        return;
+                    }
+                };
+            let fh = self.handles.allocate(ino, flags, None, pinned_target);
+            self.handles
+                .with_handle_mut(fh, |entry| entry.transformed = Some(content));
```
yummypeng: **P1: Bound memory retained by captured handles**

The 8 MiB LRU limit only accounts for references owned by `TransformCache`; this assignment gives every eligible open handle another strong `Arc<str>` reference that is not included in that limit. An Agent can keep old handles open while repeatedly editing or atomically replacing a near-limit `SKILL.md` and reopening it. Each digest produces a new transformed payload, and LRU eviction cannot release the old versions because the handles still retain them. These virtual handles do not consume backing file descriptors, so process memory can grow until SkillFS is OOM-killed rather than naturally stopping at the usual fd limit, affecting every Agent using the mount. Please add mount-wide accounting/backpressure for handle-owned captured bytes or otherwise bound the number/size of concurrently pinned versions, and cover the repeated-version/open-handle case with a stress test.

src/skillfs/crates/skillfs-fuse/src/fs.rs
```diff
@@ -35,6 +35,7 @@ mod events;
 mod paths;
 mod policy;
 mod read_resolution;
+mod transform_cache;
```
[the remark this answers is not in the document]
kongche-jbw: Fixed in f77ab03a6: renamed the filesystem parent to src/fs.rs while retaining src/fs/ for child modules, and updated the layout documentation and callback reference. Module resolution and the public API are unchanged.

src/skillfs/crates/skillfs-fuse/src/fs/callbacks/read.rs
kongche-jbw: Fixed in bc1b0eef0: captured-handle admission now enforces a mount-wide 64 MiB and 1,024-handle budget under the handle-table lock, charging the full transformed length per handle even when its Arc is shared. Over-budget opens return ENOMEM; existing handles remain stable and release restores capacity. Added concurrent/shared/empty/oversize accounting tests and a real FUSE stress test retaining eight near-8-MiB versions while performing 24 edits/replacements, checking cross-Skill rejection and recovery after close.

yummypeng (approved):
LGTM

PR #3212, opened 2026-09-10
docs(ckpt): document OpenClaw attestation rollback boundary

### LINKED ISSUE #2812: [TestAgent][ckpt] bug: rollback 后 OpenClaw agent 硬阻断（WorkspaceVanishedError），需手动清理 attestation 才能恢复（P1）
### Component

ws-ckpt（`component:ckpt`，跨组件交互：OpenClaw workspace attestation）

### Bug Description

在 k8s sidecar 部署形态下执行一次 `ws-ckpt rollback`（核心功能）后，宿主 OpenClaw agent 的后续**所有对话**被 `WorkspaceVanishedError` 硬阻断——OpenClaw 的 workspace attestation 机制把 rollback 引起的内容重置误判为"workspace 消失"，拒绝工作，直到用户手动删除 `~/.openclaw/workspace-attestations/<hash>.attested` 才能恢复。rollback 是快照产品的招牌功能，但当前 rollback × OpenClaw 组合下**一次回滚就让 agent 瘫痪**，且报错不会让用户联想到自己刚执行过 rollback。严重程度 **P1**：核心功能组合阻断。

### Steps to Reproduce

```shell
# 前置：k8s sidecar 部署正常运行（wsckpt-daemon + OpenClaw agent 均工作正常）
WS=/root/workspace/project

# 1.（基线）agent 正常对话
kubectl exec <sidecar-pod> -c openclaw --   bash -c "cd $WS && openclaw agent --local --agent main -m 'reply one word: hi'"

# 2. 执行 rollback（ws-ckpt 核心功能，回滚到历史快照）
kubectl exec <sidecar-pod> -c openclaw --   bash -c "ws-ckpt rollback -w $WS -s <snapshot-id>"

# 3. 再次触发 agent 对话（触发 WorkspaceVanishedError）
kubectl exec <sidecar-pod> -c openclaw --   bash -c "cd $WS && openclaw agent --local --agent main -m 'reply one word: hi'"

# 4. 按报错指引清理 attestation 后恢复
kubectl exec <sidecar-pod> -c openclaw --   bash -c "rm -rf ~/.openclaw/workspace-attestations/ && cd $WS &&            openclaw agent --local --agent main -m 'reply one word: hi'"
```

### Actual Behavior

步骤 1（基线）：正常 ✅

```
Hey — I just came online for the first time. Fresh workspace, no memories yet...
[agents/agent-command] [agent] run e1c043d3... ended with stopReason=stop
```

步骤 2（rollback）：成功 ✅

```
✓ Rolled back workspace ws-86c00a to snapshot: ckpt-20260824T124949.850
```

步骤 3（rollback 后 agent）：**硬阻断** ❌

```
WorkspaceVanishedError: OpenClaw workspace appears to have disappeared after a
recent initialization: /root/workspace/project. Refusing to reseed BOOTSTRAP.md
over a recently attested workspace. Restore the workspace or remove
/root/.openclaw/workspace-attestations/86c00adc0d426ced....attested if this
reset was intentional.
```
agent 拒绝执行任何对话任务（rc=0 但无对话输出）。

步骤 4（清 attestation 后）：恢复 ✅（正常回复）

### Expected Behavior

- rollback 的设计语义是"把 workspace 内容恢复到指定快照"——作为 agent 的基础设施，回滚后 agent 应能继续正常工作（workspace 内容变了但 workspace 本身存在且健康）；
- 至少不应出现需要用户手动深挖 `~/.openclaw/workspace-attestations/` 内部文件才能恢复的硬阻断；
- 若两组件的保护假设确需冲突，应有显式的联动出口（如 rollback 时清理/重放 attestation，或 OpenClaw 侧提供一键确认交互）。

### Root Cause

两个组件的保护假设冲突（Not fully confirmed 交互细节）：
- ws-ckpt rollback 重置 workspace 内容（这是它的功能）；btrfs 子卷整体替换；
- OpenClaw workspace attestation 机制把"最近 attestation 过的 workspace 发生内容级消失"（BOOTSTRAP.md 等基线文件随回滚消失）判定为数据丢失，触发 `Refusing to reseed` 保护；
- rollback 后旧 attestation 记录仍在、内容已回滚 → 机制必然误判 → 硬阻断。

自测报告 Phase 8（rollback）与 Phase 9（L4 轮次快照）顺序执行均 PASS，未暴露本组合——推测报告环境的 attestation 状态恰好未触发（如 attestation 记录在 rollback 前不存在）。测试侧在 rollback 用例先于 agent 对话用例执行时 100% 稳定复现。

### Suggested Fix

任一方向（供产品决策）：
1. ws-ckpt 侧：rollback 感知宿主 attestation——回滚时同步清理/失效相关 attestation 记录（如 rollback hook 或对 ~/.openclaw/workspace-attestations/ 的定向处理）；
2. OpenClaw 侧：把 WorkspaceVanishedError 转为可确认的交互（"检测到 workspace 被外部重置（可能来自快照回滚），是否接受？"），而非拒绝工作；
3. 至少在 ws-ckpt 文档/CHANGELOG 中说明 rollback 后需清理 attestation 的限制。

### Environment

- OS: Alibaba Cloud Linux 4.0.5 (OpenAnolis Edition), kernel 6.6.102-7.alnx4.x86_64
- Component version: ws-ckpt 0.4.3（插件 0.4.3）+ OpenClaw 2026.7.1-2
- Package/RPM: ws-ckpt-0.4.3-1.alnx4.x86_64（镜像内）
- 部署形态: k8s sidecar（ACK 标准集群，hostPath ws-state）

### Additional Context

- 稳定性: 四步闭环 100% 稳定复现（rollback→agent 报错→清 attestation→恢复，多次实测）；基线对话与恢复后行为均正常，证明是 rollback×attestation 交互问题而非环境问题
- 相关: #2034（k8s sidecar 部署形态下暴露）；自测报告坑点 10 记录过相邻行为（agent_end 钩子不依赖模型成功）但未覆盖本组合
- 影响: rollback 是快照产品的招牌功能，当前组合下一次回滚即让 agent 停止服务；k8s sidecar 形态下用户排障成本高（需理解 OpenClaw 内部 attestation 机制）

yummypeng:
## Description

`ws-ckpt rollback` may restore OpenClaw workspace content that conflicts with recent setup or attestation state, causing `WorkspaceVanishedError`. This PR adds equivalent English and Chinese recovery guidance verified specifically against OpenClaw 2026.7.1 and 2026.8.1. For 2026.7.1, the command requires the exact absolute workspace, effective-home, and effective-state-directory paths and removes only recognized attestation marker files from the three verified locations. For 2026.8.1, the guide recommends rolling back to a known-good checkpoint without editing SQLite or using a destructive full reset.

The guide also recommends creating and recording a baseline checkpoint after a successful agent conversation. Recovery is complete when the restored workspace no longer triggers `WorkspaceVanishedError`; later provider, credential, or runtime failures are treated separately.

## Related issue

Closes #2812

## Risk and compatibility

Documentation-only change. The 2026.7.1 cleanup command validates the OpenClaw marker header, regular-file type, and maximum marker size before deletion, and exits nonzero if it removes no valid record or encounters a failure.

## Validation

- Verified the documented 2026.7.1 file-backed paths, `path.resolve` hash input, marker header, maximum size, and final-component symlink behavior against the released source.
- Exercised the 2026.7.1 command with normal records, a symlinked state directory, zero matches, relative inputs, shell metacharacters, unrelated same-name files, oversized files, and partial deletion failure.
- Exercised OpenClaw 2026.7.1 and 2026.8.1 with effective profile/state paths, empty and BOOTSTRAP-only workspaces, known-good snapshots, and agent retries. A restored workspace is considered recovered once `WorkspaceVanishedError` disappears; later runtime/provider failures are separate.
- Confirmed the English and Chinese command blocks are byte-identical and the surrounding requirements have equivalent strength.
- Did not run `openclaw reset --scope full` and did not modify OpenClaw SQLite.

## Documentation and rollback

This PR is the documentation update. Revert commit `a50f911c9` to roll it back.


kongche-jbw (commented):
Review baseline: `7ad9bb53c47dd0471465721513966a950a1fec6e...032396461ad993df1137eb5552a0af9f75458823`

**[P1] `doctor --fix` does not perform the required reset**

`docs/user-guide/en/runtime/ws-ckpt.md:225` and
`docs/user-guide/zh/runtime/ws-ckpt.md:218`

On OpenClaw 2026.8.1+, the recent attestation is stored in `workspace_setup_state` inside
`openclaw.sqlite`. The v2026.8.1 doctor flow repairs and migrates state but never deletes that
row, so `openclaw doctor --fix` leaves the guard active and the next agent run raises the same
`WorkspaceVanishedError`. In that release, the actual full reset command is
`openclaw reset --scope full`, which also removes configured workspaces, so it cannot be silently
substituted here. Please document a recovery verified on 2026.8.1+ and its destructive
prerequisites (or the 24-hour expiry fallback), and add a reproduction check that retries the
agent.

**[P2] Key the safety promise to workspace evidence, not snapshot timing**

`docs/user-guide/en/runtime/ws-ckpt.md:219` and
`docs/user-guide/zh/runtime/ws-ckpt.md:213`

OpenClaw does not know when a ws-ckpt snapshot was taken; its guard checks whether recent setup
state remains and whether recognized workspace evidence survived. A snapshot made after the
first conversation can therefore still trigger the guard if the generated bootstrap/profile,
memory, and skill evidence was removed before checkpointing. The current "always safe" and
"every later rollback" promises misclassify that reproducible state. Please state the actual
evidence-based condition and cover a post-first-conversation snapshot with that evidence removed.

docs/user-guide/en/runtime/ws-ckpt.md
```diff
@@ -202,6 +202,49 @@ receives a plain copy of the mount's contents — subsequent writes land in the
...
+  ```bash
+  openclaw doctor --fix
+  ```
+
+- OpenClaw 2026.7.x and earlier — remove the attestation record for this
+  workspace (the file name is the SHA-256 of the workspace's absolute path):
+
+  ```bash
+  HASH=$(printf '%s' "/path/to/workspace" | sha256sum | cut -d' ' -f1)
+  rm -f ~/.openclaw/workspace-attestations/${HASH}.attested
```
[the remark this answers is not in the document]
yummypeng: Valid catch — fixed in d5e8f228a (amended into the original commit). The 2026.7.x recovery command now resolves the state directory via `${OPENCLAW_STATE_DIR:-$HOME/.openclaw}` instead of hard-coding `~/.openclaw`.

While fixing this I also verified against the 2026.7.1 dist that `resolveWorkspaceAttestationPaths` checks **three** locations — the current state dir, the legacy `~/.clawdbot` state dirs, and the workspace-adjacent `<workspace>.attested` — and `findRecentWorkspaceAttestationPath` trips the guard if any of them holds a recent record. So the command now removes all three, not just the state-dir one.

kongche-jbw (commented):
Review baseline: `fd2646ab223a...d5e8f228ad12`

**[P1] Match OpenClaw's effective-home and path normalization rules**

`docs/user-guide/en/runtime/ws-ckpt.md:237` and
`docs/user-guide/zh/runtime/ws-ckpt.md:226`

The recovery still misses live records in two supported cases. With
`OPENCLAW_HOME=/srv/oc` and no `OPENCLAW_STATE_DIR`, OpenClaw resolves both its
current and `.clawdbot` directories under `/srv/oc`, while this script deletes
under `$HOME`. Also, OpenClaw hashes `path.resolve(workspaceDir)`, but this
script hashes `WS` verbatim; an absolute value ending in `/` produces a
different hash and makes `${WS}.attested` point inside the workspace. In either
case all real attestations survive and the next agent run throws the same
`WorkspaceVanishedError`. Possible direction: derive the effective home and
normalize `WS` with OpenClaw-equivalent semantics, and exercise both
`OPENCLAW_HOME` and a trailing-slash workspace path.

**[P2] Describe the actual survival-evidence predicate**

`docs/user-guide/en/runtime/ws-ckpt.md:222` and
`docs/user-guide/zh/runtime/ws-ckpt.md:214`

The documented "contains those files" predicate is not equivalent to the
guard. After setup completes, OpenClaw intentionally removes `BOOTSTRAP.md`,
and 2026.8.1 treats `openclaw-workspace-state.json` as a legacy migration input
rather than a baseline file. Normal post-conversation snapshots therefore do
not contain the listed set, yet the guard accepts them via configured-profile
or intact-generated-file evidence. This leaves the promised "exact" condition
misleading. Possible direction: describe the accepted survival evidence rather
than requiring the whole list, and verify a normal post-bootstrap 2026.8.1
workspace plus a partially retained workspace.

kongche-jbw (commented):
Review baseline: `3f1292c23105...7db1266ed153`

**[P1] Normalize every path used by the recovery command**

`docs/user-guide/en/runtime/ws-ckpt.md:250` and
`docs/user-guide/zh/runtime/ws-ckpt.md:235`

`WS="${WS%/}"` removes only one trailing slash; it does not reproduce
OpenClaw's `path.resolve`. For an allowed absolute path such as
`/srv/app/../workspace` or `/srv/workspace//`, OpenClaw hashes
`/srv/workspace`, while this command hashes a different string and leaves the
state record intact. Quoted `OPENCLAW_STATE_DIR=~/.openclaw-a` also keeps the
tilde literal, although OpenClaw expands that supported override. The next run
therefore remains blocked. Possible direction: resolve `WS`, `OPENCLAW_HOME`,
and `OPENCLAW_STATE_DIR` with OpenClaw-equivalent rules, then exercise dot
segments, repeated slashes, and a tilde-prefixed state override.

**[P1] Preserve the non-destructive expiry recovery**

`docs/user-guide/en/runtime/ws-ckpt.md:278` and
`docs/user-guide/zh/runtime/ws-ckpt.md:259`

In 2026.8.1, `clearExpiredWorkspaceStateForVanishedWorkspace` deletes expired
setup and attestation rows after their 24-hour recent windows, after which
`ensureAgentWorkspace` reseeds the workspace. A user can therefore keep the
pre-baseline snapshot, wait for the state to expire, and retry without a full
reset. Calling the destructive full reset the only option may unnecessarily
discard every workspace, credentials, sessions, and plugins. Possible
direction: document the 24-hour fallback, or remove the exclusivity claim,
and verify a retry after both timestamps have expired.

**[P2] Do not count `.git` or arbitrary files as survival evidence**

`docs/user-guide/en/runtime/ws-ckpt.md:224` and
`docs/user-guide/zh/runtime/ws-ckpt.md:216`

The 2026.8.1 full startup path does not accept `.git` or an arbitrary user file
as survival evidence: `workspaceProfileLooksConfigured` omits Git evidence,
and the new-workspace test recognizes bootstrap/profile files, memory, or
installed skills only. With a recent attestation, a snapshot containing only
`.git` or `notes.txt` still raises `WorkspaceVanishedError`, contrary to the
documented safe condition. Possible direction: list only evidence accepted by
the full startup predicate and cover both negative snapshots in verification.

kongche-jbw (commented):
Review baseline: `c28f04b83d02...90f357f7193d`

**[P1] Handle named-profile state directories in the 7.x recovery**

`docs/user-guide/en/runtime/ws-ckpt.md:270` and
`docs/user-guide/zh/runtime/ws-ckpt.md:254`

When an affected 2026.7.x agent runs as `openclaw --profile team ...`, OpenClaw startup sets
its state directory to `$OPENCLAW_HOME/.openclaw-team`. That value exists only inside the CLI
process. Running this pasted block in the parent shell therefore leaves `SD` empty and removes
records from `.openclaw`, `.clawdbot`, and the workspace-adjacent path, but not the live
`.openclaw-team/workspace-attestations/...` record. The next run remains blocked. Possible
direction: derive the state directory from `OPENCLAW_PROFILE` (including `--dev`), or require
the user to export the effective `OPENCLAW_STATE_DIR`, and cover both profile forms.

**[P2] Make the survival-evidence list match each version's predicate**

`docs/user-guide/en/runtime/ws-ckpt.md:207`,
`docs/user-guide/en/runtime/ws-ckpt.md:224`,
`docs/user-guide/zh/runtime/ws-ckpt.md:203`, and
`docs/user-guide/zh/runtime/ws-ckpt.md:216`

On 2026.8.1+, a completed workspace with no BOOTSTRAP, memory, or skills and pristine
AGENTS/SOUL/IDENTITY/USER files is accepted when their stored generated hashes still match.
The current “none of the recognized evidence” bullet says that snapshot triggers, contradicting
the following normal-post-conversation example. The version details also differ: 7.1 accepts
customized TOOLS/HEARTBEAT, while 8.1 no longer seeds those files. This can steer users away
from a usable snapshot toward waiting or a full reset. Possible direction: describe the
hash-intact generated-file branch and split the evidence wording by version; regression-check
pristine completed workspaces on both sides of the storage migration.

docs/user-guide/en/runtime/ws-ckpt.md
```diff
@@ -202,6 +202,166 @@ receives a plain copy of the mount's contents — subsequent writes land in the
...
+  `<effective home>/.openclaw-<name>` (`--dev` behaves like `--profile dev`),
+  and that choice exists only inside the agent process — exporting
+  `OPENCLAW_PROFILE` in the shell does **not** move the state directory, only
+  the command-line flag does. Rather than guess and silently clean the wrong
+  directory, the command aborts when `OPENCLAW_STATE_DIR` is unset and it
+  finds more than one `.openclaw*` state directory under the effective home.
+  In that case, export the state directory the agent actually used and rerun:
+
+  ```bash
+  export OPENCLAW_STATE_DIR="$HOME/.openclaw-team"   # agent runs `openclaw --profile team ...`
```
kongche-jbw, line 281: [P2] Could the example use the agent's actual effective state directory?

With OPENCLAW_HOME=/srv/oc and an agent started using --profile team (without a state-dir override), the live directory is /srv/oc/.openclaw-team. This example instead exports $HOME/.openclaw-team. Because an explicit override bypasses discovery, following it makes the command return 0 while leaving the real record intact. I reproduced that path/deletion behavior in an isolated fixture.

Could we show an explicitly chosen absolute path and clarify that any configured override takes precedence? That would keep the example consistent with the preceding effective-home explanation without adding more inference. The Chinese example at line 258 needs the same adjustment.

docs/user-guide/en/runtime/ws-ckpt.md
```diff
@@ -202,6 +202,166 @@ receives a plain copy of the mount's contents — subsequent writes land in the
...
+    const HASH = crypto.createHash("sha256").update(WS).digest("hex");
+    const sdOverride = (env.OPENCLAW_STATE_DIR || "").trim();
+    let SD;
+    if (sdOverride) {
+      SD = path.resolve(sdOverride.replace(/^~(?=$|[\\/])/, OC_HOME));
+    } else {
+      let candidates = [];
+      try {
+        candidates = fs.readdirSync(OC_HOME, { withFileTypes: true })
+          .filter((e) => e.isDirectory() && (e.name === ".openclaw" || e.name.startsWith(".openclaw-")))
```
kongche-jbw, line 306: [P2] Directory symlinks are omitted from the new discovery step

A .openclaw-team entry that links to a state directory is reported as a symlink by readdirSync(..., { withFileTypes: true }), so isDirectory() excludes it. With that as the only profile entry and no OPENCLAW_STATE_DIR in the recovery shell, candidates is empty and SD falls back to .openclaw. I reproduced exit code 0 with the profile's attestation still present.

Could this case be made explicit in the recovery contract? Requiring the actual OPENCLAW_STATE_DIR for profile recovery would be sufficient; there is no need to build a general profile detector for this documentation change. The corresponding Chinese code is at line 283.

docs/user-guide/en/runtime/ws-ckpt.md
```diff
@@ -202,6 +202,166 @@ receives a plain copy of the mount's contents — subsequent writes land in the
...
+    so a completed workspace that was never customized also passes.
+- 2026.7.x only: a required bootstrap file whose content no longer matches
+  the generated one (AGENTS.md, TOOLS.md, or HEARTBEAT.md).
+- 2026.8.1 and later only: a customized AGENTS.md (content differs from both
+  the generated one and the template). These versions no longer seed
+  TOOLS.md or HEARTBEAT.md, and neither file ever counts as evidence there.
+
+Two consequences:
+
+- The error triggers only when the snapshot has lost **all** of the evidence
```
kongche-jbw, line 245: [P2] The evidence list still does not cover the full startup ordering

The hash-intact addition is correct, but “only when ... all” still treats BOOTSTRAP.md as sufficient by itself. In both 2026.7.1 and 2026.8.1, a workspace containing only BOOTSTRAP.md, with a recent attestation, reaches the earlier isBrandNewWorkspace rejection before the later bootstrapExists survival check. Such a snapshot retains an item in this list and nevertheless triggers WorkspaceVanishedError.

Could we present these as examples and remove the exhaustive guarantee, or include the earlier guard's precondition? The saved setup marker is also required for the attested-generated-files branch discussed in your reply. This is a remaining pre-existing wording issue, not a regression from your hash-intact fix. The same wording appears in Chinese at lines 220 and 233.

kongche-jbw (commented):
Thanks for the detailed verification and for improving the recovery command. The generated-hash branch and the 7.1/8.1 distinction address the substance of the previous evidence-list finding. Your clarification about OPENCLAW_PROFILE is also helpful: my earlier suggestion to derive the state directory from that variable should not be used for 7.1.

Review baseline:
- Fetched base (`main`): `355d55eaa0a322f7b28d5f485135c37d14bf565f`
- Previous reviewed head: `90f357f7193d66a102ab7fd52baf941391bfa6b8`
- Current reviewed head: `969fe934f27b798aacfa182a3ed227b211896185`
- Actual merge-base for the full PR diff: `2e14494de03dff7d8521ea490a832878b83fb98c`

This follow-up covers the previous-head-to-current-head diff, both language versions, your latest reply, and interactions with the original startup checks. The two recovery-path comments below concern additions in this revision. The BOOTSTRAP-only case already existed in the previous revision; it was missed in the earlier posted review, rather than introduced by this update.

Overall, documenting the limitation and a practical recovery path remains a reasonable scope for #2812. I do not think this PR needs a ws-ckpt workaround that writes OpenClaw's internal SQLite state. The remaining corrections can stay small: make the effective state directory explicit where it cannot be reliably inferred, and describe survival evidence as examples rather than an exhaustive safety guarantee. In particular, there is no need to keep expanding directory discovery solely to accommodate these review comments.

Validation: I exercised the current recovery code in six isolated filesystem scenarios. Default, single-profile, multiple-directory refusal, and explicit-profile selection behaved as intended. The export example with a different effective home, and a profile directory represented by a symlink, both returned exit code 0 while leaving the intended record in place. The English and Chinese Node bodies match, and the incremental diff passes `git diff --check`. The BOOTSTRAP-only observation comes from the previously inspected 7.1/8.1 startup code and isolated guard checks, not a live agent session.

Please retain the distinction between these checks and the live 7.x/8.1+ recovery verification you have already planned. Before merge, it would be helpful to record the actual agent retry results and refresh the PR description's “completely avoids” wording to match the final documented limitations. At review time, the website build passed; the remaining listed checks were queued.

kongche-jbw (commented):
Review baseline: `355d55eaa0a3...969fe934f27b`

**[P2] Fail closed when the effective profile state directory is unknown**

`docs/user-guide/en/runtime/ws-ckpt.md:281`,
`docs/user-guide/en/runtime/ws-ckpt.md:306`,
`docs/user-guide/zh/runtime/ws-ckpt.md:258`, and
`docs/user-guide/zh/runtime/ws-ckpt.md:283`

With `OPENCLAW_HOME=/srv/oc`, an agent started with `--profile team` uses
`/srv/oc/.openclaw-team`, but the example exports `$HOME/.openclaw-team`.
That explicit override bypasses discovery and targets the wrong directory.
Discovery also excludes a `.openclaw-team` directory symlink because
`Dirent.isDirectory()` is false. In both cases the block exits 0 while the live
attestation remains, so the next agent run is still blocked.

Possible direction: require the actual effective `OPENCLAW_STATE_DIR` for profile
recovery, or validate every candidate and fail on discovery errors. Use the
effective-home path in the example, and cover both cases with command-level tests.

**[P2] Do not present the evidence list as a sufficient condition**

`docs/user-guide/en/runtime/ws-ckpt.md:221`,
`docs/user-guide/en/runtime/ws-ckpt.md:225`,
`docs/user-guide/en/runtime/ws-ckpt.md:245`,
`docs/user-guide/zh/runtime/ws-ckpt.md:214`,
`docs/user-guide/zh/runtime/ws-ckpt.md:216`, and
`docs/user-guide/zh/runtime/ws-ckpt.md:233`

With a recent attestation, both 2026.7.1 and 2026.8.1 classify a workspace
containing only `BOOTSTRAP.md` as brand new and reject it before reaching the
later `bootstrapExists` survival check. The intact-generated-files branch also
requires a saved setup marker. A snapshot can therefore retain an item in this
list and still raise `WorkspaceVanishedError`, contrary to the “safe as long as”
and “only when ... all” guarantees. Creating one post-conversation snapshot also
does not prevent a later rollback from selecting an older pre-conversation one.

Possible direction: present the items as examples or include the earlier guard
preconditions, and tell users to select the new snapshot or a later one. Add
7.1/8.1 regression coverage for the `BOOTSTRAP.md`-only case.

docs/user-guide/en/runtime/ws-ckpt.md
yummypeng: Fixed in a50f911c9. The command no longer derives or overrides the profile state directory: it prompts for the exact absolute effective home and effective state directory used by the agent. The guide now uses `/srv/oc` and `/srv/oc/.openclaw-team` for the `OPENCLAW_HOME=/srv/oc --profile team` example and states that an explicit `OPENCLAW_STATE_DIR` takes precedence; the Chinese section matches.

docs/user-guide/en/runtime/ws-ckpt.md
yummypeng: Fixed in a50f911c9. Discovery has been removed entirely, so a symlinked `.openclaw-team` entry is no longer omitted or misresolved: the operator supplies the exact effective state directory, and filesystem access follows an intermediate directory symlink. The cleanup also uses `lstat` and refuses to delete a final-component symlink as an attestation marker. The symlinked-state-directory fixture passed, and the Chinese command is identical.

docs/user-guide/en/runtime/ws-ckpt.md
yummypeng: Fixed in a50f911c9. The exhaustive evidence guarantee is removed. Both guides now say that OpenClaw combines workspace content with version-specific setup state and that no single file, including `BOOTSTRAP.md`, proves acceptance. Live 2026.7.1 and 2026.8.1 runs confirmed that a BOOTSTRAP-only restored workspace remains blocked; recovery is judged only by the absence of `WorkspaceVanishedError` when the restored workspace is exercised.

kongche-jbw (commented):
LGTM for the documentation scope of this PR. Thanks for working through the feedback and simplifying the recovery contract.

Review baseline:
- Fetched base (`main`): `9a4a30e3fb533f32d8e64d3d92643741d68511bb`
- Previous reviewed head: `969fe934f27b798aacfa182a3ed227b211896185`
- Current reviewed head: `a50f911c9351ba286644c9e31e5d9160c0db295e`
- Actual merge-base: `f76f76a0286a1a174f27ffd628bae6d303815aaa`

I reviewed the incremental changes in both guides, the updated PR description, and the replies to the three inline comments. The previous findings are addressed:
- Recovery requires the actual absolute workspace, effective-home, and state-directory paths, removing the incorrect shell-home assumption.
- Directory discovery is gone, so a symlinked state directory is no longer omitted. Final-component symlinks are deliberately preserved.
- The exhaustive survival-evidence guarantee is removed. The guide explains that no individual file, including BOOTSTRAP.md, proves acceptance and asks users to exercise the restored workspace.

The resulting scope is coherent: release-specific recovery guidance for 2026.7.1 and 2026.8.1, without a ws-ckpt workaround that depends on OpenClaw's private SQLite schema or a promise that the underlying integration issue has been eliminated. I found no new blocking issue in this revision.

Validation: nine isolated command scenarios passed, covering normal records at all three locations, a symlinked state directory, zero matches, relative input, shell metacharacters, unrelated files, oversized files, final-component symlinks, and an injected partial deletion failure. English/Chinese command blocks match; shell syntax and incremental diff checks passed. The marker header, size limit, and path hashing were also checked against the released 2026.7.1 source.

The author's reported live 7.1/8.1 runs and agent retries address the previously pending recovery verification. I have not independently repeated the complete ws-ckpt rollback and agent-conversation flow; my independent execution was limited to the command scenarios above. Current CI reports success, with component tests skipped for this documentation-only change.

Submitting this as a COMMENT review; formal approval will be handled separately.

kongche-jbw (approved):

PR #3250, opened 2026-09-14
test(tokenless): retire the stale PostToolUse replacement shell suite

Forrest-ly:
## What

Retires `src/tokenless/tests/test-posttooluse-replacement.sh` (331 lines) and ports its one unique assertion into the maintained Python suite.

## Why

The script has been dead weight since the external hook entry was unified behind `tokenless compress` (67f8dafa, Protocol v2, 2026-08-25). It was last touched on 2026-08-05 (13817938), before that migration, and its stub CLI still speaks the legacy contract:

```bash
case "${1:-}" in
  compress-response) ... ;;
  compress-toon) exit 0 ;;
  *) exit 0 ;;        # <- `compress` lands here: exit 0, empty stdout
esac
```

`hook_utils.run_compress()` spawns `[tokenless_bin, "compress"]` and requires a JSON envelope with `protocol_version == 2`. Empty stdout means "returned no output", so the hook fails open with `{}`. On `main` @ `429c9a19`:

```text
$ bash src/tokenless/tests/test-posttooluse-replacement.sh
...
  Summary: 7/15 passed        # exit 1
```

The 8 failures are **not** a product regression - they are the stub answering a subcommand the hook no longer issues.

Nobody noticed because the script is orphaned: no Makefile target and no workflow step references it (`grep -rn posttooluse-replacement` matches only the file itself). The `test:` aggregate runs `test-tokenless test-hooks test-integration test-hook-parity test-adapters test-claude-code-detect test-raw-package test-npm-package test-toon-cleanup` - this file is in none of them, so it kept looking like coverage while asserting a contract the hook stopped speaking 3 weeks ago.

## Coverage is not lost

`tests/test_compress_response_hook.py` - which CI does run, via `make -C src/tokenless test-integration` in the `Test tokenless` job - already exercises the same surface end to end: the hook is spawned as a subprocess with a mock `tokenless` / `claude` on `PATH` and an isolated `HOME`.

| shell suite section | Python equivalent |
| --- | --- |
| claude-code >= 2.1.121 replaces via `updatedToolOutput`, no duplicate in `additionalContext`, schema kept | `TestReplacementProtocol` (`test_claude_code_uses_updated_tool_output` + replacement-semantics tests) |
| old claude version fails open (known versions) | `test_no_savings_and_old_claude_do_not_emit_replacements` |
| **failed `claude --version` probe fails open** | **was not covered here - lands in #3301 as `TestClaudeVersionProbeFailure`** |
| non-beneficial compression passes through for every agent | `TestPassthrough` |
| qoder-cli replaces with JSON *text* | `test_qoder_cli_uses_updated_tool_output` |
| generic agents pass through; `additionalContext` is reserved for env-attribution diagnostics | `TestNonReplacementAdapters` |
| version-skewed CLI response fails open | `test_version_skewed_response_fails_open` |
| one `tokenless` subprocess per invocation | `TestSingleSubprocess` |
| **6.1 version cache is 0600 under a 0700 `~/.tokenless`** | **new: `TestClaudeVersionCacheHardening`** |

Only the last row was unique to the shell script, so it is ported (with the shared-HOME rationale spelled out) instead of dropped.

## Testing

```text
$ python3.12 src/tokenless/tests/test_compress_response_hook.py
Ran 35 tests in 3.806s
OK                      # 34 before this change
```

`make -C src/tokenless test-integration` runs the same file; the suite self-skips below Python 3.9. No workflow files are touched.

## Follow-up (not in this PR)

`src/tokenless/tests/test-toon-full.sh` is the only other script with no Makefile or CI reference. It is not stale in the same way - it validates a *deployed* install (OpenClaw plugin config present, COSH PostTool hook installed, `rtk-rewrite` / `schema-compression` / `response-compression` / `toon-compression` activated), so on a clean checkout it reports 26/35 here. Wiring or retiring it needs a separate decision on whether it is a release-verification script or a test.


kongche-jbw (approved):
LGTM. Retiring the orphaned shell suite is reasonable: its mock still implements the legacy CLI subcommands, while the hook now uses `tokenless compress` with Protocol v2. Moving the cache permission assertion into the maintained Python suite brings that check under the existing CI integration target without changing production behavior.

Validation against 6658ad6717cf10bb86948b570aa8e75ed734da2d:
- Python 3.12.3: all 35 tests in `test_compress_response_hook.py` passed.
- Reproduced the obsolete shell suite's 7/15 result.
- Independently widened the cache file and directory modes to 0644 and 0755 in a temporary snapshot; the new test caught both regressions.
- `git diff --check` passed; the current Test tokenless CI check passed.

Non-blocking: the coverage table slightly overstates equivalence. The Python suite checks known old/new Claude versions but does not exercise a failed version probe; adding that case would close an existing coverage gap. The generic-agent row should also describe current passthrough behavior, with additionalContext reserved for diagnostics, rather than preservation of the old additive compression fallback. Neither point indicates a runtime regression introduced by this PR.

Forrest-ly:
Follow-up on the two non-blocking points from the review. This PR merged as `7f99c69f` before they could be answered on-branch, so there is no new head SHA here: point 1 lands as a separate PR, point 2 is a description correction.

**1. "the coverage table slightly overstates equivalence ... does not exercise a failed version probe" - accepted, and it was a real gap rather than a wording problem.**

`_cached_claude_version()` has three ways to come back empty - the CLI exits non-zero, it prints something with no `x.y.z` in it, or the PATH entry resolves but cannot be spawned - and all three feed `_claude_supports_replacement() -> False`, which is the fail-open half of the 2.1.121 gate. The suite only ever fed the probe a mock answering with a *known* version (`2.1.120` / `2.1.210` / `2.1.259`), so that half was asserted by reading the code rather than by a test.

#3301 adds `TestClaudeVersionProbeFailure`: one broken mock per branch, asserting per branch that the hook exits clean and emits `{}`, declares `replace_output=false` while `replace_with_text` stays `true` (the payload was otherwise fully eligible, so the version gate is the only thing holding it), spawns exactly one `tokenless` subprocess, warns that the gate is off, and leaves `~/.tokenless/.claude-version` unwritten so a transient probe failure cannot pin a stale verdict. The payload is compressible and the mock *would* replace it, so the test cannot pass vacuously - both regressions it guards were injected into a scratch copy of the hook and each was caught:

| injected regression | result |
| --- | --- |
| `_claude_supports_replacement()` returns `True` when the version is `None` | FAIL, all 3 subtests (`updatedToolOutput` emitted) |
| `_cached_claude_version()` caches even when `parse_version()` returns `None` | FAIL, `unparseable` subtest (`.claude-version` written) |

A hanging probe shares the `except Exception` branch with the unspawnable case, so it is documented in the helper rather than exercised at a 5 s cost.

Validation on that branch (Linux x86_64, CPython 3.12.9; CI runs 3.11 and the file self-skips below 3.9, no post-3.9 syntax used):

```text
Ran 2 tests in 4.563s
OK
==> Testing hook integration (Python)...
Ran 36 tests in 4.086s        # test_compress_response_hook.py, was 35
OK
... 13 further Python suites ...
# 15 suites, 192 tests, 0 failures, 0 skips
```

**2. "the generic-agent row should also describe current passthrough behavior" - accepted.**

`main()`'s fall-through branch sets `can_replace = False` for additionalContext-only hosts, and `_emit_attribution_or_skip()` only ever puts env attribution into `additionalContext`. There is no additive compression fallback left to preserve, so the old row described a contract that no longer exists. `TestNonReplacementAdapters` already asserts the current behaviour (passthrough, attribution retained, no `updatedToolOutput`) - the test was right and the row was wrong.

Both rows in the description above are now corrected: the version row no longer claims "undetectable", the failed-probe case is split onto its own line pointing at #3301, and the generic-agent row reads as passthrough with `additionalContext` reserved for diagnostics.

PR #3228, opened 2026-09-11
fix(anolisa): release displaced plugin tool names on adapter enable

Forrest-ly:
## Why

`anolisa adapter enable agent-memory openclaw` installs and activates `memory-anolisa`, but OpenClaw's bundled `memory-core` stays loaded — the loader exempts it from memory-slot exclusivity — and OpenClaw's plugin tool registry is first-wins. So the adapter's own `memory_get` / `memory_search` are dropped (`plugin tool name conflict (memory-anolisa): memory_get`) and reads answer from `memory-core`, which reports `disabled:true` for every `~/.anolisa/memory` path (#3218).

`verify_runtime` passes, because `memory-anolisa` really is `loaded`; nothing in the enable path observes the name collision. Only the bundle's `install.sh` performed the hand-off, and `OpenClawDriver::apply_enable` builds its own argv and never runs that script — so the documented adapter-manager entry point left the bug fully reproducible (#3225).

## What changed

Declare the hand-off in the component contract and let the built-in driver execute it:

- **Contract** — `[adapters.openclaw].displaces` names the bundled plugins an adapter displaces, each with the exclusive slot the plugin re-takes when restored. Publishing metadata only, like `[[adapters.openclaw.config]]`: every command still comes from the driver.
- **No self-displacement, no duplicates** — a contract may not displace the adapter's own plugin, checked **twice**: once against the manifest's `plugin_id`, and again against the id the driver resolved from the bundle. That manifest key is optional, so the first check on its own compares against `""`; without the second, a self-displacing contract would install and verify its own plugin, disable it, and still report success with an `Enabled` receipt. It also may not name one plugin twice, or promise one exclusive slot to two plugins. Each entry is well-formed on its own, so only a cross-entry check can see either, and nothing downstream sees two of them at once until the receipt is already on disk: a repeated id persists a receipt its own driver cannot consume (a `cleanup_failed` record that neither `status` nor `disable` can read), and a repeated slot fails *silently* — on disable the first restore makes that plugin the slot's owner, the guard then reads it as a **third** owner for the second one and steps aside, removing the receipt while the second plugin stays disabled by this adapter. All three gates run before any host call, on the dry-run and the real path alike.
[...]

PR #3266, opened 2026-09-15
fix(memory): fall back to a writable session dir instead of losing it

Forrest-ly:
## What

`start_session` used the configured session directory unconditionally and propagated the first `create_dir_all` error. The shipped default that lands there — `/run/anolisa/sessions`, which is also the exact value the OpenClaw plugin forwards as `MEMORY_SESSION_DIR` on every spawn — is only writable by root, while the server always runs unprivileged. So on a stock install the session log never existed.

Resolve the base through a preference chain instead and take the first candidate that can be created **and** written to.

## Why this is a bug, not a configuration preference

| Install path | `/run/anolisa/sessions` state | Non-root server |
| --- | --- | --- |
| RPM | created `0700 root:root` by `config/systemd/anolisa-memory-tmpfiles.conf` (system instance, `%tmpfiles_create` in `%post`) | cannot traverse the parent (no `x` for others), cannot create `<sid>` |
| `make install` / container / dev box | absent — no tmpfiles snippet is installed, and `/run` is `drwxr-xr-x root root` | `create_dir_all` fails on the first component with EACCES |
| `systemctl --user anolisa-memory@$USER` | as above | unit declares `RuntimeDirectory=anolisa/sessions/%i` but never tells the server about it, and `ReadOnlyPaths=/` makes it unwritable from inside the namespace anyway |

Failure was silent: `MemoryService::new` degraded to `session = None` behind a single `warn!` in the child's stderr, and every `mem_promote` returned `SessionUnavailable` while every `mem_session_log` returned `NotImplemented`. Session-scoped consolidation had no log to read either. `mem_promote` is the capability the plugin's `sessionId` pinning exists to support (see `a3e641a3`), so this removes the one property the user guide says a fixed session id is for.

## Fix

`resolve_session_base` tries, in order, and takes the first usable candidate:

1. the configured directory (`resolved_session_dir()`), unchanged;
2. `$XDG_RUNTIME_DIR/anolisa/sessions` — already `0700` and user-owned, and what the user unit's `RuntimeDirectory=` creates;
3. `<$TMPDIR|/tmp>/anolisa-sessions-<uid>` — uid-suffixed so two users on one host never share a base.

A fallback is reported with `warn!` naming both directories; only when every candidate fails does the service degrade, and the aggregated error then lists each one with its reason. Treating the configured value as a *preference* rather than a requirement is what makes this work for every caller — including the plugin, which always sends an explicit `MEMORY_SESSION_DIR`, so changing only the compiled-in default would not have reached it.

Two details worth calling out:

- **Existence is not a usable probe.** A pre-existing read-only directory passes `create_dir_all` and then fails on the first session, so each candidate gets a real create-and-remove.
- **Directories we pick are hardened; the operator's are not.** Chosen candidates must be non-symlinks owned by us and are created `0700`. Without that, any local user could plant `/tmp/anolisa-sessions-<victim uid>` as a symlink and redirect the probe, every session root, `log.jsonl` and the `mem_promote` source tree into a directory of their choosing. The configured directory is exempt from those two checks so a symlink or a group-shared mount there keeps working.

The user unit is fixed alongside: `Environment=MEMORY_SESSION_DIR=%t/anolisa/sessions/%i` makes the `RuntimeDirectory=` it already creates live, and `%t/anolisa` is added to `ReadWritePaths` (`ReadOnlyPaths=/` would otherwise make it read-only inside the namespace). `/run/anolisa` is left listed for operators who point `MEMORY_SESSION_DIR` back at it.

## Verification

Driven over stdio exactly as the plugin does (`initialize` → `notifications/initialized` → `tools/call mem_session_log`), non-root uid 1359527, no RPM installed, `/run/anolisa` absent, `MEMORY_SESSION_DIR=/run/anolisa/sessions`:

| build | `mem_session_log` | server stderr |
| --- | --- | --- |
| `main` @ `69b6869a` | `isError: true` — `session log unavailable; check MEMORY_SESSION_DIR / /run/anolisa permissions` | `WARN session log unavailable (io: Permission denied (os error 13)); mem_promote / mem_session_log will return errors` |
| this branch | `isError: false` — `(session log is empty)` | `WARN session dir /run/anolisa/sessions is not usable by uid 1359527; using /run/user/1359527/anolisa/sessions instead (set MEMORY_SESSION_DIR to override)` |

With `XDG_RUNTIME_DIR` unset the same fixed build falls through to `$TMPDIR/anolisa-sessions-1359527` and still answers `isError: false`.

Suites:

| Gate | Result |
| --- | --- |
| `cargo test --locked` | **353 passed / 0 failed**, 16 suites (lib 191 → 198) |
| `cargo fmt --all --check` | clean |
| `cargo clippy --all-targets --locked -- -D warnings` | clean |
| `systemd-analyze verify anolisa-memory@.service` | identical warning set before and after (only line numbers shift); `%t/anolisa` accepted as absolute |

New `service::session_base_tests` (7): preference order, uncreatable configured dir, existing read-only dir, symlink refused for chosen dirs but honoured for configured ones, `0700` on a created fallback, all-candidates-failed error text, and the shape of the candidate chain.

`tests/session_test.rs::session_log_degrades_gracefully_when_session_dir_unavailable` pinned the old behaviour with `assert!(svc.session.is_none())` and now fails by design. It is rewritten as `unusable_session_dir_falls_back_instead_of_losing_the_session`, which keeps the "service still builds" half of the old contract and adds: session root is outside the unusable path, and `mem_promote` + `mem_session_log` work end to end through the fallback. **Reverting only `src/service/mod.rs` makes it fail**, so it reproduces the bug rather than describing it.

## Scope

Three files, no overlap with the in-flight `component:memory` PRs — #3228 (`anolisa-core` adapter), #3238 (plugin `config.ts` / `index.ts` / manifest / user guide), #3256 (`Makefile`), #3222 (adapter `install.sh`).

Deliberately left for follow-ups:

- `config/systemd/anolisa-memory-tmpfiles.conf` still creates `/run/anolisa{,/sessions}` root-only. That is now *unused* rather than broken; changing its mode is a security-relevant packaging decision that should not ride along here.
- `docs/user-guide/{en,zh}/token-saving/agent-memory.md` and the plugin's `DEFAULT_SESSION_DIR` / `sessionDir` descriptions still present `/run/anolisa/sessions` as the effective location. Both files are being edited by #3238, so touching them here would only create a conflict.

Related to Multica issue AGE-7763.


src/agent-memory/src/service/mod.rs
```diff
@@ -413,9 +413,170 @@ impl MemoryService {
...
+
+    if !md.is_dir() {
+        return Err(MemoryError::Other(format!(
+            "{} is not a directory",
+            dir.display()
+        )));
+    }
+    if ours {
+        let me = nix::unistd::Uid::current();
+        if md.uid() != me.as_raw() && !me.is_root() {
```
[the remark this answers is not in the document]
Forrest-ly: Confirmed, and fixed in `a13fdab5`. The exemption was in the comparison itself — `md.uid() != me.as_raw() && !me.is_root()` — so a root server accepted a fallback owned by anybody. `/tmp/anolisa-sessions-0` is exactly the predictable name in a world-writable directory that makes that worth attacking: whoever plants it owns the base, therefore owns every `<sid>` root the server builds inside it, and can swap a known `MEMORY_SESSION_ID` entry for a symlink that the server's own `create_dir_all` / chmod / metadata writes then follow.

The check is now `fallback_owner_is_us(owner, me) = owner == me`, compared against the effective uid with no root branch, and the message says the directory is being refused rather than only naming the mismatch. The operator-configured candidate stays exempt (`ours == false`) — that one is the operator's own choice, and the doc comment on `probe_session_base` now spells out which three rules it is exempt from and which one it is not.

Pinned by two tests:

- `fallback_owner_check_has_no_root_exemption` asserts the predicate on all four `(owner, me)` combinations, `(1000, 0)` included. Reverting to `owner == me || me.is_root()` fails it.
- `rejects_a_foreign_owned_fallback_even_as_root` covers the wiring end to end through `pick_session_base`: it `chown`s a planted `anolisa-sessions-0` to `65534` numerically (so it does not depend on `nobody` existing in the image) and asserts the chain falls through to the next candidate. It self-skips when the runner is not root, which is the case on `ubuntu-latest`, so the predicate test above is the one guarding the regression in CI — hence extracting the comparison rather than leaving it inline.

src/agent-memory/src/service/mod.rs
```diff
@@ -413,9 +413,170 @@ impl MemoryService {
...
+            )));
+        }
+    }
+
+    let probe = dir.join(format!(
+        ".anolisa-probe-{}-{}",
+        std::process::id(),
+        PROBE_SEQ.fetch_add(1, Ordering::Relaxed)
+    ));
+    std::fs::write(&probe, [])?;
```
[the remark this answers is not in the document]
Forrest-ly: Confirmed, and fixed in `a13fdab5`. `std::fs::write` is `O_CREAT|O_TRUNC` with no `O_EXCL`, so it followed the planted link and truncated the target. The name is `pid` plus a counter that starts at 0 in every process, so a fresh server's first probe is guessable, and — as you note — the operator-configured base is explicitly permitted to be group- or world-writable, which is precisely the case where a neighbour can plant it.

The probe moved into `probe_writable(dir, seq_base)` and now opens with `create_new(true)` (`O_CREAT|O_EXCL`) and `mode(0o600)`, so it refuses to open, let alone follow, whatever already occupies the name. Handling of a taken name splits in two:

- a **symlink** rejects the candidate outright, because somebody is actively racing us for this base and a file we do not own is not ours to unlink;
- a **regular file** is treated as a stale probe from a killed server, so the loop takes the next name, bounded by `PROBE_ATTEMPTS = 8`.

`PROBE_SEQ` now reserves that whole block per call (`fetch_add(PROBE_ATTEMPTS, Relaxed)` instead of one per attempt), so a retry can never step onto a name a concurrent prober was just handed. The `O_EXCL` probe applies to configured and fallback candidates alike; only the symlink / owner / mode rules on the base itself are the operator's business.

Pinned by two tests, both of which fail if `create_new(true)` is swapped back for `create(true).truncate(true)`:

- `probe_refuses_to_follow_a_planted_symlink` plants a link at `.anolisa-probe-<pid>-7` pointing at a file holding `precious`, calls `probe_writable(&base, 7)`, and asserts the error names the symlink, the target still reads back `precious`, and the link is still there.
- `probe_steps_over_a_stale_file_and_removes_only_its_own` covers the other half: a stale regular file at the first name survives untouched, the next name is used for the real probe, and only our own probe is removed. Under the old `fs::write` this failed with `NotFound` on reading the stale file, because the probe had truncated and then deleted it.

`skips_an_existing_read_only_dir` still asserts no `.anolisa-probe-*` is left behind in a directory the probe rejected.

src/agent-memory/src/service/mod.rs
```diff
@@ -413,9 +413,170 @@ impl MemoryService {
...
+    let md = match std::fs::metadata(dir) {
+        Ok(md) => md,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            std::fs::create_dir_all(dir)?;
+            if ours {
+                // Fallbacks can land under a world-writable /tmp. Each
+                // `<sid>` is already forced to 0700 by
+                // `SessionLogService::start`, but the base should not be
+                // listable by other users either.
+                let _ = std::fs::set_permissions(dir, std::fs::Permissions::from_mode(0o700));
```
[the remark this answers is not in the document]
Forrest-ly: Confirmed, and fixed in `a13fdab5`. The `let _ =` swallowed the chmod error and the candidate was used anyway, so a base could sit at its umask-derived mode while holding session data. It was narrower than that too: the tightening only ran in the `NotFound` arm, so a fallback created before this hardening was never tightened at all.

The permission work now lives in the `ours` branch *after* the owner check, so we never chmod a directory we do not own:

- if `mode() & 0o077 != 0`, chmod to `0700` and propagate failure as a candidate rejection (`cannot tighten <dir> to 0700: <e>`), which makes `pick_session_base` move on to the next fallback and, failing that, surface it in the aggregate `no writable session directory; tried …` message;
- the mode is then re-`stat`ed and the candidate rejected if it is *still* loose. Some FUSE and network mounts return success from `chmod` without applying it, which reaches the same "silently listable base" outcome you are describing with no error left to propagate.

Only widening is guarded — a base the operator deliberately made `0500` is left at `0500`.

Pinned by two tests:

- `tightens_a_pre_existing_fallback_to_0700`: a pre-existing `0755` fallback ends up `0700`. Restoring the `let _ =` in the create arm and dropping the block leaves it at `0755` (`493 != 448`) and fails.
- `leaves_an_operator_configured_dir_mode_alone`: an operator's `0770` configured directory is still `0770` afterwards, so the tightening did not leak onto the exempt path.

`creates_a_fallback_0700` still covers the fresh-create case. A chmod that *errors* is not injectable from the unit-test sandbox — it needs a read-only mount or a foreign owner, and the foreign-owner case is now rejected earlier by design — so that arm rests on `?` propagation rather than on a test; flagging it rather than leaving the gap implied.

Verified on the branch: `cargo test --locked` 359 passed / 0 failed (6 new), `cargo fmt --all -- --check` and `cargo clippy --all-targets --locked -- -D warnings` both clean.

kongche-jbw (commented):
Reviewed the full merge-base-to-head diff, the e334264d -> a13fdab5 follow-up, affected callers, tests, and all existing review threads and replies. The three previously reported issues have direct fixes, but three additional findings remain below. All three were already present in e334264d; they are gaps in the previous review, not regressions introduced by a13fdab5.

Baseline: declared base 2e1ee6585d42eb4dc2f6b5756136e5734c94c4c1; merge base 69b6869a98b04c45e048926e947121216024ceb1; reviewed head a13fdab5181c5a370f7f6d112eb7707f896982e5. The current agent-memory CI check passed.

Validation was primarily static. `git diff --check` passed. A bounded Python simulation in a private temporary directory confirmed the filesystem-operation interleaving described in the first finding; this was not a cross-user exploit or an end-to-end execution of the Rust server. No services, namespaces, mounts, or network/security configuration changes were performed. The temporary directory was removed and its absence verified.

src/agent-memory/src/service/mod.rs
```diff
@@ -413,9 +413,255 @@ impl MemoryService {
...
+/// every session root into a directory of their choosing. The
+/// operator-configured directory is exempt from those three: a symlink, a
+/// shared owner or a group-writable mount there is the operator's own
+/// decision. It still gets the `O_EXCL` probe, because "the operator chose
+/// a world-writable location" is not "the operator chose to let neighbours
+/// clobber our files".
+fn probe_session_base(dir: &Path, ours: bool) -> Result<()> {
+    use std::os::unix::fs::{MetadataExt, PermissionsExt};
+
+    if ours && std::fs::symlink_metadata(dir).is_ok_and(|md| md.file_type().is_symlink()) {
```
kongche-jbw: [P1] Bind fallback validation and use to the same directory

The symlink check and the following `metadata(dir)` resolve the pathname separately. A local user can pre-create the predictable fallback directory under `/tmp`, let `symlink_metadata` observe that ordinary directory, then rename their directory and replace it with a symlink to an existing directory owned by the server user before `metadata` runs. The latter follows the symlink, so the ownership and mode checks can pass. `create_new(true)` protects only the final probe filename, not the directory components, and `SessionLogService::start` subsequently follows the redirected base for directory creation and metadata writes. This bypass remains after the owner-check fix. A private-directory simulation of that exact call interleaving accepted the redirected base and wrote the session beneath its target. Validation and subsequent session operations need to remain anchored to the same directory opened without following symlinks; another pathname check alone leaves the race.

src/agent-memory/src/service/mod.rs
```diff
@@ -413,9 +413,255 @@ impl MemoryService {
...
+    let mut out: Vec<(PathBuf, bool)> = vec![(configured.to_path_buf(), false)];
+
+    if let Ok(xdg) = std::env::var("XDG_RUNTIME_DIR") {
+        let xdg = xdg.trim();
+        if !xdg.is_empty() {
+            out.push((Path::new(xdg).join("anolisa").join("sessions"), true));
+        }
+    }
+
+    let uid = nix::unistd::Uid::current().as_raw();
```
kongche-jbw: [P2] Preserve the host UID for the temporary-directory suffix

This UID is read after the default `auto` strategy may have entered a user namespace: `main::early_enter_userns` runs before service construction, and `LinuxUserNsMount::enter` installs the mapping `0 <host uid> 1`. Consequently, different host users all compute `/tmp/anolisa-sessions-0` when the configured base is unusable, XDG_RUNTIME_DIR is absent, and TMPDIR is unset/shared. The first user's directory is then foreign-owned from the next user's namespace, so the new ownership check rejects the last candidate and that user still loses the session. Use a host identity captured before namespace entry for the path suffix, while keeping ownership validation in the namespace appropriate to the filesystem metadata.

src/agent-memory/tests/session_test.rs
```diff
@@ -235,16 +242,24 @@ fn session_log_degrades_gracefully_when_session_dir_unavailable() {
...
+    // Both session-dependent tools work end to end through the fallback.
+    std::fs::write(session.scratch_root().join("draft.md"), "fallback scratch").unwrap();
+    let n = svc.promote("draft.md", "notes/from-fallback.md").unwrap();
+    assert!(n > 0);
+    assert!(svc.mount.root.join("notes/from-fallback.md").exists());
+    svc.session_log().unwrap();
 
-    let err = svc.promote("x.md", "y.md").unwrap_err();
-    assert!(matches!(err, MemoryError::NotImplemented(_)));
+    let _ = std::fs::remove_dir_all(session.root());
```
kongche-jbw: [P2] Isolate the fallback directory and session ID in this integration test

Only the configured base and persistent store are confined to this test's temporary directories. The fallback still comes from the real XDG_RUNTIME_DIR/TMPDIR, and `start_session` reads the inherited MEMORY_SESSION_ID directly. If tests run with an ID that already exists in that fallback, service construction reopens that session, the test writes `scratch/draft.md`, and this cleanup recursively removes the entire pre-existing session, including unrelated scratch data. Even without a pinned ID, the test probes and may chmod the user's real fallback base. Run this case in an isolated child process with a test-owned runtime/tmp directory and a test-specific session ID so creation and cleanup cannot reach a live session.

src/agent-memory/src/service/mod.rs
Forrest-ly: Confirmed, and fixed in `d259a8e2`. You are right that the owner-check fix did not touch this: `symlink_metadata` and `metadata` were two independent pathname resolutions, and `create_new(true)` only ever protected the probe's own filename, never the directory components.

The check is now a single resolution. `probe_session_base` opens the candidate once — `O_RDONLY|O_DIRECTORY|O_CLOEXEC`, plus `O_NOFOLLOW` for the candidates we pick — and every check after that is an `fstat`/`fchmod` on the descriptor it got back:

- ownership and mode come from `fstat(fd)`, so there is no second lookup left to race against;
- the `0700` tightening is `fchmod(fd)`, which cannot be redirected at all;
- the writability probe is `openat(fd, ".anolisa-probe-*", O_CREAT|O_EXCL|O_WRONLY|O_NOFOLLOW|O_CLOEXEC, 0600)` followed by `unlinkat(fd, …)`, and the "is this a squatter's symlink" test is `fstatat(fd, name, AT_SYMLINK_NOFOLLOW)`;
- when the candidate does not exist yet, `create_dir_all` runs and then we **re-open** instead of re-checking — `O_NOFOLLOW` on that second open is what catches a swap that lands during the mkdir.

The descriptor is what gets returned. The new `SessionBase` (`src/agent-memory/src/session/base.rs`) owns it and hands out `/proc/self/fd/<n>` as the path to use; the kernel resolves that magic symlink to the descriptor's *inode* rather than re-walking the pathname, so the rename-then-symlink swap you reproduced no longer redirects anything built afterwards. `SessionBase::from_fd` verifies the dev/ino agreement once at construction, so a host without a usable `/proc` fails there with an explanation instead of turning into a mysterious `ENOENT` partway through session setup. `pick_session_base` / `resolve_session_base` return the `SessionBase`, `SessionLogService::start_in` keeps it alive next to the paths derived from it (dropping it would dangle every one of them), and `<sid>` itself is created with `mkdirat` on the base descriptor and then verified with `fstatat(…, AT_SYMLINK_NOFOLLOW)`, so a planted name cannot smuggle in a symlink either.

The operator-configured candidate still honours a final symlink — that part is the operator's own decision — but it is likewise resolved exactly once, into the descriptor the winner carries.

Pinned by three tests:

- `validation_and_use_stay_anchored_when_the_pathname_is_swapped` is your interleaving end to end: validate, `rename` the base away, `symlink` a server-owned directory into its place, then build the session through the returned handle. It asserts the session's canonicalised root lands under the renamed directory and that the redirect target stays empty.
- `probe_writes_into_the_validated_directory_not_the_swapped_name` occupies the predictable probe name *inside the swapped-in target* with a symlink. A probe that resolved the pathname would hit it and refuse, so the `.unwrap()` succeeding is the assertion.
- `a_symlink_to_a_directory_we_own_is_still_refused` covers the static shape — `metadata()` would have followed the link and reported a directory we own with a mode we chose.

One behaviour note worth flagging: a directory can only be opened read-only, so the base is now opened `O_RDONLY|O_DIRECTORY`. A session base the server cannot even list is therefore rejected where it previously might have scraped through; write access, which is what actually matters, is still proven by the `O_EXCL` probe.

Related API note: `root()` returns the anchored path and is documented as valid only while the service is alive; `display_root()` is the operator-facing pathname and is what the `serve` startup log now prints. `end_discard_removes_dir` / `end_keep_preserves_dir` assert on `display_root()` because `end()` consumes the service and closes the descriptor.

src/agent-memory/src/service/mod.rs
Forrest-ly: Confirmed, and fixed in `d259a8e2`. Exactly the failure mode you describe: `main::early_enter_userns` runs first, the mapping is `0 <host uid> 1`, so `Uid::current()` is 0 for every user on the box by the time the chain is built — and the ownership check added in the last round then rejects the directory the first user created, so the last candidate fails and that user loses the session anyway.

New `crate::host` (`src/agent-memory/src/host.rs`):

- `capture_host_uid()` records the launching uid in a `OnceLock` — first call wins, so it is idempotent. It is called from `main` before `early_apply_cgroup` / `early_enter_userns`, from `LinuxUserNsMount::enter` immediately before the `unshare` (for library consumers that never go through `main`), and at the top of `MemoryService::new` before `pick_strategy` gets a chance to unshare.
- `host_uid()` returns the captured value; when nothing was captured it recovers the answer from `/proc/self/uid_map` (`<inside> <outside> <count>`, mapping the current uid back through whichever range covers it) rather than silently handing back the namespace uid.
- `session_base_candidates` is now a thin wrapper over `candidates_for_host_uid(configured, host_uid())`, so the naming rule is testable without entering a namespace.

On the second half of your point: the ownership validation deliberately keeps using `Uid::current()`. That is the namespace `st_uid` is reported in — inside our own userns a base we own on `/tmp` shows 0 and a neighbour's shows the overflow uid — so comparing it against the host uid would reject every fallback we own. The split is now "host uid for the *name*, namespace uid for the *metadata*", and `fallback_owner_is_us`'s doc comment says so explicitly and cross-references `candidates_for_host_uid`.

Tests:

- `tmp_fallback_is_named_after_the_host_uid_not_the_namespace_uid` asserts `candidates_for_host_uid(…, 1000)` and `(…, 1001)` yield `anolisa-sessions-1000` / `anolisa-sessions-1001` and therefore never collide on a shared `TMPDIR`.
- `the_live_candidate_chain_uses_the_host_identity` pins the wrapper to `host_uid()`.
- In `src/host.rs`: `recovers_the_host_uid_from_a_single_id_mapping` (the exact map `enter` writes), `is_the_identity_outside_a_user_namespace`, and `handles_a_multi_range_map_and_ignores_junk`, which also covers the unmapped-uid case returning `None`.

Verified end to end on this host: `agent-memory serve` as a non-root user, inside the userns (`entered userns : true`), with `XDG_RUNTIME_DIR` unset and `MEMORY_SESSION_DIR=/run/anolisa/sessions` → falls back to `<$TMPDIR>/anolisa-sessions-1359527`, i.e. the host uid, not 0, base created `0700`, session tree written, and the startup log reports the operator-facing path.

src/agent-memory/tests/session_test.rs
Forrest-ly: Confirmed, and fixed in `d259a8e2`. Both halves were real: the fallback came from the inherited `XDG_RUNTIME_DIR` / `TMPDIR`, so the test probed — and could `chmod 0700` — the developer's real runtime dir; and `start_session` reads the inherited `MEMORY_SESSION_ID`, so an id that already had a session in that fallback would be reopened, have `scratch/draft.md` written into it, and then be recursively deleted by the cleanup at the end of the test.

The case now re-executes itself in a child process. `unusable_session_dir_falls_back_instead_of_losing_the_session` checks a marker env var: unset → `fallback_parent()`, set → `fallback_child()`.

The parent creates four tempdirs and gives the child a completely private environment — `XDG_RUNTIME_DIR`, `TMPDIR`, `MEMORY_SESSION_ID=ses_fallback_child`, plus the store dir, the blocking file and the expected fallback base as `ANOLISA_TEST_*` — then spawns `current_exe() --exact <test name> --nocapture`. Nothing inherited is used, so nothing inherited can be damaged.

It also plants a decoy: `<runtime>/anolisa/sessions/ses_fallback_decoy/scratch/keep.md` at `0700`, i.e. a live session sitting in the very base the child is about to fall back into. After the child exits the parent asserts that the decoy is byte-identical, that the child's own `ses_fallback_child` is gone (so cleanup really ran and really was scoped to its own id), and that no `anolisa-sessions-*` appeared under the child's `TMPDIR` (so the fallback stopped at the runtime dir).

The child does the work and asserts: the pinned session id is honoured; `session.root()` and `session.display_root()` both canonicalise to the test-owned fallback base; `promote` and `session_log` succeed end to end through it; then cleanup goes through the real `svc.try_end_session(cfg.memory.session.end_action)` shutdown path rather than a bare `remove_dir_all`. It prints `fallback child ok: <path>` at the end and the parent requires that line in stdout — a child that matched no test would also exit 0, and that must not pass.

Two neighbouring tests needed adjusting for the anchored base from the P1 fix: `end()` consumes the service, which closes the descriptor that `root()`'s `/proc/self/fd/<n>` prefix resolves through, so `end_discard_removes_dir` / `end_keep_preserves_dir` now assert on `display_root()` — with `end_discard` additionally checking `root().exists()` while the service is still alive, so the anchored path is exercised too.

Forrest-ly:
Thanks for the second pass, @kongche-jbw — all three findings confirmed and fixed in `d259a8e2`. Each thread has the detail; the short version:

**[P1] `service/mod.rs` — validation and use are now bound to one descriptor.** `probe_session_base` resolves the candidate exactly once (`O_RDONLY|O_DIRECTORY|O_CLOEXEC[|O_NOFOLLOW]`) and does every check with `fstat`/`fchmod` on that descriptor; the probe uses `openat`/`unlinkat`/`fstatat`. The winner is returned as a new `SessionBase` (`src/agent-memory/src/session/base.rs`) that owns the descriptor and hands out `/proc/self/fd/<n>` — the kernel resolves that magic symlink to the descriptor's inode rather than re-walking the pathname, so the rename-then-symlink swap you reproduced cannot redirect the session any more. `SessionLogService::start_in` keeps the handle alive next to the paths derived from it, and creates `<sid>` with `mkdirat` + an `AT_SYMLINK_NOFOLLOW` verification. New `display_root()` carries the operator-facing pathname for logs.

**[P2] `service/mod.rs` — the tmp suffix uses the host uid.** New `crate::host` captures the launching uid into a `OnceLock` before any `unshare` (from `main`, from `LinuxUserNsMount::enter`, and from the top of `MemoryService::new`), and falls back to recovering it from `/proc/self/uid_map` when nothing was captured. The ownership check deliberately keeps using the *current* uid, since that is the namespace `st_uid` is reported in — the split is now "host uid for the name, namespace uid for the metadata".

**[P2] `tests/session_test.rs` — the fallback case runs in an isolated child.** Private `XDG_RUNTIME_DIR`, private `TMPDIR`, test-owned `MEMORY_SESSION_ID`, everything else passed in as `ANOLISA_TEST_*`. The parent plants a decoy live session in the fallback base and asserts it survives the child's cleanup, that the child's own session is gone, and that the tmp fallback was never needed. The child must print a completion marker, so a child that matched no test cannot pass by exiting 0.

Verification:

| check | result |
| --- | --- |
| `cargo fmt --all --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | clean |
| `cargo test --all-targets` | 368 passed, 0 failed (lib 213, up from 208) |
| `agent-memory serve`, non-root, inside userns, `XDG_RUNTIME_DIR` unset, `MEMORY_SESSION_DIR=/run/anolisa/sessions` | falls back to `<$TMPDIR>/anolisa-sessions-1359527` (host uid, not 0), base `0700`, session tree written, log reports the operator-facing path |

One deliberate behaviour change to flag: a directory can only be opened read-only, so the base is opened `O_RDONLY|O_DIRECTORY`. A session base the server cannot even *list* is now rejected where it previously might have scraped through; write access — the property that actually matters — is still proven by the `O_EXCL` probe.

Ball back with you and @ikunkun-sys for re-review.

kongche-jbw (approved):
LGTM. Re-reviewed a13fdab5181c -> d259a8e24a90, including the author replies, affected callers, descriptor lifetime, and session cleanup. The three findings from my previous review are addressed:

- Fallback validation, probing, and subsequent session operations stay anchored to the validated directory descriptor.
- Temporary-directory naming uses the UID captured before namespace entry, while ownership validation uses the current namespace's UID.
- The fallback integration test runs in a child process with test-owned runtime/tmp directories and a pinned test session ID, and verifies that a neighbouring session survives cleanup.

No new actionable findings in this revision. `git diff --check` passed, and a bounded local filesystem simulation confirmed that probing, writes, and cleanup remain in the original directory after its pathname is replaced. The current agent-memory CI check passed. Local validation was primarily static; I did not run the full Rust suite or namespace/mount tests. The documented requirement for usable `/proc/self/fd` and a readable session base was considered in this review.

PR #3277, opened 2026-09-15
test(tokenless): run the Protocol v2 golden parity suite from test-integration

Forrest-ly:
## What

One line in `src/tokenless/Makefile`: `test-integration` now depends on `test-hook-parity`, so the Protocol v2 golden parity suite runs wherever the integration suite runs — including CI.

```diff
-test-integration:
+test-integration: test-hook-parity
 	@echo "==> Testing hook integration (Python)..."
```

(plus a comment above the target explaining the prerequisite)

## Why

`make test-hook-parity` is the documented entry point for `tests/test_hook_parity.py` — the suite that replays `tests/contract/fixtures` through the real hooks against the real `tokenless` debug binary and diffs every envelope with the Protocol v2 goldens. **No CI job calls it.** The `Test tokenless` job runs `make -C src/tokenless test-integration` for the hook suites and nothing else; grepping every workflow for `test-hook-parity` / `hook_parity` returns zero hits. So a hook change that breaks the v2 envelope keeps CI green.

The gap is also silent, which is why simply appending `python3 tests/test_hook_parity.py` to the recipe is not enough. The suite guards its only class with:

```python
@unittest.skipUnless(
    os.path.exists(corpus.DEBUG_TOKENLESS_BIN),
    "tokenless debug binary not built",
)
```

With `target/debug/tokenless` absent it prints `ss` / `Ran 2 tests in 0.000s` / `OK (skipped=2)` and **exits 0** — a green job with a dark gate. Depending on `test-hook-parity` instead keeps that target's existing `cargo build -p tokenless-cli`, so the binary is always there and the goldens always run for real.

Two properties worth calling out:

- **No extra CI cost.** The job's earlier `Run tests` step (`cargo test --workspace`) already leaves `target/debug/tokenless` in place, so the prerequisite's `cargo build` is a no-op there. That the binary exists at that point is visible in the job log today: `tests/test_rewrite_hook.py`'s `RealCorePreToolTest` is guarded by the same `skipUnless`, and the log reports `Ran 10 tests ... OK` rather than `OK (skipped=1)`.
- **No double run.** make expands a prerequisite once per invocation, so the aggregate `test` target — which lists both `test-integration` and `test-hook-parity` — still runs the parity suite a single time (`make -n test | grep -c test_hook_parity.py` → `1`).

Trade-off, stated plainly: `make test-integration` is no longer runnable in a Python-only environment, it now needs the Rust toolchain like every other tokenless test target. In exchange the parity gate cannot be skipped by accident.

[...]

PR #3282, opened 2026-09-15
test(tokenless): run the benchmark rtk probe suite from test-rtk-integration

Forrest-ly:
## What

`src/tokenless/Makefile` only: a new `test-benchmark-rtk-probe` target that runs `tests/test-benchmark-rtk-version-probe.sh`, plus `test-rtk-integration` depending on it (and the matching `.PHONY`, aggregate `test:` and `help` entries).

```diff
-test-rtk-integration:
+test-benchmark-rtk-probe:
+	@echo "==> Testing benchmark RTK version probe..."
+	bash tests/test-benchmark-rtk-version-probe.sh
+
+test-rtk-integration: test-benchmark-rtk-probe
 	bash scripts/setup-rtk.sh
```

No test, script or Rust source is modified — the suite itself is untouched and already green on `main`.

## Why

`tests/test-benchmark-rtk-version-probe.sh` is the regression suite for the best-effort `rtk --version` probe that fills `rtk_version` in `benchmark/l1-compressor/benchmark_identity.json`. It covers the parts of that probe that were reworked recently:

- a bare `$RTK_BIN` (e.g. `RTK_BIN=rtk`) is resolved through `PATH`, matching the Rust `find_rtk_binary` convention, instead of being reported `unavailable`;
- the probe is bounded by `RTK_VERSION_TIMEOUT_SECS` under `timeout(1)`, including a stub that ignores `SIGTERM` (needs the unignorable `--kill-after` / `-s KILL` follow-up), and the abandoned child is reaped;
- only a **successful** probe records a version — output printed before an expired deadline, and stdout from a non-zero exit, collapse to `unavailable`;
- the same rules hold on a host without a `timeout(1)` helper.

**Nothing runs it.** A repository-wide `grep -rn test-benchmark-rtk-version-probe` at `b0fa1a0f` matches only the file itself: no Makefile target, no workflow (all 22 files under `.github/`), no other script. It is not part of `make test`, so even a developer running the full aggregate never executes it. A probe regression therefore reaches the L1 benchmark job silently — either as a run that stalls on a hung `rtk`, or as an identity file that quietly publishes `"rtk_version": "unavailable"` while the job still reports green.

Wiring it through `test-rtk-integration` puts it in CI without a workflow edit: that target is the one the `Test tokenless` job calls (`Test bundled RTK integration`), and it is the natural neighbour — both are RTK-contract suites. The ordering is deliberate: the probe suite drives stub binaries only, needs no RTK checkout and no cargo build, and finishes in ~4.5 s, so it runs **before** the pinned fetch and release build and fails fast. make expands a prerequisite once per invocation, so the aggregate `test` target (which now lists both names) still runs the suite a single time.

## Test report

Environment: Linux x86_64; GNU bash 4.4.20; GNU Make 4.2.1; Python 3.8.17 (not used by this suite); git 2.x. No network access is needed by any command below.

| # | Command | Result |
| --- | --- | --- |
| 1 | `grep -rn test-benchmark-rtk-version-probe .` (unpatched `main` @ `b0fa1a0f`) | only `src/tokenless/tests/test-benchmark-rtk-version-probe.sh` itself — the orphan reproduced |
| 2 | `bash tests/test-benchmark-rtk-version-probe.sh` (unpatched) | exit 0, 19 checks `ok`, ~4.5 s — baseline: the suite is green, so the wiring is purely additive |
| 3 | `make test-benchmark-rtk-probe` (patched) | exit 0, `==> Testing benchmark RTK version probe...`, 19 checks `ok`, real 4.5 s |
| 4 | `make -n test-rtk-integration` (patched) | probe first, then `bash scripts/setup-rtk.sh` → `cargo build --release --locked …` → `python3 tests/test_rtk_integration.py` |
| 5 | `make -n test \| grep -c test-benchmark-rtk-version-probe.sh` (patched) | `1` — prerequisite deduped in the aggregate |
| 6 | `make help` (patched) | new `test-benchmark-rtk-probe` line renders next to `test-rtk-integration` |

Mutation checks — the suite is load-bearing, not vacuous. Each mutation was applied to `benchmark/l1-compressor/run-benchmarks.sh`, the suite run, and the file restored (`git status --porcelain` empty before committing):

| # | Mutation | Result |
| --- | --- | --- |
| 7 | `resolve_rtk_bin()` no longer resolves a bare name through `PATH` (`command -v` branch stubbed out) | exit 1 — `FAIL bare command name resolved through PATH: expected 'rtk 0.1-test', got 'unavailable'`, `FAIL PATH resolution without a timeout(1) helper: …` |
| 8 | probe records stdout even when `--version` exits non-zero (success-only rule removed) | exit 1 — `FAIL partial output dropped when the probe times out: expected 'unavailable', got 'rtk partial'`, plus both `non-zero … degrades to unavailable` checks (with and without a helper) |
| 9 | probe never bounded (`RTK_TIMEOUT_PREFIX` left empty even when `timeout(1)` exists) | exit 1 — `FAIL deadline honoured (SIGTERM-responsive rtk): took 30s for a 1s deadline`, same for the SIGTERM-ignoring stub, and `FAIL partial output dropped when the probe times out`. The suite completes in ~60 s rather than hanging: the stubs `exec sleep 30`, so a real regression costs three bounded stalls and then a red job, not a stuck runner |

Not run, with reasons:

- `make test-rtk-integration` end-to-end — it fetches the pinned RTK commit and runs `cargo build --release --locked` on the vendored tree, which is out of proportion for a Makefile-only change. The new prerequisite was verified by the `make -n` ordering (row 4) and executed standalone (row 3); the CI job will exercise the full target on this PR.
- `cargo test --workspace`, the L1/L2 benchmark suites, `test-integration`, `test-adapters`, `test-hooks`, `test-claude-code-detect`, `test-raw-package`, `test-npm-package`, `test-toon-cleanup`, `test-python-runtime`, `test-agentscope-integration` — the diff touches no Rust, Python or shell source and no other target's recipe; `make -n test` still expands every one of them (row 5).
- The CI job itself.

## Notes for the reviewer

- Same class of gap as #3277 (a suite whose target no CI job calls) but a different mechanism: that one adds a prerequisite to `test-integration`, this one adds a target and a prerequisite to `test-rtk-integration`. The hunks do not overlap — #3277 touches only the `test-integration:` line and its comment, this PR touches the `.PHONY` list, the aggregate `test:` line, the block between `test-integration` and `test-hook-parity`, and `help` — so the two should merge in either order.
- The alternative is one word in `.github/workflows/ci.yaml` (`make -C src/tokenless test-rtk-integration test-benchmark-rtk-probe`). The Makefile route was preferred because it also covers every non-CI caller (`make test`, a developer running the RTK target locally) and keeps the ordering guarantee — probe first, expensive pinned build second — inside the component that owns it.
- Worth a look while you are here, not part of this diff: the same orphan scan leaves `tests/test-toon-full.sh` referenced only from documentation (#3251 fixes the script itself) and `tests/test-posttooluse-replacement.sh` referenced from nowhere (#3250 retires it). `test-benchmark-rtk-version-probe.sh` was the only suite that was both unreferenced and green.


src/tokenless/Makefile
```diff
@@ -274,7 +274,22 @@ test-integration:
...
+# collapses to "unavailable"). tests/test-benchmark-rtk-version-probe.sh guards
+# exactly that contract with stub binaries, but no target ever ran it, so a
+# probe regression reached CI unnoticed: the benchmark job would either stall on
+# a hung rtk or silently publish an "unavailable" identity. The suite needs no
+# rtk build and finishes in a few seconds, so it runs ahead of the pinned build
+# below and fails fast. `test-rtk-integration` is the only RTK target a CI job
+# calls ("Test bundled RTK integration"), which is what puts this in CI.
+test-benchmark-rtk-probe:
+	@echo "==> Testing benchmark RTK version probe..."
+	bash tests/test-benchmark-rtk-version-probe.sh
```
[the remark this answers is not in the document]
Forrest-ly: Adopted — this one reproduces, thanks. Fixed in 9f6c895.

**Confirmed mechanism.** `timeout` moves itself *and* the monitored command into a fresh process group (`setpgid (0, 0)`), and when the `--kill-after` deadline fires, `cleanup()` ends with `send_sig (0, sig)` — a `kill(0, SIGKILL)` on that group. So `timeout` dies by SIGKILL without reaping the stub it just killed; the stub is reparented to PID 1 and, where PID 1 has no reaper, stays a zombie that `kill -0` reports as present. I read that path in v9.4's `src/timeout.c`; my runs below were on coreutils 8.30, which behaves the same.

**Reproduction.** Fresh PID namespace whose PID 1 is a `python3` process that only ever reaps its own child (`os.waitpid(child, 0)`, never `waitpid(-1)`), pre-fix script:

```
  ok   SIGTERM-ignoring rtk still bounded
  ok   deadline honoured (SIGTERM-ignoring rtk) (returned after 2s, deadline 1s)
  FAIL SIGTERM-ignoring probe child was force-killed and reaped
benchmark rtk version probe test FAILED
--- zombies still present under non-reaping PID 1 ---
    PID    PPID STAT COMMAND
     40       1 Z    sleep <defunct>
```

Only the SIGTERM-ignoring case fails, because it is the only one that reaches the `--kill-after` SIGKILL path — the other stubs die on SIGTERM and are reaped by `timeout` normally. That also explains why this did not show up on runners whose PID 1 has a reaper.

**Fix (test-only).** The liveness test now goes through a `pid_is_gone` helper that reads the process state (`/proc/<pid>/stat`, falling back to `ps -o state=`) and counts `Z` as gone; where neither is available it degrades to the previous `kill -0` behaviour. A live child still fails the check, so the guard keeps its meaning. A new check 8 holds a dead child under a parent that never `wait()`s and asserts the tolerance, so a regression of the helper turns the suite red on any host rather than only inside a reaper-less container.

**Verification** (Linux x86_64, GNU bash 4.4, GNU Make 4.2.1, GNU coreutils 8.30, python3 3.8):

- `make test-benchmark-rtk-probe` — 20/20 checks pass on a normal host, and 20/20 inside the non-reaping PID-1 namespace above (1 failure there before the fix); the `sleep <defunct>` zombie is still present in that namespace afterwards, so the tolerance is what carries it.
- Mutation: deleting the `[ "$state" = "Z" ]` branch of `pid_is_gone` → exit 1 with `FAIL unreaped (zombie) child counts as gone` on a plain host. A live `sleep` is still reported as not gone and a reaped one as gone, so the helper is not vacuous.
- `shellcheck` 0.10.0 clean on the changed file; `make -n test-rtk-integration` still runs the probe ahead of `scripts/setup-rtk.sh`; `make -n test | grep -c test-benchmark-rtk-version-probe.sh` is still 1.

No production code changed: the probe bounded and killed its child correctly in every run — the fault was purely in the test's liveness predicate.

src/tokenless/Makefile
```diff
@@ -274,7 +274,22 @@ test-integration:
...
+
+test-rtk-integration: test-benchmark-rtk-probe
```
ikunkun-sys, line 292: **[P2] Preserve the no-timeout-helper path when gating RTK integration**

This new unconditional prerequisite makes the existing RTK integration target fail on macOS hosts without `timeout` or `gtimeout`. The production benchmark runner explicitly permits that environment and runs the version probe without a deadline, but the newly gated suite always expects the hanging stubs to time out. I ran the unmodified `make test-benchmark-rtk-probe` in that environment and reproduced:

```text
FAIL deadline honoured (SIGTERM-responsive rtk): took 31s for a 1s deadline
FAIL deadline honoured (SIGTERM-ignoring rtk): took 30s for a 1s deadline
FAIL partial output dropped when the probe times out: expected 'unavailable', got 'rtk partial'
```

`make` exited 2. Because this is now a prerequisite, both `test-rtk-integration` and `make test` inherit the failure before the RTK integration recipe runs. Please explicitly skip the timeout-dependent assertions when neither helper is available, while retaining the path-resolution and version/exit-status checks.

ikunkun-sys (changes requested):
Found one reproducible P2 compatibility regression in the new prerequisite. On macOS without timeout or gtimeout, the unmodified probe target fails and consequently blocks test-rtk-integration and the aggregate test target. Please make the timeout-specific assertions conditional on helper availability.

Validation: ran make test-benchmark-rtk-probe on macOS without either helper (make exited 2), checked shell syntax, and verified the test-rtk-integration dry-run ordering. The successful Linux CI run was inspected; the full RTK integration build was not rerun locally.

src/tokenless/Makefile
Forrest-ly: Adopted — reproduced here too, fixed in 132f3d0. Thanks for the exact failure lines; they matched one for one.

**Reproduction.** Linux x86_64 with `timeout` and `gtimeout` hidden from `PATH`, which drives the same branch your macOS host hits (`command -v timeout` → `command -v gtimeout` → `RTK_TIMEOUT_CMD=""` → `RTK_TIMEOUT_PREFIX` empty → unbounded probe). Unmodified `9f6c895`:

```text
FAIL deadline honoured (SIGTERM-responsive rtk): took 30s for a 1s deadline
FAIL deadline honoured (SIGTERM-ignoring rtk): took 30s for a 1s deadline
FAIL partial output dropped when the probe times out: expected 'unavailable', got 'rtk partial'
benchmark rtk version probe test FAILED     # exit 1 (2 through make), 90.6s wall
```

Your diagnosis is the right one: this is not a probe regression. With no helper the runner deliberately probes unbounded, so there is no deadline to honour and no abandoned probe whose partial output must be dropped — the stub prints its line, exits 0, and the runner records it, which is the documented behaviour. Only the assertions were wrong for that environment; because the target is a prerequisite they took `test-rtk-integration` and `make test` down with them.

**Fix (test-only, `tests/test-benchmark-rtk-version-probe.sh`).** The suite now mirrors the runner's own discovery: `HAVE_TIMEOUT_HELPER` probes for `timeout` then `gtimeout`, and when neither exists the three hanging-stub groups (4a, 4b, 5a) print an explicit `skip` with the reason instead of running, with the count echoed in the summary:

```text
  skip deadline honoured (SIGTERM-ignoring rtk) (no timeout(1) or gtimeout(1) helper on this host)
  ...
benchmark rtk version probe test passed (8 checks skipped: no timeout(1) or gtimeout(1) helper on this host)
```

Everything you asked to retain still runs on every host: PATH resolution for a bare `$RTK_BIN` (both with the helper and against the derived `prelude-no-helper.sh`), absolute paths, missing / non-executable / unknown-on-PATH / silent rtk, first-line-only, the non-zero-exit rule in both preludes, and the JSON validity + zombie self-checks. On a helper-less host the suite is now 12 passed / 8 skipped / 0 failed in 0.45s, instead of 3 failed in 90.6s.

I left the Makefile prerequisite as it is: with the gate inside the suite, `test-rtk-integration` keeps the fail-fast ordering this PR was after, and macOS hosts still get every check that applies to them.

One related portability fix while in the same file: `pid_is_gone` now matches the **leading** state char (`case Z*`) instead of the whole field, because BSD/macOS `ps -o state=` prints a sequence (`Z+`, `SN`) where procps prints a single char — and with no `/proc` on macOS that `ps` fallback is exactly the path your platform takes.

**Verification** (Linux x86_64, GNU bash 4.4.20, GNU Make 4.2.1, GNU coreutils 8.30, python3 3.8, shellcheck 0.10.0):

- helper-less `PATH`: exit 0 — 12 `ok`, 8 `skip`, 0 `FAIL`, 0.45s (was 3 FAIL / exit 1 / 90.6s at `9f6c895`).
- normal host: `make test-benchmark-rtk-probe` exit 0, 20/20 checks, no skips — unchanged from `9f6c895`.
- fresh PID namespace whose PID 1 never reaps orphans: 20/20 — the previous round's zombie tolerance is intact.
- mutation: replacing the `Z*` branch of `pid_is_gone` with a never-matching pattern still turns the suite red (`FAIL unreaped (zombie) child counts as gone`, exit 1), so the skip gate did not make the reaping checks vacuous.
- `shellcheck` 0.10.0 clean; `make -n test-rtk-integration` still runs the probe ahead of `scripts/setup-rtk.sh`; `make -n test | grep -c test-benchmark-rtk-version-probe.sh` is still 1.
- CI on `132f3d0`: 24 checks, 0 failures (9 SUCCESS + 15 SKIPPED). `Test tokenless` job log shows all 20 probe checks green — deadline assertions included, since the runner does have `timeout(1)` — followed by `setup-rtk.sh`.

Not run locally: macOS itself (no such host here — the helper-less `PATH` run is the closest equivalent, and it exercises the same `RTK_TIMEOUT_CMD` branch), and the full RTK integration recipe (needs the RTK checkout plus a release build), which the `Test tokenless` job covers and passed.

ikunkun-sys (approved):
Re-reviewed 132f3d0. The previously reported P2 regression is fixed; no new actionable findings in the update.

I reran the unmodified make test-benchmark-rtk-probe on the same macOS host with neither timeout nor gtimeout: exit 0 in 2.72 seconds, with 12 checks passed and 8 timeout-dependent checks explicitly skipped. Path resolution, version/exit-status checks, JSON validity, and the zombie self-check still execute. The helper detection matches the runner, and the leading-Z process-state match preserves the intended zombie handling.

bash -n passed, and make -n test-rtk-integration still puts the probe suite before RTK setup/build. I also inspected the latest Linux CI job (104647218848): all 20 probe checks passed with no skips, followed by a successful RTK build and 9 passing integration tests. The full RTK integration build was not rerun locally.

Issue #3267, opened 2026-09-15
[TestAgent][Nightly][tokenless] bug: detect.sh re-list budget (DETECT_PLUGIN_RELISTS=2) exhausted under load — installed claude-code intermittently reported not-installed (recurrence of #3082 after merged fix #3085)


zhangtaibo:
### Component

tokenless — claude-code adapter `detect.sh`

### Bug Description

`detect.sh` intermittently reports the claude-code framework as **not installed** (exit `1`) on first-run detection, even though the `claude` CLI and its config directory are both present. The failure is a registry-index timing race: right after a fresh install/registration, `claude plugin list` succeeds (exit 0) but has not yet indexed the newly-added plugin, so `detect.sh` concludes "plugin install: not installed".

This is a **recurrence of #3082** after the mitigation in **#3085 (merged 2026-09-07)**. #3085 added a settle + re-list loop (`settle_plugin_listed`) to retry when the plugin list succeeds but omits the plugin. However the re-list budget `DETECT_PLUGIN_RELISTS=2` is too small: under full concurrent system load (e.g. a full build + test suite running in parallel), the registry index lag exceeds 2 re-list attempts, the budget is exhausted, and the false "not installed" verdict is returned again. Isolated re-runs (no load) always pass.

### Steps to Reproduce

On an ECS with the anolisa tokenless package installed:

1. Ensure the claude-code adapter files are present:
   `/usr/share/anolisa/adapters/tokenless/claude-code/scripts/detect.sh`
2. Trigger a fresh claude plugin registration so the `claude plugin` registry index
   is momentarily stale (e.g. reinstall / re-add the claude-code plugin, or run on a
   freshly provisioned host where the plugin was just registered).
3. Immediately — while the registry index is still catching up, and under concurrent
   system load (several CPU-bound processes running) — invoke the detect script
   directly:
   ```bash
   bash /usr/share/anolisa/adapters/tokenless/claude-code/scripts/detect.sh; echo "rc=$?"
   ```
4. Repeat step 3 a few times in a row under load.

### Actual Behavior

Intermittently (roughly 4 of 5 attempts under load) `detect.sh` exits `1` and prints a
verdict equivalent to:

```
[tokenless] claude-code detect
[tokenless]   claude CLI                 present (/usr/local/bin/claude, v2.x)
[tokenless]   claude config dir          present (~/.claude)
[tokenless]   plugin install             not installed
rc=1
```

[...]

PR #3297, opened 2026-09-16
fix(tokenless): stop the toon cleanup stub from breaking setup-rtk, and run both shell suites in CI

Forrest-ly:
## What this PR does

Two changes; the first is what makes the second safe.

### 1. `fix(tokenless)`: the toon cleanup suite poisoned the vendored rtk checkout

`make test-toon-cleanup` stubs `third_party/rtk/target/release/rtk` when the vendored rtk has not been built, because `install-helpers` copies that artifact unconditionally. The exit trap removed the stub **file** but left the directories it had created, and `scripts/setup-rtk.sh` refuses to fetch into a destination that already exists without a `Cargo.toml`:

```text
ERROR: RTK destination is incomplete: <repo-root>/src/tokenless/third_party/rtk
```

So on a fresh checkout, running that one suite broke every later rtk entry point — `just setup-rtk`, `make build`, `make build-tokenless` and `make test-rtk-integration` (the CI step *Test bundled RTK integration*) — until `just clean-rtk` removed the directory by hand. The `make test` aggregate hides it, because `test-rtk-integration` runs before `test-toon-cleanup` there; a standalone run of the suite, or any target that orders the two the other way round, hits it.

The suite now records which directories its `mkdir -p` is about to create (walking up from the artifact to the first directory that already exists) and `rmdir`s exactly those on exit, deepest first. `rmdir` only removes empty directories, so a real vendored checkout — cloned, built, or both — can never lose anything, and the stub file is removed as before.

### 2. `test(tokenless)`: two members of `make test` that no CI job ever called

`make test` aggregates nine targets and the *Test tokenless* job calls four of them. Grepping every file under `.github/` for these two target names, and for the two script names behind them, returns nothing:

| Target | What it covers |
| --- | --- |
| `test-claude-code-detect` | `detect.sh`'s first-run plugin re-list retry — the path a nightly run found exhausting its relist budget (#3267) |
| `test-toon-cleanup` | the `clean-stale-toon` ownership rules: `install-helpers` / `uninstall` drop only Tokenless-owned `toon` symlinks and preserve an unrelated user-installed `toon` executable |

A regression in either reached `main` with a green build. Both are plain bash plus a recursive make over a `mktemp` sandbox with stub binaries — no cargo, node, jq or rtk build — and finish in about a second, so they ride along on `test-integration`: the hook-contract target a CI job already calls (*Run Python integration tests*) and the same host the golden parity suite uses. They cannot go on `test-rtk-integration` instead, because that target's own recipe is what fetches and builds the vendored rtk and it runs later in the job. `make` runs each target once per invocation, so the aggregate still executes both a single time (`make -n test` lists each script once).

**Ordering note:** this puts `test-toon-cleanup` ahead of the CI step that calls `setup-rtk.sh`, which is only safe because of commit 1 — hence one PR rather than two.

**Deliberately not wired:** `test-hooks`, the third uncovered member of the aggregate. Its `run-all-tests.sh` states above test 5.0b that it drives the *installed release* binary found on `PATH` on purpose (release-layout verification), CI has no installed `tokenless`, and the pending change to that suite is still settling how the binary is resolved and guarded. Wiring it here would preempt that review.

No `.github/workflows/` file is touched.

## Test report

Environment (public information only): Linux x86_64, GNU Make 4.2.1, bash 4.4.20, rustc/cargo 1.96.0, Python 3.12, node v22.21.1, jq 1.6. Base = upstream `main` at `d940a663` (tokenless 0.8.2, vendored rtk v0.49.0); the workspace CLI used below is built from this tree, and the rtk used is the vendored one built via `scripts/setup-rtk.sh` + `cargo build --release --locked --manifest-path third_party/rtk/Cargo.toml`.

### The bug: reproduce, then verify the fix

Run in a clean worktree of `main` with no `third_party/rtk`:

| Step | Before the fix | After the fix |
| --- | --- | --- |
| `make test-toon-cleanup` | passes | passes |
| `find third_party -maxdepth 3` afterwards | `rtk/target/release/` left behind | only `patches/` |
| `bash scripts/setup-rtk.sh` afterwards | exit 1, `RTK destination is incomplete` | exit 0, `RTK v0.49.0 setup complete` |

Other tree states, after the fix:

| Tree state | Result |
| --- | --- |
| rtk cloned but not built | suite passes, no `target/` residue, `setup-rtk.sh` reports *already set up* |
| rtk built (developer tree, and CI after the RTK step) | no stub taken; artifact checksum identical before and after |

### The wiring

| Check | Result |
| --- | --- |
| `make test-integration` | exit 0 |
| ... golden parity prerequisite | `Ran 2 tests ... OK` |
| ... `test-claude-code-detect` | `claude-code detect retry test passed` |
| ... `test-toon-cleanup` | `tokenless stale-toon cleanup tests passed` |
| ... 14 Python integration suites | 15 `OK` blocks, 0 failures, 0 errors |
| `make -n test-integration` | order: parity → detect → toon → Python |
| `make -n test` | each of the two scripts listed exactly once |
| `bash -n tests/test-toon-cleanup.sh` | clean |
| Gap proof | 0 hits for both target names and both script names anywhere under `.github/` |

**Negative control (proves the wiring is load-bearing):** with `clean-stale-toon` mutated to `rm -f $(DESTDIR)$(BINDIR)/toon` unconditionally — i.e. the regression these suites exist to catch — `make test-integration` now exits 2 with `FAIL: install-helpers removed an unrelated regular-file toon executable` at the `test-toon-cleanup` prerequisite. Before this PR the same mutation left CI green. The mutation was reverted; the branch contains only the two commits.

### Not run

- The full `make test` aggregate: it needs a release build of the workspace plus npm packaging, which this change does not touch.
- `test-hooks`: not wired by this PR, for the reason given above.


kongche-jbw:
LGTM for f6cef6f59841fcd7a2fb1342e2e960ac238d1ec6.

Reviewed the full diff and affected Makefile targets, cleanup lifecycle, detection script, and CI wiring. No actionable correctness or regression findings.

Validation:
- Reproduced the directory residue and subsequent setup-rtk failure on the base revision.
- Verified cleanup with an absent RTK directory, an existing checkout, existing empty build directories, and an existing artifact; pre-existing directories and artifact contents/mode/mtime were preserved.
- Both shell suites passed locally; Make dry-runs confirmed each runs once through test-integration and the aggregate test target.
- Confirmed CI logs show both suites passing, followed by successful RTK setup.

The full make test aggregate was not rerun locally.

kongche-jbw (approved):

PR #3301, opened 2026-09-16
test(tokenless): cover claude probe failure

Forrest-ly:
## Why

`_cached_claude_version()` in `compress_response_hook.py` can come back empty three ways:

1. `claude --version` exits non-zero;
2. it prints something with no `x.y.z` in it, so `parse_version()` returns `None`;
3. the `claude` PATH entry resolves but cannot be spawned, so `subprocess.run()` raises.

`_claude_supports_replacement()` maps all three to "no replacement capability". That is the fail-open half of the 2.1.121 gate: an undetectable version must never receive `updatedToolOutput`, and must never get the compressed copy appended through `additionalContext` beside the still-visible original.

The suite only ever fed the probe a mock answering with a *known* version (`2.1.120` / `2.1.210` / `2.1.259`), so that half of the gate was asserted by reading the code rather than by a test. This was raised in review on #3250, whose coverage table listed "old / undetectable claude version fails open" against `test_no_savings_and_old_claude_do_not_emit_replacements` - a test that covers the known-version side only.

## What changed

One file, `src/tokenless/tests/test_compress_response_hook.py`. No product code changes.

- **New `TestClaudeVersionProbeFailure`** - drives one broken mock per failure branch (`exit-nonzero`, `unparseable`, `unspawnable`) and asserts, per branch, that:
  - the hook exits clean and emits `{}` (fail open, nothing replaced);
  - the captured request declares `replace_output=false` while `replace_with_text` stays `true` - the payload was otherwise fully eligible for the text slot, so the version gate is the only thing holding it;
  - `content_origin` is unchanged (`api_response`);
  - exactly one `tokenless` subprocess is spawned (section 5.6);
  - the hook warned `Claude Code < 2.1.121 (or version unknown): ...`, which is what stops the test from passing vacuously through some unrelated early return;
  - `~/.tokenless/.claude-version` was **not** written - only a parsed version is cached, so one transient probe failure cannot pin a stale verdict for later invocations.
- **New `_create_broken_mock_claude()` helper** - sibling of `_create_mock_claude()`; raises `ValueError` on an unknown mode, matching `_create_mock_tokenless()`. The `unspawnable` mode writes a non-image file with the executable bit set, so `shutil.which()` resolves it and `execve` then fails; a hanging probe lands in the same `except Exception` branch, so it is documented rather than exercised at a 5s cost.
- **`_run_hook()` grows an optional `stderr_sink`** list so a caller can assert on `warn()` diagnostics. Default behaviour and return type are unchanged; no existing caller is touched.
- Module docstring gains one bullet.

## Related issue

`no-issue: follow-up to the non-blocking review feedback on #3250`

## User / Agent impact

None. Test-only.

## Risk and compatibility

Low risk: one new test class, one new mock helper and one optional keyword argument on a test-local helper.

## Validation

Environment: Linux x86_64, CPython 3.12.9. CI runs this suite on 3.11; the file self-skips below 3.9 and the new code uses no post-3.9 syntax.

```text
Ran 2 tests in 4.563s
OK
==> Testing hook integration (Python)...
Ran 36 tests in 4.086s        # test_compress_response_hook.py, was 35
OK
... 13 further Python suites ...
# 15 suites, 192 tests, 0 failures, 0 skips
```

The new test is not vacuous. Both regressions it guards were injected into a scratch copy of the hook and each was caught, then the hook was restored:

| injected regression | result |
| --- | --- |
| `_claude_supports_replacement()` returns `True` when the version is `None` | FAIL - all 3 subtests, `updatedToolOutput` emitted |
| `_cached_claude_version()` caches even when `parse_version()` returns `None` | FAIL - `unparseable` subtest, `.claude-version` written |

Post-restore tree state:

```text
$ git status --short
 M src/tokenless/tests/test_compress_response_hook.py
$ git diff --check
(clean)
```

Not run locally: the `cargo fmt --check` / `clippy` / `cargo test --workspace` steps of the `Test tokenless` job. This PR touches no Rust source; `cargo build -p tokenless-cli` was run to produce the binary the golden-parity suite above needs.

## Documentation and rollback

No documentation changed - the coverage claim this corrects lived in the #3250 description, not in the tree. Revert the single commit to roll back.


kongche-jbw:
Follow-up to [the response on #3250](https://github.com/agentic-os-org/ANOLISA/pull/3250#issuecomment-5693943564): this PR addresses the failed-version-probe coverage gap raised there. I independently confirmed that the new test catches both mutations described in that response (allowing an unknown version and caching an unparseable version).

One remaining finding against e78b2ab3769275cb9e3fd5de4c0333aec428512d:

**[P2] Make the non-zero-exit fixture discriminate the return-code check** — [`test_compress_response_hook.py:195`](https://github.com/agentic-os-org/ANOLISA/blob/e78b2ab3769275cb9e3fd5de4c0333aec428512d/src/tokenless/tests/test_compress_response_hook.py#L195-L196).

The `exit-nonzero` mock only calls `sys.exit(1)`, leaving stdout empty. Removing `if proc.returncode != 0: return None` from `_cached_claude_version()` still leaves all 36 tests passing: `parse_version("")` returns `None`, masking the missing exit-status check. A failed probe that prints a parseable supported version could then incorrectly enable replacement and populate the version cache without this suite detecting the regression.

In an isolated snapshot, changing this fixture to `print("2.1.210"); sys.exit(1)` made the same mutation fail in the `exit-nonzero` subtest; the unchanged production hook passed with that fixture. Please have this mode emit a supported version before exiting non-zero so it independently protects the exit-status contract, while `unparseable` continues to cover parsing failure.

Validation: Python 3.12.3, the original 36-test file passed, and the mutations above were tested only in a temporary snapshot. No production-code regression was found in this test-only PR.

kongche-jbw (approved):
LGTM for d39c3df199a1743e1b32788e8dd2e0515bd01210.

Re-reviewed the incremental change from e78b2ab3769275cb9e3fd5de4c0333aec428512d. The previously reported P2 is resolved: the non-zero-exit mock now prints a supported version before failing, so the test independently protects the exit-status guard rather than relying on empty stdout failing to parse. No new actionable findings.

Independent validation on Python 3.12.3:
- All 36 tests in test_compress_response_hook.py passed.
- Removing the production exit-status guard in a temporary snapshot caused exactly the exit-nonzero subtest to fail on unexpected updatedToolOutput, confirming the regression is now detected.
- git diff --check passed.

Test tokenless CI is still running at approval time; its final result remains pending. The full make test aggregate was not rerun locally.

Forrest-ly:
**[P2] non-zero-exit fixture — adopted; it discriminates the return-code check now.**

Reproduced the gap first on `e78b2ab`: deleting `if proc.returncode != 0: return None` from `_cached_claude_version()` left all 36 tests green, exactly as reported — `parse_version("")` returns `None`, so both the fail-open assertion and the "cache not written" assertion passed vacuously.

The `exit-nonzero` mock now prints a supported version (`2.1.210`) before `sys.exit(1)`, so the exit status alone has to void an otherwise parseable answer. Same mutation, re-checked against that fixture:

| production hook | `exit-nonzero` subtest |
| --- | --- |
| guard removed | **FAIL** — hook emits `updatedToolOutput` and writes `~/.tokenless/.claude-version` |
| untouched | pass (36/36) |

`unparseable` keeps covering parse failure and `unspawnable` the spawn exception, so the three modes stay one-to-one with the three branches of `_cached_claude_version`; the fixture docstring now records why the failing probe has to print something parseable.

Validation: Linux x86_64, Python 3.12.9. `python3 tests/test_compress_response_hook.py` → 36 tests, OK. `make -C src/tokenless test-integration` → 192 tests across 15 suites, OK (0 failures, 0 skips). Test-only change, no production code touched. CI was green on the pushed head `d39c3df`, which is what landed here — thanks for squashing it in.

PR #3326, opened 2026-09-17
feat(anolisa): explain raw platform mismatches

### LINKED ISSUE #2087: [anolisa] explain available platforms when no artifact matches the host
## Component

anolisa

When a component is published but has no artifact for the current host,
`anolisa install` reports the same generic resolver error used for an unknown
component:

```text
no distribution entry matches the query
```

For example, installing a component on `linux/aarch64` is difficult to
diagnose when the raw distribution index contains only a `linux/x86_64`
artifact.

If the package exists in the installable raw index but no entry supports the
requested operating system and architecture, show the requested platform and
the published alternatives:

```text
error: cosh-ng is not available for linux/aarch64

available platforms:
  - linux/x86_64
```

The platform list should be deduplicated and sorted so the output is stable.

## Implementation pointers

- Resolver and distribution entries:
  `src/anolisa/crates/anolisa-core/src/distribution.rs`
- Raw install error mapping:
  `src/anolisa/crates/anolisa-cli/src/commands/tier1/install/raw.rs`
- Raw install tests:
  `src/anolisa/crates/anolisa-cli/src/commands/tier1/install/tests/raw_e2e.rs`

Keep `DistributionIndex::resolve` selection behavior unchanged. When
`ResolveError::NotFound` is returned, inspect entries for the same
backend-native package and channel:

- Report a platform mismatch only when that package is published but has no row
  matching the requested `os` and `arch`; an `arch = "any"` row counts as
  a match.
- Build the available list from artifact types supported by the raw backend.
- If the requested platform exists and a later filter such as version, install
  mode, libc, or package base caused the failure, retain the existing
  diagnostic.
- If the package is unknown, retain the existing diagnostic.

## Acceptance criteria

- [ ] A published package with no host-platform artifact reports the requested
      `os/arch`.
- [ ] The diagnostic lists the available `os/arch` pairs without duplicates
      and in stable order.
- [ ] An unknown package retains the current generic not-found diagnostic.
- [ ] Failures caused by version, install mode, libc, or package base are not
      misreported as platform mismatches.
- [ ] Resolver selection rules and repository fallback behavior do not change.
- [ ] Human-readable and `--json` error output remain valid.
- [ ] Unit or raw end-to-end tests cover both the mismatch and unknown-package
      cases.
- [ ] The relevant checks from `src/anolisa/AGENTS.md` pass.

## Non-goals

- Publishing new artifacts
- Adding automatic cross-platform or cross-backend fallback
- Redesigning every `ResolveError::NotFound` variant
- Changing artifact selection priority

Forrest-ly:
## Why

Through the raw backend, `anolisa install <component>` answers two very different failures with one sentence. When the index carries no row for this host, the resolver's `NotFound` is rendered as:

```text
error: cannot resolve package 'cosh-ng' (component 'cosh-ng', version latest, linux/aarch64, system mode) from <index>: no distribution entry matches the query
```

That is the same wording an unknown component name produces, so an operator on `linux/aarch64` against an `x86_64`-only index cannot tell a wrong build host from a typo without reading the index by hand.

## What changed

- `DistributionIndex::published_platforms` (anolisa-core): the distinct `os`/`arch` pairs the index publishes for the query's component and channel, sorted and deduplicated. It sits beside `matching_versions` — the same question asked about targets instead of versions — and ignores every selector `resolve` applies later, so it reports what the index carries rather than what this one query could take. `resolve` and its selection rules are untouched.
- `resolve_raw` (anolisa-cli): on `ResolveError::NotFound`, when the **installable** index does publish the package but no row matches the requested `os`/`arch`, refuse with

  ```text
  error: component 'cosh-ng' (package 'cosh-ng') is not available for linux/aarch64 from <index>; nothing was changed

  available platforms:
    - linux/x86_64
  ```

  The list comes from the installable view, so a platform published only with artifact types the raw backend cannot place is never offered as an alternative. `arch = "any"` counts as a match for the requested arch, exactly as `resolve` reads it.

The ordering is deliberate: the pinned-version diagnostics still run first, and the platform branch fires only when no row matches `os`/`arch`. A refusal a later filter caused — version, install mode, libc, package base — keeps its existing wording, and an unknown package keeps the generic one.

## Related issue

closes #2087

## User / Agent impact

One raw-install refusal branch changes its message: it now names the requested platform and lists the published ones. Exit code (`INVALID_ARGUMENT`, 2), the `--json` envelope shape, resolver selection rules and repository fallback behavior are unchanged. The reason stays a single string, so the multi-line list is escaped normally in JSON output.

## Risk and compatibility

The visible change is the reason text of one error branch. A script that matched `no distribution entry matches the query` to detect a platform gap now sees the new wording; it still sees the old one for an unknown package, which is the case that message was always accurate for. Otherwise: one new read-only method on `DistributionIndex`, no serialization, receipt, plan/apply, or lock contract touched, and nothing is fetched or written on this path — the refusal happens during resolution, so the `nothing was changed` claim is the same one the pinned-version refusals already make.

## Validation

Non-root Alibaba Cloud Linux 3 x86_64, rustc 1.93.0 (the pinned 1.93.1 could not be fetched from the configured rustup mirror — 404 — so the gates ran on the installed 1.93.0), tests run without network:

- `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean
- `cargo test --workspace --locked` — 2581 passed, 0 failed, 2 ignored (the two isolated JSON-renderer children)
- `cargo doc --workspace --no-deps --locked` — clean

New tests, against the acceptance criteria in #2087:

- core: `published_platforms_dedup_sort_and_ignore_narrowing_selectors` (stable order, one line per platform, and install mode / libc / package base must *not* narrow the list), `published_platforms_scope_to_component_and_channel` (another component's or channel's targets are not offered; an unknown component publishes nothing), `published_platforms_report_any_arch_verbatim`
- raw e2e: `install_platform_mismatch_lists_published_platforms` (requested platform named, list deduplicated and sorted, generic resolver wording gone, and the multi-line reason rendered byte-identical through the production `--json` path in an isolated child), `install_unknown_package_keeps_generic_not_found_diagnostic`, `install_host_platform_published_keeps_generic_not_found_diagnostic` (a `pkg_base` refusal is not misreported as a platform gap), `install_pinned_version_on_unpublished_platform_lists_platforms` (the platform gap wins over the pin wording)

Both directions were checked by mutation, so the tests are not vacuous: reverting the `resolve_raw` branch makes the two mismatch tests fail with the old generic text; separately dropping `dedup` duplicates a platform line, and dropping the `host_published` guard makes the `pkg_base` test fail.

## Documentation and rollback

No documentation change: neither the user guide nor the anolisa README enumerates raw resolver diagnostics — the pinned-version refusal this branch is modelled on (d2af9b33) changed no docs either. Rollback is a revert of this single commit; no state, index, or receipt format is involved.


kongche-jbw (commented):
Review baseline: `8c7aac3994aa...b733e4fa1e83`
未发现阻塞问题。

Remaining risks:
- 未在真实 aarch64 或 macOS 主机运行；平台判断已由生产调用链测试覆盖。

Verification scope:
- 完整 diff、相邻 resolver/install 实现、Issue 目标及全部评论线程。
- 本地检查使用 rustc 1.93.1 并通过；当前非 skipped CI 均通过。

ikunkun-sys (approved):
结论：APPROVE

已审查当前提交 b733e4fa1e83743b9177687bb1cedb4521c1ad3a 的完整变更及 install/update 调用路径，未发现需要修改的问题。平台列表来自可安装制品视图，按包名及 channel 限定范围，去重排序稳定；arch = "any" 与现有解析器一致，已有版本诊断优先级及后续筛选失败的诊断保持不变。

本地验证（Linux x86_64，Rust 1.93.1）：
- cargo +1.93.1 test --locked -p anolisa-core published_platforms：3 项通过。
- cargo +1.93.1 test --locked -p anolisa-cli commands::tier1::install::tests::raw_e2e：54 项通过，2 个 ignored 的 JSON 渲染子测试由父测试调用并验证。
- cargo +1.93.1 fmt --all -- --check：通过。
- cargo +1.93.1 doc --locked -p anolisa-core --no-deps：通过。

未在本地运行完整 workspace 测试或其他平台实机测试。提交前已刷新 head、CI、mergeability 和 review threads：head 未变化，CI 14 项成功、15 项跳过，无未解决 review 线程。

Issue #3255, opened 2026-09-14
ci(memory): wire the OpenClaw plugin TS unit suite into CI — the Test agent-memory job has no Node step


Forrest-ly:
### Component

- CI: `.github/workflows/ci.yaml` → job `test-agent-memory` ("Test agent-memory", line 1031), steps at lines 1043–1075 on `main` @ `69b6869a`
- plugin suite: `src/agent-memory/Makefile:156-160` (`test-openclaw-plugin`), `src/agent-memory/adapters/agent-memory/openclaw/tests/unit/*.ts` (7 files), `package.json` scripts `test` / `test:coverage` / `smoke`

### Background

`make test-openclaw-plugin` is the only entry point for the OpenClaw plugin's TypeScript unit suite — `npm test` → `npx tsx --test tests/unit/*-test.ts` over `config-test.ts`, `corpus-routing-test.ts`, `corpus-test.ts`, `keyword-extract-test.ts`, `manifest-config-schema-test.ts`, `mcp-client-test.ts`, `safety-test.ts`. **No CI job calls it.** Grepping every workflow for the target and for the seven file names returns nothing.

What `Test agent-memory` does run:

| Step | Command |
| --- | --- |
| Test OpenClaw adapter installer script | `make test-openclaw-install` (ci.yaml:1046) — bash only, deliberately before any toolchain setup |
| Check formatting | `cargo fmt --all --check` (ci.yaml:1065) |
| Lint | `cargo clippy --all-targets --locked -- -D warnings` (ci.yaml:1070) |
| Run tests | `cargo test --locked` (ci.yaml:1075) |

All four are Rust or bash. The job has no `actions/setup-node` step, so nothing in it can execute a line of the plugin's TypeScript — and the plugin is where most of the component's recent churn is (`a3e641a3`, `2ca89281`, `a706fa4b`, `0c8b4fea` all edit `adapters/agent-memory/openclaw/src/*.ts` together with these very unit tests).

#3134 closed on 2026-09-09 having landed only the bash installer test at ci.yaml:1043. The Makefile comment above `test-openclaw-plugin` still reads *"(see #3134 for wiring it into the `Test agent-memory` CI job, which has no Node step today)"*, so that pointer now leads to a closed issue that never covered this suite.

### The suite is green on main today, so this is purely additive

`make -C src/agent-memory test-openclaw-plugin` at `69b6869a`: exit 0.

| Stage | Result |
| --- | --- |
[...]

PR #3333, opened 2026-09-18
feat(sight): default the dashboard to Agent Health

### LINKED ISSUE #2723: feat(sight): Dashboard 默认落地页改为 Agent 看板（/health）
## 背景

当前 Dashboard 的默认落地页（`/`）渲染的是 **Agent 可观测**（`ConversationList`），即 LLM 调用会话列表。但对于日常运维场景，用户打开 Dashboard 第一眼更想看的是 **Agent 看板**（`/health`）——它展示所有 Agent 的在线/离线/挂起状态概览，信息密度更高、更适合作为首页。

## 期望行为

- 访问 Dashboard 根路径 `/` 时，默认展示 Agent 看板（`/health` 对应的 `AgentHealthPage`）
- 如果 `agent_health` capability 不可用，回退到当前行为（`agent_observability` → `/`，否则 `/sessions`）

## 当前实现

`dashboard/src/App.tsx` 中的路由逻辑：

```tsx
const LOCAL_DEFAULT_PATH = "/sessions";

function defaultPath(capabilities: string[]): string {
  return capabilities.includes("agent_observability") ? "/" : LOCAL_DEFAULT_PATH;
}

// Route "/"
<Route
  path="/"
  element={
    capabilities.includes("agent_observability")
      ? <ConversationList />
      : <Navigate to={fallbackPath} replace />
  }
/>
```

## 修改建议

1. 将 `defaultPath()` 的优先级改为：`agent_health` → `/health`，`agent_observability` → `/`，兜底 `/sessions`
2. 根路径 `/` 的 Route element 改为 `<Navigate to={fallbackPath} replace />`（统一走 redirect），或直接在 `/` 渲染 `AgentHealthPage`
3. NavBar 中可考虑将 Agent 看板排到第一位（当前排第二）

## 验收标准

- [ ] 全新打开 Dashboard（`http://localhost:3004/`）直接看到 Agent 看板
- [ ] `agent_health` capability 关闭时，回退到 Agent 可观测页
- [ ] 现有导航链接不受影响
- [ ] `cargo test` + 前端构建通过

Forrest-ly:
## Why

Opening the Dashboard at `/` renders Agent Observability (the LLM conversation list), while the NavBar already lists Agent Health first — and `/health` is the page an operator reads at a glance, because it shows every agent's online / offline / hung state instead of one session's token breakdown. #2723 asks for the root URL to land on the Agent dashboard, falling back to Agent Observability when `agent_health` is not advertised.

`/` cannot both render Agent Observability and redirect somewhere else, so this moves the conversation list to `/observability` and turns the bare root into a redirect to the landing page.

## What changed

- **New `dashboard/src/utils/navigation.ts`** — one table (`NAV_ROUTES`) holding path, label key, icon and required capability for every page, in navigation order, plus the two pure decisions built on it: `pathAllowed()` (the capability guard) and `defaultPath()` (the landing page). `App.tsx` and `NavBar.tsx` both read it, so the nav order, the guard and the landing page cannot drift apart; the offline `DEFAULT_CAPABILITIES` in `App.tsx` is derived from it too, instead of being a fourth hand-maintained copy of the capability list.
- **Landing page** — `defaultPath()` returns the first route in navigation order whose capability is advertised: `agent_health` → `/health`, then `agent_observability` → `/observability`, then the rest. Scanning the table instead of the old `agent_observability ? '/' : '/sessions'` test also keeps the redirect target inside the advertised set; that pair could name `/sessions` on a host that does not advertise it, and the guard would then reject its own redirect target and bounce between guard and redirect.
- **Routes** — `/` is now `<Navigate to={fallbackPath} replace />`, `/observability` serves `ConversationList`, and `/health` plus `/observability` are referenced by constant so the redirect contract and the nav entry cannot disagree.
- **Guard** — sub-paths match on a `/` boundary, so `/audit` no longer claims a sibling such as `/auditorium`; `/` is always allowed because it renders no page of its own (its target is capability-checked in its own right).
- **NavBar** — renders `NAV_ROUTES`. The exact-root special case in the active-link test is gone because no entry maps to `/` any more. Agent Health was already first, so the visible order is unchanged.

Per-route capability gating is unchanged, and Agent Observability is the only page whose address moved.

## Related issue

closes #2723

## User / Agent impact

- A fresh visit to `http://localhost:3004/` lands on the Agent dashboard and the URL settles on `#/health`. The post-login redirect sets `#/`, so it lands there too.
- Agent Observability moved from `#/` to `#/observability`. Its NavBar entry still works and still highlights correctly; bookmarks or shared links to `#/` now land on the Agent dashboard rather than on the conversation list. No other page changed address, and no backend route, API path or capability name changed — the SPA uses `HashRouter`, so `#/health` never reaches the server's own `/health` liveness probe.
- With `agent_health` absent from `/api/auth/status`, a fresh visit lands on Agent Observability, which is the fallback the issue specifies.

## Risk and compatibility

The documented behavior change is the Dashboard landing page and the `#/` → `#/observability` move for Agent Observability, both requested by #2723. Nothing privileged, no cross-component contract, no state or schema to migrate. Rollback is a straight revert; there is no flag to flip.

## Validation

Linux x86_64, Node v22.21.1 / npm 10.9.4, from `src/agentsight/dashboard`:

| Command | Result |
| --- | --- |
| `npm run typecheck` | clean |
| `npm run test:navigation` (new) | 12 passed, 0 failed |
| `npm run test:api-client` | 24 passed (unchanged) |
| `npm run test:i18n` | 5 passed (unchanged) |
| `npm run test:product-branding` | 2 passed (unchanged) |
| `npm run build:embed` | webpack compiled successfully, bundle emitted into `frontend-dist/` |

From `src/agentsight`, `cargo test --workspace --exclude ebpf-ifc-engine --exclude actplane-ifc-compiler --locked --no-fail-fast`: 1823 lib tests plus every integration target pass. The one failure is `enforcement::target::tests::is_init_pid_namespace_returns_true_in_test_runner`, which asserts the runner sits in the init PID namespace and fails only because this sandbox cannot `readlink /proc/1/ns/pid` (exit 1). This PR changes no Rust.

The new suite covers the landing-page priority (health → observability → nav order → `/sessions`), that the landing path is always one the guard accepts, the `/` redirect contract, per-route gating including sub-paths, and the prefix-boundary fix. Since the router itself is not unit-testable here, it also asserts at source level that `App.tsx` wires `/` to the redirect, `OBSERVABILITY_PATH` to `ConversationList` and `HEALTH_PATH` to `AgentHealthPage`, and that `NavBar.tsx` no longer keeps its own copy of the table.

Mutation-checked — each of these fails exactly one test: reverting the nav order (health after observability), re-gating `/` on `agent_observability`, dropping the `/` boundary from the prefix match, restoring the conditional `<ConversationList />` at `/`.

Not run here: no browser is installed in this environment, so the acceptance item "open `http://localhost:3004/` and see the Agent dashboard" is verified through the route table and the production bundle rather than a rendered page. Also, `npm run test:navigation` is not wired into `.github/workflows/ci.yaml` — the agent token has no `workflow` scope, the same blocker as #3255. `ci.yaml` runs only `test:i18n` of the four dashboard suites today, so this matches the existing convention rather than regressing it.

## Documentation and rollback

No documentation file describes the Dashboard route table — `README.md`, `docs/DEVELOPMENT.md` and `AGENTS.md` only mention `npm run dev` on :3004 — so there was nothing to update. Per `AGENTS.md` §13 the `CHANGELOG.md` entry belongs to the version-bump commit, not to this one. Rollback: revert the single commit.


kongche-jbw:
@codex review

src/agentsight/dashboard/src/App.tsx
```diff
@@ -156,14 +126,11 @@ const AppShell: React.FC<{ status: AuthStatusResponse | null }> = ({ status }) =
...
-                capabilities.includes('agent_observability')
-                  ? <ConversationList />
-                  : <Navigate to={fallbackPath} replace />
-              }
-            />
+            {/* The bare root renders no page: a fresh visit lands on the
+                Agent dashboard (#2723). `/observability` keeps serving the
+                conversation list, so the nav entry still has an address. */}
+            <Route path="/" element={<Navigate to={fallbackPath} replace />} />
+            <Route path={OBSERVABILITY_PATH} element={<ConversationList />} />
```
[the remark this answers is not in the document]
Forrest-ly: Fixed in `fc1dade1`.

You're right, and the drift was worse than one sentence: the two paragraphs and both screenshots that follow that line describe the observability page (session rows with Token totals and interruption badges, a row expanded into its conversations, then **Details** into the trajectory). With `/` redirecting, a reader who followed the guide was looking at the Agent Dashboard while the prose walked them through a different page.

Both QUICKSTART guides now name the real landing page and give the session analysis page its own address, so the walkthrough still has a page to point at:

- **en** — "…and you land on the Agent Dashboard (`http://127.0.0.1:7396/#/health`) — live Agent health plus the unresolved interruption inbox.", then "The session analysis page keeps an address of its own: pick **Agent Observability** in the navigation bar, or open `http://127.0.0.1:7396/#/observability` directly."
- **zh** — the same two steps, `Agent 看板` / `#/health` then `Agent 可观测` / `#/observability`.

Labels and paths are read off the UI rather than invented: `nav.agentHealth` renders as "Agent Dashboard" / "Agent 看板" and `nav.agentObservability` as "Agent Observability" / "Agent 可观测" in `i18n.tsx`, and the addresses are `HEALTH_PATH` / `OBSERVABILITY_PATH` from `utils/navigation.ts`. No screenshot was regenerated — both still show the observability page, which is now what the surrounding prose says they show.

I synced `dashboard.md` (en + zh) as well, beyond the two guides you named. `src/agentsight/AGENTS.md` maps navigation and `capabilities`-probing changes to that page, and it is the page-by-page reference where a reader looks for an address. Its "Navigation and page availability" section now records where a fresh visit lands, that the bare root renders no page of its own and only redirects, and that the landing page follows navigation order over the advertised capabilities.

One correction worth recording, since it changes what the docs may promise: on a real host the landing page is the Agent Dashboard unconditionally. `detect_capabilities` in `src/agentsight/src/server/capabilities.rs` extends the list with `agent_health` outside every probe, so the fallback chain only ever bites for the offline `DEFAULT_CAPABILITIES` path in `App.tsx`. I documented the fallback anyway — it is the router's actual contract, `defaultPath()` is what enforces it, and a reader on a host that does stop advertising `agent_health` should not find the guide contradicting the redirect.

Verification, per the doc rules in `src/agentsight/AGENTS.md`:

- `bash scripts/docs-lint.sh` — naming convention OK, en/zh tree parity OK
- `python3 scripts/docs-link-check.py` — all relative links resolve
- `npm run build --prefix website` — en and zh both compile; the built `quickstart` and `dashboard` pages carry `#/health` and `#/observability` in both locales, and `prepare:docs` reports 90 English / 90 Chinese documents paired

Dashboard suites re-run green against the documented contract: `typecheck` clean, `test:navigation` 12/12 — including "a fresh visit lands on the Agent dashboard", "the landing page falls back to Agent observability, then to the nav order" and "App.tsx redirects the bare root and serves observability at its own path" — plus `test:api-client` 24, `test:i18n` 5, `test:product-branding` 2.

Same honest caveat as the PR description: no browser exists in this sandbox (no chrome/chromium/firefox), so the post-login page was never observed rendered. The addresses come from the route table, the capability probe in Rust, and the production site build — not from a live session.

src/agentsight/dashboard/package.json
```diff
@@ -8,6 +8,7 @@
     "typecheck": "tsc --noEmit",
     "test:api-client": "node tests/run-api-client-regression.cjs",
     "test:i18n": "node tests/run-i18n-regression.cjs",
+    "test:navigation": "node tests/run-navigation-regression.cjs",
```
jfeng18, line 11: P2: this script is not referenced by any workflow — `ci.yaml`'s test-agentsight job only runs `npm run test:i18n` for the dashboard, and there is no aggregate `test` script, so the new navigation regression suite never gates a PR. Please wire `npm run test:navigation` into the dashboard steps (or add an aggregate script) in this PR; otherwise the suite is dead infra from day one.

src/agentsight/dashboard/src/utils/navigation.ts
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Agent observability.
+ *
+ * Moved off `/` so the bare root can redirect to the landing page instead of
+ * rendering a page of its own; the conversation list keeps its own address.
+ */
+export const OBSERVABILITY_PATH = '/observability';
+
+/** Landing path of last resort, for a capability set that advertises no route. */
+export const LOCAL_DEFAULT_PATH = '/sessions';
```
jfeng18, line 42: P3 (latent): `defaultPath([])` returns this `/sessions` fallback, but `pathAllowed('/sessions', [])` is `false` — the doc comment above `defaultPath` claims the landing path is always one the guard accepts, which fails exactly when the fallback is used (the guard-acceptance test skips `[]`, and the sibling test only asserts the returned value). Unreachable in production today because `capabilities.rs` always advertises `sessions`, but if that ever changes, the guard redirects to the current path and the user gets a blank page. Suggest either stating the non-empty-by-construction assumption in the comment, or making the fallback guard-accepted by construction.

jfeng18 (commented):
Solid change overall — the nav model as a single source of truth, the boundary-matched guard, and the honest regression suite are the right shape. Verified the motivation against base: the old `observability ? '/' : '/sessions'` default could indeed hand the guard a target it would reject (the bounce was real). Two findings, details inline:

- **P2**: `test:navigation` is not wired into any workflow, so the new regression suite never runs in CI — dead infra from day one. Suggest adding `npm run test:navigation` to the dashboard steps in `ci.yaml`'s test-agentsight job. (Pre-existing siblings `test:api-client`, `test:product-branding`, and `typecheck` share the gap; wiring them all is welcome, but the new one belongs in this PR.)
- **P3 (latent)**: `defaultPath([])` returns `/sessions`, which `pathAllowed(..., [])` rejects — contradicting the stated invariant that the landing path is always guard-accepted. Unreachable in production today (the server's capability list always includes `sessions`), so a comment-level fix is enough.

Everything else checks out: boundary matching fixes the `/audit` vs `/auditorium` sibling claim, the exact-root special case removal is consistent end to end, the docs sync bilingually (and the Codex-flagged QUICKSTART drift is fixed in fc1dade1), and the tests discriminate (per-route `without(capability)` assertions, plus source-level wiring checks).

src/agentsight/dashboard/src/utils/navigation.ts
Forrest-ly: Fixed in b6a3e59d — took your first option.

`LOCAL_DEFAULT_PATH`'s comment now states that the fallback is deliberately not guard-accepted and why (both servers seed every capability list with `sessions`: `detect_capabilities` in `src/server/capabilities.rs`, and the fixed list in `src/local/server.rs`), `defaultPath`'s claim is scoped to non-empty capability sets, and the regression test pins both halves of the exception instead of skipping it:

```js
assert.equal(defaultPath([]), LOCAL_DEFAULT_PATH);
assert.equal(pathAllowed(LOCAL_DEFAULT_PATH, []), false);
```

I did not make the fallback guard-accepted by construction: that needs either weakening `pathAllowed` for one path or special-casing the guard in `App.tsx`, and both would let a page render and query a capability the host never advertised — worse than an honest comment for a case production cannot currently produce.

Standalone commit rather than a `--fixup` into c7927541, where the overclaim originated: squashing it in means a force-push, which would mark your review comments outdated and drop their line anchors mid-review.

src/agentsight/dashboard/package.json
Forrest-ly: Confirmed — `ci.yaml`'s `test-agentsight` job runs `npm ci` + `npm run test:i18n` only, and `package.json` has no aggregate `test` script, so `test:navigation` gates nothing as it stands.

I cannot wire it from this branch: `.github/workflows/ci.yaml` is a workflow file, and the credential this branch pushes with has `gist, read:org, repo` — no `workflow` scope — so GitHub rejects the push outright. Same blocker as #3255.

The step a maintainer needs in that job, right after "Run dashboard i18n regression":

```yaml
      - name: Run dashboard navigation regression
        working-directory: src/agentsight/dashboard
        run: npm run test:navigation
```

The pushable half is an aggregate script — `"test": "npm run test:i18n && npm run test:navigation && npm run test:api-client && npm run test:product-branding"`, with that job switched to `npm test`. Say the word and I'll add it; on its own it changes nothing in CI, because the job names `test:i18n` explicitly, so I left it out rather than land a script that looks wired up but isn't.

Forrest-ly:
Merged `main` (67743fb6) to clear the conflict with #3365, which landed while this branch was open. The reasoning is in the merge commit message; short version:

- `navigation.ts` gains the `/reuse` entry (label, icon, `reuse_labels`) in main's navigation position — after Risk Enforcement, before Trajectory Viewer — so the Reuse Labels page keeps both its nav entry and its capability gate now that NavBar renders the shared table.
- `App.tsx` and `NavBar.tsx` keep this branch's side: main's hand-maintained `DEFAULT_CAPABILITIES` list, its local `pathAllowed` / `defaultPath`, and its `navItems` copy are exactly the duplication this PR removes. `reuse_labels` reaches the offline default through `ALL_CAPABILITIES` instead of a second list to keep in sync.
- Both user guides merged cleanly — main's Reuse Labels rows and this branch's landing-page paragraph sit in the same sections without overlap.

The nav model and the `AppCapability` union still cover each other exactly (12 routes / 12 capabilities), and every route in `App.tsx` has a nav entry.

Branch head is now b6a3e59d, which also carries the P3 review fix; GitHub reports the PR `MERGEABLE` again. Validated from `src/agentsight/dashboard`: `npm run typecheck` clean; `test:navigation` 12, `test:api-client` 24, `test:i18n` 5, `test:product-branding` 2 — all passing; `npm run build:embed` compiles.

src/agentsight/dashboard/package.json
jfeng18: Wired in 56af69b7 (pushed directly to this branch with workflow-scoped credentials, since the author's token lacks the `workflow` scope): ci.yaml's test-agentsight dashboard steps now run `npm run test:navigation` right after `test:i18n`.

kongche-jbw (commented):
Review baseline: `2c276dd6ad21...56af69b7c805`

**[P2] Preserve capability fallback after token auto-login**

`src/agentsight/dashboard/src/App.tsx:25`

When a visitor opens `/?token=...`, the token branch returns before `fetchAuthStatus()`, so
`status` stays `null`. The new root redirect therefore expands `ALL_CAPABILITIES` and selects
`/health`, even if the host status omits `agent_health` but advertises
`agent_observability`. This misses the required observability fallback on the documented remote
login path. Possible direction: fetch and store auth status after token login, and add a routing
test for a token URL with `agent_health` absent.

jfeng18 (approved):
Approve. Both findings from my earlier review are now closed with evidence:

- **P3 (landing-path invariant)**: fixed in b6a3e59d — the fallback is documented as deliberately not guard-accepted, the claim is scoped to non-empty capability sets, and the regression test pins the exception explicitly (`pathAllowed(LOCAL_DEFAULT_PATH, []) === false`). The cited evidence checks out (`capabilities.rs` base vec and `src/local/server.rs` both seed `sessions`).
- **P2 (CI wiring)**: the author's token lacks the `workflow` scope, so I pushed the wiring commit (56af69b7) directly to this branch. Step-level proof from run 35508651649: "Run dashboard navigation regression" completed successfully.

Everything else verified in the first round still holds: boundary-matched guard, single-source nav model, bilingual docs in sync, discriminating tests. Ship it.

PR #3347, opened 2026-09-18
fix(memory)!: namespace OpenClaw memory tools

### LINKED ISSUE #3225: fix(openclaw): adapter enable path never releases memory_get/memory_search from memory-core (#3218 still reproducible)
Filed from #3222 review (`chatgpt-codex-connector` P1 on `src/agent-memory/adapters/agent-memory/openclaw/scripts/install.sh`): #3222 releases the `memory_get` / `memory_search` tool names by disabling OpenClaw's bundled `memory-core`, but it only does so on the `install.sh` entry point. The documented adapter-manager path never reaches that code, so #3218 stays reproducible there.

`docs/user-guide/{en,zh}/token-saving/agent-memory.md` offer two entry points side by side:

```bash
bash /usr/share/anolisa/adapters/agent-memory/openclaw/scripts/install.sh
```

```bash
anolisa adapter enable agent-memory openclaw
```

Only the first one performs the hand-off:

| Entry point | Runs `plugins install` | Disables `memory-core` | Restores it on removal |
|---|---|---|---|
| agent-memory `install.sh` | yes | yes, after #3222 | yes, `uninstall.sh` from the marker |
| `anolisa adapter enable` (anolisa-core driver) | yes — `apply_enable`, `openclaw.rs:507-516` | **no** | **no** |

`OpenClawDriver::apply_enable` (`src/anolisa/crates/anolisa-core/src/adapter/openclaw.rs:409`) builds the argv itself via `install_argv` (`openclaw.rs:2314`) and runs it at `openclaw.rs:516`, then proceeds through config apply and `verify_runtime`. It never executes the adapter bundle's `install.sh`, so an operator who follows the second block gets a loaded `memory-anolisa` whose `memory_get` / `memory_search` are dropped by OpenClaw's first-wins tool registry — the exact symptom #3218 reports, gateway log `plugin tool name conflict (memory-anolisa): memory_get` included. `verify_runtime` passes, because `memory-anolisa` itself really is `loaded`; nothing in the driver's enable path observes the name collision.

`anolisa adapter enable agent-memory openclaw` releases the two tool names the same way `install.sh` does, and `anolisa adapter disable agent-memory` hands the memory slot back — with the same guarantee #3222 pins for the script path: a `memory-core` the operator disabled themselves is never re-enabled, and a `memory-core` the driver disabled is never left behind.

## Possible approach

The receipt is the right place for the "who caused this transition" evidence, replacing the script path's marker file, but the existing config machinery cannot carry it:

- `disable` deliberately does **not** reverse config entries — `openclaw.rs:757`, *"Config entries are NOT reversed on disable (framework-wide config should persist)"*. A `plugins.entries.memory-core.enabled=false` entry declared through `ctx.declared_config` would therefore stay off forever after `adapter disable`, which is the harm the marker exists to prevent.
- So this needs its own receipt-tracked resource, not a declared config entry: a `ClaimResourceKind::FrameworkPlugin` (`claim.rs:537`) recording the displaced plugin id, referenced from `OpenClawClaim` (`claim.rs:604`) by resource id per the payload's "id references only" rule, with `DRIVER_SCHEMA_VERSION` bumped from 3 (`claim.rs:42`) and the round-trip/validation tests extended (`adapter_claim_toml_round_trip`, `claim.rs:1380`).
- Enable must read the prior state before disabling, because `plugins disable` is idempotent and exits 0 either way — its status cannot distinguish a real transition from a no-op. #3222 reads `plugins.entries.memory-core.enabled` through `openclaw config get` and treats only a positive false as somebody else's choice; the driver should apply the same rule and the same asymmetry (an unreadable answer records the disable, since skipping it strands the host with no memory plugin at all after removal).
- `disable` then runs `plugins enable memory-core` only when the receipt carries that resource, mirroring `build_uninstall_cmd` (`openclaw.rs:2419`) for argv construction, and reports it in `DisableReport.messages`.

Related: #3167 tracks the same *shape* of divergence — install entry points disagreeing on capability consent — across the same paths. If that lands a shared entry-point policy, the memory-core hand-off should ride the same decision rather than being special-cased again.

Until then the docs say which entry point performs the hand-off and give the manual commands for the driver path.

ikunkun-sys:
## Why

OpenClaw's bundled memory-core can remain loaded as a dreaming sidecar and own
`memory_get` / `memory_search`, hiding the agent-memory tools. Give the component's
OpenClaw tools distinct names so both backends can coexist through either
installation entry point, without adding plugin displacement to anolisa-core.

## What changed

- Expose `anolisa_memory_search` and `anolisa_memory_get`; keep native MCP methods,
  stored data, `memory_observe`, and `memory_get_context` unchanged.
- Update the manifest, memory prompt, name mapping and regression tests together.
- Declare all four contract tools in manifest `toolMetadata` for the `coding` profile.
  OpenClaw 2026.9.2 conversation policy recognizes them without installer writes
  to operator allowlists; explicit allow/deny and sandbox restrictions still apply.
- Remove #3222's explicit memory-core disable/restore commands from the scripts.
  Keep existing disable markers and warn with manual migration guidance: restoring
  a plugin with `plugins enable` also changes the selected slot.

## Related issue

Closes #3225. Addresses the tool-name collision reported in #3218.
Supersedes the displacement-based solution in #3228, which is now closed at the
requester's direction. Replaces the script takeover approach merged in #3222.

## User / Agent impact

Agents use the two `anolisa_` names to access ANOLISA memories. The unprefixed
names remain available to the host's own backend when that backend is loaded.
Existing prompts, skills, tool allowlists and direct callers of the old plugin
names must migrate; restart the gateway and start a new conversation after upgrade.
No old-name aliases are registered because they would recreate the collision.

## Risk and compatibility

The breaking change concerns the two OpenClaw-facing tool names and matching
policy entries. The manifest gives all four plugin contracts `coding` profile availability; it
does not override explicit restrictions. `group:memory` still means only the host
tools. Default sandbox sessions did not allow the old memory tools either; users
who intend ANOLISA access there must add the new names to their sandbox policy.
Hosts/tool surfaces that ignore manifest profile metadata can use explicit
`tools.alsoAllow`, as documented. No
anolisa-core changes or new displacement contract are included. OpenClaw still
owns memory-slot selection and its install/enable/uninstall side effects; removing
our explicit takeover does not redefine those host semantics. Legacy markers are
not silently discarded, and no automatic restoration overrides operator choices.

## Validation

Linux x86_64, Node.js 24.15.0:

- Plugin `npm run build:check`, `npm run build`, and `npm test`: 154 tests passed.
  The registration test now verifies every registered memory contract declares
  coding eligibility; this guard runs in the existing packaging test gate.
- Additional real OpenClaw 2026.5.7 loader/policy probe: profile metadata is
  ignored, but explicit `tools.alsoAllow` exposes all four plugin contracts.
  No new host floor is imposed; older-host configuration remains documented.
- `bash src/agent-memory/tests/test-openclaw-adapter-install.sh`: 53 cases passed,
  including fresh installs, reinstall/uninstall, legacy-marker retention, explicit
  off/third-party slots, and both enabled/disabled memory-core flags in the CLI stub.
- Real OpenClaw 2026.9.2 policy regression (12 cases): from the plugin directory,
  `npm run test:host-policy -- /path/to/openclaw-2026.9.2/node_modules/openclaw`.
  Uses the production manifest loader, plugin resolver and conversation policy
  pipeline. Reproduces the pre-fix coding omission, verifies coexistence after the
  fix (all four plugin contracts plus both host tools), preserves tool/plugin
  denies and core-only allowlists, and covers default
  sandbox, `group:memory`, explicit full/read-only sandbox grants, deny precedence, and the
  explicit `alsoAllow` fallback when metadata is absent. This separate integration
  test requires a 2026.9.2 installation; it is not part of `npm test`.
- ShellCheck on both scripts and the installer suite; docs lint, relative-link check,
  and `git diff --check`: passed.
- Actual OpenClaw 2026.9.2 plugin loader and tool resolver, the built plugin bundle,
  and an existing native agent-memory 0.2.7 binary in an isolated state/data directory:
  both host names and both namespaced names registered once; `memory_observe` wrote
  a unique marker, `anolisa_memory_get` read it back, and `anolisa_memory_search`
  returned the written file path through the unchanged MCP backend.

The real-host check directly exercises the production loader/resolver and tool
callbacks; it is not a long-running Gateway/model-session test. The installer suite
uses a CLI stub; a complete live AdapterManager installation was not run. Rust
sources are unchanged, so the Rust workspace suite was not rerun.

## Documentation and rollback

Updated both component READMEs and both user-guide locales with names, installation
behavior, policy configuration, and legacy-marker migration. Custom allowlists
and sandbox permissions remain operator-owned. The release bump must carry the
no-alias rename warning into the changelog; this feature PR does not edit release
changelogs under the repository rules. Internal MCP names and stored memories need
no migration. Reverting restores the old names and #3222's takeover behavior;
callers would need to revert their tool-name changes too.


Forrest-ly (commented):
## Code review — `fix(memory)!: namespace OpenClaw memory tools` @ `5bbe41e`

Verdict: **COMMENT (non-blocking)**. The code itself is correct, well-tested and
documentation-consistent — on code quality alone it is approve-worthy. I am
deliberately *not* filing a formal `--approve`, because this PR and the in-flight
user PR #3228 are mutually exclusive solutions to the same collision (#3218 /
#3225), and choosing the route is a human maintainer decision that should not be
pre-empted by a review approval that satisfies `REVIEW_REQUIRED`.

### What I verified (independently re-ran the suites)

- `npm run build:check` (tsc, against the real `openclaw` 2026.5.7 typings): clean.
- `npm test`: **154 pass / 0 fail** — including the new `tool-names-test.ts`
  ("registers distinct tools and routes ANOLISA reads to the existing MCP backend").
- `bash src/agent-memory/tests/test-openclaw-adapter-install.sh`: **53 cases pass**,
  including the new 12-case coexistence matrix (legacy marker × slot owner × memory-core flag).
- ShellCheck on `install.sh`, `uninstall.sh` and the installer suite: clean.

Correctness cross-check of the rename:

- `openclaw.plugin.json` `contracts.tools`, the two `api.registerTool` names +
  their `{ names: [...] }` option, the memory prompt, and `TOOL_NAME_MAP` all agree
  on `anolisa_memory_search` / `anolisa_memory_get`. `tool-names-test.ts` pins
  registered-tool set == manifest `contracts.tools`, and `options.names == [spec.name]`.
- Routing is right: `callTool("anolisa_memory_search")` → MCP `memory_search`,
  `callTool("anolisa_memory_get")` → MCP `mem_read`. Timeouts are keyed on the
  *resolved* MCP name, so the 120 s search budget is preserved.
- The corpus-read path correctly switched from `callTool("memory_get")` to
  `callToolByName("mem_read")` — necessary, since `memory_get` is no longer a map
  key and would otherwise fall through to a non-existent MCP `memory_get`.
- `CONTRACT_TOOLS_HIDDEN_BY_EXPERT` intentionally keeps the MCP-native names, and
  `config-test.ts` now maps contract→MCP (`resolveMcpToolName`) before intersecting
  with the child's Tier-B set — the profile-gate invariant still holds.
- `install.sh` / `uninstall.sh`: the #3222 disable/restore blocks are removed
  cleanly (no dangling `MEMORY_CORE_ID` / `RESTORE_MEMORY_CORE` / `MEMORY_SLOT_*` /
  `slot_value` references). Legacy markers are detected, warned about and **retained**;
  config cleanup only strips `memory-anolisa` keys/slots and preserves other backends
  and the operator's slot choice. The docstring slot value fix
  (`"agent-memory"` → `"memory-anolisa"`) matches the actual plugin `id`.
- Docs (EN + ZH) match the code: mapping table, migration steps, and the
  legacy-marker recovery procedure. No stale "disable memory-core" takeover text remains.

### Breaking-change hazard (please keep prominent in release notes)

There are **no old-name aliases**, so after upgrade a stale prompt / skill / tool
allowlist that calls bare `memory_search` / `memory_get` will now bind to the
**host's own** memory-core backend (not agent-memory) and can silently return
wrong or `disabled:true`-style results — i.e. the #3218 symptom, but for
un-migrated callers. This is inherent to the namespacing approach and is mitigated
by the updated prompt + migration docs, but operators MUST update callers, restart
the gateway and start a new conversation. Worth a line in the changelog/release notes.

### Non-blocking nits (optional)

1. `tests/smoke-test.ts:72` calls `callTool("memory_search")` (MCP-native name)
   while `:78` calls `callTool("anolisa_memory_get")` (contract name). Both resolve
   correctly, but the mixed convention reads oddly — consider using the contract
   name in both, or a one-line comment on why line 72 uses the native name.
2. `src/index.ts:159` (auto-recall) and `:528` (corpus supplement) call
   `callTool("memory_search")`, which now relies on the implicit `?? contractName`
   fallback in `resolveMcpToolName` (the explicit `memory_search: "memory_search"`
   identity entry was dropped). It is correct and covered by `mcp-client-test.ts:22`,
   but a short comment ("MCP-native name; passes through the map fallback") would
   prevent future confusion.
3. The `expert`-profile rejection message (`config.ts:154`) names `memory_search`
   (the child/MCP name) while the operator sees the registered `anolisa_memory_search`.
   Accurate from the child's perspective, but spelling both would read clearer.

### Route decision: #3347 vs #3228 (explicitly requested)

Both resolve the same collision and cannot both land:

- **#3347 (this PR, namespacing):** self-contained in `src/agent-memory/**`, no
  anolisa-core change, preserves OpenClaw's slot/host semantics, cleanly supersedes
  the #3222 script takeover, and is fully green (154 + 53 + tsc + ShellCheck).
  Cost: a breaking tool rename with no aliases and the un-migrated-caller hazard above.
- **#3228 (Forrest-ly, displacement):** adds a plugin-displacement contract in
  anolisa-core (heavier, cross-component), currently `CHANGES_REQUESTED`, with ~12
  in-flight Multica cards. Lets agent-memory keep owning the bare `memory_*` names.

My recommendation, on code-risk grounds: **#3347 is the more contained and
merge-ready fix** — smaller blast radius, no core contract, host semantics intact,
evidence reproduces. The trade-off is the migration burden of a no-alias breaking
rename, which the docs handle. That said, whether the project wants namespaced
coexistence (this PR) or core-level displacement (#3228) is an architecture call,
and #3228 is a user PR with substantial work invested — so a human maintainer
should adjudicate and close the losing route to avoid divergence. From a pure
code-quality standpoint this PR is good to go once the route is chosen.

kongche-jbw (commented):
Review baseline: `67dd3a1b916b...5bbe41ebba98`
**[P1] 命名空间工具会被默认 coding profile 过滤**
`src/agent-memory/adapters/agent-memory/openclaw/openclaw.plugin.json:12`
OpenClaw 2026.9.2 会把新本地配置默认设为 `tools.profile: "coding"`，
而该 profile 的 `group:memory` 只包含 `memory_search` / `memory_get`。
改名后两个 `anolisa_` contract 名不再命中 allowlist，会在模型调用前被过滤；
实际结果是默认本地会话仍只能看到 `memory-core` 的宿主工具，无法手动搜索或读取
ANOLISA 记忆。新增单测直接 mock `registerTool`，未经过 host resolver，因此会漏掉
这个路径；sandbox session 的 plugin-tool allowlist 也有同样问题。
Possible direction: 让两个安装入口在不覆盖 operator policy 的前提下把新名字加入
有效的 `tools.alsoAllow`，或采用 host 支持的 `group:memory` 扩展；补一条用
OpenClaw 2026.9.2 resolver、`tools.profile: "coding"` 和 sandbox policy 的回归。

ikunkun-sys:
Addressed the review feedback in **0794c7b83c51b277b8c2c07d86b2cdf4c70a71b6** (amended the original change per repository commit discipline).

**[kongche-jbw's P1](https://github.com/agentic-os-org/ANOLISA/pull/3347#pullrequestreview-5246443374): confirmed and fixed for the default coding conversation.** The original head really did expose only the host's `memory_search` / `memory_get` through OpenClaw 2026.9.2's loader plus conversation policy. The plugin now declares `toolMetadata.<new-name>.profiles: ["coding"]` for both renamed tools. The host's supported manifest profile mechanism preserves their visibility through either installer without adding policy mutation to anolisa CLI or install/uninstall scripts.

Sandbox needs a distinction: its default allowlist excluded the old memory tools too. An operator's explicit `group:memory` still expands only to the host names, so a sandbox intended to access ANOLISA needs the two new names in its effective `tools.sandbox.tools.alsoAllow`. Both user-guide locales now explain this migration, explicit deny precedence, and the `tools.alsoAllow` fallback for hosts/tool surfaces that do not consume profile metadata. We deliberately preserve operator restrictions rather than grant sandbox access at install time.

The new `tests/host-policy-test.mjs` runs **11 cases against actual OpenClaw 2026.9.2 modules**, including the pre-fix omission, post-fix coexistence, explicit tool/plugin denies, core-only allowlists, minimal profile, default sandbox, sandbox `group:memory`, explicit sandbox additions and deny precedence. It uses the production manifest loader, plugin resolver and conversation policy pipeline; it does not mock registration. Run after building the plugin with `node tests/host-policy-test.mjs /path/to/openclaw-2026.9.2/node_modules/openclaw`. This is a separate version-pinned integration test, not part of `npm test` and not a full Gateway/model-session test.

**[Forrest-ly's nonblocking suggestions](https://github.com/agentic-os-org/ANOLISA/pull/3347#pullrequestreview-5246402458): addressed.** The smoke test now uses the exposed contract names consistently; auto-recall and corpus search use explicit native `callToolByName("memory_search")`; the expert-profile error distinguishes MCP operations from `anolisa_memory_search`. The no-alias migration warning remains in both READMEs and user guides. The release bump should carry it into the changelog, which repository rules reserve for release commits. The requester already selected the component-owned namespacing route and instructed closing #3228; that PR is closed, so it is no longer a competing open route.

Validation: type-check/build passed, **154 plugin tests**, **53 installer cases**, and **11 real-host policy cases** passed; ShellCheck, docs lint/link checks and `git diff --check` passed. The separate real-host/native-MCP observe/read/search round trip also passed again. PR description updated with the policy behavior and validation boundaries.

kongche-jbw (commented):
Review baseline: `b0915370c81d...0794c7b83c51`
**[P1] coding profile 仍隐藏两个已声明的记忆工具**
`src/agent-memory/adapters/agent-memory/openclaw/openclaw.plugin.json:18`
在 OpenClaw 2026.9.2 以 `tools.profile: "coding"` 加载当前 manifest 时，
生产 policy pipeline 最终只保留两个 `anolisa_` 工具及宿主的 search/get；
`memory_observe` 和 `memory_get_context` 都不在可见工具列表中。
但插件仍把两者列入 `contracts.tools`，并在 prompt 中要求模型调用
`memory_observe`；默认安装因此无法显式记录观察或获取上下文。
`host-policy-test.mjs` 只用两项 `names` 过滤结果，所以当前 11 个 case 会全部通过。
Possible direction: 给这两个 contract 也声明 `profiles: ["coding"]`，并让宿主回归
断言默认 coding 会话可见全部四个插件 contract，同时保留显式 deny 测试。

Forrest-ly (commented):
## Code review — `fix(memory)!: namespace OpenClaw memory tools` @ `0794c7b`

**Verdict: COMMENT (non-blocking).** The [P1] raised against the previous head `5bbe41e` is genuinely closed, and every validation claim in the PR description reproduces exactly here. Two [P2] items below are worth acting on — one is a six-line change in the manifest block this PR already owns; the other is a version-floor / CI-hardening gap that is fine as a follow-up.

### What I verified independently (not just read)

I installed real `openclaw` hosts from npm and exercised this PR against them.

**P1 closure — mechanism confirmed in host source and host docs:**

- `openclaw@2026.9.2` `dist/conversation-capability-profile-B6PkMXFD.js`:

  ```js
  function resolveManifestToolProfileNames(snapshot, profile) {
    if (!profile) return [];
    return uniqueStrings((snapshot?.plugins ?? []).flatMap((plugin) =>
      (plugin.contracts?.tools ?? []).filter((toolName) =>
        plugin.toolMetadata?.[toolName]?.profiles?.some((candidate) => candidate === profile))));
  }
  // ...
  const profilePolicy = mergeAlsoAllowPolicy(
    resolveToolProfilePolicy(effective.profile),
    resolveManifestToolProfileNames(params.pluginMetadataSnapshot, effective.profile));
  ```

  The manifest declaration is merged into the profile policy as an `alsoAllow` contribution and flows into `policy.explicitToolAllowlist`; operator config is never rewritten. That is exactly the "host-supported mechanism other than `group:memory`" the previous review asked for.
- Host docs agree — `node_modules/openclaw/docs/plugins/manifest.md:506`: *"`profiles`: built-in tool profiles that expose the plugin tool by default. Valid values are `minimal`, `coding`, `messaging`, and `full`. These contributions merge into the corresponding profile allowlist; explicit operator allowlists and deny rules remain authoritative."*
- `"group:memory": ["memory_search", "memory_get"]` (host `dist/register-DDw_2wfh.js:3398`), so the guide's statement that `group:memory` does not expand to the ANOLISA names is exact rather than approximate.
- **Declaring only `coding` is parity-correct, not an under-scoped guess.** The host core catalog (`dist/tool-catalog-79RBtNnN.js`) defines `memory_search` and `memory_get` with `profiles: ["coding"]` and no other profile, and `CORE_TOOL_PROFILES.full = { allow: ["*"] }`. So `minimal` and `messaging` never exposed the old names, and `full` still exposes everything. No profile regresses.
- The rename premise is itself host-documented (`docs/plugins/building-plugins.md`): *"Tool names must not conflict with core tool names; conflicts are skipped and reported in plugin diagnostics."* `memory_search` / `memory_get` are core catalog ids, so #3218 was unavoidable under the old names.

**Ran the new integration test against a real host:** `node tests/host-policy-test.mjs ./node_modules/openclaw` with `openclaw@2026.9.2` on Node 24.15.0 → **all 11 cases PASS**. All eight hashed internal chunks it imports exist at 2026.9.2, and the `version === "2026.9.2"` guard at `tests/host-policy-test.mjs:13-16` is accurate.

**Reproduced the rest of the stated validation** (head `0794c7b`, pinned `openclaw@2026.5.7`, Node 24.15.0):

- `npm run build:check` — clean.
- `npm test` — **154 tests, 154 pass, 0 fail** (matches the PR exactly).
- `bash src/agent-memory/tests/test-openclaw-adapter-install.sh` — **53 PASS, 0 FAIL**.
- `shellcheck -s bash scripts/install.sh scripts/uninstall.sh` — clean.

**Previous round's non-blocking suggestions — all addressed:** auto-recall (`src/index.ts:159`) and corpus search (`src/index.ts:528`) now use `callToolByName("memory_search")`, corpus read uses `callToolByName("mem_read")` (`src/index.ts:555`); the `expert` refusal now names *"the MCP operations behind anolisa_memory_search, memory_observe and memory_get_context"* with `tests/unit/config-test.ts:283` pinning that wording; smoke-test titles use the exposed contract names; the no-alias migration warning is in `src/agent-memory/README.md`, `README_zh.md`, and both user-guide locales.

**Legacy `memory-core` marker — the riskiest part of dropping #3222's takeover — is handled correctly.** The marker is retained, both scripts warn, nothing auto-restores, and no `plugins enable` runs behind the operator's back. The installer suite now covers the `legacy={yes,no} × memory-core={true,false} × slot={memory-anolisa, memory-lancedb, none}` matrix, which is the right shape for this.

Two smaller things worth calling out: the `src/index.ts:7` header comment now says `plugins.slots.memory: "memory-anolisa"` (it previously named a plugin id that does not exist), and both `src/index.ts` and `tests/smoke-test.ts` gained their missing EOF newline.

### [P2] 1 — `memory_observe` / `memory_get_context` are still filtered out of the default `coding` conversation, while the injected prompt tells the model to call one

`openclaw.plugin.json:18-25` declares `profiles: ["coding"]` for the two renamed tools only. I probed the same real-host pipeline `host-policy-test.mjs` uses and printed the whole resolved tool set for the `coding` case:

```
CASE: coding keeps both ANOLISA tools and host tools
  visible: [ 'anolisa_memory_get', 'anolisa_memory_search', 'memory_get', 'memory_search' ]   (total 4)
```

`memory_observe` and `memory_get_context` are absent — in every one of the 11 cases.

The gate is name-based, in host `dist/tools-D1ohN2ZT.js`:

```js
function pluginToolNamesMatchAllowlist(params) {
  return !params.optional && params.allowlist.includesDefaults
      || params.allowlist.size > 0 && (params.names.length === 0
         || params.names.some((name) => params.allowlist.allowsTool(params.pluginId, name)));
}
```

`coding`'s allowlist is `listCoreToolIdsForProfile("coding") + "bundle-mcp"` plus manifest-profile contributions, and `memory_observe` / `memory_get_context` appear **nowhere** in the 2026.9.2 dist — no core catalog entry, no `group:memory` membership. `allowsPlugin("memory-anolisa")` is false because the allowlist holds neither `*`, `group:plugins`, nor the plugin id. So both are dropped: the same failure mode the P1 fix exists to prevent, applied to the other two contract tools.

Consequence: `src/index.ts:103`, `:104` and `:109` inject into the system prompt

- ``- `memory_observe` — Record an observation. The OS picks `notes/observed/<ulid>.md` and writes it.``
- ``- `memory_get_context` — Retrieve recently modified memory files as a preview, capped by tokens.``
- ``- After learning new information about the user, call `memory_observe` to persist it.``

in a default-`coding` conversation where neither tool is in the model's tool list. Auto-capture and auto-recall are unaffected (they call the MCP child directly at `src/index.ts:511` / `:159`), so this is a prompt-vs-tool-list inconsistency and a standing dead instruction for the model, not data loss.

**This is not a regression** — before the rename these two names were equally absent from the `coding` allowlist, so the 2-of-4 outcome is unchanged. But this PR owns the exact manifest block that fixes it, and I verified the fix:

```json
"memory_observe":     { "profiles": ["coding"] },
"memory_get_context": { "profiles": ["coding"] }
```

With those two entries added, my probe reports 6 visible tools under `coding` (`anolisa_memory_search`, `anolisa_memory_get`, `memory_observe`, `memory_get_context`, plus the host's `memory_search` and `memory_get`); the `pre-fix coding drops renamed tools` case still yields `[]`, and every other host-policy case is unchanged.

Worth fixing in the same pass: `tests/host-policy-test.mjs:43` hardcodes `const names = ["anolisa_memory_search", "anolisa_memory_get"]` and every assertion filters to `names`, so the suite structurally cannot see this. Deriving the expectation from `openclaw.plugin.json` `contracts.tools` would cover all four.

**Ask:** either add the two `toolMetadata` entries and widen the test, or scope it out explicitly — a follow-up issue plus one sentence in the user guide stating that `coding`-profile conversations expose only search/read. I'd prefer the former: six lines in a block this PR already rewrites, and it removes a false instruction from the prompt.

### [P2] 2 — the fix silently depends on a host ≥ 2026.8, but no floor is declared and CI validates against a host that ignores the metadata

I checked `resolveManifestToolProfileNames` across published hosts:

| openclaw | `toolMetadata.<tool>.profiles` honored |
|---|---|
| 2026.5.7 (this repo's pinned devDep) | **absent** |
| 2026.6.11 | **absent** |
| 2026.7.1 | **absent** |
| 2026.8.1 | present |
| 2026.9.2 | present |

So on any host ≤ 2026.7.x with a conversation profile configured, the new `toolMetadata` block is inert and both renamed tools vanish from the tool list — silently, with no alias and no diagnostic. The user guide does document the `tools.alsoAllow` fallback, so the case is mitigated; my concern is that the failure mode is silent and this repo already owns the loud-failure mechanism.

`src/agent-sec-core/openclaw-plugin/package.json:19` is the in-repo precedent:

```json
"openclaw": { "install": { "minHostVersion": ">=2026.4.14" }, "compat": { "pluginApi": ">=2026.4.14" } }
```

with `peerDependencies.openclaw: ">=2026.4.14"` and a unit test pinning all three (`src/agent-sec-core/openclaw-plugin/tests/unit/registration-test.ts:142-158`). Its README states the intent plainly: *"makes older OpenClaw hosts fail clearly during plugin install or non-bundled manifest loading."* The agent-memory plugin declares `peerDependencies.openclaw: "*"` and no floor.

**Ask:** declare a floor the same way (`openclaw.install.minHostVersion` + `openclaw.compat.pluginApi` + `peerDependencies.openclaw`, pinned by a unit test), at whatever version the team considers supported — `>=2026.8.1` is where the metadata starts working, `>=2026.9.2` is what the new integration test pins. If a floor is deliberately unwanted, say so next to the `toolMetadata` block and in both guides ("requires OpenClaw ≥ 2026.8; on older hosts use `tools.alsoAllow`").

**Related, and the reason this can't simply be pushed into CI today:** `devDependencies.openclaw` is still `2026.5.7`, which is what `npm test` and the packaging gate actually run against. I tried the bump so nobody has to discover it mid-change — at `2026.9.2` the suite is **not** green:

- `npm run build:check` → `tests/unit/corpus-routing-test.ts(21,32): error TS7016: Could not find a declaration file for module 'openclaw/plugin-sdk/memory-core-host-engine-storage'`.
- `npm test` → 153 pass / 1 fail: `tests/unit/corpus-routing-test.ts:187` *"is answered by the workspace file when one exists"* — the host's `readMemoryFile` returns `"builtin one\nbuiltin two"` at 2026.9.2 versus `"builtin one\nbuiltin two\n"` at 2026.5.7.

That divergence is pre-existing and out of scope here, but it means the plugin is currently regression-tested against a host four months older than the one whose behavior the P1 fix targets. Worth its own issue.

### [P3] 3 — the only automated guard for the P1 path is manual and hash-pinned

`tests/host-policy-test.mjs:17-38` imports eight minified internal chunks by content hash and single-letter exports (`{ o: loadPlugins }`, `{ n: resolvePluginTools }`, …). All eight resolve at 2026.9.2 today, and the version assert fails loudly rather than confusingly on any other host — that guard is well placed. But there is no npm script for it, no CI entry point, and `devDependencies` cannot currently supply a host that satisfies it, so a manifest regression that re-breaks the `coding` profile would ship unnoticed.

Cheapest durable guard, and it needs no host install: assert in the existing unit suite (which *does* run in packaging CI via `make test-openclaw-plugin`) that every name in `openclaw.plugin.json` `contracts.tools` either carries a `toolMetadata.<name>.profiles` entry containing `"coding"` or appears in an explicit exemption list. That catches manifest drift at PR time. Adding `"test:host-policy": "node tests/host-policy-test.mjs"` to the `package.json` scripts would also make the manual invocation discoverable instead of living only in the PR description.

### [P3] 4 — nits

- `scripts/install.sh:275` and `scripts/uninstall.sh:17` each hardcode `.anolisa-memory-anolisa-disabled-memory-core`. Deriving it as `"${OPENCLAW_STATE_DIR}/.anolisa-${PLUGIN_ID}-disabled-memory-core"` produces the byte-identical path the old installer wrote (legacy detection preserved) while keeping the two scripts from drifting if `PLUGIN_ID` ever changes. Separately, since nothing deletes the marker now, an upgraded host re-prints this WARNING on every future install *and* uninstall; that is intentional per the guide, but saying so in the warning text itself ("repeats until you remove the marker") saves an operator a documentation trip.
- `tests/unit/mcp-client-test.ts:21` — the case titled *"passes other OpenClaw contract names through"* still asserts `resolveMcpToolName("memory_search") === "memory_search"`. `memory_search` is no longer a contract name; it now exercises the identity fallback at `src/mcp-client.ts:101`. The assertion is still correct but the title is stale. That fallback also means a future stale `callTool("memory_get", …)` would be forwarded verbatim to the MCP child, which has no such tool, instead of failing fast — a drift guard asserting `Object.keys(TOOL_NAME_MAP)` equals `manifest.contracts.tools` would close that.
- `.github/actions/package-source/action.yaml:228` still describes the suite as "7 files / 140 tests"; after this PR it is 9 files / 154 tests. Stale before this PR, cheap to correct while the area is in flux.
- `src/benchmark/claweval-runner/src/ce_runner/tool_injector.py:73` denies `memory_get` / `memory_search` as "built-in gateway tools". Nothing under `src/benchmark` references agent-memory, so there is no live breakage — just note that this deny list no longer covers ANOLISA's memory tools by name if a benchmark ever enables `memory-anolisa`.

### CI at review time

26 checks on `0794c7b`: 10 passed, 16 skipped, 0 failed.

Forrest-ly:
Cross-reference for triage: my review above landed ~20 min after @kongche-jbw's review on this same head, and our main finding is the **same defect** — under `tools.profile: "coding"` on OpenClaw 2026.9.2 the resolved plugin tool set is only `anolisa_memory_search`, `anolisa_memory_get` plus the host's `memory_search` / `memory_get`; `memory_observe` and `memory_get_context` are dropped even though they are declared in `contracts.tools` and the injected prompt tells the model to call `memory_observe`. `tests/host-policy-test.mjs:43` filters every assertion to the two renamed names, so all 11 cases pass regardless.

One severity note so the two labels don't read as a disagreement: I marked it **[P2]** only because it is not a regression — those two names were equally absent from the `coding` allowlist before the rename (`coding` is `listCoreToolIdsForProfile("coding") + "bundle-mcp"`, and neither name exists anywhere in the 2026.9.2 host dist). Functionally it is the same gap @kongche-jbw calls [P1], the fix is the same one I verified (`profiles: ["coding"]` on both, → 6 visible tools under `coding`, pre-fix case still `[]`, all other cases unchanged), and it should be fixed in this PR since it owns that manifest block. Treat it as blocking if you prefer; the substance is identical.

Everything else in my review is additive and independent of that item: verification that the [P1] from `5bbe41e` is genuinely closed (host source, host docs, and running the 11 cases against a real 2026.9.2), reproduction of the stated validation (154/154 unit, 53/53 installer, clean `build:check` and ShellCheck), the host-version floor for `toolMetadata.<tool>.profiles` (absent at 2026.5.7 / 2026.6.11 / 2026.7.1, present at 2026.8.1 / 2026.9.2) with the in-repo `minHostVersion` precedent from `src/agent-sec-core/openclaw-plugin`, and the measured cost of bumping `devDependencies.openclaw` to 2026.9.2 (one TS7016 in `corpus-routing-test.ts:21`, one trailing-newline behavior change at `:187`).

Forrest-ly:
One correction to my review above, so the merge decision isn't ambiguous: I hedged with "treat it as blocking if you prefer", and that was too soft. Having cross-read @kongche-jbw's [P1] on this same head, **my position is that the `memory_observe` / `memory_get_context` gap should block this PR** and be fixed here rather than deferred.

To be explicit about the two labels, since they look like a disagreement and aren't:

- My **[P2]** tag stated one technical fact — *the gap is not a regression*. Before the rename those two names were equally absent from the `coding` allowlist, so the 2-of-4 outcome is unchanged by this diff. That is a statement about blame, not about severity.
- On **whether it should block**, I agree with [P1]. This is a breaking-change PR sitting at `reviewDecision=REVIEW_REQUIRED` with 0 approvals, the fix lives inside the one manifest block this PR already rewrites, and it costs roughly six lines plus widening `tests/host-policy-test.mjs:43` to derive its expectation from `contracts.tools` instead of a hardcoded pair of names. Merging as-is ships a default install whose injected prompt instructs the model to call `memory_observe` in conversations where that tool is not exposed — and splitting that into a second breaking-change window buys nothing.

Concretely, what I'd want before approving:

1. `openclaw.plugin.json` — add `"memory_observe": { "profiles": ["coding"] }` and `"memory_get_context": { "profiles": ["coding"] }`. Verified on real `openclaw@2026.9.2`: `coding` then resolves 6 tools (all four plugin contracts plus the host's `memory_search` / `memory_get`), the `pre-fix coding drops renamed tools` case still yields `[]`, and the other ten cases are unchanged.
2. `tests/host-policy-test.mjs` — assert all four `contracts.tools` names are visible in a default `coding` conversation, keeping the existing explicit-deny cases.

My **[P2] 2** (host-version floor: `toolMetadata.<tool>.profiles` is absent at 2026.5.7 / 2026.6.11 / 2026.7.1 and present at 2026.8.1 / 2026.9.2, while this plugin declares no `openclaw.install.minHostVersion` unlike `src/agent-sec-core/openclaw-plugin/package.json:19`) and the **[P3]** items are genuinely fine as follow-ups and are not reasons to hold the PR.

For the record, everything else in my review stands as verified: the [P1] from `5bbe41e` is closed via the host-sanctioned mechanism, and the stated validation reproduces exactly (154/154 unit, 53/53 installer, clean `build:check`, clean ShellCheck, 11/11 host-policy against a real 2026.9.2).

ikunkun-sys:
Addressed in **9e176afb206ba39e5e1d7b79b24bd9115f2edeb6**, folded into the original commit.

The shared finding from [kongche-jbw](https://github.com/agentic-os-org/ANOLISA/pull/3347#pullrequestreview-5259083444) and [Forrest-ly](https://github.com/agentic-os-org/ANOLISA/pull/3347#pullrequestreview-5259158443) is **confirmed and fixed**. Before changing the manifest, I widened the real-host regression to use `contracts.tools`; it failed on this head with `memory_observe` and `memory_get_context` missing. Both now declare `profiles: ["coding"]`. Default coding exposes all four plugin contracts plus the two host tools. The gap predated the rename, but it belongs in the complete contract fix here.

Regression coverage now checks the entire manifest contract set, not just the renamed pair. **12 real OpenClaw 2026.9.2 cases pass**, including explicit tool/plugin denies, restricted profiles, and a new sandbox read-only grant that must not expose observe/context. The existing registration unit test also checks coding eligibility for every registered tool, so packaging's normal `npm test` gate catches missing declarations without installing another host. Added the discoverable command `npm run test:host-policy -- /path/to/openclaw-2026.9.2/node_modules/openclaw`; the real-host suite remains a separate, explicitly version-pinned check.

Other feedback dispositions:

- **Host floor:** the compatibility concern is valid. I verified the actual 2026.5.7 manifest parser ignores `profiles`, then exercised its real loader/resolver and policy filter: coding alone hides the plugin tools, while explicit `tools.alsoAllow` exposes all four. We therefore deliberately retain this older-host configuration path instead of rejecting those installations with a new minimum version. Both user guides now state the distinction and show all four explicit grants; sandbox additions remain operator-controlled. The SDK/devDependency upgrade is a separate compatibility change and is not included here; this reply does not claim to have independently reproduced the reported 2026.9.2 typing/newline failures.
- **Stale MCP test title:** corrected to distinguish native MCP names from identity-mapped contracts. The passthrough behavior itself is intentional; the exposed search/read mapping and registered/manifest contract agreement remain tested. No new public export or speculative rejection path was added solely to inspect the private mapping table.
- **Legacy marker:** retain the literal historical filename. A future plugin-id change must still discover markers written by the old installer, so deriving this migration key from a potentially changed current id would be undesirable. Repeated warnings until manual recovery/removal remain intentional and documented.
- **Packaging comment / benchmark:** the static test-count comment is stale, but changing it does not affect execution; it is left outside this component fix. The benchmark denies host tools and has no current agent-memory integration, so there is no demonstrated benchmark regression to patch. These do not block the tool-contract correction.

Validation for this revision: `npm run build:check`, `npm run build`, **154/154 unit tests**, **12/12 real-host policy cases**, the separate 2026.5.7 explicit-grant probe, docs lint/link checks, and `git diff --check` passed. Installer scripts are unchanged from the previously validated 53-case revision. Updated the PR description and both README/user-guide locales. No anolisa CLI change or installer policy mutation was introduced.

kongche-jbw (commented):
Review baseline: `4accf530c8dc...9e176afb206b`

未发现阻塞问题。

Remaining risks:

- OpenClaw 2026.5.7 的显式授权回退未运行完整 Gateway/model session。
- 真实宿主测试依赖 2026.9.2 的私有 bundle hash，且不在常规 `npm test` 中。

Verification scope:

- 复查完整 diff、相邻实现、双语文档、测试及全部 PR 评论、Review 和 thread。
- `build:check`、`build`、154/154 单测、53/53 安装器用例均通过。
- OpenClaw 2026.9.2 production policy pipeline 的 12/12 用例通过。
- 当前 PR checks 无失败（10 passed，16 skipped）；`git diff --check` 通过。
- 本机未安装 ShellCheck，因此未单独复跑该项。

kongche-jbw (approved):

Forrest-ly (commented):
## Code review — `fix(memory)!: namespace OpenClaw memory tools` @ `9e176af`

**Verdict: no blocking issues found. Everything the previous two rounds asked for is genuinely closed.** This review is posted post-merge (APPROVED + merged by @kongche-jbw at 03:16:11Z, ~30s after my review card was opened), so the items below are follow-ups rather than gates.

Review baseline: `67dd3a1b` (merge-base) → `9e176afb206ba39e5e1d7b79b24bd9115f2edeb6`. I also confirmed the rebase onto main was clean: merge commit `71490005` has `4accf530` as its single parent, and the `src/agent-memory/` + `docs/user-guide/` tree at `71490005` is byte-identical to `9e176af` — what I reviewed is exactly what landed.

### What I verified independently (not just read)

I installed real `openclaw@2026.9.2` from npm and ran this PR against it on Node 24.15.0.

**1. The [P1] gap is closed, and the host regression is not re-hardcoded.**

`npm run test:host-policy -- <openclaw-2026.9.2>` → **12/12 PASS**, including the pre-fix reproduction case. To confirm the assertions really derive from `contracts.tools` rather than restating four names — the exact weakness in the previous version — I mutated the manifest back to its pre-fix shape (deleting `toolMetadata` for `memory_observe` and `memory_get_context` only) and re-ran against the real host:

```
PASS missing profile metadata hides all plugin contracts
AssertionError [ERR_ASSERTION]: coding keeps every plugin contract and both host tools
-   'memory_get_context',
-   'memory_observe'
  actual: [ 'anolisa_memory_get', 'anolisa_memory_search' ]
```

The suite fails naming precisely the two tools that regressed. That is the guard working as intended.

**2. The CI-side guard really fails `npm test`.** Since the real-host suite is version-pinned and out of CI, this was the load-bearing question. On head: `build:check` clean, `npm test` **154/154**. Then two mutations:

- delete `toolMetadata.memory_observe` → **153 pass / 1 fail**, `registered memory contract memory_observe must be available in coding conversations`
- append an unregistered 5th name to `contracts.tools` → **153 pass / 1 fail** on the tools↔manifest set equality

So `tests/unit/tool-names-test.ts` catches both a missing profile declaration and manifest/registration drift without any host installed. The `Test agent-memory` job is a real gate here.

**3. No residual name collision for the two tools that stayed un-namespaced.** The host suite asserts tool *name* + *count*, and the resolved tool objects carry no plugin identity, so name+count cannot distinguish "ours" from "memory-core's" — the #3218 failure mode would look identical. I probed the real resolver directly with `memory-core` and `memory-anolisa` both enabled and our plugin holding the memory slot:

```
total resolved plugin tools: 6
{"anolisa_memory_search":1,"anolisa_memory_get":1,"memory_observe":1,
 "memory_get_context":1,"memory_search":1,"memory_get":1}
```

memory-core contributes only `memory_search` / `memory_get`, so `memory_observe` / `memory_get_context` are collision-free on 2026.9.2 today. Leaving them un-prefixed is safe as of this host release.

**4. Installer suite + shellcheck.** `test-openclaw-adapter-install.sh` → **53/53 PASS**, including the new 12-case legacy × slot-owner × memory-core-enabled matrix. ShellCheck (`-S warning`) clean on `install.sh`, `uninstall.sh` and the suite — worth recording since the approving review noted it could not run ShellCheck locally.

**5. Rename completeness.** No consumer was missed. The anolisa CLI driver hardcodes zero memory tool names and writes no tool policy, so both entry points stay correct. Every remaining bare `memory_search` / `memory_get` hit in the repo is either an MCP-layer name (unchanged by design), historical CHANGELOG/spec text, or a different component that does not consume this plugin (`benchmark/claweval-runner`, the qwenpaw adapters). en/zh user guides are at parity.

**6. Sandbox path.** All four sandbox cases pass on the real host: default restricted, `group:memory` does not alias the ANOLISA names, a read-only grant exposes only the two read tools, and deny beats `alsoAllow`. The per-case `JSON.stringify(cfg)` invariant confirms operator policy is never rewritten. The docs give a copy-pasteable `tools.sandbox.tools.alsoAllow` block and state plainly that installers grant nothing — acceptable tradeoff.

### Follow-ups (non-blocking)

**[P2] A hermeticity self-test was deleted as collateral damage.**
`src/agent-memory/tests/test-openclaw-adapter-install.sh`

The displacement removal is correct, but the block that went with it at base lines ~1002–1010 was *not* memory-core-specific:

```sh
# The suite must be hermetic: hostile ambient switch values must neither
# leak into baseline scenarios nor break the suite itself.
if [ -z "${TEST_HOSTILE_AMBIENT:-}" ]; then
    env AGENT_MEMORY_ACCEPT_CAPABILITIES=0 AGENT_MEMORY_SAFE_INSTALL=1 TEST_HOSTILE_AMBIENT=1 \
        bash "$0" > "$SANDBOX/output" 2>&1 || rc=$?
    ...
fi
```

`TEST_HOSTILE_AMBIENT` no longer appears anywhere in the repo. `scenario()` and `channel_scenario()` still `env -u` those two switches, so the suite is probably still hermetic in practice — but the guard that *proved* it is gone, and it uniquely covered "hostile ambient env does not break the suite itself", which the per-scenario unsets cannot. Suggest restoring it as a standalone tail block. While there: `TEST_UNSAFE` and `TEST_INSTALL_BIG` are ambient-sensitive and were never covered by it nor unset by the helpers — worth folding in.

**[P3] Assert tool provenance in the host suite, not just name + count.**
`tests/host-policy-test.mjs:~155`

Per finding 3 above, name+count cannot detect a future host release that starts registering `memory_observe` / `memory_get_context` in memory-core: first-wins would bind those names to the host backend, the name would still appear exactly once, and both this suite and `tool-names-test.ts` would stay green while #3218 silently returns for the two un-namespaced contracts. Provenance *is* assertable today — I confirmed the resolved objects expose `label`/`description`, and ours are distinguishable:

```
memory_observe     | label: "Memory Observe (agent-memory)"
memory_get_context | label: "Memory Get Context (agent-memory)"
memory_search      | label: "Memory Search"   <- memory-core, no suffix
```

One extra assertion that all four contract names carry the `(agent-memory)` label suffix would close this permanently.

**[P3] The pinned host suite fails confusingly on an under-versioned Node.**
`tests/host-policy-test.mjs:13`

It asserts the host *package* is exactly 2026.9.2 but not the Node runtime. On Node 22.21.1 it dies inside a dynamic import with a SQLite stack trace (`OpenClaw requires SQLite 3.51.3+ ... Upgrade to Node 22.22.3+, 24.15.0+, or 25.9.0+`) that points at the host, not at the test's own precondition. A `process.versions.node` check up front would make the failure self-explanatory. Related and pre-existing: the 8 hash-pinned bundle filenames plus minified export names (`o`, `n`, `t`, `r`) break on every host release, and npm already publishes 2026.9.3/.4/.5 — this suite has a short shelf life. The new `npm run test:host-policy` entry point does fix the discoverability half of that, which was the actionable part.

**[P3] Doc-comment drift after the rename.**
`src/config.ts:~118` — `CONTRACT_TOOLS_HIDDEN_BY_EXPERT` is documented as "Tier B tools ... that this plugin registers for the host's memory contract", but it now holds *MCP* names (`memory_search`), not registered contract names (`anolisa_memory_search`). The runtime rejection message was correctly updated to "the MCP operations behind anolisa_memory_search, ..."; the comment above the constant was not. Also `scenario()`'s comment still says "start every scenario from 'install.sh has never disabled memory-core here'" though install.sh no longer disables anything. Cosmetic only.

### On the host-floor decision ([P2]2 from the previous round)

The disposition — no `minHostVersion`, documented `tools.alsoAllow` fallback instead — is defensible and I verified its factual basis: 2026.5.7 is the pinned devDependency, its manifest parser ignores `toolMetadata.profiles`, and a hard floor would reject installs that genuinely can work via explicit grants. The docs cover it in both languages with a concrete JSON block, plus sandbox and legacy-marker guidance. So this is a deliberate tradeoff, not an oversight, and I don't think it warrants reopening.

One residual worth naming for a maintainer rather than re-litigating here: on a pre-2026.8 host a default install loads the plugin and injects the memory prompt — which instructs the model to call `anolisa_memory_search` and `memory_observe` — while those tools are invisible, so the model will attempt calls to tools that do not exist, with nothing warning at runtime. `src/agent-sec-core/openclaw-plugin/package.json:19` takes the opposite stance (`install.minHostVersion: ">=2026.4.14"` + `compat.pluginApi`) and fails loudly. Whether the two OpenClaw plugins in this repo should agree on that philosophy is a product-consistency call for a human, not something an automated reviewer should settle.

### Summary

The design is right: namespace the two colliding names, declare all four contracts for `coding` through the host's own `toolMetadata` mechanism instead of rewriting operator policy, and drop the memory-core displacement takeover that #3222 introduced. Both prior [P1]s are closed with regression coverage at two independent layers — a real-host integration suite that provably fails on the pre-fix manifest, and a unit-level guard that provably fails `npm test` without a host. Follow-up A (restore the hermeticity self-test) is the only one I'd want tracked as an issue.

PR #3349, opened 2026-09-18
test(cosh-ng): [shell] stop the ecs probe configure watchdog flaking under load

Forrest-ly:
## What

Test-only change in `src/cosh-ng/crates/cosh-shell/tests/protocol/ecs_probe.rs`
(+29 / -3). Splits one overloaded 5 s timing budget into two named budgets with
different meanings, so `configure_response_does_not_wait_indefinitely_for_process_exit`
stops flaking on loaded CI runners.

## Why

[`Test cosh-ng fast checks` failed on PR #3342](https://github.com/alibaba/anolisa/actions/runs/35322291094/job/105527452739)
— a zh-CN `/help` localization change that touches nothing this test can see:

```
thread '<unnamed>' (16717) panicked at crates/cosh-shell/tests/protocol/ecs_probe.rs:53:9:
fixture never received the request

thread 'ecs_probe::configure_response_does_not_wait_indefinitely_for_process_exit' (16714)
panicked at crates/cosh-shell/tests/protocol/ecs_probe.rs:193:42:
called `Result::unwrap()` on an `Err` value: Any { .. }

test result: FAILED. 78 passed; 1 failed
```

This is the only test in the file that polls for the fixture's pid **before** the
request is issued — the watchdog thread is spawned first, and only then does the
main thread call `registry_query`. So `started_pid`'s 5 s deadline had to cover
the main thread being scheduled into `registry_query`, the fork of `/bin/sh`, the
shell reading the request line, `mkfifo`, and the `printf` that records the pid —
while `run-test-gates.sh fast` runs `--test-threads=4`, i.e. three other
process-spawning protocol tests beside it, on a shared self-hosted runner. When
that chain overran 5 s the watchdog panicked, and `join().unwrap()` collapsed it
into an `Any { .. }` that named neither the test nor the cause.

## How

| Budget | Was | Now | Meaning |
| --- | --- | --- | --- |
| `FIXTURE_PID_BUDGET` | 5 s (inline) | 30 s | host scheduling only — asserts nothing about the product |
| `CONFIGURE_RETURN_BUDGET` | 2 s (inline) | 6 s | the actual assertion: configure must return without waiting for a child that never exits |

`CONFIGURE_RETURN_BUDGET` at 2 s was tight for a reason worth recording: the
production path in `cosh_core_registry.rs` already spends up to ~500 ms on
purpose — the 250 ms `wait_timeout` in the `auth`/`configure` branch, then the
250 ms SIGTERM grace in `terminate_and_reap_process`, which the fixture's
`trap '' TERM` always exhausts — leaving ~1.5 s of scheduling slack. 6 s keeps
the assertion sharp (a regression blocks until the watchdog kills the group, so
it cannot be mistaken for a pass) while tolerating a loaded runner.

`watchdog.join().unwrap()` becomes `.expect(...)` so a watchdog panic names the
watchdog instead of printing `Any { .. }`.

Neither budget change weakens a failure: `started_pid` still panics on expiry,
and it panics on its **first** call, so a fixture that genuinely never starts
fails as fast as before rather than once per iteration of the
`fifty_cancellations_reap_each_owned_child` loop.

## Verification (Linux x86_64, base `main` 67dd3a1b, rustc 1.96.0)

- **Reproduction of the exact CI failure.** A temporary 6 s sleep inserted
  between `thread::spawn(watchdog)` and `registry_query` — i.e. the starvation
  this change tolerates — fails the unpatched test with the CI signature
  verbatim (`ecs_probe.rs:53` "fixture never received the request", then
  `Any { .. }` at the `join().unwrap()`), and passes with the patch applied.
  Both experiments were reverted; they are not in this diff.
- **Regression control.** Replacing the configure branch's `wait_timeout(250 ms)`
  guard with a bare `child.wait()` still fails the patched test at 6.01 s with
  `configure waited for the fixture watchdog to kill its child`. The wider
  budget did not blunt the assertion. (Also reverted.)
- `cargo test -p cosh-shell --test protocol -- --test-threads=4` → **79 passed,
  0 failed** in 15.5 s, the same wall time as before the change: no happy-path
  cost.
- **Same suite under contention** (48 busy loops pinned to 2 CPUs, suite pinned
  to the same 2 CPUs): the failure set is *identical* before and after this
  change — `provider_lifecycle::cosh_core_sync_drains_child_output_while_writing_large_prompt`
  plus four `session_management::*` tests. All are pre-existing load
  sensitivities in modules this diff does not touch, and **no ecs_probe test is
  among them**. They are deliberately left alone here; they are a separate
  conversation.
- `cargo fmt --all -- --check` clean; `cargo clippy -p cosh-shell --all-targets -- -D warnings` clean.
- `scripts/check-test-inventory.sh`, `scripts/check-source-layout.sh` and
  `crates/cosh-shell/scripts/check-layout.sh` all pass. No test was added,
  removed or renamed, so the necessity registry input is unchanged.

## Impact

Unblocks the unrelated `Test cosh-ng fast checks` failure on #3342 and removes a
re-roll-the-dice failure mode for every future PR that touches cosh-ng.


SunnyQjm:
PR number: #3349
head_sha: 0815152f96bbc0b334bb2a5e06e6dfb2d494c272
reviewed_at: 2026-09-18T10:29:27Z


未发现 blocking package/module/public API 组织问题。

剩余风险：fixture 真正无法启动时，失败反馈最长延迟至 30 秒。未独立运行测试；建议以 CI 中 protocol 测试及 layout gate 结果为准。

Forrest-ly:
The red `Test cosh-ng` on this PR was unrelated to its diff (this PR only touches `tests/protocol/ecs_probe.rs`) and I have re-run the failed job.

What it actually was: `relay::raw_relay_bash_preserves_carriage_return_debug_output_before_guard_shadow` failed at `relay.rs:700` with `the guard filter dropped unrelated partial output`. Root-caused and reproduced locally (6 CPU hogs pinned to 2 cores → 8/8 failures): the fixture waited 200 ms between submitting its `trap ... DEBUG` line and submitting `/mode`, and that wait is a host-scheduling budget for the *gated* Bash slash route. The submit lowers the prompt gate; on a slow runner `/mode` is typed before the parser raises it again, so the relay takes the Rust intercept, Bash never runs the readline assignment and the fixture's DEBUG trap never fires — the assertion then blamed a filter that never saw the bytes. A `trap -p DEBUG` dump proved the trap was installed, and a file appended from inside the trap body was never created, so it never fired.

Fix is test-only and in #3402 (documented budget + an assertion message that names both causes). The same assertion flaked on 2026-09-15 from `fix/cosh-ng/token-plan-preflight` (job `104292506869`), also unrelated to that branch.

Separately, the same contention harness shows `renders_wrapped_ascii_slash_submission_exactly_once` failing the `*_exactly_once` oracle (command on screen twice, and an 8 s pre-submit wait does not help) — that is a product-side erase/redraw race, not a fixture budget, filed as #3401 with the reproduction recipe and the CI occurrence from 2026-09-14.

PR #3372, opened 2026-09-20
ci(memory): run plugin unit tests on PRs

ikunkun-sys:
## Why

Agent-memory's OpenClaw TypeScript unit suite runs during packaging but does not run in the per-PR `Test agent-memory` job. Add the existing test target after the Rust checks so plugin regressions fail PR CI, using Node 22.23.2 to match packaging and satisfy the imported OpenClaw runtime's engine requirement.

## What changed

- Set up Node 22.23.2 and run `make test-openclaw-plugin` in `test-agent-memory`.
- Refresh comments describing the suite's CI entry points.

## Related issue

Closes #3255. Supplements #3281.

## User / Agent impact

PRs changing agent-memory now receive plugin unit-test feedback before packaging. No runtime behavior changes.

## Risk and compatibility

- [ ] Public CLI, API, configuration, or documented behavior changed
- [ ] Privileged or security-sensitive behavior changed
- [ ] Cross-component contract changed
- [ ] Migration or rollback guidance is needed

Low risk. Adds npm dependency installation and test time to the existing component job. Node 20 cannot run the OpenClaw runtime imported by this suite; the version matches the established packaging test pin. Rust check ordering and existing nightly/release gates are preserved.

## Validation

Linux x86_64, Node 22.23.2, initially absent plugin node_modules:

- `make -C src/agent-memory test-openclaw-plugin`: 154 tests passed, 0 failed.
- `actionlint -shellcheck= -config-file /tmp/issue3255-actionlint.yaml .github/workflows/ci.yaml`: passed; temporary configuration declares the repository's two existing custom runner labels.
- Both changed YAML files parse successfully; `git diff --check` passed.
- Rust tests were not rerun: no Rust code changes. Remote CI is pending.

## Documentation and rollback

Updated inline CI/Makefile comments; no user documentation changes. Revert this commit to remove the PR gate while retaining the packaging gate.

PR #3377, opened 2026-09-20
ci(memory): pin plugin gate deps, add type check

Forrest-ly:
## Why

Forrest-ly's post-merge review on #3372 ([review 5259973673](https://github.com/alibaba/anolisa/pull/3372#pullrequestreview-5259973673)) pointed out that the per-PR plugin gate added by `0b9ea9f3` resolves its dependencies at run time and never type-checks, and that two comments describe constraints which do not actually hold. This PR lands three of those four points.

The fourth point — `cache: 'npm'` + `cache-dependency-path` on the per-PR `setup-node` step, plus a pointer comment for the four hardcoded `22.23.2` pins — lives in `.github/workflows/ci.yaml`. Pushing a workflow file requires an OAuth token with the `workflow` scope (`remote: refusing to allow an OAuth App to create or update workflow .github/workflows/ci.yaml without workflow scope`), so it is delivered as a ready-to-apply patch on the tracking issue instead and will be re-submitted here once that scope is available.

## What changed

- **`src/agent-memory/Makefile` (`test-openclaw-plugin`)** — installs with `npm ci --legacy-peer-deps --no-audit --no-fund` from the committed lockfile instead of `npm install`, and runs `npm run build:check` (`tsc --project tsconfig.json --noEmit`) between the install and the tsx suite. This closes the type-gate gap: `npm test` is `tsx --test`, and tsx strips types without checking them exactly like the esbuild in `build-openclaw-plugin` does, so until now a PR that broke the plugin's types merged green and only blew up when nightly/release packaging ran tsc.
- **`src/agent-memory/Makefile` (`build-openclaw-plugin`)** — the type-check rationale no longer says the PR CI job merely "runs the plugin unit suite"; it now records that `test-openclaw-plugin` carries the same `build:check` (which is what keeps type errors out of main) and that this copy still gates the bundle that actually ships.
- **`.github/actions/package-source/action.yaml`** — dropped the false ordering claim. It asserted that `sync-versions` runs first and `manifest-config-schema-test.ts` asserts on the rewritten `openclaw.plugin.json`, so the suite had to follow the build step. Re-verified here: `sync-versions` rewrites only `.version`, and none of the three tests that read `openclaw.plugin.json` (`manifest-config-schema-test.ts`, `tool-names-test.ts`, `config-test.ts`) mention `version` at all — `grep -rni version` over those files returns nothing. The suite is self-contained and passes on a bare checkout, which is precisely what the per-PR job already proves. Only the real rationale (fail packaging before the archive is created) is documented now, together with the deliberate loss of the warm `node_modules`.

## Related issue

no-issue: follow-up to the review comments on #3372 (merged as `0b9ea9f3`); tracked internally as AGE-8564.

## User / Agent impact

No runtime behaviour change. Agent-memory PRs now get type-error feedback before merge instead of at packaging time, and the plugin gate resolves dependencies from the committed lockfile rather than from whatever the npm registry serves that day — a breaking patch to `tsx` / `esbuild` / `c8` / `typescript` can no longer redden every agent-memory PR with an error that reads like a plugin regression.

Developer-visible trade-off: `make test-openclaw-plugin` now requires `package.json` and `package-lock.json` to be in sync (standard `npm ci` contract), and on the packaging path it discards the `node_modules` that `build-openclaw-plugin` just warmed, adding one clean install (~15 s measured locally) to nightly and release. Accepted in exchange for a deterministic gate; `build-openclaw-plugin` keeps `npm install` so `make build` on a dev box still tolerates a freshly edited `package.json`.

## Risk and compatibility

Low risk. No shipped artifact, target name, or documented interface changes (`make test-openclaw-plugin` is not referenced by any agent-memory doc — the only README hits for that name belong to `src/agent-sec-core`). Both callers of the target are updated in lockstep by this PR's semantics: `.github/workflows/ci.yaml:1101` (per-PR) and `.github/actions/package-source/action.yaml:258` (nightly/release). The PR job still needs no `jq`: `build:check` requires nothing beyond the devDependencies installed by `npm ci`, and the target deliberately does **not** call `build-openclaw-plugin`, whose `sync-versions` prerequisite needs `jq` and the PR job installs only `libsystemd-dev` + `cmake`.

## Validation

Linux x86_64, Node v22.21.1 / npm 10.9.4, base `a35d3d2e` (current `main`):

- `make -C src/agent-memory test-openclaw-plugin` — full new recipe green end to end: `npm ci` added 594 packages, `tsc --noEmit` clean, `154 pass / 0 fail` (26 suites), ~22 s total.
- Type gate actually gates: dropped `export const probe: number = "not a number";` into `src/__typegate_probe.ts`, `npm run build:check` reported `error TS2322: Type 'string' is not assignable to type 'number'`; probe removed, tree clean again.
- Packaging-path lockfile consistency (the point the tracking issue asked to confirm): `make sync-versions VERSION=9.9.9` rewrote `.version` in `package.json` and `package-lock.json` together (plus the three other derived JSON files), after which `npm ci --legacy-peer-deps --no-audit --no-fund` still succeeded (594 packages). Tree restored with `git checkout --` afterwards.
- `actionlint 1.7.7 -shellcheck= -config-file <custom-runner-labels>` on `.github/actions/package-source/action.yaml`: byte-identical findings before and after this change (no new ones). Same check on `.github/workflows/ci.yaml` for the pending cache patch: only the pre-existing `workflow_dispatch` 11-inputs warning.
- `python3 -c "yaml.safe_load(...)"` on both YAML files: parse clean.

## Documentation and rollback

Comment-only documentation changes inside the two files above; no user docs, no CHANGELOG entry (per AGENTS.md §13 the changelog is touched only by version-bump commits). Rollback is a straight `git revert` of the single commit — nothing persists state, and reverting restores the previous `npm install`-based target.


kongche-jbw (commented):
Review baseline: `70059ae1a496...953179d9722f`

未发现阻塞问题。

Remaining risks:

- PR CI 的 npm cache 不在本轮 diff 内，registry 可用性风险按 PR 描述留待后续。
- 本地为 Node 24/npm 11；目标 Node 22.23.2 已由本轮 `Test agent-memory` CI 覆盖。

Verification scope:

- 完整检查两文件的 diff、相邻 Make/CI/打包调用链及前序 Review。
- `make -C src/agent-memory test-openclaw-plugin`：类型检查通过，154/154 测试通过。
- `sync-versions VERSION=9.9.9` 后 `npm ci` 通过；YAML 解析和 diff check 通过。

PR #3387, opened 2026-09-20
fix(sight): rebuild prebuilt objects for inode_guard_val

### LINKED ISSUE #3379: [sight] 3fefd1364 prebuilt objects do not match its C source: inode guard silently fails open on fresh installs
### [sight] 3fefd1364 prebuilt objects do not match its C source: inode guard silently fails open on fresh installs

**Component:** agentsight (ebpf-ifc-engine)

`3fefd1364` changed `te_inode_guard`'s value from `__u32` (4B) to `struct inode_guard_val {flags, domain_id}` (8B) in `process.bpf.c`, and updated the vendored Rust side (`InodeGuardVal`, `guard_inode(ino, dev, flags, domain_id)`) — but **did not rebuild the committed prebuilt objects**.

**Evidence** (kernel 6.6.102, root, fresh pin root):

- `bpftool btf dump file prebuilt/process.bpf.o | grep inode_guard_val` → empty; same for `process-inode-only.bpf.o`. Both objects predate the struct change.
- A fresh `PinnedEngine` install pins `te_inode_guard` with `value 4B` (created from the stale object), while the userspace `HashMap<FileId, InodeGuardVal>` writes 8B values → every `bpf_map_update_elem` fails with EINVAL.
- The failure is **silent at the binding level**: `apply()` still reports `Enforced`, but no guard entry ever lands, so `te_inode_guarded()` never matches and **unlink/rename/write on "guarded" files is NOT blocked** — a fail-open.
- The read side confirms the mismatch: `failed to clear inode guards during shutdown: pinned map te_inode_guard: invalid value size 8, expected 4`.
- Cross-check: swapping in prebuilts rebuilt from the current C source (the pair carried by #3264) restores the 8B map and the guard path works.

**Impact:** every deployment that fresh-installs the engine (no pre-existing pin) from current main gets enforcement that claims Enforced but does not block file operations guarded via the inode path — the 5.10/6.6 code path. Hosts with a pre-3fefd1364 pin keep the old consistent (4B) layout and are unaffected.

**Fix:** rebuild both prebuilt variants from the current C source and commit them. #3264 already carries the correctly rebuilt pair (plus the drain gate and the attested-build staging that keeps the object and the patch queue in sync); merging it fixes this desync as well. The e2e test in #3260 fails closed against this defect and stays red until one of the two lands.

**Reproduction:**

```
sudo rm -rf /sys/fs/bpf/actplane/v1
# install a binding with a `block unlink file "X" if AGENT` rule, then rm X
# → rm succeeds (exit 0) despite binding state Enforced
```

jfeng18:
## Description

Fixes #3379. `3fefd1364` changed `te_inode_guard`'s value from `__u32` (4B) to `struct inode_guard_val {flags, domain_id}` (8B) in the C source, but did not rebuild the committed prebuilt objects. A fresh `PinnedEngine` install pins `te_inode_guard` with a 4B value while userspace writes 8B values, so every `bpf_map_update_elem` fails with EINVAL — the binding reports `Enforced` but never blocks, a silent fail-open on the 5.10/6.6 path.

## Fix

Rebuilt both prebuilt objects from the current C source:

- Kernel 6.6.102, clang 17.0.6, via the crate's own Makefile (`make process`; the inode-only variant compiled with `-DINODE_GUARD_ONLY` through the same flags).
- No C source, Rust, or build-script change — binaries only.

## Validation

Verified on the built artifacts:

- `grep -c inode_guard_val`: **2** in both `process.bpf.o` and `process-inode-only.bpf.o` (previously 0 — the stale 4B-value objects).
- `bpf_d_path` (helper 147) call sites: `process.bpf.o` **3** (full variant, 7.1.x+), `process-inode-only.bpf.o` **0** (verifier-safe for 5.10/6.6).
- Sizes: `process.bpf.o` 1,649,880B; `process-inode-only.bpf.o` 1,489,312B.

## Risk and compatibility

Hosts with a pre-`3fefd1364` pin (4B layout) are unaffected by this defect and keep working. After this lands, a fresh install (or a cleared pin root) gets the consistent 8B layout. #3260's enforcement e2e should turn green against this fix.


viron-xiao:
Thanks for the quick stopgap for #3379 — rebuilding the objects is the right minimal fix, and the variant direction is correct (full object with bpf_d_path for 7.1.x+, inode-only without it for 5.10/6.6).

Two points before this lands:

1. **Coordination with #3264** — both PRs rebuild `process.bpf.o` from the same base blob into different objects, and #3264's prebuilt is a superset (it includes the `cap_pending_submitter` gate map). Suggested handling: if #3264 lands soon, close this as superseded; if main's 5.10/6.6 fail-open needs an immediate stopgap, land this first and #3264's rebase will overwrite its objects. Both orders work — the one to avoid is landing this AFTER #3264 (object downgrade).

2. **Follow-up, not blocking:** the root cause of #3379 — C source changed while the committed objects were not rebuilt — currently has no CI guard. A cheap assertion in CI (for example a symbol/BTF grep on the prebuilt objects for `inode_guard_val`) would make 'object does not match source' fail loudly instead of silently loading stale logic. Worth a small follow-up PR regardless of which stopgap lands.

jfeng18:
Hi @viron-xiao — gentle nudge for a review when you have a moment (following your approval on #1219).

This is a binaries-only rebuild of the two prebuilt BPF objects, fixing the `inode_guard_val` silent fail-open in #3379. I verified the artifacts on ECS with `pahole`: main's objects have no `inode_guard_val` (the stale 4-byte value), while both objects in this PR carry the 8-byte `struct inode_guard_val { flags, domain_id }` — the claimed fix is real at the binary level. No C/Rust source changes, CI is green (25/25).

Thanks!

jfeng18:
Superseded by #3264 (merged as `bee1da7d`).

The drain-fix PR carries prebuilt objects rebuilt from the same 8B `inode_guard_val` C source plus its gate changes (staged hash `db8b3a82`), and its staging mechanism also fixes the attested-build path (`build-enforcer.sh`), which this PR did not cover. With #3264 in main, the vendored objects are already the 8B pair — closing this as redundant. #3379 is fixed by that merge.

PR #3394, opened 2026-09-21
docs(tokenless): install SDK wheels from a published release

Forrest-ly:
## Why

The SDK user guide pins its copy-pasteable `pip install` examples to whatever version the working
tree carries, so every version bump rewrites them. A bump lands on `main` before the matching
`tokenless/vX.Y.Z` release is published, and only a published release serves wheel assets — so
since the 0.8.3 bump all four SDK install snippets have pointed at URLs that answer HTTP 404 until
the publish workflow uploads the wheels. Running the command exactly as documented on current
`main`:

```text
Collecting anolisa-tokenless==0.8.3
  ERROR: HTTP error 404 while getting https://github.com/alibaba/anolisa/releases/download/tokenless/v0.8.3/anolisa_tokenless-0.8.3-cp311-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
ERROR: Could not install requirement anolisa-tokenless==0.8.3 ... HTTP error 404 Client Error: Not Found
```

The release is missing, not just one asset: `GET /releases/tags/tokenless%2Fv0.8.3` answers 404,
while `tokenless/v0.8.2` publishes both wheels plus `SHA256SUMS-python-wheels.txt`. The release
*tag* page the guide linked answers 200 but carries no downloadable assets, so a reader following
the guide has nowhere to go.

This is the reader-facing half of the bump-before-publish window; #3392 closes the packaging half.
The docs have no gate at all, and nothing in the bump flow records that the URLs it rewrites must
already be served.

## What changed

Documentation only — `sdk.md` and `sdk/agentscope.md` under
`docs/user-guide/{en,zh}/token-saving/tokenless/`:

- the version moves out of the URL into a `TOKENLESS_VERSION` variable that the reader sets to the
  newest entry on the filtered releases page, so the commands keep working across a bump window and
  a version bump no longer has to touch the user guide;
- both install sections state the publish-before-download rule and link the existing
  troubleshooting section that already explains the same 404 for adapter bundles.

en and zh were edited together and stay mirrored. No production code, script, manifest, test or
workflow file is touched.

## Related issue

no-issue: documentation defect found while checking the 0.8.3 bump window. #3392 covers the
packaging side of the same window; this is the docs side.

## User / Agent impact

Readers who copy the SDK or AgentScope install command get wheels that download instead of an
HTTP 404, and the guide stops needing an edit at every release boundary. No runtime behavior
changes.

## Risk and compatibility

[...]

PR #3399, opened 2026-09-21
test(tokenless): run install-method suites in CI

Forrest-ly:
## Why

`make test-install-methods` — the three suites guarding the public npm / curl / Skill install routes added in 0a6fc322 — is a member of the `test` aggregate that no workflow ever called. Grepping `.github/` for `test-install-script`, `test-install-docs` and `test-npm-postinstall` returns nothing, so the only automated guard on `scripts/install.sh` / `scripts/uninstall.sh` (the receipt, the staged rollback, and the ownership marker that decides what an uninstall may delete), on the assertions that keep the four install routes and their docs agreeing, and on the npm postinstall that shares `~/.local/share/anolisa/adapters/tokenless` with the anolisa CLI ran on developer machines only. A regression in the route both READMEs point a first-time user at reached a release with the **Test tokenless** job green.

## What changed

One Makefile hunk: `test-install-methods` becomes a prerequisite of `test-adapters`, with the comment block recording why (the neighbouring `test-npm-package` block is the precedent). No workflow edit, no production code, no test code, and no change to what `make test` runs — make dedups the shared prerequisite, so each suite still runs exactly once.

`test-adapters` is the target the **Test tokenless** job calls as “Test adapter installers”, and the first one after `actions/setup-node@v4`. That matters: `test-npm-postinstall` drives the real `npm/scripts/postinstall.js` and *skips itself* when no interpreter can run it, so hanging the group off `test-integration` — which the job runs before node is set up — would have left the postinstall contract silently unenforced behind a passing step.

## Related issue

no-issue: CI-wiring follow-up for the install routes added in 0a6fc322, which shipped the suites but no caller.

## User / Agent impact

None at runtime. The **Test tokenless** job grows by about a minute (and only runs when `src/tokenless/` changes); installer and install-doc drift now fails CI instead of reaching a release.

## Risk and compatibility

Low risk — build graph only. The three suites are offline bash over a `mktemp -d` sandbox with `HOME` redirected (curl, npm and cargo are stubs inside them, and the jq-less scenarios are deliberate), need no cargo or rtk build, and touch the repo tree read-only, so unlike `test-npm-package` they share no directory with the plugin builds and add nothing to serialize under `make -j`. `test-install-docs` reads outside `src/tokenless/` (`docs/user-guide/*/token-saving/tokenless/`, `src/os-skills/`), read-only.

Known limit, unchanged by this PR: `detect-changes` sets `tokenless` only for `^src/tokenless/`, so a docs-only PR that edits the tokenless user guide or the install Skill without touching `src/tokenless/` still does not run the job.

## Validation

Environment: Linux x86_64, GNU Make 4.2.1, GNU bash 4.4.20, node v22.21.1 / npm 10.9.4 (the suites themselves need only bash + coreutils + node).

| Check | Result |
|---|---|
| `make -C src/tokenless test-adapters` (whole target, with the new prerequisite) | exit 0, 1m29s — 1160 `ok` + 73 `[PASS]`, 0 failures |
| ↳ `test-npm-postinstall` | exit 0, 0.9s, 52 ok |
| ↳ `test-install-script` | exit 0, 51.7s, 757 ok |
| ↳ `test-install-docs` | exit 0, 1.2s, 336 ok |
| `make -n test-adapters` ordering | openclaw build → dsh validate → npm packaging suite → postinstall → install-script → install-docs → adapter lifecycle suites |
| `make -n test` dedup | `test-npm-postinstall.sh`, `test-install-script.sh`, `test-install-docs.sh`, `test-package-npm-prebuilt.sh` each listed exactly once; exit 0 |
| Negative control A — doc drift (`Node.js 16.7+` → `16.8+` in the en QUICKSTART) | `make test-install-docs` exit 2: `FAIL en QUICKSTART states the npm-path prerequisite` |
| Negative control B — installer contract (`OWNER_MARKER_FILE=".tokenless-owner"` → `.tokenless-owner-x`) | `make test-install-script` exit 2 after 51 ok: `FAIL removes the tokenless link: unexpected file <sandbox>/npm-install/home/.local/bin/tokenless` |
| Negative control C — propagation through the target CI calls (control A plus `make -o build-openclaw-plugin -o build-dsh-plugin -o test-npm-package test-adapters`) | `make test-adapters` exit 2 at `test-install-docs`, i.e. the “Test adapter installers” step fails |
| Mutations reverted, tree clean apart from the Makefile | re-run of `test-install-methods` exit 0, 1145 ok, 0 FAIL |
| Base sanity on the merge base (not required by this change) | `cargo test --workspace --quiet` → 846 passed / 3 ignored (21 suites); `test-run-hook-install-scope`, `test-tool-ready-readable-fixer`, `test-claude-code-detect-retry`, `test-toon-cleanup`, `test-benchmark-rtk-version-probe` → all exit 0 |

Not run: the **Test tokenless** job itself (no GitHub runner here). This PR changes `src/tokenless/Makefile`, so `detect-changes` sets `tokenless=true` and the job — including the newly wired suites — runs on this PR; that run is the end-to-end evidence. `tests/test_rtk_integration.py` was not run either (it needs a local `cargo build --release --manifest-path third_party/rtk/Cargo.toml`), and three pytest modules that import `enum.StrEnum` need Python ≥ 3.11 while the local interpreter is 3.10 — none of them are touched by this change.

## Documentation and rollback

No user-facing docs change; the Makefile comments now record the CI caller. Rollback = revert this commit, which restores the previous `test-adapters` prerequisite list; the suites stay runnable directly via `make test-install-methods`.


Forrest-ly:
Closing this one: the premise it was built on turned out to be wrong, and merging it would run the three suites twice per tokenless CI run.

**They already reach CI.** `npm-package` has carried them as prerequisites since `0a6fc322`:

```make
npm-package: prepare-npm-native-binaries \
	test-install-script test-install-docs test-npm-postinstall-strict
```

and `CI / Components` calls `make -C src/tokenless npm-package` in the *npm packaging smoke test (native Linux)* step of the Test tokenless job. Grepping `.github/` for the target names misses that, because the wiring is transitive — which is what the comment block added here claims away.

**Evidence, main branch at `1e47735a` (this PR is not in it).** All three banners appear inside the `npm packaging smoke test (native Linux)` step of the Test tokenless job:

```
==> Testing standalone install/uninstall scripts...
==> Testing install-method documentation consistency...
==> Testing npm postinstall adapter ownership (a usable node is required)...
```

**Evidence, this PR's own run.** The same three banners appear twice in the same job — once under `make -C src/tokenless test-adapters`, again under `make -C src/tokenless npm-package`:

```
07:47:26  ==> Testing npm postinstall adapter ownership...            (test-adapters step)
07:47:26  ==> Testing standalone install/uninstall scripts...         (test-adapters step)
07:47:43  ==> Testing install-method documentation consistency...     (test-adapters step)
07:48:22  ==> Testing standalone install/uninstall scripts...         (npm packaging step)
07:48:43  ==> Testing install-method documentation consistency...     (npm packaging step)
07:48:44  ==> Testing npm postinstall adapter ownership (a usable node is required)...
```

That is ~55 s of duplicate work on every tokenless PR, plus a Makefile comment that contradicts the one already sitting directly above `npm-package` (which documents exactly this CI wiring).

**The one real thing left here is ordering, and it has a cheaper fix.** On main the suites run *after* `prepare-npm-native-binaries`, i.e. after a release build, so an installer or docs regression is reported several minutes later than it could be. If that fail-fast is wanted, reordering `npm-package`'s prerequisites (the three test targets before `prepare-npm-native-binaries`) gets it with zero duplication — say the word and I'll open that instead.

**Related hole found while re-checking this, different fix, not covered by this PR.** `tests/test-install-docs.sh` reads thirteen files that live outside `src/tokenless/` — the en/zh `docs/user-guide/*/token-saving/tokenless/` pages, `src/os-skills/{component.toml,README.md,README_zh.md,ai/install-tokenless/SKILL.md}`, `src/anolisa/manifests/components/{index.toml,os-skills/component.toml}` and `specs/documentation-standard.md`. `detect-changes` routes the Test tokenless job on `^src/tokenless/` only, so a PR that edits just those docs reports no component at all and the suite never runs. Reproduced locally: changing `Node.js 16.7+` to `16.8+` in the en QUICKSTART makes the suite exit 1 (`FAIL en QUICKSTART states the npm-path prerequisite`), while feeding that single path to the `detect-changes` script from `ci.yaml` leaves all eleven flags `false`, and the two checks a docs-only PR does run (`scripts/docs-lint.sh`, `scripts/docs-link-check.py`) both exit 0 — so that drift merges green. #3394 is a live example of the skip. Fixing it needs a workflow change (a path-filtered job for the suite, or a tokenless-docs route in `detect-changes`), which is outside what this PR was doing.

Forrest-ly:
Closing per the analysis above: the three install-route suites already reach this job through the `npm-package` prerequisites, so this change only executes them a second time. The ordering benefit (failing before the release build) is available without duplication by reordering those prerequisites instead — happy to open that on request. The separate docs-only-PR trigger hole described above needs a workflow change and is not addressed here.

Issue #3401, opened 2026-09-21
[cosh-ng] bug: a slash submission can render twice when the host is slow — readline echo not erased before the intercept panel draws (*_exactly_once oracle counts 2)


Forrest-ly:
### Component

cosh-ng — `crates/cosh-shell` raw relay: Bash slash route (`slash_via_shell`) plus intercept-panel rendering.

### Bug Description

When the host is slow to schedule the relay's reader/parser threads, one slash submission can reach the screen **twice**: once as the readline echo of the typed (or history-recalled) line, and once as the intercept panel's own copy. The `*_exactly_once` oracles in `crates/cosh-shell/tests/shell_host/relay.rs` catch it — `exact_screen_text_occurrences(&screen, command)` returns 2 instead of 1.

This does not look like a fixture timing budget: widening the fixture's pre-submit wait to 8 s does not change the outcome (see below), so the erase/redraw of the echoed line appears to lose a race against the panel draw rather than the fixture submitting too early.

### Steps to Reproduce

Deterministic locally — 6 CPU hogs pinned to the same 2 cores the test runs on:

```shell
cd src/cosh-ng
cargo test --package cosh-shell --test shell_host --no-run
BIN=$(ls -t target/debug/deps/shell_host-* | grep -v '\.d$' | head -1)
printf '#!/bin/bash\nwhile true; do :; done\n' > /tmp/hog.sh && chmod +x /tmp/hog.sh
for i in 1 2 3 4 5 6; do taskset -c 0,1 /tmp/hog.sh & done
taskset -c 0,1 "$BIN" \
  relay::raw_relay_bash_renders_wrapped_ascii_slash_submission_exactly_once \
  --exact --test-threads=1
```

3/3 failures at the shipped 200 ms wait, and still failing with that wait raised to 8 s.

CI, no contention harness needed — job `103911710429` (run `34823885735`, 2026-09-14, branch `test/cosh-ng/terminal-screen-2598`, whose diff does not touch this file):

```text
thread 'relay::raw_relay_bash_renders_recalled_slash_submission_exactly_once'
panicked at crates/cosh-shell/tests/shell_host/relay.rs:150:5:
assertion `left == right` failed: screen: [
    ...
    "[root@host ~]#  /mode",
    "[root@host ~]# /mode",
```

### Actual Behavior

Wrapped variant, local, 40 columns:

```text
"[root@host ~]# /mode aaaaaaaaaaaaaaaaaaa",
[...]

PR #3402, opened 2026-09-21
test(cosh-ng): [shell] stop the guard-shadow slash-route flake under load

Forrest-ly:
## What

Test-only. `assert_bash_guard_preserves_partial_debug_output` waited a fixed 200 ms between submitting the `trap ... DEBUG` setup line and submitting `/mode`. That wait is a host-scheduling budget for the relay's **gated** Bash slash route, and 200 ms is not enough on a contended runner.

With `slash_via_shell`, the relay hands a slash line to Bash only while the prompt gate is up (`shell_host/raw_runner.rs`: `prompt_ready raises the gated route; submits lower it`). Submitting the trap-setup line lowers the gate, and it rises again once the parser thread has consumed the following prompt. When the runner is slow, `/mode` is typed while the gate is still down, the relay takes the Rust intercept instead, Bash never runs the readline assignment — and the fixture's own DEBUG trap never fires. The assertion then reported `the guard filter dropped unrelated partial output`, pointing at a filter that had never seen those bytes.

## Why this is the gate and not the filter

Both CI failures are on this assertion, from branches that do not touch this file:

| date | test | CI job |
| --- | --- | --- |
| 2026-09-15 | `preserves_partial_debug_output_before_guard_shadow` | `104292506869` (`fix/cosh-ng/token-plan-preflight`) |
| 2026-09-18 | `preserves_carriage_return_debug_output_before_guard_shadow` | `105567153067` (#3349, diff = `tests/protocol/ecs_probe.rs` only) |

Reproduced locally rather than guessed — 6 CPU hogs pinned to the same 2 cores the test runs on, then `shell_host before_guard_shadow --test-threads=1`: **8/8 failures** at 200 ms.

A throwaway probe (not committed) separated the two candidate causes:

- `trap -p DEBUG` dumped to a file → the trap **was** installed verbatim, so the setup line landed fine;
- a second file appended from inside the trap body → **never created**, so the trap never fired and Bash never saw the slash line.

That rules out "the filter dropped it". Moving the *pre*-trap wait to 3 s changed nothing (still 8/8), which pins the race to the submit that lowers the gate, not to shell startup.

## Change

- `SLASH_ROUTE_GATE_BUDGET = 3 s`, documented as a host-scheduling budget with the measurement above, replacing the 200 ms wait between the trap line and `/mode`.
- The assertion message now names both candidate causes instead of blaming the filter outright.

No product code is touched, so the gate semantics #1721/#1718 asked for are unchanged.

## Verification

- 6-hog / 2-core contention: `before_guard_shadow` **3/3 iterations pass** (was 8/8 fail); both tests also pass unloaded.
- `cargo test --package cosh-shell --test shell_host -- --test-threads=4`: 216 passed / 1 failed — the same single failure as a pristine-`main` baseline run on this machine (`termios::scripted_shell_exit_timeout_kills_foreground_group`, which asserts `elapsed < 8 s` and fails on unmodified main here too).
- `cargo fmt --all --check` and `cargo clippy --package cosh-shell --tests`: clean.
- `scripts/check-test-inventory.sh`, `scripts/check-test-necessity.sh`, `scripts/check-source-layout.sh`, `crates/cosh-shell/scripts/check-layout.sh`: pass. No test was added, removed or renamed, so the necessity registry is unaffected.

## Not fixed here

Under the same contention `raw_relay_bash_renders_wrapped_ascii_slash_submission_exactly_once` fails 3/3 at the `exact_screen_text_occurrences == 1` oracle — the wrapped slash command reaches the screen twice (readline echo + panel copy), and an 8 s pre-submit wait does not change it, so it is not this budget. CI hit the same oracle on 2026-09-14 through `renders_recalled_slash_submission_exactly_once`. Filed with the reproduction recipe and both screen dumps as #3401; left alone here to keep this diff to the failure it proves.

Related to #3401

no-issue: test-only CI flake fix — no product behaviour change, and the defect found on the way is tracked in #3401 instead


SunnyQjm:
PR number: #3402
head_sha: 9e31d73bfd2517c08a56a2dec5f68f2c08184558
reviewed_at: 2026-09-21T10:17:16Z


未发现 blocking package/module/public API 组织问题。

剩余风险：3 秒固定等待仍依赖调度时序，只是扩大了容差；当前 CI 尚未全部完成。未在本次静态评审中运行测试。

PR #3444, opened 2026-09-22
fix(sight): add reuse fields to local test fixtures

jfeng18:
## Why

The macOS-only `local` module's lib test target no longer compiles on main: three `LocalState` literals in `local/server` tests are missing the `reuse_store` and `reuse_llm_judge_enabled` fields added when the reuse feature landed (E0063). macOS CI only runs `cargo build`, which never type-checks test code, so the drift is invisible there; it surfaces as soon as anyone runs the test target on macOS.

## What changed

Fill the two missing fields with `None` / `false` in the three fixtures (`local/server.rs` ×1, `local/server/trajectories.rs` ×2). The fixtures assert trajectory listing and auth endpoints, not labelling, so disabled reuse is the correct value.

## Related issue

no-issue: trivial test-fixture drift behind a platform-gated test target

## User / Agent impact

None — test code only, and on a module that is macOS-only in production.

## Risk and compatibility

Low risk: two added fields in test literals, no production code touched, no contract change.

## Validation

Force-compiled the macOS-only `local` module on Linux (temporary local gate + platform deps, reverted afterwards): `cargo test --lib local::` → 79/79 passed. Before the fix the same target failed with E0063 at all three sites, so the fix is exactly covering. `cargo fmt --all -- --check` clean.

## Documentation and rollback

No documentation changes. Roll back by reverting the commit.


chengshuyi (approved):
LGTM. 3 处测试 fixture 补齐 reuse_store/reuse_llm_judge_enabled，与全仓 LocalState 字面量核对无遗漏；local 模块为 macOS 门控，CI 只 build 不编测试导致漂移，修复恰好覆盖。纯测试代码，无生产改动。

